PR #9, opened 2026-04-08
feat(build): stage 4+5 — AST-based auto-globals scanner as default

JumpLink:
## Summary

**Stages 4 + 5 of the 4-stage globals tree-shaking refactor.** Stacks on top of #8.

Combines the Stage 4 experiment (an AST-based scanner for `--auto-globals`) and its positive conclusion in Stage 5 (promote AST to default, drop the experimental flag, add stable user docs and contributor rules). All in one PR, two commits per stage.

## Motivation

The Stage 3 regex scanner works, but has structural false positives:
- Shadowed identifiers: `const fetch = mock; fetch('…')` triggers an unnecessary `@gjsify/fetch/register` injection.
- Property access: `api.Buffer('hello')` triggers `@gjsify/buffer/register` even though the global `Buffer` is never used.

These false positives are harmless at runtime (register modules are idempotent) but add a few KB of dead weight per unused global. A real parser can do strictly better with negligible overhead.

### Commit 1 — `experiment(build): stage 4 — AST-based scanner for --auto-globals`

- New `packages/infra/esbuild-plugin-gjsify/src/utils/scan-globals-ast.ts` — full `acorn` + `acorn-walk` scanner. Strips TypeScript via `esbuild.transform` before parsing. Walks identifier references with a scope-aware check that respects:
  - Module-level bindings (imports, top-level `var`/`let`/`const`, function and class declarations)
  - Function parameters (declaration, expression, arrow)
  - Catch-clause parameters
  - Block-scope declarations
  - Reference positions only (skips member access property names, object literal keys, declaration identifiers, method names, labels)
- New `acorn` + `acorn-walk` + `@types/acorn` dependencies (~20 KiB total).

### Commit 2 — `experiment(build): stage 4 — add hidden --ast-scan flag`

(Merged into commit 1 historically; kept here for narrative.)

### Commit 3 — `refactor(build): promote AST scanner to default, drop --ast-scan flag`

After running the Stage 4 measurements on 3 reference showcases:

| Showcase | Regex | AST | Delta |
|---|---:|---:|---:|
| `express-webserver` | 498 ms | 514 ms | +16 ms (+3.2 %) |
| `canvas2d-fireworks` | 497 ms | 510 ms | +13 ms (+2.6 %) |
| `three-geometry-teapot` | 507 ms | 523 ms | +16 ms (+3.2 %) |

And on a synthetic shadowing test:

| Scanner | Injected registers | Verdict |
|---|---|---|
| **Regex** | `@gjsify/buffer/register`, `@gjsify/node-globals/register`, `fetch/register` | 2 false positives |
| **AST** | `@gjsify/node-globals/register` | Correct — only `process.env` |

The experiment concluded positively. This commit:

- Removes the `--ast-scan` CLI flag and all `astScan` plumbing through `CliBuildOptions`, `ConfigData`, `PluginOptions`, `config.ts`, and `actions/build.ts`.
- Changes `gjs.ts` to call `scanFileForGlobalsAst` unconditionally. On any exception it silently falls back to the regex scanner in `scan-globals.ts` — defence-in-depth for files acorn cannot parse (unusual TS syntax, etc.).
- Simplifies `scan-globals-ast.ts` — no more `try/catch` returning silently; any error propagates to the caller. The walk phase runs only after all parsing succeeds, so a thrown error cannot leave the accumulator in a partial state.
- Removes the "experimental" language from the `autoGlobals` JSDoc and updates the Stage 4 CHANGELOG entry to reflect the promotion.

The regex scanner in `scan-globals.ts` is kept as the silent internal fallback. It never appears in user-facing docs.

### Commit 4 — `docs(globals): user docs + AGENTS.md rules for tree-shakeable globals`

**User docs** (`website/src/content/docs/`):

- `cli-reference.md`: new `--auto-globals` and `--globals` rows in the `gjsify build` table, plus a new "Auto-globals" subsection documenting the scope-aware scanner, the modifier syntax (`+crypto`, `-fetch`), and the transitive-npm-dep escape hatch (`--globals +Buffer`).
- `how-it-works.md`: expanded from 2 mechanisms to 3 by adding "Automatic global registration". Describes the acorn-based scanner, its scope awareness, the regex fallback, and the entry-points-only limitation.
- `getting-started.md`: updated the "Using Node.js and Web APIs" section to explicitly state that users do NOT need `import '@gjsify/node-globals'` in typical projects.

**Contributor rules** (`AGENTS.md`):

New section **"Tree-shakeable Globals — `/register` subpath convention"** with 9 hard rules for every `@gjsify/*` package that registers globals:

1. No side-effects in `src/index.ts`
2. Side-effects live in `src/register.ts`
3. `package.json` declares both subpaths + `sideEffects` array
4. Every registered identifier must be in `globals-map.mjs`
5. Alias layer must have entries for both the bare specifier and the fully-qualified `/register` path
6. Tests that need globals use explicit `/register` imports
7. User code never needs manual register imports (auto-globals handles it)
8. Documented exception: intra-package class inheritance (only `@gjsify/eventsource` currently)
9. New-global checklist

Plus updates to:
- The build section documenting the scanner architecture
- The package-convention rules adding the `register.ts` layout note
- Fixed the `ensureMainLoop` reference to reflect the current re-export chain

**Stale reference cleanup**:

- `STATUS.md`: updated the `@gjsify/globals` and `@gjsify/web-globals` rows to reflect the split between pure root exports and `/register` side-effects. Added two completed entries for the `/register` refactor and the auto-globals scanner.
- `packages/{node/globals,web/web-globals}/README.md`: rewritten usage examples to explain that manual imports are unnecessary in typical projects.
- Memory file `feedback_globals_strategy.md` updated to the new subpath convention.

## Test plan

- [x] Full workspace build (`yarn build`) succeeds
- [x] All 53,310 workspace tests pass (`yarn test`, exit 0)
- [x] `--ast-scan` flag is gone from `gjsify build --help`
- [x] `--auto-globals` / `--globals` flags still visible
- [x] Express webserver still builds identically (1,690,214 bytes)
- [x] Minimal GTK hello-world still under 6 KB
- [x] AGENTS.md has the new Tree-shakeable Globals section with hard rules
- [x] Website docs build without broken links

## Refactor roadmap — complete

1. ✅ Stage 1 — Quick Wins (#6)
2. ✅ Stage 2 — `/register` subpath refactor (#7)
3. ✅ Stage 3 — Hybrid auto-injection (#8)
4. ✅ Stage 4 — AST-parser experiment ← **merged into this PR**
5. ✅ Stage 5 — Promote AST scanner to default, drop experimental flag ← **also merged into this PR**

## Known limitation — transitive npm dependencies (future work)

Both scanners only walk entry-point files, not external npm dependencies. If an npm package references globals internally that the user's own code never touches (e.g. a client library that calls `Buffer.from` inside its body), auto-globals won't inject the matching `/register`. Today's escape hatch: `gjsify build --globals +Buffer`.

A proper fix would be a two-pass build (first pass observes all loaded files via `onLoad`, second pass injects the collected register set). This is a separate future PR — **Stage 6** — not blocked by any of the 1–5 commits.


JumpLink:
Closing this PR without merging — the Stage 4+5 AST scanner approach is being superseded by a simpler explicit design.

## Postmortem

Stage 3 introduced a regex-based entry-point scanner. Stage 4 experimented with an AST scanner (acorn-based, scope-aware) and Stage 5 promoted it to the default. Both stages worked well for the three reference showcases.

Stage 6 attempted to extend the scanner to transitive npm dependencies via a two-pass build. Implementation kept hitting architectural dead-ends:

- **onLoad-based scanning** produced false positives from isomorphic libraries guarded by `typeof document !== 'undefined'` (e.g. `debug/src/browser.js`, which still member-accesses `document.documentElement`).
- **Scope split** (Node-runtime vs user-only globals) fixed the DOM case but left other edge cases open.
- **Metafile-based scanning** to only scan files surviving tree-shaking left the scanner set empty in ways that were hard to debug.

Each fix added a new special case without closing the underlying gap. Heuristic scanners on arbitrary npm dependency graphs are fundamentally unreliable — dynamic imports, `globalThis['X']` patterns, isomorphism checks, and runtime feature detection cannot be statically distinguished.

## New approach — see follow-up PR

A new PR based on `refactor/globals-stage-3` will:

- **Keep Stages 1-2 completely** (the `/register` subpath architecture, `sideEffects` arrays, pure root exports). This is the real framework win and is unchanged.
- **Remove all scanner infrastructure** — no acorn, no regex scanner, no AST walker, no two-pass build, no metafile parsing, no `--auto-globals` flag.
- **Keep a simplified `--globals fetch,Buffer,process,URL,crypto,structuredClone` CLI flag** — plain comma-separated whitelist, no modifier syntax, no scan baseline.
- **Default the flag in the `@gjsify/create-app` scaffolder** so typical projects get a working bundle out of the box without the user thinking about globals.
- **Document every known global identifier → register path** so users can look up what to add if they need something the default doesn't cover.

Net result: simpler codebase, fewer dependencies, predictable CLI, the full `/register` tree-shaking architecture from Stages 1-2 intact.

PR #70, opened 2026-05-05
fix(cli,esbuild-plugin-gjsify): merge config sources, outfile safety, PnP-zip warning

JumpLink:
## Summary

Three independent fixes that surfaced while validating the gjsify v0.3.8 PnP relay against [ts-for-gir PR #378](https://github.com/gjsify/ts-for-gir/pull/378). All three forced the consumer (ts-for-gir) to keep workarounds that should not be needed:

- A `gjsify.bin` field in package.json silently disabled `.gjsifyrc.js`
- `gjsify build` could overwrite the user's `src/index.ts` source
- The PnP rewriter could bake `.yarn/__virtual__/<pkg>.zip/...` paths into the GJS bundle that don't exist at runtime

### 1. Config precedence — merge `package.json#gjsify` with `.gjsifyrc.*`

**File:** `packages/infra/cli/src/config.ts`

cosmiconfig's default first-match-wins behaviour silently drops one source when both `package.json#gjsify` and an explicit `.gjsifyrc.*` / `gjsify.config.*` are present. A project that adds `gjsify.bin` to package.json (so `gjsify dlx` resolves the GJS bundle) was unintentionally disabling its `.gjsifyrc.js`.

The loader now:
1. Searches for the explicit file with `package.json` excluded from `searchPlaces`
2. Reads `package.json#gjsify` independently
3. Merges them — package.json is the lower layer, the explicit file wins on key collisions

### 2. Outfile safety — refuse to overwrite TypeScript sources

**File:** `packages/infra/cli/src/actions/build.ts`

`gjsify build` previously defaulted `--outfile` to `package.json#main`/`module`. For TS-direct projects that resolves to `src/index.ts`, and esbuild silently overwrites the source.

Added `isUnsafeDefaultOutput()` — refuses to default to any TypeScript extension or any path under `src/` and throws a clear error pointing at `--outfile`/`--outdir` or `gjsify.esbuild.outfile`.

### 3. PnP virtual-zip rewriter warning

**File:** `packages/infra/esbuild-plugin-gjsify/src/utils/rewrite-node-modules-paths.ts`

When the rewriter resolves `import.meta.url` for code inside a Yarn-PnP virtual zip cache, the relative URL points at a `.yarn/__virtual__/<pkg>.zip/...` path that doesn't physically exist outside the PnP runtime — any `readFileSync(URL-relative)` crashes with ENOENT.

Now emits a build-time warning (once per build) pointing at the mitigation (`nodeLinker: node-modules`) until the rewriter learns to inline zip-resident reads. This is the regression that forced ts-for-gir PR #378 to keep `nodeLinker: node-modules` despite v0.3.8 being designed to make that workaround unnecessary.

A real fix (rewriting zip-resident reads to use the bundled `__embed_<id>` content) is a follow-up — tracked in STATUS.md Open TODOs.

## Test plan

New e2e suite `tests/e2e/cli-config/` covers (1) and (2):

- `merges package.json#gjsify with .gjsifyrc.js` — installs a project with `gjsify.bin` in package.json AND `outfile` in `.gjsifyrc.js`, runs `gjsify build`, asserts the `.gjsifyrc.js` outfile is honoured (was silently dropped before)
- `refuses to default --outfile to src/index.ts` — installs a TS-direct project (`main: src/index.ts`), runs `gjsify build src/index.ts`, asserts non-zero exit + clear error message + source intact

Wired into root `package.json#test:e2e` and exposed as `yarn test:e2e:cli-config`. Both tests pass locally.

- [x] `yarn test:e2e:cli-config` passes (2/2)
- [x] Existing e2e suites unaffected
- [x] STATUS.md updated


JumpLink:
@SocketSecurity ignore npm/es5-ext@0.10.64

Accepted: pre-existing transitive devDep chain (`@gjsify/integration-webtorrent` → `webtorrent-fixtures` → `brfs` → `static-module` → `scope-analyzer` → `es6-map` → `es5-ext`). Lives only inside a `private: true` workspace; no published tarball ships this code. The postinstall hook is a political statement (peace message broadcast in Russian timezones, no-op elsewhere) with no security implications. `yarn.lock` is unchanged in this PR.

PR #73, opened 2026-05-06
feat(esbuild-plugin-gjsify,cli): rewriter asset extraction for portable bundles

JumpLink:
## Summary

Implements `extractNodeModulesAssets` — an opt-in build mode that makes the GJS bundle portable across machines and layouts (`gjsify dlx` cache, manual tarball extracts, Yarn-PnP node-modules linker). Replaces the closed-as-design-only #72.

Surfaced by [#70](https://github.com/gjsify/gjsify/pull/70) and confirmed by cross-testing ts-for-gir against v0.3.9 in a packed-tarball + dlx-cache layout — even outside PnP, the rewriter's source-relative paths break under any layout that isn't the original build site:

```
ENOENT: …/node_modules/typedoc/package.json
```

## How it works

When the option is on, the rewriter switches the URL form it emits for `import.meta.url` in `node_modules/` files:

```
Default (off):   new URL(\"../../node_modules/typedoc/dist/x.js\", import.meta.url)
Asset-extracted: new URL(\"./_node_modules/typedoc/dist/x.js\",   import.meta.url)
```

For every `node_modules/<pkg>/` whose files the rewriter touches, the package root is registered in a per-build `AssetRegistry`. At `build.onEnd`, each registered package root is recursively copied into `<bundleDir>/_node_modules/<safe-id>/`. The bundle plus that companion directory is then re-locatable to anywhere on disk.

Both the file-namespace rewriter (in `@gjsify/esbuild-plugin-gjsify`) and the PnP-relay's `transformContents` callback (in `@gjsify/cli`) write into the same registry via a `WeakMap<PluginBuild, AssetRegistry>` so a single `build.onEnd` hook covers every entry regardless of which onLoad ran for a given file.

## Off by default

Local-dev re-runs would otherwise re-copy ~5 MB of node_modules every build for no benefit (the source-relative URLs already resolve back to the live `node_modules/` tree). Enable for distribution bundles via:

- `gjsify build … --extract-node-modules-assets`
- `gjsify.extractNodeModulesAssets: true` in package.json
- or `extractNodeModulesAssets: true` in `.gjsifyrc.*` / plugin options

## Test plan

- [x] New `tests/e2e/portable-bundle/` suite (2/2 ✓): builds a tiny bundle that imports a fixture `node_modules` package whose entry reads a sibling JSON file via `import.meta.url`. Default mode keeps the legacy form and runs in-place; extracted mode produces `_node_modules/`, runs in-place AND from a moved location AND from a deeply-nested layout.
- [x] Cross-tested against ts-for-gir's GJS bundle:
  - 5.5 MB companion directory (typedoc + mini-shiki + yargs + yargs-parser)
  - Bundle runs from the build directory ✓
  - From `/tmp/ts-for-gir-portable/` (simple move) ✓
  - From `/tmp/dlx-sim/cache/abc/node_modules/@ts-for-gir/cli/bin/` (deeply-nested dlx-style layout) ✓
- [x] All 23 existing e2e tests still pass (cli-only + cli-config).

## What this unblocks

- ts-for-gir distributable via `gjsify dlx @ts-for-gir/cli` from npm — no more \"unpack into the same layout we built in\" constraint.
- Eventually drops ts-for-gir's `nodeLinker: node-modules` workaround once the smaller PnP-zip-extraction follow-up lands (still tracked as a single Open TODO in STATUS.md).


JumpLink:
Closing in favour of build-time inlining. Asset extraction was a pragmatic catch-all but not the architecturally correct solution — every real failure mode in ts-for-gir's bundle is a statically-resolvable `readFileSync(new URL(<literal>, import.meta.url), <encoding>)` pattern that we should inline as a string constant at build time, keeping the bundle a single self-contained file. Tracking the replacement work as the next follow-up.

PR #94, opened 2026-05-08
feat: gjsify dlx native-prebuild fix + minify-default + showcase shape

JumpLink:
## Summary

Three coupled changes — context: `npx @gjsify/cli showcase express-webserver` was crashing with `Typelib file for namespace 'GjsifyHttpSoupBridge' not found`. Root cause spans gjsify-side detection logic, showcase package shape, and a bonus build-default policy decision the user wanted in the same pass.

### 1. Bug fix — `gjsify dlx` / `gjsify run` finds transitive native typelibs

Previous detection ran two overlapping detectors:

- `detectNativePackages(cwd)` — exhaustive `node_modules` walker, started from `process.cwd()`. Empty when user runs from a dir with no node_modules (e.g. `npx … showcase`).
- `resolveNativePackages(bundlePath)` — `require.resolve` from the bundle, but iterated **only direct** `dependencies` of the bundle's `package.json`. Transitive typelibs invisible.

DLX layout (`~/.cache/gjsify/dlx/<sha>/.../node_modules/<pkg>/dist/bundle.js`) hits both blind spots, so `GI_TYPELIB_PATH` stayed unset and `gjs` couldn't resolve the typelib import.

**Fix**: collapse to one algorithm — `detectNativePackages(startDir)` — invoked from both `cwd` and `dirname(bundlePath)`, deduped by package name (CWD shadows bundle). Bundle-side walk now reaches transitive deps automatically.

### 2. Showcase shape — `@gjsify/example-node-express-webserver`

The showcase bundles `gi://GjsifyHttpSoupBridge?version=1.0` (express → @gjsify/http → @gjsify/http-soup-bridge). The bridge package was only a build-time dep via `@gjsify/node-globals` — never declared as a runtime `dependencies` entry. So `npm install` of the published tarball never shipped the typelib package, and the detection fix alone wouldn't help published consumers.

Added `@gjsify/http-soup-bridge` to `dependencies` of the showcase. Audit confirmed only express-webserver carries `gi://Gjsify*` imports today.

### 3. ⚠ BREAKING — `gjsify build` minifies by default

`--minify` flipped from `default: false` to no default + `defaultDescription: 'true'`. Opt-out: `--no-minify` or `bundler.output.minify: false` in `package.json#gjsify`. The four orchestrators (`app/{gjs,browser,node}.ts`, `library/lib.ts`) no longer hard-code `minify: false` after the `...input.output` spread, so user config now actually takes effect (it was silently overridden before).

Backwards-compat: `easy6502/app-gnome` already pins `bundler.output.minify: false` explicitly — keeps emitting unminified output. ts-for-gir-CLI and other unconfigured consumers get minified bundles after their next `@gjsify/cli` bump.

### Refactors

- Drop `resolveNativePackages` + `findNearestPackageJson` from `utils/detect-native-packages.ts` — one walker, two startpoints.
- Extract `computeNativeEnvForBundle(bundlePath, cwd)` from `runGjsBundle` as a pure function returning `{ env, envPrefix }`. Lets the e2e tests assert the env without spawning gjs.

## Test plan

- [x] New e2e suite `dlx-native-prebuilds/` (2 tests) — synthetic consumer with `@gjsify/http-soup-bridge` installed in node_modules + bundle in nested package dir, asserts `computeNativeEnvForBundle` populates `GI_TYPELIB_PATH` / `LD_LIBRARY_PATH` from the bundle-side walk alone (CWD has no node_modules — the regression scenario).
- [x] Existing 16 e2e suites green. `inline-static-reads` got `--no-minify` (substring asserts target object property names the default-on minifier mangles).
- [x] Manual minify smoke: default-mode bundle 4 lines / 4164 B; `--no-minify` 197 lines / 7054 B.
- [x] Manual showcase smoke: detection fix verified end-to-end via `gjsify run` against the workspace tree — `GI_TYPELIB_PATH` correctly contains `@gjsify/http-soup-bridge/prebuilds/linux-x86_64`. (A pre-existing express-CJS-interop error `superCtor.prototype must not be undefined` reproduces against the v0.3.16 baseline too — out of scope for this PR.)
- [ ] CI: full e2e matrix on PR run.

## Follow-up (separate PR)

`@gjsify/example-node-express-webserver` hits a `superCtor.prototype` error in `inherits()` from express's `send` module after this PR's detection fix gets the bundle past the typelib import. Reproduces against v0.3.16 main → not a regression introduced here. Needs investigation into the CJS-ESM interop layer (likely `@gjsify/stream` or `@gjsify/util` namespace-vs-default unwrap) — separate PR.


JumpLink:
@SocketSecurity ignore npm/markdown-it@14.1.1

PR #144, opened 2026-05-10
feat(cli): Phase D-2.B.5b — gjsify build CLI wire-up to native rolldown

JumpLink:
## Summary

Phase D-2.B.5b — wires the \`gjsify build\` CLI to use \`@gjsify/rolldown-native\` when \`GJSIFY_BUNDLER=native\` is set, otherwise stays on npm rolldown.

- New \`packages/infra/cli/src/bundler-pick.ts\` encapsulates the engine choice behind \`runBundle(finalOpts) → Promise<RolldownOutput>\`. \`buildApp\` and \`runOneLibraryBuild\` both call it instead of \`rolldown(opts).write()\`.
- Plugin-shape adapter \`toNativePlugin\` translates rolldown's per-hook \`{filter, handler}\` form to \`NativePlugin\` shape; \`filter.id\` regex/string sources become \`idFilter.<hook>\` regex strings exercising B.2's short-circuit.
- \`@gjsify/cli\` declares \`@gjsify/rolldown-native\` as an optional peer dep so npm consumers without the prebuild aren't broken.

> **Base branch is \`feat/rolldown-native-plugins-b5\` (PR #143)** — this PR should land after #143.

## Test plan

- [x] \`yarn check\` passes (TS strict)
- [x] \`yarn build\` passes
- [x] \`yarn test\` passes (full workspace; npm-default path unchanged)
- [ ] B.6 self-host smoke test — next PR exercises this wire-up under GJS

PR #145, opened 2026-05-10
test(integration): Phase D-2.B.6 — rolldown-native integration suite

JumpLink:
## Summary

Phase D-2.B.6 — permanent regression coverage for the Phase B plugin-bridge contract.

New \`tests/integration/rolldown-native/\` (\`@gjsify/integration-rolldown-native\`, private). 6 specs / 29 assertions, all green under GJS 1.88, exercising \`bundleWithPlugins()\` end-to-end:

- All 12 hook positions fire on a single multi-hook plugin (B.2)
- \`idFilter.load\` regex short-circuits non-matching ids (B.2)
- \`this.resolve()\` re-enters the resolver pipeline (B.3)
- \`this.resolve()\` from one plugin triggers another plugin's \`resolveId\` (B.3 re-entrancy)
- \`this.error()\` throws synchronously in the JS handler (B.3)
- rolldown-shaped \`{filter, handler}\` hooks dispatch correctly via \`toNativePlugin\` (B.5b adapter)

GJS-only because the native bridge needs the GjsifyRolldown typelib. Replaces the ad-hoc \`/tmp/rolldown-b{2,3,5}-*.js\` smoke scripts built up across B.1–B.5; permanent regression coverage now lives in version control.

> **Base branch is \`feat/rolldown-native-plugins-b5b\` (PR #144)** — this PR should land after #144.

The end-to-end self-host loop the original plan sketched (CLI bundles itself for GJS, then bundles a test project, diff vs Node-CLI output) needs the gjsify CLI's own runtime portability story to land first — tracked in STATUS.md "Open TODOs" rather than blocking the B-series merges.

## Test plan

- [x] \`yarn check\` passes
- [x] \`yarn build:test\` produces \`dist/test.gjs.mjs\`
- [x] \`yarn test:gjs\` — 29/29 assertions pass


JumpLink:
Replaced by clean-history version on rebased branch — see new PR.

PR #191, opened 2026-05-18
feat(worker_threads): cross-process SharedBuffer transfer via @gjsify/sab-native

JumpLink:
## Summary

Phase 1 of the `@gjsify/sab-native` plan. Wires the Phase-0 bridge (PR #190) into `@gjsify/worker_threads`:

- `Worker.postMessage(value, [sab])` and `workerData: { sab }` route `SharedBuffer` instances across the subprocess-Worker boundary via SCM_RIGHTS on an inherited fd 3 side-channel.
- In-process `MessagePort.postMessage(value, [sab])` accepts SharedBuffer in transferList (same-process pass-through; no fd-passing needed).
- Cross-process futex (drop `_PRIVATE` flag) so `atomics.wait32` / `notify32` work between Workers + parent on the same memfd-backed pages.

Closes the SharedArrayBuffer cross-process gap end-to-end.

> **Base**: this PR targets `feat/sab-native-bridge` (PR #190). Merge PR #190 first, then this becomes mergeable against `main`.

## Architecture

**Spawn (`worker.ts`):** `FdChannel.makePair()` → retain parent end, `launcher.take_fd(childFd, 3)` to inherit child end as fd 3 in the worker.

**Send (`worker.ts:_serializeWithSabTransfer`):** Walk the value tree (`extractSharedBuffers` helper in `sab-transfer.ts`), substitute each `SharedBuffer` with `{__sab: tag, size: N}`, `sendFd(parentFd, sb.fd, tag)` over the side-channel, then write the JSON line on stdin. Tags come from a per-Worker sequence counter.

**Receive (bootstrap template):** Per inbound JSON message, `countSabPlaceholders(msg.data)`, `drainSabFds(count)` calls `recv_fd` exactly N times — synchronous but never blocks because the parent's `sendmsg` always completed before the corresponding `write(stdin)`, so the fds are already buffered in the kernel by the time the JSON line arrives. Then materialise each placeholder via `SharedBuffer.fromFd(fd, size)`.

**Lifecycle:** `Worker._cleanup()` calls `fdChannel.closeFd(parentFd)` on shutdown. New `closeFd` API in sab-native is a thin wrapper over `close(2)` so consumers don't drag in GioUnix for a single syscall.

**Cross-process futex:** `sab-helpers.c` switched from `FUTEX_*_PRIVATE` to plain `FUTEX_*`. PRIVATE keys by per-process virtual address; SharedBuffer regions are mmap'd at different addresses in each Worker but back the same physical page via `memfd_create`, so PRIVATE would never match across processes.

**In-process MessagePort (`message-port.ts`):** transferList entry of type SharedBuffer is a no-op pass-through — same JS heap, structuredClone passes the instance through via the new branch in `packages/gjs/utils/src/structured-clone.ts` (same shape as the SharedArrayBuffer branch).

## Test plan

- [x] 4 new GJS-only worker_threads unit tests (`worker-shared-buffer.gjs.spec.ts`): atomic-store/load round-trip via child Worker; typed accessors mid-message; postMessage SharedBuffer delivery mid-stream; atomic add round-trip parent + child.
- [x] 1 new GJS-only worker-stress integration test (`sab-native-parallel-hash.gjs.spec.ts`): 4 subprocess Workers each SHA-256 a 256 KiB slice of a 1 MiB SharedBuffer via `GLib.Checksum`, write 32-byte digest into the region, bump barrier via `atomic_store_i32`; parent polls `atomics.load32` per slot, XOR-folds the 4 digests, asserts equality against in-process `createHash('sha256')` reference.
- [x] 275/275 worker_threads GJS unit + 270/270 Node unit (no regressions; sab-native suite wrapped in `on('Gjs')` so it's skipped on Node).
- [x] 1034/1034 worker-stress GJS integration + 1169/1169 Node integration.
- [x] Single-process sab-native suite still 18/18 (no regression from the futex-flag swap; the non-blocking probes use timeout=0 or value-mismatch paths so they don't observe the PRIVATE-vs-not distinction).

## Backwards compatibility

When `@gjsify/sab-native`'s prebuild is absent, `fdChannel` resolves to `null`. The Worker spawn skips the FdChannel setup, `_serializeWithSabTransfer` is a no-op, and `Worker.postMessage` falls back to the original JSON-only IPC. Only SharedBuffer-shaped payloads ever need sab-native — every existing consumer keeps working unchanged.

## STATUS.md

- `worker_threads` row: cross-process `SharedBuffer` in `Worker.postMessage` / `workerData` listed under Working; Missing column updated.
- Open TODOs: Workstream C section renamed to "SharedArrayBuffer constructor opt-in (Mozilla pref)" — Phase 0 ✓ Phase 1 ✓. What's left is just the SAB constructor (an upstream-GJS pref) + generic ArrayBuffer cross-process transfer (would need a binary IPC frame format; SharedBuffer covers the high-bandwidth case).
- Metrics: +5 tests across unit + integration (4 + 1).
- worker-stress section: third suite row for `sab-native-parallel-hash.gjs.spec.ts`.


JumpLink:
Re-targeted to `main` after PR #190 merged.

PR #247, opened 2026-05-22
docs(website): add Patterns section with GObject classes page

JumpLink:
## Why a new section

The existing docs split — *Documentation / Guides / Packages / Contributing* — has no good home for *idiom* / *recipe* content. **Guides** is task-oriented (*\"how do I ship a Flatpak\"*); **Packages** is API reference. Things like *\"how should I write a `GObject.registerClass` class?\"* don't fit either.

This adds a **Patterns** section in between. Each page lists preferred patterns with a short rationale, then names the rough edges so readers can recognise them before they bite.

## First page: GObject classes

Driven by:

- Philip Chimento's matrix thread about `static { GObject.registerClass() }` type safety — confirms the static-block pattern is the right call AND flags the `static $gtype: GType<Foo>` rough edge (TS static inheritance is invariant).
- The [GNOME/gjs#704](https://gitlab.gnome.org/GNOME/gjs/-/work_items/704) init-order trap (Gary Li's `vfunc_init breaks` reproducer).
- The two example directories we already have in ts-for-gir (`gobject-register-class-inference`, `gobject-param-spec`), plus a third one (`gobject-static-block-ordering`) coming in a follow-up PR.

Page covers: **three registerClass forms** (static-block-inline / static-block-with-fields / functional) with concrete code + when to pick which, **rough edges** (init-order trap, `$gtype` static narrowing, subclass re-declaration), a **quick checklist**, and references back to the upstream issue + working examples.

## Planned follow-ups

Index page lists the open patterns roadmap — bridges, signal handlers, `/register` convention, async on GJS, native bridges. Will land as the source material accumulates.

## Test plan

- [x] `yarn build` clean — 20 pages built in 6.29 s
- [x] both new HTML files render — `dist/patterns/index.html` + `dist/patterns/gobject-classes/index.html`
- [x] section anchors resolve (\"TL;DR\", \"Three forms\", \"Static-block ordering trap\", \"Quick checklist\")

### DIFF
--- a/website/src/content/docs/patterns/gobject-classes.md
+++ b/website/src/content/docs/patterns/gobject-classes.md
@@ -81,7 +81,7 @@ class Foo extends GObject.Object {
 }
 ```
 
-Works **iff** every `static [GObject.*] = …` initializer appears *above* the static block. ES class evaluation is strict source-order, so `static [GObject.interfaces] = …` must be assigned *before* `registerClass()` reads it. **Form A is preferable** because the rule isn't enforceable at the type level — a refactor that moves a static field around breaks Form B silently at runtime.
+Works **only if** every `static [GObject.*] = …` initializer appears *above* the static block. ES class evaluation is strict source-order, so `static [GObject.interfaces] = …` must be assigned *before* `registerClass()` reads it. **Form A is preferable** because the rule isn't enforceable at the type level — a refactor that moves a static field around breaks Form B silently at runtime.
 
 ### Form C — functional (no static block)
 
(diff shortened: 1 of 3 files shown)


website/src/content/docs/patterns/gobject-classes.md
```diff
@@ -0,0 +1,186 @@
...
+    static [GObject.interfaces] = [Gio.Initable];        // ← must come BEFORE registerClass
+    static [GObject.properties] = { … };
+
+    vfunc_init(): void { /* … */ }
+
+    static { GObject.registerClass(Foo); }               // ← still last
+}
+```
+
+Works **iff** every `static [GObject.*] = …` initializer appears *above* the static block. ES class evaluation is strict source-order, so `static [GObject.interfaces] = …` must be assigned *before* `registerClass()` reads it. **Form A is preferable** because the rule isn't enforceable at the type level — a refactor that moves a static field around breaks Form B silently at runtime.
```
JumpLink: typo

PR #316, opened 2026-05-26
test(e2e): gjsify gresource (glib-compile-resources wrapper)

JumpLink:
## Summary

- Adds `tests/e2e/gresource/run.mjs` — 3 `node:test` cases that exercise `gjsify gresource <xml>` end-to-end against a real `glib-compile-resources` invocation
- Registers the suite in `package.json`: appended to the `test:e2e` chain immediately after `gsettings`, plus a standalone `test:e2e:gresource` script
- Updates `STATUS.md`: Completed bullet, E2E-suites count 25+ → 26+, `gresource` added to the suite list, Current state header refreshed

## Interface exercised

`gjsify gresource <xml> [--sourcedir <dir>] [--target <path>]`

Fixture: a minimal `app.gresource.xml` referencing `hello.txt` (a single `.txt` resource), both written into a temp project that installs `@gjsify/cli` from a local tarball.

## Assertions

| Test | What is asserted |
|---|---|
| Default target | `data/app.gresource` exists and `size > 0` (`.xml` suffix stripped, blob placed next to descriptor) |
| `--target` | Explicit output path exists and is non-empty; parent dir auto-created by the command |
| `--sourcedir` | XML descriptor and resource file in separate directories; compiled `.gresource` exists and is non-empty |

Each test calls `t.skip()` when `glib-compile-resources` is not on PATH — mirrors the `gsettings` e2e pattern exactly (`hasCommand` + guard in `before` + per-test `if (!projectDir) return t.skip()`).

## Not verified locally

`glib-compile-resources` is available on this machine, so the skip path was verified by reading the source rather than running it. CI will exercise the skip path on any runner lacking the GNOME toolchain.


JumpLink:
Superseded by #323 (consolidated into one integration PR + single CI run to avoid the test:e2e/STATUS.md merge cascade). Branch kept until #323 merges.

PR #318, opened 2026-05-26
test(e2e): self-update binary-path detection + update flow

JumpLink:
## Summary

- Adds `tests/e2e/self-update/run.mjs` (5 tests, `node:test`) guarding the `gjsify self-update` command's prefix-detection guard and version-comparison short-circuits — the production-bug surface where the "no @gjsify/cli install found" warning (and associated crash path in earlier versions) was introduced.
- Registers the suite in `package.json` (`test:e2e` chain + dedicated `test:e2e:self-update` script).
- Updates `STATUS.md`: new `### Completed` entry, `self-update` added to the E2E suites table row, header refreshed.

## What it covers

| Test | Coverage |
|---|---|
| prefix populated → no warning | Regression guard: `installedAtPrefix = true` → command proceeds without the "no @gjsify/cli install found" message |
| prefix empty → warning, still exits 0 | `installedAtPrefix = false` → warning printed but "Already up to date" (same version) prevents crash |
| Same version → "Already up to date" | `currentVersion === target` short-circuit, exit 0 |
| `--check` + newer version | Exit 1 + "Update available" — the `--check` exit-code contract |
| `readCurrentVersion()` null fallback | `readFileSync` patched via `--import` preload to poison `@gjsify/cli` package.json → version shows `(unknown)`, `--check` exits 1 with "Install required" |

## Mock strategy

`fetchPackument('@gjsify/cli')` in `self-update.ts` calls `globalThis.fetch` with no registry-URL override option (unlike `gjsify upgrade` which reads `npm_config_registry`). The only available interception seam is a Node `--import` preload that patches `globalThis.fetch` before the CLI ESM graph loads. The preload is written to a temp file and routes `registry.npmjs.org` requests to an in-process `http.createServer()` mock.

## Limitation

The `installPackages` + `linkGlobalBins` path (actual binary download and replacement) is **not** exercised. That would require a full tarball server and a real install tree — out of scope for a unit-level e2e. The prefix-detection and version-comparison branches are the production-bug surface and are fully covered by this suite.


JumpLink:
Superseded by #323 (consolidated into one integration PR + single CI run to avoid the test:e2e/STATUS.md merge cascade). Branch kept until #323 merges.

PR #343, opened 2026-05-27
fix(showcase): repair #336 EnemyActor narrowing in excalibur-jelly-jumper

JumpLink:
## Summary

PR #336 over-narrowed `makeSpawner`'s `spawn` callback return type to `unknown` in `showcases/dom/excalibur-jelly-jumper/src/classes/level-scene.ts:326`, but `EnemySpawnerArgs.spawn` (`actors/enemy-spawner.ts:7`) expects `EnemyActor`. The inner wrapper `() => spawn({x,y,z}, props)` then produced `() => unknown` → tsc TS2740 with 130+ missing-properties errors on construction.

Same narrow class of fix as #340 (which closed the integration-spec side of the same #336 regression — this is the showcase-side that #340 missed because the showcase check is only exercised by CI).

## Changes

- `level-scene.ts:326` — `makeSpawner` `spawn` callback typed `=> EnemyActor` (was `=> unknown`)
- `level-scene.ts:10` — added `import type { EnemyActor } from './enemy-actor.js'`
- `STATUS.md` — leading current-state entry + Completed entry

## Verification

- `cd showcases/dom/excalibur-jelly-jumper && npm run check` → 0 errors locally
- No `any` reintroduced; no test assertions weakened

## Why this unblocks other PRs

PRs #335, #337, #339 are all rebased onto current main but their CI runs hit this showcase failure (which is on main itself). Once this merges + they rebase, they go green.


JumpLink:
Closed — fix already landed on main via PR #341 (which covered the same #336 over-narrowing class).

PR #489, opened 2026-06-10
feat(cli): GJS-first bin-shim default + in-process gjsify run dispatch

JumpLink:
Makes gjsify **GJS-first by default** (the Node-free build chain milestone), with the in-process `gjsify run` dispatch that eliminates the nested-gjs build thrash.

## What
- **In-process dispatch** (`cli-app.ts` + `commands/run.ts`): the yargs surface is factored into `runCli(argv)`; under GJS, a single `gjsify <cmd>` script (no shell ops/subs/unquoted globs) runs in-process via `await import('../cli-app.js')` instead of spawning another gjs. Collapses the `gjsify run build:gjsify → gjsify build` (and `→ gjsify tsc`) chains. Node path unchanged (shell spawn).
- **GJS-first bin-shim** (`install.ts buildBinShim`): the `.bin/gjsify` wrapper now prefers `gjs -m dist/cli.gjs.mjs` when gjs is on PATH + the bundle exists, falling back to the Node entry otherwise. So the whole workspace build runs node-free under GJS by default.

## Verified
- No thrash/hang: `foreach build -tp --jobs 4` under GJS builds the whole workspace with **0 failures** (was a multi-hour hang).
- `install --immutable` under GJS: exit 0, tar path clean (1 web-streams copy, on top of #487's fix).
- Native rolldown loads from any package cwd (the wrapper exports an absolute `GI_TYPELIB_PATH`).

## Known perf caveat (accepted, follow-up)
Over all 216 workspaces the GJS build is **slower than the Node build** (per-package native-rolldown multi-pass + `gjsify tsc` overhead). Landing per the "ship node-free default now, optimize after" decision; a perf follow-up will bring the time down. Not a correctness issue.


JumpLink:
Parking as draft pending **Step 3 (native rolldown perf)**. The flip is functionally correct (everything builds under GJS locally; packages build green in ~4 min, examples pass with `--jobs 1`), but running the *entire* CI under GJS (build + examples + every package's test-bundle build via native rolldown) exceeds the 2-hour job cap. Band-aids fix individual hangs but can't close the cumulative slowness — that needs the native-rolldown `--globals auto` multi-pass + memory optimization. Will un-draft + rebase once Step 3 lands.

JumpLink:
Un-drafting: Step 3 landed (#490 — `--globals auto` closure-map fast path, 2 analysis bundles + final instead of 4–5, −32 % per GJS build). This branch is rebased onto it; this CI run is the measurement of whether the full GJS-first pipeline now fits the 2-hour job cap.

PR #512, opened 2026-06-14
fix(cli): mirror npm-cli OIDC exchange request envelope

JumpLink:
## Summary

npm tightened its OIDC token-exchange endpoint around **2026-06-14**: a Trusted-Publishing release that succeeded on **2026-06-08** (v0.4.44, 108 packages) began returning `401 — "OIDC token exchange error - unauthorized"` with an **unchanged** GitHub OIDC JWT.

## Root cause

The GitHub OIDC JWT is identical regardless of which client requests it (same audience `npm:registry.npmjs.org` + same workflow context). So when npm-cli's exchange succeeds and ours 401s **with the same token**, the rejection is about the **request envelope**, not the token. gjsify sent a non-npm `user-agent`, no `npm-command`, and a `Content-Type: application/json` + `'{}'` body; npm-cli (`npm-registry-fetch`) sends an `npm/<ver>` user-agent + `npm-command: publish` and **no body**.

## Fix

`npm-oidc.ts` now mirrors npm-cli's exchange request exactly:
- `user-agent: npm/11.12.1 node/v24.0.0 linux x64 workspaces/false` (npm derives the `>= 11.5.1` Trusted-Publishing requirement from this prefix),
- `npm-command: publish`,
- **no request body** (drops the prior `Content-Type` + `'{}'`).

Pinned to the `refs/npm-cli` version gjsify tracks.

## Context / validation

Companion to #507 (the `%2F` URL-decode fix). Without #507 the exchange POST 404'd (decoded `/@scope/name`); with it the POST reaches npm and exposed this 401. The OIDC exchange can only be exercised in CI (needs the GitHub Actions id-token), so this is being validated via a `verify_only` audit run on this branch — a 401→✓ transition confirms the fix. PR CI validates no regression in parallel.

Unblocks OIDC Trusted-Publishing releases under GJS.


JumpLink:
Closing: mirroring npm-cli's OIDC exchange envelope (user-agent + npm-command + no body) did NOT fix the 401 — the verify_only audit still returns 0/110 'unauthorized', and this PR's own CI failed. The 401 is npm-side (correct TP config + re-saved + npm-cli-identical request all still 401), not a gjsify request bug. The %2F URL fix (#507) remains the real, necessary fix.

PR #556, opened 2026-06-21
Auto-bundle by-name plugins under GJS (exports-map workaround)

JumpLink:
## Problem

GJS's native ESM loader does **not** follow `package.json#exports` subpath maps for bare specifiers. So a `package.json#gjsify.bundler.plugins` entry whose source imports an internal subpath — e.g. `@mdx-js/rollup` → `@mdx-js/mdx/internal-create-format-aware-processors` — throws `Module not found` when the CLI runs under GJS, even though the file is physically on disk. This blocked running real bundler plugins under the Node-free GJS build chain (the concrete blocker was porting Learn6502's MDX build off esbuild).

## Fix

`resolveUserPlugins` now takes an injectable `loadModule` strategy:

- **Node** (default): plain dynamic `import()` of the resolved path — unchanged.
- **GJS** (`isGjs()`): `BuildAction.loadPluginViaGjsBundle` bundles the plugin to a **single self-contained ESM** via a nested `BuildAction.buildApp('gjs')`. Rolldown resolves the exports map at bundle time, so the emitted file has no unresolvable bare-specifier imports left; the CLI then imports that.

The nested config carries no `bundler.plugins`, so it never recurses. The bundle is cached under `node_modules/.cache/gjsify/plugins/`, invalidated when the **plugin entry OR the project lockfile** (`gjsify-lock.json`/`package-lock.json`/`yarn.lock`/`pnpm-lock.yaml`) is newer than the cached bundle — a dep bump rewrites the lockfile but not the plugin entry's mtime. `GJSIFY_NO_PLUGIN_CACHE=1` forces a rebuild (the residual window: a transitive dep edited in place).

## Also: host-detection consolidation

Centralizes the 9 byte-identical inline `typeof globalThis.imports?.gi !== 'undefined'` probes (across 5 files in 2 packages) into a shared, pure `@gjsify/rolldown-plugin-gjsify/runtime` module exporting `isGjs()` + `isNode()`.

- `isNode()` guards on `isGjs()` first, because `@gjsify/process` sets `process.versions.node` under GJS (a false Node positive).
- Exposed via a dedicated `./runtime` subpath so lightweight callers (`bundler-pick.ts`) don't pull the full plugin index under Node.
- `@gjsify/utils` was deliberately **not** chosen as the home: it's a gjs-runtime (`@girs/*`-importing) package, and these consumers are infra/build-tooling — routing through it would be a layering inversion.

## Tests

- `resolve-plugin-by-name.spec.ts` — the `loadModule` injection contract, pass-through, named export, error paths, default direct-import loader.
- `runtime.spec.ts` — `isGjs`/`isNode` against an independent host probe (mutual exclusivity + correct host).
- `tests/e2e/plugins-by-name-gjs/` — a **real exports-subpath plugin** built under `gjs -m cli.gjs.mjs`, asserting the transform ran + the bundle was cached. Skip-guarded (Linux + gjs + rolldown-native prebuild); wired into `test:e2e`.

## Validation (local)

- CLI type-check clean · 570 unit tests pass (10 new) · GJS e2e passes.
- Cache behavior verified end-to-end: fresh build → reuse → rebuild on lockfile touch → rebuild on `GJSIFY_NO_PLUGIN_CACHE=1`.
- Reviewed adversarially; confirmed: no recursion, Node path byte-identical, subpath inlined (not externalized) in the GJS bundle, library mode unaffected (never supported by-name plugins).


JumpLink:
Follow-up: proactively validated the **real `@mdx-js/rollup`-under-GJS** path (the actual driver for this feature) before merge, which surfaced a gap the synthetic e2e missed — `@mdx-js/rollup`'s factory is its **default** export, and its toolchain touches `document` at module-eval, so the plugin bundle needs `--globals auto`. But auto's entry wrapper re-exports with `export * from <entry>`, which drops `default` → the plugin imported as "no default export".

Fixed by a `preserveDefaultExport` option on the `--app gjs` entry wrapper (commit `3a4309f`): keeps globals injection **and** re-exports `default` (zero impact on normal app builds — flag defaults off). 

**Verified end-to-end**: a real `.mdx` compiled by `@mdx-js/rollup` under `gjs -m cli.gjs.mjs` builds *and executes* (default `MDXContent` + named export both resolve). Added `gjs-entry-wrapper.spec.ts` for regression coverage (577 unit tests green).

PR #615, opened 2026-06-27
ci: split build/test/e2e/examples into parallel jobs

JumpLink:
> **Draft** — the big structural change. Needs a CI validation pass (could not be runtime-validated locally) and a **branch-protection update** before merge. Stacked on top of #614 (D) → #612 (C).

## Why

The monolithic `linux` job ran install → build → check → test → e2e **serially** on one 2-core runner. The big costs couldn't overlap: **test ~14m + e2e ~18m + examples ~8m = ~40m serial**.

## What

Split into parallel jobs after a shared `build`, plus the `foreach --shard` feature from the previous commit:

- **`.github/actions/gjsify-setup`** (composite) — shared setup (node, corepack, node_modules cache, install, bundle verifications, build-cache restore). GitHub Actions has no YAML anchors, so a composite is the only way to share steps across jobs. Each job checks out first (a local composite can't resolve pre-checkout).
- **`build`** (matrix fedora) — compiles + type-checks packages (selective from #614 when warm), leaves `lib/`+`dist/` in the build-output cache.
- **`test`** (matrix fedora × **shard 1..4**) — `gjsify foreach test --shard i/4`; WebGL conformance on shard 1. `needs: build`.
- **`e2e`** (matrix fedora) — CLI end-to-end suite, parallel. `needs: build`.
- **`examples`** (matrix fedora) — builds + checks + tests example/showcase workspaces self-contained (their `dist/` isn't in the build cache). `needs: build`.
- **Concurrency group** cancels superseded PR runs (not main/schedule).

Downstream jobs restore the build job's output from the build cache (`needs: build` guarantees its post-save completed first).

## Expected

`test + e2e + examples` (~40m serial) overlap as parallel shards/jobs → wall-clock ≈ `build + max(slice)` instead of the sum. Biggest single wall-clock win of the whole effort.

## ⚠️ Breaking (CI ops) — required before merge

Job/check names change: no more `Fedora 44 (GJS 1.88 / SM 140)`. New checks: `Build Fedora 44 …`, `Test 1/4 Fedora 44`, `E2E Fedora 44 …`, `Examples Fedora 44 …`. **Branch protection's required status checks must be updated** or merges will block on a check that no longer runs.

## Validation status

- `foreach --shard` feature: unit-tested + e2e (`tests/e2e/foreach-command`), validated locally (234 workspaces → 59/59/58/58, disjoint+complete).
- Workflow: `actionlint` clean, YAML valid. The multi-job + composite + cross-job cache wiring is **not** locally runtime-testable — first CI run is the real validation. Recommend validating C → D first (this rebases onto them).

Part 5 (final) of the CI-speedup effort.


JumpLink:
**Paused** (per decision: ship C+D first, defer the job-split).

### What's validated
The hard parts of the split are **green**: `Build`, `Test 1/4–4/4` (sharding + cross-job build-cache passing), and the `foreach --shard` core feature (unit + e2e). The composite action + `needs: build` cache hand-off work.

### Why it's paused — the remaining blocker
`E2E` and `Examples` fail from **one root cause**: `build:examples` (app-mode, bundles everything) only succeeds when it runs **right after a full `gjsify run build` in the same job**. Proven:
- ✅ builds fine locally (real node_modules)
- ✅ green in PR #612's warm re-run (full build + build:examples in one monolith job, against a *restored* node_modules cache)
- ❌ fails in this PR's `examples` job, which builds examples off the build-cache-only state (no full build in-job) — an example `--app browser` build can't resolve a workspace dep (`@gjsify/adwaita-fonts` from `adwaita-web/src/index.ts`).

`test:e2e` runs `tests/lint-engines.mjs`, which asserts every `@gjsify/example-*` has built `dist/` → it fails too once examples aren't built.

### Fix plan (for when resumed)
- Move `build:examples` (+`check:examples`) into the **build** job, after the full Build (the proven-working condition).
- Upload `showcases/*/*/dist` + `examples/*/*/dist` as an **artifact**; download it in the `e2e` + `examples` jobs (avoids the build-cache staleness on example-only changes).
- Force a **full** build whenever examples are needed (gate build:examples + full-build on `run-e2e || example-closure || global || empty`).
- Optional follow-up: shard `test:e2e` too (it's the ~18m long pole).

Rebase needed before resuming (stacked on D #614 → C #612, both moving).

PR #685, opened 2026-07-02
feat(cli): content-hash build cache for workspace/foreach

JumpLink:
Implements **ADR 0006 phase 1** — the CLI-owned, content-hash per-package build cache targeting the last big CI cost block (the ~27 min full Rolldown re-bundle of every package).

## What

`gjsify workspace <name> <script> --cached` and `gjsify foreach <script> --cached` skip re-running a build-ish script for workspaces whose inputs are unchanged and restore the previously stored outputs instead. `GJSIFY_BUILD_CACHE=1` is the env default (so a whole root script chain like `gjsify run build` opts in without editing every nested invocation); an explicit `--no-cached` wins. Script mode only — an explicit `--cached` with `--exec` is rejected; the env default silently ignores `--exec` invocations.

New module: `packages/infra/cli/src/utils/build-cache.ts` (pure, unit-testable key computation + store/restore/prune + a `BuildCacheRunner` orchestrator), thin wiring in `commands/foreach.ts` / `commands/workspace.ts`.

## Cache key (per workspace package)

sha256 (32 hex) over, in order:

- layout version (`v1` — bump to invalidate everything),
- script name + forwarded args,
- **toolchain salt**: workspace-resolved versions of `@gjsify/cli`, `@gjsify/tsc`, `rolldown`, `typescript` (read from `<root>/node_modules/<pkg>/package.json`; `@gjsify/cli` falls back to the running CLI's own version) — a CLI/bundler/compiler bump invalidates every entry,
- the package's **own input hash**: every file under `src/**` + `package.json` + root `tsconfig*.json`, streamed sha256 of real file contents, sorted paths, `node_modules` excluded, mtime-independent,
- the input hashes of the package's **full transitive workspace-dep closure** (prod + dev + optional + peer `workspace:` edges, via `@gjsify/workspace`'s `buildDependencyGraph`). A dep edit invalidates every transitive dependent.

Design notes:
- Composing the closure's *own hashes* (instead of recursing into dep *keys*) yields identical invalidation semantics while staying well-defined on the cyclic dev-dependency graph this monorepo actually has.
- All relevant input hashes are computed **eagerly at run start**, so a build that mutates its own `src/` mid-run (see adwaita-web below) cannot make a sibling's key depend on scheduling order.

## Storage

`node_modules/.cache/gjsify/build/<pkg-name-sanitized>/<key>/` — `manifest.json` (package, script, key, recorded dirs, content hash of outputs, created/used stamps) + `outputs/<dir>/…` copies. Stores are atomic (tmp dir + rename). Growth is bounded: at most **2 keys per package**, oldest by store/use evicted; stray tmp dirs are reaped.

## Output contract + safety invariants

- After a **miss**, the candidate output dirs (`lib/`, `dist/`, `dist-templates/`) are diffed against a pre-run snapshot and **only the dirs the script actually modified** are stored.
- A **hit** replaces exactly the recorded dirs (delete-then-copy within the package) — and skips the copy entirely when the on-disk content already matches (fast warm path).
- **The v0.x clobber incident cannot recur**: a package that doesn't define the script is filtered out before the cache ever sees it, and a dir the build never touches (a committed no-build `lib/` — `@gjsify/tsc`, `@gjsify/resolve-npm`) is never *recorded*, so it can never be *restored over*.
- **Per-package opt-out**: `package.json#gjsify.buildCache: false` — set on `@gjsify/adwaita-web`, whose build generates `src/styles.generated.ts` + per-file JS inside `src/` that consumers resolve from source; a cached skip would leave those missing, so it always runs uncached.
- Every cache error (unreadable entry, EACCES, full disk) fails soft to an uncached run.

## CI wiring (`main.yml` + `gjsify-setup`)

- The build job's **Build** and **Build examples** steps set `GJSIFY_BUILD_CACHE=1` (both the full `gjsify run build` chain and the selective `gjsify foreach build --include …` path inherit it through the process env).
- `actions/cache/restore` of `node_modules/.cache/gjsify/build` before the build (key `pkg-build-cache-v1-fedora<v>-<run_id>-<run_attempt>`, prefix restore-keys fallback → rolling cache) + a `chown` hand-off to `testuser` (a root-owned restore would silently disable stores).
- An **explicit `actions/cache/save` with `if: always()`** right after the build steps — the implicit post-save skips failed jobs, and red runs profit most from warm entries on retry.
- The node_modules cache now excludes `!node_modules/.cache/gjsify/build` so the two caches don't double-store.
- The existing whole-tree lib/dist cache is unchanged (check/test/e2e jobs still consume it).

## GJS gotcha found & fixed during development

The hash-record template literals initially contained raw NUL bytes as field separators. Node/V8 parses a raw NUL inside a template literal fine, but **GJS truncates module source at NUL** — the rebuilt `cli.gjs.mjs` failed to parse under SpiderMonkey with ``SyntaxError: `` literal not terminated before end of script``. Separators are plain spaces now; the committed bundle is verified working (`gjs -m dist/cli.gjs.mjs --version` → 0.14.0, `--cached` exercised under GJS).

## Evidence

- **Unit** (`build-cache.spec.ts`, registered in `test.mts`): key determinism + dep-order independence; sensitivity to script/args/toolchain/own-hash/dep-hash; input hashing ignores `lib`/`dist`/`node_modules` but reacts to `src/**`/`tsconfig*`/`package.json` content; transitive dependent invalidation (incl. devDep edge, unrelated package unaffected); modified-dir detection never captures a committed untouched `lib/`; store→delete→restore round-trip; prune-to-N; runner hit/miss integration. `gjsify workspace @gjsify/cli test`: **672 completed, 0 failed**.
- **e2e** (`tests/e2e/build-cache/run.mjs`, wired into `test:e2e` + `test:e2e:build-cache`): 9/9 pass — (1) cold cache builds all; (2) second run hits for all, markers byte-identical; (2b) hit restores deleted outputs; (3) editing A's src re-runs A **and** dependent B but not C; (4) a package without the script stays untouched (no cache dir, committed lib intact); (5) ≤2 keys per package; (6) env toggle + `--no-cached` override; (7) `gjsify workspace --cached` hits the foreach-stored entry; (8) explicit `--cached --exec` rejected.
- **Neighbors**: `foreach-command`, `workspace-with-dependencies`, `run-command` e2e — 29/29 pass locally.
- **Measured** (3 real packages — utils/events/assert, local): uncached **10.5s**, cached-cold **11.1s** (+6% store overhead), cached-warm **0.4s**.
- `gjsify workspace @gjsify/cli build -d && check` green; changed files oxlint-clean + oxfmt-formatted.

## Docs

- AGENTS.md: `--cached` subsection in Build (key ingredients, storage path, output contract, safety rule, opt-out, CI wiring).
- `website/src/content/docs/cli-reference.md`: `--cached` rows for foreach/workspace + a "Build cache" section.
- STATUS.md: ADR 0006 phase 1 struck through (phase 2 source-direct spike stays open); ADR status line updated.

## Out of scope / follow-ups

- ADR 0006 **phase 2** (publish-time-only lib builds spike) — untouched, stays in the backlog.
- The first CI run on this PR is a cold cache (expect no speedup); the measurable win lands from the second run / after merge once main-scope caches exist.


JumpLink:
⚠ **Update — CI wiring reverted, feature kept.** The first CI run failed: enabling `GJSIFY_BUILD_CACHE=1` on the `main.yml` `Build examples` step blew it past the 45-min timeout on a **cold** cache. Root cause: the serial (`--jobs 1`) examples build with per-package transitive-closure hashing recomputed from scratch per package is O(pkgs × closure-files) — pathological at monorepo scale; examples have deep `@gjsify/*` closures.

Per ADR 0006 ("acceptable to leave main.yml unchanged and note it as the remaining sub-item"), this PR now **reverts all `main.yml` build-cache wiring** and ships only the CLI feature (`--cached` / `GJSIFY_BUILD_CACHE=1`, opt-in, off by default; unit + `tests/e2e/build-cache` green). CI is unchanged.

**Follow-up before CI re-enablement:** memoize input hashes across a single `foreach` run (compute each package's own-input hash once, reuse when composing dependents' keys) so a cold-cache full build is O(total-files), not O(pkgs × closure-files). Tracked in STATUS.md + the ADR.

PR #721, opened 2026-07-14
refactor(node-gi): split addon.cc into translation units

JumpLink:
## Summary

Purely mechanical split of `packages/node-gi/node-gi/src/addon.cc` (5601 lines, one translation unit) into 11 cohesive TUs + a shared `common.h` — **zero behavior change**. This lands FIRST so the upcoming review-gated fixes (toggle-ref/multi-env teardown crash, vfunc OUT/INOUT chain-up — ADR 0005 graduation gate) have reviewable, contained diffs instead of edits buried in a 5.6k-line file.

| File | Lines | Contents |
|---|---|---|
| `common.h` | 301 | provenance header, includes, `namespace nodegi`, shared structs, cross-TU decls, extern globals |
| `repo.cc` | 295 | repository/introspection entry points + per-env instance data |
| `marshal.cc` | 996 | GIArgument↔JS, boxed/GType handles, container marshalling |
| `toggle.cc` | 571 | toggle-ref GC bridge + test-only stress block |
| `calls.cc` | 913 | ffi-closure callbacks, GError→throw, `InvokeFunctionInfo`, call* entry points |
| `object.cc` | 500 | GValue↔JS, construct/newObject, property/type entry points |
| `class.cc` | 879 | registerClass machinery incl. vfunc trampoline + chain-up |
| `template.cc` | 108 | Gtk composite-template API (dlsym), MaybeInitTemplate, getTemplateChild |
| `variant.cc` | 579 | GVariant build/unpack |
| `signals.cc` | 416 | JsClosure, connect/emit/disconnect, template-callback scope |
| `loop.cc` | 161 | uv↔GLib mainloop bridge |
| `addon.cc` | 97 | env-shutdown hook + `Init` + `NODE_API_MODULE` only |

## Mechanics

- The single anonymous namespace became `namespace nodegi`; symbols used across TUs dropped `static` and are declared in `common.h` (all 40 N-API entry functions + ~30 helpers). Everything else stays TU-local, incl. all C/ffi trampolines.
- Cross-TU globals became `extern` (`g_owner_env`, `g_toggle_shutdown`, `g_queue_mutex`, `g_drain_tsfn`, `g_drain_async_inited`, `g_syncEmitDepth`; handle type tags as `extern const`).
- `Init` stays at global scope (`NODE_API_MODULE` token-pastes its regfunc argument).
- A line-level diff verified every original line survives verbatim apart from the ~50 linkage/signature edit lines.

## Test plan

- [x] `npm test` (Node): 261 tests, 0 fail — identical to pre-split baseline (incl. the 11 `--expose-gc`-gated skips)
- [x] `npm run test:gc`: 261/261
- [x] `npm run test:bun`: 26/26 conformance files green (Bun 1.3.13)
- [x] `npm run test:deno`: 26/26 conformance files green (Deno 2.9.1)
- [x] Build: zero warnings; `binding.gyp` lists all 11 sources, flags unchanged


JumpLink:
**Verification addendum:** the original local test runs (baseline + post-split) were unknowingly executed against a stale, untracked `prebuilds/linux-x64/node_gi.node` (Jul 2) — `nativeCandidates()` prefers the prebuild over `build/Release`, so both runs loaded the same old binary and were trivially identical. After removing the stale prebuild, the split was re-verified against the actual freshly built split addon:

- `npm test` (Node): 261 tests, 0 fail (250 pass + 11 `--expose-gc`-gated skips)
- `npm run test:gc`: **261/261**
- `npm run test:bun`: 28/28 conformance files (incl. the two files added in #722)
- `npm run test:deno`: 28/28

CI was never affected (the prebuild is not committed; the cross-runtime job stages it fresh from the job's own build), so the green CI on this PR is independently authoritative. The local-dev footgun gets its own small fix PR (`NODE_GI_NATIVE` override + test scripts pinning to the fresh build).

PR #822, opened 2026-07-27
Make a green PR predict a green main

JumpLink:
Makes "this PR is green" predict "main will stay green".

## The incident this is specified from

`Deploy Documentation` ran on `push: main` only and appeared in **no PR check**. #812 merged fully green; the docs site broke and stayed broken on main for **eight hours** (13:08 → 22:25 UTC on 2026-07-26). Nobody was careless — the signal was structurally incomplete.

It is still broken. 7fe6225e3 fixed the root cause (`build:deps` now builds adwaita-web *with* its workspace deps) and the very next run failed differently:

```
before 7fe6225e3:  [commonjs--resolver] Failed to resolve entry for package "@gjsify/adwaita-core"
at     7fe6225e3:  gjsify build: no usable bundler engine under GJS —
                   `@gjsify/rolldown-native` is not loadable
```

`build:deps` now runs `gjsify build --library`, and under GJS the only bundler engine is `@gjsify/rolldown-native`, whose JS facade `lib/` is a build output `gjsify install` does not produce. Nothing in that job produced it. The same gap hit `release.yml` the same night (748e722e9, `@gjsify/napi` missing from the 0.23.0 train). Both were found post-merge, for the same reason.

## Inventory — every workflow, per job, on `pull_request`?

| Workflow | PR? | Notes |
|---|---|---|
| `main.yml` | yes | all 10 jobs; **PR-gated legs below** |
| `audit-runtimes.yml` | yes | no paths filter — full parity |
| `commitlint.yml` | yes | PR-only by nature |
| `node-gi.yml` | yes | identical `paths` on push + PR; all 17 jobs incl. macOS/Windows |
| `napi.yml` | yes | identical `paths`; `windows` is `workflow_dispatch` only — documented, blocked at gjs-on-Windows |
| `deploy-docs.yml` | **no** | ← the incident. **Fixed here** (build half) |
| `build-ci-image.yml` | **no** | ← same class, worse: `main.yml` keeps using the last good tag, so a broken Dockerfile is invisible *even after* merge. **Fixed here** |
| `prebuilds.yml` | **no** | Vala/meson/Rust natives first compiled post-merge. **Left open** — see below |
| `cli-cross-platform.yml` | effectively no | `paths` = only its own file; weekly + dispatch. Tests a *published* CLI version, every step `continue-on-error` — cannot gate a PR by construction |
| `release.yml` | no | `release: published` — post-merge by definition |

Gated *within* `main.yml` on a PR (all deliberate, all kept):

| Leg | Gate |
|---|---|
| `macos` | `ci:macos` label absent → skipped (10x billing) |
| `build`/`check`/`test`/`cross-runtime`/`e2e`/`examples`/`browser` | affected classifier (`skip-all` / `global` / `include-args` / `run-e2e`) |
| `Build/Check examples`, WebGL conformance, macOS + browser bundle staging | step-level `if:` on the same outputs |
| integration suites | wired into **no** workflow, PR or main |

The AGENTS.md "matrix reduction: ONE Fedora leg on PR, FULL matrix (43 + 44) on push" no longer exists — the 43 leg was dropped, so Fedora is one leg on every event. Corrected here.

## What changed

**(a) Unintentional — `deploy-docs.yml` split.** Build runs on `pull_request`; deploy stays main-only. `pages: write` + `id-token: write` move from the workflow level onto the deploy job, so the build can never come to need a token a fork PR lacks. On a PR the Pages artifact upload is skipped and concurrency is per-ref instead of the shared `pages` group.

Two things the split exposed, both fixed:

- `paths` named `packages/web/adwaita-web/**` — a second copy of the dependency graph, which is why it could not see `@gjsify/adwaita-core`. Now coarse and directory-level (`website/**`, `packages/**`, `showcases/**`, root manifests). Over-triggers by design; cannot silently miss.
- The job hand-rolled a subset of the shared setup, so it shared no cache with anything (**cold ~17 min install on every run**) and drifted from the environment the build needs. Now uses `.github/actions/gjsify-setup`, plus `gjsify run build:infra` — the same bootstrap the root build and (since 748e722e9) release.yml use. **This is also what un-breaks main.**

**(b) Cost controls — kept, and made visible.** New `ci-summary` job, `if: always()`, writes to `$GITHUB_STEP_SUMMARY`:

```
| Job                           | Result    |
| Unit tests (GJS, 4 shards)    | success   |
| Cross-runtime (node+bun+deno) | skipped   |
| macOS GJS suite               | skipped   |
...
### Why
- classifier: `skip-all=false` `global=false` `run-e2e=false`
- classifier reason: `2 workspace(s) affected`
- affected closure: `--include @gjsify/fetch --include @gjsify/http`
- macOS leg: `enabled=false` — pull_request without the `ci:macos` label …
- integration suites are not wired into any workflow yet

### Workflows with no run for this commit
| Build Native Prebuilds | `.github/workflows/prebuilds.yml` |
| Release to NPM         | `.github/workflows/release.yml`   |
```

Everything is **derived**, nothing restated:

- rows are `needs.<job>.result` — GitHub's own evaluation of each `if:`. Change a gate, the table follows.
- the "why" values are the gating jobs' own outputs, printed verbatim. `setup` gains a `macos-reason` emitted by the same branch that sets `macos-enabled`, so gate and explanation cannot desync.
- the cross-workflow table is the repo's active workflows minus the workflows GitHub created a run for at this commit (`gh api …/actions/workflows` vs `…/actions/runs?head_sha=`). A new main-only workflow appears the day it is added.

There is deliberately **no hand-maintained "what we skip" list** — that is the artefact that drifts and then lies.

## Cost

- `main.yml`: one extra `ubuntu-latest` job at the end. No test job changed.
- `build-ci-image.yml`: fires only when the Dockerfile changes.
- `deploy-docs.yml`: the one real addition. Runs on PRs touching `packages/**`/`website/**`/`showcases/**` — but going through `gjsify-setup` turns its cold ~17 min install into a warm-cache no-op, so it is now *cheaper* than it was on main. Runs in parallel with `main.yml`'s 60–90 min; on a small selective PR it may become the long pole.
- Full Fedora matrix on PRs: **not** added. macOS label gate, classifier gates, `napi.yml`'s Windows dispatch gate: all untouched.

## Fork PRs

The docs build needs no secret: `packages: read` for the container, no Pages token, no artifact upload. The CI-image PR leg neither logs in nor pushes. `ci-summary` degrades to a note if `actions: read` is unavailable.

Independent pre-existing gap, now in STATUS.md: every container job pulls the **private** `ghcr.io/gjsify/ci-fedora`, so fork PRs get no working run at all. Publishing the image (a Fedora + GNOME devel stack, nothing secret) would close it.

## Gaps left open, with reasons

- **`prebuilds.yml`** — a PR leg means QEMU-emulated aarch64/ppc64/s390x/riscv64 builds plus Rust builds of rolldown/oxc/lightningcss out of `refs/`. Violates the fast-PR constraint. STATUS.md records the cheap 80%: a `pull_request` leg on the **native x86_64** runner for the **Vala-only** packages, leaving QEMU + Rust on main.
- **`cli-cross-platform.yml`** — tests a *published* CLI version with `continue-on-error` everywhere. Cannot gate a PR's code; a PR leg would be theatre.
- **`release.yml`** — post-merge by definition. `workflow_dispatch -f verify_only=true` already audits Trusted Publishers without publishing.
- **integration suites** — still on no event. Pre-existing, tracked.

## Verification

- All three edited workflows parse (`python3 -c "import yaml; yaml.safe_load(...)"`).
- Both summary step bodies dry-run locally against representative and empty/unset env values.
- Both `gh api` queries verified against this repo: for `77d5ca54b` they correctly report only `main.yml` + `audit-runtimes.yml` ran.
- `node scripts/audit-runtimes.mjs --check --strict` and `scripts/audit-test-scripts.mjs` still exit 0.

**Not verifiable without a CI run** — and this PR is the run, since it touches `main.yml`, `deploy-docs.yml` and `build-ci-image.yml` and therefore triggers all three:

1. that `gjsify run build:infra` + `gjsify-setup` actually turns the docs build green (the diagnosis is from the failure logs and 748e722e9's identical fix, not from a local repro — a local repro needs a ~2 GB install);
2. real timings for the docs job through the shared caches;
3. `needs.<job>.result` values and the step-summary rendering under real matrix jobs;
4. that the `pull_request` `head_sha` union covers how GitHub records runs for this event.

Please do not merge until the `Build Documentation` check on this PR is green.


JumpLink:
### Verified on this PR

The PR triggers all three workflows it changes, so it checks itself.

| Check | Result |
|---|---|
| **Build Documentation** (new PR leg) | **pass, 3m23s** |
| Deploy to GitHub Pages | correctly `skipping` on a PR |
| **Build & push Fedora 43 / 44 image** (new PR leg) | pass, `push: false` |
| Detect runtime-triplet drift | pass |
| Lint commit messages | pass |

The docs step that has been red on main since #812 — `gjsify build: no usable bundler engine under GJS` — is **green here**. So this also un-breaks main on merge.

Step timings for the docs job:

```
Initialize containers                          42s
Checkout                                        8s
gjsify-setup (caches + install + verifies)     49s   ← was a cold ~17 min install
gjsify run build:infra                         73s   ← mtime-skipped on a warm cache
Build documentation site                       ~60s
Upload artifact                            skipped   (PR)
```

**3m23s total, against ~19 min per run before.** Moving the job onto the shared `gjsify-setup` made the new PR leg cheaper than the old main-only one.

### One thing the first run found

The first attempt died before running anything:

```
su: cannot open session: Permission denied
```

`gjsify-setup` raises testuser's `nofile` via `/etc/security/limits.conf`, and `pam_limits` refuses to open a session asking for more than the **container's** hard limit. Every `main.yml` job that uses the composite passes `--ulimit nofile=1048576:1048576`; the docs job did not, because that was never part of what the composite documented. Fixed in 247e4b5e7, which also states the requirement in the composite's own description — an undocumented precondition on the shared setup is the same drift as everything else in this branch, it just failed loudly this time.

### Still to observe

`ci-summary` runs last in the (full, `global=true`) `GJS` run on this PR — its rendering under real matrix jobs is the one piece not yet seen. Worth a look at the run summary before merging.

PR #823, opened 2026-07-27
Check the headless package contract in CI; stage a suite's test/ assets

JumpLink:
Two invariants that were missing, not two features.

## 1. `gjsify.headless` — the intra-GJS layering contract (ADR 0015)

`@gjsify/canvas2d-core` documents itself as headless with **"NO GTK dependency in
the ROOT entry"** — the property that makes the split out of `@gjsify/canvas2d`
worth its cost. It imported `gi://Gdk` at five call sites, and in GTK4 GDK lives
inside `libgtk-4.so`, so the "headless" core dlopened the entire GTK stack on
import. Fixed in d88b38c5d; nothing prevented it recurring.

**No existing check could have caught it**, and both blind spots were verified
against the pre-fix tree rather than reasoned about:

- The **drift check** feeds the `gi_url` signal into `suggestRuntimes()` as an
  **input**, so the Gdk import made the declaration agree *better*, not worse. It
  produced agreement, not drift — declaration and code were consistent, and both
  were wrong about the same thing.
- The **ADR-0014 reachability pass** is gated on `polyfill`/`partial` slots over
  `browser`/`nativescript`/`node`. canvas2d-core declares `node:"none"`,
  `browser:"native"`, `nativescript:"none"` — every slot was skipped, so `src/**`
  was never examined.

The gap is categorical. Every audit in the script models **cross-runtime** reach.
What canvas2d-core promises is **intra-GJS layering** — headless vs
toolkit-bound — which had no declaration and no check anywhere. The contract
lived only in prose, and prose does not fail a build.

### The declaration mechanism

`package.json#gjsify.headless`, **declared, never inferred**. A `*-core` name
heuristic is guessable but wrong by construction: `adwaita-core` and
`storybook-core` are headless for different reasons, `webrtc-native` is a
`-native` package that is the most toolkit-bound thing in the tree, and a rename
would silently change meaning. Two spellings, because there are two genuinely
different promises:

| Spelling | Promise | Declared on |
|---|---|---|
| `"headless": true` | the root entry reaches **no typelib at all** (no `gi://`, no `@girs/*` value import, no bare `cairo`/`system`/`gettext`, no `imports.*`) | `@gjsify/adwaita-core` — "pure TS, NO platform imports" |
| `"headless": ["Gdk","GdkPixbuf","Gsk","Gtk","Adw"]` | the root entry reaches **none of these** | `@gjsify/canvas2d-core` — headless *of GTK*, legitimately binds Cairo + PangoCairo |

Boolean-only would force canvas2d-core — the package the invariant exists for —
to lie or opt out. List-only would force adwaita-core to enumerate an open-world
set, and the namespace nobody lists is exactly the one that leaks. The field
carries the **shape** of the promise; the check reads it literally.

Only these two packages declare it: they are the two whose AGENTS.md row *and*
`package.json#description` make the headless claim in those words.
`storybook-core` / `stories` / `devtools-protocol` describe themselves as
*renderer-* / *runtime-* / *transport-agnostic* — a cross-runtime claim ADR 0014
already enforces through their all-`polyfill` slots, so adding a redundant field
there would only dilute what `headless` means.

### The check

Walks the ROOT import graph from `exports["."]`, across relative imports **and**
`@gjsify/*` workspace edges (a headless root that reaches GTK through a sibling
is no less GTK-bound). **Root-entry-only is the point**, and it is the subtlety
that hid the original bug: a side-effect *subpath* may legitimately reach the
forbidden typelibs — the fix put the GDK code behind `@gjsify/canvas2d-core/gdk`,
imported explicitly by `@gjsify/dom-elements/register/canvas` and
`@gjsify/canvas2d` — so scanning `src/**` would turn the **fixed** tree red.

Runs in every `--check`, not behind `--strict`: declaration-driven, ~0.9 s for
the whole workspace, and a guard CI does not execute is not a guard. An
unresolvable root entry and a malformed declaration are both hard failures — a
check that silently passes when it cannot find what it is meant to inspect is
decorative.

### Proof it fires

Restoring the three pre-fix sources
(`git checkout d88b38c5d^ -- canvas-pattern.ts context/drawing.ts context/pixels.ts`):

```
$ node scripts/audit-runtimes.mjs --check   # exit 1
HEADLESS-CONTRACT FAILURES (ADR 0015) on 4 package(s):
  - @gjsify/canvas2d-core: gjsify.headless=[Gdk, GdkPixbuf, Gsk, Gtk, Adw] but the ROOT entry
    graph reaches gi://Gdk — via src/index.ts → src/canvas-pattern.ts (headless-contract-violated).
  - … reaches gi://Gdk — via src/index.ts → src/canvas-rendering-context-2d.ts
    → src/context/index.ts → src/context/drawing.ts
  - … reaches gi://Gdk — via … → src/context/pixels.ts
  - … reaches gi://GdkPixbuf — via … → src/context/pixels.ts
```

One finding per `gi://` **value** import, each with its chain from
`src/index.ts`; the two `import type GdkPixbuf` sites are correctly ignored. On
that same pre-fix tree **the drift, tier, platform and reachability audits all
still report OK** — direct confirmation that the two structural blind spots are
real and that this is what closes them. Reverted afterwards; the clean tree is
green.

Also proven per spelling: `headless: true` fires on a `gi://GLib` import; a
cross-package workspace edge into `@gjsify/utils` is followed and reported with
a repo-relative path; `headless: []` fails as `headless-declaration-invalid`.

## 2. Harness gap 7 — a suite's `test/` assets never reached the bundle

`stageFixtures` bridged `dist/fixtures → ../fixtures` and nothing else.
`@gjsify/fs`'s `sync.spec.ts` writes its watch scratch file to
`join(__dirname, 'test/watch.js')`, which under the harness resolves to
`<pkg>/dist/test/` — a directory the harness itself failed to create. The suite
was recorded in the committed survey as a hard node-gi failure for that reason
alone: a false negative in the report that prioritises engine work.

`stageTestAssets` now bridges every on-disk asset dir
(`STAGED_ASSET_DIRS = ['fixtures', 'test']`) rather than one hard-coded name, as
symlinks rather than copies — these suites *write* into the dirs, so a copy would
diverge from the tree the package's own `test:gjs` run uses.

**Re-measured** (`node scripts/node-gi-consumer-harness.mjs @gjsify/fs --runtimes node`),
Fedora 44 · node 24.15.0 · node-gi built from this tree:

| | before | after |
|---|---|---|
| status | `fail` | **`pass`** |
| counts | — | **657/657** |
| reason | `other` — `Failed to create file "…/packages/node/fs/dist/test/watch.js.UIV0S3": No such file or directory` | — |

So `@gjsify/fs` runs its entire GJS suite unchanged on Node through the reverse
bridge. Its bun/deno `ran-no-summary` **survived** the fix, which settles what
gap 1 had listed as a guess: it cannot be about missing assets, and is the same
process-lifetime shape as `node-globals` and `worker_threads`. Survey (md + json)
and STATUS.md updated to the measurement — node passes 25 → 26.

## Also

`ci(node-gi)`: the `consumer-suites` setup comment still said canvas2d-core is
out of the proof set *"until the Gdk dependency is removed from the headless
core"*. It was removed, and is now machine-checked — the comment described a
state that no longer exists, right where a reader goes to decide whether the
package can be promoted. Corrected to record what actually blocks promotion (the
gtk4 typelib's graphene/vulkan-loader cost in a 20-minute job): a cost decision,
not a contract violation.

## Verification

- `node scripts/audit-runtimes.mjs --check` and `--check --strict` exit 0 on the clean tree; `--json` / `--markdown` / `--platforms` / table modes unaffected.
- The new invariant fails on the pre-fix state, and on each spelling (above).
- `node --test tests/e2e/node-gi-consumer-harness/run.mjs` — 13/13. The three new staging cases fail when `STAGED_ASSET_DIRS` is reverted to `['fixtures']`.
- `gjsify lint` on the changed scripts reports only findings that reproduce on `main`; `oxfmt --check` is clean on the e2e test and makes no change inside the new audit section.

---


JumpLink:
Superseded by #826 — identical commits on a fresh branch. GitHub stopped creating workflow runs for this branch (zero check-runs for the last two commits; close/reopen and an empty commit both produced none, while other branches triggered normally), so it could not be verified at its head.

PR #832, opened 2026-07-27
fix(prebuilds): stage + load-test the macOS sibling cdylib

JumpLink:
Repo task #61 — *on macOS the Vala bridge cannot load the Rust cdylib next to it*.

## What the error actually was

From the experimental darwin-arm64 leg (run 30266508948):

```
Library not loaded: @rpath/libgjsify_rolldown.dylib
  Referenced from: …/prebuilds/darwin-arm64/libgjsifyrolldown.dylib
  Reason: tried: '…/prebuilds/darwin-arm64/libgjsify_rolldown.dylib' (no such file),
          '/opt/homebrew/Cellar/glib/2.88.2/lib/…' (no such file), … (many Homebrew paths)
```

This reads like a missing rpath and is not one. dyld's **first** candidate is the
prebuild directory itself — that IS `@loader_path` expanding correctly. The Homebrew
paths after it are the other `LC_RPATH` entries meson bakes for the glib dependency,
tried only because the first candidate did not exist.

The file was genuinely absent. The two libraries have different leaf names on purpose
(`libgjsifyrolldown.dylib` = the Vala library the typelib names; `libgjsify_rolldown.dylib`
= the cargo cdylib, underscore), and the experimental job's `cp` listed only the first.
Nothing else in the pipeline mentions the second, so nothing noticed.

Verified structurally on the committed artifacts, from Linux — no naming mismatch, no
missing rpath:

| artifact | records | search paths |
|---|---|---|
| `oxfmt-native/prebuilds/darwin-arm64/libgjsifyoxfmt.dylib` | `@rpath/libgjsify_oxfmt.dylib` | …glib…, …gettext…, **`@loader_path`** |
| `lightningcss-native/prebuilds/darwin-arm64/libgjsifylightningcss.dylib` | `@rpath/libgjsify_lightningcss.dylib` | …glib…, …gettext…, **`@loader_path`** |
| `rolldown-native/prebuilds/linux-x64/libgjsifyrolldown.so` | `libgjsify_rolldown.so` | **`$ORIGIN`** |

So the link side (`install_name_tool -id @rpath/<leaf>` on the cdylib + `build_rpath`/
`install_rpath` `@loader_path`, from #812) is already right on all three bridges and is
unchanged here. `@loader_path` over a bare added rpath is the correct tool and stays:
`DYLD_LIBRARY_PATH` from `buildNativeEnv()` only has to find the *bare leaf the typelib
records* (dyld has no notion of "next to the typelib"); the Vala→Rust hop must need
nothing at all, or the prebuild works under exactly one launcher.

## Changes

1. **Stage the cdylib** (`+ .gir`) in the experimental rolldown leg — the actual bug.
2. **`scripts/check-prebuild-loader-path.mjs`** — for every `libgjsify*` dependency a
   shipped library records: the file must be present in the same directory AND the
   recording library must carry `@loader_path`/`$ORIGIN`. Parses Mach-O *and* ELF
   directly (no `otool`/`readelf`), so a Linux host verifies a committed darwin artifact
   it cannot run. Wired into `stage-prebuild.mjs` (staging and verifying are one step)
   and run explicitly in both macOS jobs, which hand-copy.
3. **The half that matters — the required job now loads what it builds.**
   `build-prebuilds-macos` built six darwin-arm64 bridges, copied them and uploaded them
   for `commit-prebuilds` to commit to main *without ever loading one*; the only job that
   loaded anything was the manual-dispatch experimental one, which is precisely why this
   hid. It now ends in a real `gjs -c 'imports.gi.<Ns>.<Class>'` per bridge (resolving a
   GObject class forces GI to call its `…_get_type` symbol → dlopen; a typelib that is
   merely *found* passes the namespace lookup and throws there), plus an env-free
   `ctypes.CDLL(<abs>)` under `/usr/bin/python3` with `DYLD_*` unset — which proves the
   sibling hop rides `@loader_path` alone. Both fail the job.

No trigger or job-graph changes.

## Verification boundary

Locally (Linux, gjs 1.88) — the analogous ELF case, all real:
- all seven namespaces resolve from the committed `linux-x64` prebuilds under `gjs`;
- all three Vala `.so`s `dlopen` by absolute path with `LD_LIBRARY_PATH` unset (`$ORIGIN`);
- the checker reproduces the rolldown failure when the sibling is withheld, and passes on
  every committed `darwin-arm64` + `linux-*` directory.

macOS itself is CI-only: `prebuilds.yml` dispatched on this branch, which runs both the
required and the experimental darwin legs. Log evidence in a comment below.


JumpLink:
## CI evidence — run [30271998319](https://github.com/gjsify/gjsify/actions/runs/30271998319), both darwin legs green

### Required job — `Build prebuilds (darwin-arm64)` ✅

Structural check on the real artifacts (`@loader_path` present, both siblings staged):

```
[check-prebuild-loader-path] packages/infra/oxfmt-native/prebuilds/darwin-arm64
  libgjsify_oxfmt.dylib [macho] needs (no sibling) | search: (none)
  libgjsifyoxfmt.dylib [macho] needs @rpath/libgjsify_oxfmt.dylib | search: /opt/homebrew/Cellar/glib/2.88.2/lib, /opt/homebrew/opt/gettext/lib, @loader_path
[check-prebuild-loader-path] OK — 6 directories sound
```

**The load test — every bridge the job builds now actually loads:**

```
GjsifyLightningcss.Engine resolved — shared library loaded
GjsifyOxfmt.Formatter resolved — shared library loaded
GjsifyTerminal.Terminal resolved — shared library loaded
GjsifyTls.Tls resolved — shared library loaded
GjsifyHttpSoupBridge.Server resolved — shared library loaded
GjsifyHttp2.SessionBridge resolved — shared library loaded
```

**Env-free proof of the `@loader_path` hop** (`dlopen` by absolute path, `DYLD_LIBRARY_PATH` + `DYLD_FALLBACK_LIBRARY_PATH` unset, under SIP-stripped `/usr/bin/python3`):

```
loaded /Users/runner/work/gjsify/gjsify/packages/infra/lightningcss-native/prebuilds/darwin-arm64/libgjsifylightningcss.dylib
loaded /Users/runner/work/gjsify/gjsify/packages/infra/oxfmt-native/prebuilds/darwin-arm64/libgjsifyoxfmt.dylib
```

**This answers the worry that motivated the task:** the `darwin-arm64` prebuilds `commit-prebuilds` put on main in 6120dcdf are NOT broken — they carry `@rpath/libgjsify_<name>.dylib` + `LC_RPATH @loader_path` with both files staged, and they load. Nothing had proven it before; now the required job does, on every run.

### Experimental job — `[experimental — rolldown + webgl + webrtc]` ✅

The leg that produced the original failure, now passing with the cdylib staged:

```
[check-prebuild-loader-path] packages/infra/rolldown-native/prebuilds/darwin-arm64
  libgjsify_rolldown.dylib [macho] needs (no sibling) | search: (none)
  libgjsifyrolldown.dylib [macho] needs @rpath/libgjsify_rolldown.dylib | search: …/glib/2.88.2/lib, …/gettext/lib, …/json-glib/1.10.8/lib, @loader_path
[check-prebuild-loader-path] OK — 1 directory sound

rolldown typelib loaded OK
loaded /Users/runner/work/gjsify/gjsify/packages/infra/rolldown-native/prebuilds/darwin-arm64/libgjsifyrolldown.dylib
```

The last line is the env-free one — `@loader_path` alone found `libgjsify_rolldown.dylib`, no `DYLD_*` set.

### Mechanism, confirmed rather than reasoned

The rpath was never broken. dyld's error listed the prebuild directory **first** because that is `@loader_path` expanding correctly; the Homebrew entries after it are the glib `LC_RPATH`s meson bakes, tried only because the first candidate did not exist. The two leaf names differ deliberately (`libgjsifyrolldown` = Vala/typelib, `libgjsify_rolldown` = cargo cdylib) and both are correct — the experimental job's `cp` simply never staged the second. `install_name_tool -id @rpath/<leaf>` + `build_rpath`/`install_rpath` `@loader_path` (from #812) is the right construction and is unchanged.

### Follow-up recorded, not smuggled in

`rolldown-native`'s darwin leg now passes every gate the required job applies to its siblings, but promoting it means a matrix + `gjsify.platforms` + `commit-prebuilds` change — a job-graph edit this PR deliberately does not make. Logged in STATUS.md → CI coverage follow-ups.

Issue #655, opened 2026-06-28
@gjsify/vite-plugin-blueprint: `types/` excluded from npm tarball


Luoyayu:
This was introduced in commit df2c038 (PR #179 — Phase D.7d), which added `"files": ["lib"]` to packages. 
The fix was correct for packages where all types live under `lib/`, 
but **vite-plugin-blueprint** is unique in having a `types/` directory referenced via its `exports` map.

PR #837, opened 2026-07-27
fix: fail builds that emit no declared output

JumpLink:
Closes #67.

## The bug

`gjsify tsc` reports success while emitting nothing. Reproduced verbatim in `packages/web/adwaita-core`:

```
# with tmp/.tsbuildinfo present
$ gjsify tsc ; echo $?
0
$ ls lib/
esm                 ← lib/types MISSING

$ rm tmp/.tsbuildinfo && gjsify tsc ; echo $?
0
$ ls lib/
esm  types          ← now emitted
```

Mechanism: `composite` + `incremental` + a `tsBuildInfoFile` that lives **outside the emit tree**. When the build info outlives a wiped `lib/`, tsc considers the project up to date, prints nothing, writes nothing and exits 0. The package publishes a `types` entry point that does not exist and consumers hit `TS7016: Could not find a declaration file…` — which reads like a *consumer* bug.

**Nothing in the pipeline could see it.** `check` (`gjsify tsc --noEmit`) type-checks SOURCES and never looks at `lib/`. The build cache owns per output UNIT, so a tree missing a unit is a HIT. `gjsify pack`'s type-shipping guard (#655) deliberately does **not** fire on an ABSENT declaration, because an unbuilt dev tree must still pack. And `tests/e2e/tsbuildinfo-containment` proves the build info stays inside the *package*, not that it is coupled to the *emit tree*. The only signal was "the build exited 0", and it lied.

## Blast radius — the desyncing shape

Surveyed every `tsconfig*.json` under `packages/ showcases/ examples/ tests/ templates/ website/` and evaluated TypeScript's own `getTsBuildInfoEmitOutputFilePath`. Of **103** incremental/composite configs, **93** put their build info outside the emit tree; only 10 are already contained.

**A. Explicit `tsBuildInfoFile: "tmp/.tsbuildinfo"` — 18**

`packages/dom/canvas2d-core` · `packages/framework/{adwaita-app,bridge-types,canvas2d,devtools,devtools-browser,devtools-cdp,devtools-mcp,devtools-protocol,event-bridge,iframe,stories,storybook,storybook-core,video,webgl}` · `packages/web/adwaita-core` · `packages/dom/dom-elements` (`tsconfig.tsbuildinfo` at the package root)

**B. DERIVED, `outDir: lib` + `rootDir: src` → `<pkg>/tsconfig.tsbuildinfo` (package ROOT, survives `rm -rf lib`) — 67**

`packages/gjs/{runtime,utils}` · `packages/nativescript-bridge/{adwaita,devtools,fs,platform,storybook}` · `packages/node/{assert,async_hooks,buffer,child_process,cluster,console,constants,crypto,dgram,diagnostics_channel,dns,domain,events,fs,globals,http,http2,https,inspector,module,net,os,path,perf_hooks,process,querystring,readline,sqlite,stream,string_decoder,sys,timers,tls,tty,url,util,v8,vm,worker_threads,ws,zlib}` · `packages/web/{abort-controller,compression-streams,dom-events,dom-exception,domparser,eventsource,fetch,formdata,gamepad,message-channel,streams,web-globals,webassembly,webaudio,webcrypto,webrtc,websocket,webstorage,xmlhttprequest}`

**C. DERIVED into `lib/` but as a LOOSE FILE — 8** (two configs each)

`packages/infra/{npm-registry,semver,tar,workspace}` — `tsconfig.json` and `tsconfig.build.json` derive to `lib/tsconfig.tsbuildinfo` / `lib/tsconfig.build.tsbuildinfo`. Coupled to `rm -rf lib`, but a loose file in `lib/` degrades the build cache's candidate from per-child units (`lib/esm`, `lib/types`) to whole-dir ownership (`build-cache.ts`) — the shape the v2 layout exists to avoid. Worth a separate look; not touched here.

Already contained (10): `packages/infra/{lightningcss,oxfmt,rolldown}-native`, `packages/napi/napi`, `packages/node/{http2-native,http-soup-bridge,sab-native,terminal-native,tls-native}`, `packages/web/webrtc-native` — the ones #26 pinned after the same trap escaped the package entirely.

## The fix, and why this one

`scripts/verify-package-outputs.mjs` asserts that every path a non-private workspace **declares** — `main` / `module` / `types` / `typings` / `browser` / `bin` / `gjsify.bin` and every `exports` condition, walked recursively — exists on disk, and treats a miss as a **build failure**. Package set and path set are both DERIVED (root `workspaces` globs × each `package.json`), so a new package or a new subpath is covered the day it lands. There is deliberately **no per-package opt-out** — an opt-out list is the artefact that drifts and then lies. Same shape as `scripts/verify-committed-bundles.mjs`.

**Why not co-locate the build info with the emit dir?** It fixes exactly one cause and it fights the build cache: `lib/` is a cache candidate whose per-child-unit granularity (`lib/esm` vs `lib/types`) is what stops `build:gjsify` clobbering `build:types`, and a candidate holding a **loose** file degrades to whole-dir ownership — the v0.x clobber, reintroduced (group C above is already living that). The only safe landing spot would be *inside* `lib/types`, which then makes the incremental compiler's correctness depend on cache-restore semantics. And it still would not catch a missing `main`, a build script that was never wired into `build`, or a typo in `exports`. The post-condition is cause-agnostic and covers all of them.

**What it costs.**
- It is a POST-condition, so it needs a built tree. It runs in `main.yml`'s **`build` job and only there** — the only job that has one — right after `Build` and before `Verify committed bundles` (which rebuilds and restores around itself). Both build paths leave a complete tree: FULL builds everything, and the selective path only runs when the warm-cache probe found a baseline.
- It proves **existence, not freshness**. A `.d.ts` regenerated from stale sources still passes; that class belongs to `verify-committed-bundles.mjs`.
- Wildcard subpaths (`./assets/*`) are only checked down to their static directory prefix.
- Runtime is a stat sweep (~0.1 s), so no job that merely restores the build cache needs to repeat it.

## Failing → passing, on the reproduction

```
──────── A. built state: lib/{esm,types} + tmp/.tsbuildinfo
esm
types
All declared entry points exist.

──────── B. lib/types wiped, tmp/.tsbuildinfo kept  →  gjsify tsc ; echo $?
0
ls lib/ → esm

──────── C. the guard on that state
Declared-output check: 1 workspace package(s) (non-private)
ERROR: @gjsify/adwaita-core (packages/web/adwaita-core) declares 2 path(s) that do not exist:
    types → lib/types/index.d.ts   (missing file: packages/web/adwaita-core/lib/types/index.d.ts)
    exports["."]["types"] → ./lib/types/index.d.ts   (missing file: packages/web/adwaita-core/lib/types/index.d.ts)
    produced by: gjsify workspace @gjsify/adwaita-core build · … · gjsify workspace @gjsify/adwaita-core build:types · …
ERROR: 2 declared path(s) missing across 1 package(s). A build that leaves a declared entry point
unwritten still exits 0 — most often a `.tsbuildinfo` that outlived its output tree, which makes
`gjsify tsc` a silent no-op. …
GUARD EXIT CODE: 1

──────── D. rm tmp/.tsbuildinfo && gjsify tsc ; echo $?
0
ls lib/ → esm
types

──────── E. the guard on that state
Declared-output check: 1 workspace package(s) (non-private)
All declared entry points exist.
GUARD EXIT CODE: 0
```

## Tests

`tests/e2e/package-outputs/run.mjs` (8 cases, wired into `test:e2e` + `test:e2e:package-outputs`): fixture workspace roots covering the field matrix (`types` / `main` / `bin` / `exports` conditions), subpath patterns, private packages, non-path values (`browser` remaps, `false`), `--allow-unbuilt` — plus the #67 mechanism driven **end to end through the real committed `@gjsify/tsc` bundle**: emit → wipe `lib/` → tsc exits 0 having written nothing → guard fails → drop the build info → tsc emits → guard passes. That last case fails if TypeScript ever stops no-opping *or* if the guard stops noticing.

```
✔ passes when every declared path exists
✔ FAILS when the declaration tree is missing, naming the field and the path
✔ covers main and bin, not just types
✔ ignores non-path values (bare specifiers, false, nested condition objects)
✔ checks a subpath PATTERN down to its static directory prefix
✔ skips private workspaces unless --include-private
✔ --allow-unbuilt downgrades the failure to a warning
✔ catches the stale-tsbuildinfo silent no-op end to end
ℹ pass 8  ℹ fail 0
```

`tests/e2e/tsbuildinfo-containment` (the #26 guard) stays green; `oxfmt --check` and `oxlint` clean on both new files.

## `@gjsify/adwaita-web`'s workaround — kept, with a reason

`@gjsify/adwaita-core` now emits its declarations, and `check` + `build:types` are green for both `@gjsify/adwaita-core` and `@gjsify/adwaita-web`.

The workaround does **not** become removable. An isolated probe — a consumer built with adwaita-web's exact `compilerOptions`, placed outside this repo's `node_modules` ancestry (the worktree is nested under the parent checkout, so an in-tree probe silently resolves against the parent's built `node_modules`) — pins it down:

| adwaita-core `lib/types` | flags | result |
|---|---|---|
| present | `strict: false` (as shipped) | clean |
| present | `--noImplicitAny` | clean |
| **missing** | `strict: false` (as shipped) | **clean** |
| **missing** | `--noImplicitAny` | `error TS7016: Could not find a declaration file for module '@gjsify/adwaita-core'` |

So `noImplicitAny: false` (via `strict: false`) is indeed what suppresses TS7016 — but that line dates to #3 (2026-04-03) and `@gjsify/adwaita-core` to #683 (2026-07-02), so it predates the dependency by three months and was never a workaround *for it*. Enabling it surfaces genuine unrelated strictness debt (`TS7011` in `src/source-view/adw-source-view.ts:408`), i.e. a separate migration, not a symptom cleanup — and the guard now catches the real symptom at the producing end regardless of what any consumer's tsconfig says.

## Note on local verification

The shared machine could not complete a `gjsify install --immutable` (six concurrent agent installs pinned every core; the download loop moved 9/1597 in 35 min), so the build halves that need the bundler were exercised via the committed `@gjsify/tsc` GJS bundle plus artifacts from the sibling checkout at the same commit. Everything the guard itself does is filesystem-only and was run directly.


JumpLink:
## CI finding resolved — it was **(b)**, a genuinely broken build

The guard's first run flagged `@gjsify/xmlhttprequest`. Diagnosis: **(b) a genuinely broken/unwired build** — `build:types` never produced those declarations *for anyone*, including every published release. Not (a) stale cache state, not (c) an over-broad subpath (the `./register` subpath is real and ships `lib/esm/register.js`).

### Evidence

`packages/web/xmlhttprequest/tsconfig.json` had `outDir: lib` + `rootDir: src` + `composite` and **no `declarationDir`**, so declarations land in `outDir`:

```
$ rm -rf lib tsconfig.tsbuildinfo && gjs -m packages/infra/tsc/dist/tsc.gjs.mjs   # in the package
tsc exit=0
lib/index.d.ts
lib/index.js
lib/register.d.ts
lib/register.js          ← never lib/types/**
```

Three independent confirmations:
- the **fully built sibling checkout** at this commit has no `lib/types/` under that package;
- the **published tarball** `@gjsify/xmlhttprequest@0.23.0` contains `package/lib/index.d.ts` + `package/lib/register.d.ts` while its `package.json` declares `types: lib/types/index.d.ts` — so every TypeScript consumer of the *released* package already hits `TS7016`;
- it also ships **two JS trees** for one entry (`lib/index.js` from tsc *and* `lib/esm/index.js` from rolldown).

Sibling `@gjsify/fetch` has the same no-`declarationDir` shape but *declares* `lib/index.d.ts` to match — internally consistent, which is why it never fired.

### Fix (at the cause, on the build side)

`declaration` + `emitDeclarationOnly` + `declarationDir: "lib/types"` — the layout ~100 siblings use and the one this package's `exports`/`files` have always promised. Correcting the *declaration* to `lib/*.d.ts` instead would have enshrined the outlier layout and left the stray-JS hazard (`packages/infra/workspace/tsconfig.build.json` documents exactly that trap).

```
$ rm -rf lib tsconfig.tsbuildinfo && gjs -m …/tsc.gjs.mjs
tsc exit=0
lib/types/index.d.ts
lib/types/register.d.ts
$ gjs -m …/tsc.gjs.mjs --noEmit   # check
exit=0
$ node scripts/verify-package-outputs.mjs --only @gjsify/xmlhttprequest …
All declared entry points exist.
```

### Is anything else in the same position? No — and this is the complete picture, not one CI round at a time

Three sweeps over **every** non-private workspace:

1. **Declared `types` vs emittable `.d.ts`** (each tsconfig's `declarationDir ?? outDir` × rootDir-relative sources): **0** mismatches after the fix. Before the fix it found exactly one — `@gjsify/xmlhttprequest`, all three paths. (A coarser prefix-only version missed it, because `lib/types/index.d.ts` *is* under `lib/`; the emitted-file-set comparison is what catches it.)
2. **Declared `lib/esm/**.js` vs the sources behind the `--library` mirror**: **0**.
3. **The guard itself against the fully built sibling checkout** at this commit: **119 checked, 1 broken** — this package, these three paths. All 118 others clean.

So the empirical answer is one package, not a queue.

### Not fixed here (reported, deliberately out of scope)

- `@gjsify/fetch` emits stray `lib/*.js` next to rolldown's `lib/esm/*` (same tsconfig shape, but consistent declarations — nothing is broken).
- The 8 `packages/infra/{npm-registry,semver,tar,workspace}` configs deriving a **loose** `lib/*.tsbuildinfo`, which degrades the build cache's `lib/` candidate from per-child units to whole-dir ownership (group C in the PR body).

Also in this push: `--json` no longer appends the human failure line, so the payload stays parseable under `2>&1`.

PR #838, opened 2026-07-27
ci: hold declared platforms to a real prebuild

JumpLink:
Repo task #56 — *`gjsify.platforms` can promise platforms that ship no prebuild, and nothing notices.*

## The gap

The three existing platform invariants compare **declarations to each other**: the promise (`gjsify.platforms`), the directory names a package happens to carry, and the targets a CI job produces. A target with **no** directory is simply absent from the "shipped" set they compare against, so all three stay green on an empty promise. That is how `@gjsify/oxfmt-native` declared `darwin-arm64` with no artifact behind it for weeks while `node scripts/audit-runtimes.mjs --check` exited 0 the entire time.

`--check` now also opens what is behind the promise.

## Half 1 — existence

For every package that names a committed `gjsify.prebuilds` directory, each declared target must have that directory, holding a shared library in that OS's format **and** the `.typelib` without which `GI_TYPELIB_PATH` resolves nothing. Packages with no `gjsify.prebuilds` (`@gjsify/node-gi`, node-gyp, built on install / shipped from a release artifact) are out of scope — this repo owes no committed body for them.

## Half 2 — loadability

A directory that exists proves nothing. #832's lesson was a *required* macOS job that built two bridges and only **copied** them; the bug hiding there was a missing sibling FILE that read exactly like a broken rpath.

**Structurally, on every committed artifact regardless of target** — the half that works from any host:

* the image's own machine must match the directory it sits in (ELF `e_machine` / Mach-O `cputype` / PE `Machine`);
* every `libgjsify*` sibling it records must be staged beside it and reachable via `$ORIGIN`/`@loader_path` — `scripts/check-prebuild-loader-path.mjs`, reused rather than reimplemented;
* every library leaf the **typelib** records must be present, because that leaf is what GI hands to `dlopen` the moment a consumer resolves a class. A `meson.build` rename that the staging step follows but the typelib does not was invisible to every check before this.

**Functionally, only for the checking host's own `${process.platform}-${process.arch}`** — a real `dlopen` with `LD_LIBRARY_PATH`/`DYLD_LIBRARY_PATH`/`DYLD_FALLBACK_LIBRARY_PATH` stripped, the same env-free probe the macOS jobs use, so the self-relative sibling hop is proven rather than inferred.

### What this can and cannot check from Linux

| | verified | how |
|---|---|---|
| any target, any host | machine ↔ directory, sibling staging + `$ORIGIN`/`@loader_path`, typelib-named leaves present, artifact is a real image | file headers, parsed directly (no `otool`/`readelf`) |
| host's own target only | it actually loads, env-free | `dlopen` |
| **cannot** be checked here | a cross-arch prebuild running, its symbols resolving on real hardware, GI class resolution on that OS | needs that architecture |

The summary prints both counts **separately on every run** — on this branch: `27 directories verified structurally … 11 libraries loaded on linux-x64; 16 are for OTHER targets and CANNOT be loaded here`. A bridge whose *system* dependencies the runner lacks (libsoup, GStreamer, libgda — the audit runs on a bare `ubuntu-latest`) is reported as **not load-tested**, never as broken: that would be a fact about the runner, not the artifact.

## The escape hatch

`package.json#gjsify.platformsUncommitted: {"<target>": "<why>"}` — a target that is genuinely declared and genuinely CI-built, whose artifact this repo deliberately does not commit. Awkward to abuse on purpose: the reason is **mandatory** and printed on every run, the target must already be in `gjsify.platforms` (you can only defer what you promise), and the entry becomes a **failure** the moment the directory appears, so it cannot outlive its cause. The matrix renders it `○`, distinct from `✓`.

## Two live gaps this found on its first run

**1. Every emulated-arch prebuild in the repository was x86-64.** All 24 committed `linux-{ppc64,s390x,riscv64}` directories, across all eight bridges that declare them:

```
packages/node/tls-native/prebuilds/linux-s390x/libgjsifytls.so:
  ELF 64-bit LSB shared object, x86-64, … BuildID[sha1]=79fce958…
packages/node/tls-native/prebuilds/linux-ppc64/libgjsifytls.so:
  ELF 64-bit LSB shared object, x86-64, … BuildID[sha1]=79fce958…   ← same build
```

`uraimo/run-on-arch-action` **ignores its `arch` input whenever a custom `base_image` is supplied** (its README: *"Specify a custom base image, `arch` and `distro` should be set to `none` in this case"*, with the platform carried by the image — `--platform=linux/riscv64 riscv64/ubuntu:24.04`). The job passed `arch: ${{ matrix.qemu_arch }}` next to a bare `base_image: fedora:43`, so docker ran the runner's own platform and every "emulated" leg compiled x86-64.

Nothing downstream could see it: the directory existed, the typelib sat beside it, and the artifact even *loaded* on an x86-64 runner. The `~3 min per emulated leg` figure recorded in the workflow and STATUS.md — used to argue the legs were cheap — measured no emulation at all, and contradicted the same file's note that the lightningcss Rust build takes 20-30 min under QEMU.

Fixed at the root (`base_image: --platform=linux/ppc64le fedora:43`, `arch: none`), the 24 mis-staged directories removed, and each package records the gap in `platformsUncommitted` until a `commit-prebuilds` run on `main` restores real binaries. An exotic-arch consumer now hits the guarded `imports.gi` degrade instead of an unloadable library.

**2. `@gjsify/napi` declares `darwin-arm64`** which `napi.yml` builds, load-tests and uploads for a release to ship, and no job commits back. Now stated in `platformsUncommitted` instead of being implied by an absent directory.

## Where it runs

`.github/workflows/audit-runtimes.yml` runs `--check --strict` on **every `pull_request`**, no `paths` filter — confirmed, and load-bearing here: the things that break this invariant are a `gjsify.platforms` edit, a `prebuilds/**` binary and `prebuilds.yml`'s matrix, which no per-package path list would cover together.

`parseCiPlatforms()` still reads `prebuilds.yml` as **static YAML**, and now only accepts `arch:` values that name a CPU — otherwise the `arch: none` the action requires alongside a custom `base_image` would be read as a target and fail every package the emulated job builds on a phantom `linux-none`.

## Demonstrations

Each run against the branch tip and restored afterwards (`git status` clean each time).

**Half 1 — remove a committed prebuild directory** (`packages/infra/oxfmt-native/prebuilds/darwin-arm64`, the exact historical gap):

```
PREBUILD-ARTIFACT FAILURES (does the declared platform have a loadable body?) on 1 target(s):
  - @gjsify/oxfmt-native: declares `darwin-arm64` but ships no `prebuilds/darwin-arm64/` — a
    promised platform with no artifact behind it. Either commit the prebuild …, or record the
    gap honestly in `gjsify.platformsUncommitted` with the reason.
```
exit 1 → restored → exit 0.

**Half 1 — declare a platform that does not exist** (`win32-x64` added to `oxfmt-native`): the pre-existing declared-vs-CI check *and* the new artifact check both fire. Adding `platformsUncommitted: {"win32-x64": ""}` to silence it does not work:

```
  - `gjsify.platformsUncommitted["win32-x64"]` needs a non-empty reason. An unexplained
    exemption is the silent gap this field exists to replace.
```
exit 1 → restored → exit 0.

**Half 2 — withhold the sibling cdylib from the macOS prebuild**, which this Linux host cannot run (`libgjsify_oxfmt.dylib` moved aside):

```
  - @gjsify/oxfmt-native: …/prebuilds/darwin-arm64/libgjsifyoxfmt.dylib: records a dependency
    on @rpath/libgjsify_oxfmt.dylib but libgjsify_oxfmt.dylib is NOT staged in …
    The loader resolves the directory correctly and then finds nothing there — stage both
    libraries (the Vala one the typelib names AND the Rust cdylib it links against).
```
exit 1 → restored → exit 0.

**Half 2 — the live one**: on the pre-fix tree the machine check reported **27 failures** across seven packages, e.g. `prebuilds/linux-ppc64/libgjsifyhttpsoupbridge.so is a linux/x64 image in a linux-ppc64 directory`, plus `@gjsify/napi`'s missing `darwin-arm64`. That is the finding above, and it is the one loadability defect a cross-arch target can be caught for at all from a Linux host.

`tests/e2e/prebuild-declaration-invariant/` locks all of it in — 16 cases over synthetic packages built from real linked binaries in a temp directory (the e2e suites run four-at-a-time against one shared checkout, so a suite that removed a real prebuild would break whatever else is reading it).

## Notes for review

* `scripts/check-prebuild-loader-path.mjs` stays the **one** binary parser: it gains `os`/`arch` for ELF/Mach-O/PE and a GI-typelib `shared-library` reader instead of a second parser growing beside it. PE reports `inspectable: false`, so the machine check covers a future win32 `.dll` while its unparsed import table is never mistaken for "records no sibling".
* The audit's main block moves into `main()` behind an entry guard so `auditPrebuildArtifacts` is importable — that is what makes the invariant testable without mutating the checkout. `git diff -w` shows the wrap for what it is (434 real insertions, 2 deletions).
* Formatting churn is isolated in its own `style:` commit; `scripts/audit-runtimes.mjs` was not oxfmt-clean at HEAD.
* `gjsify lint` clean (the two warnings on this file are pre-existing).

## Follow-up, not in this PR

Nothing has verified a genuinely emulated leg yet — `prebuilds.yml` build legs run on `pull_request`, so this PR's own run is the first evidence, and `commit-prebuilds` is main-only, so the real ppc64/s390x/riscv64 binaries only land after merge. Expect the QEMU legs to become the PR critical path now that they actually emulate; re-measure and decide deliberately whether they stay on `pull_request` (STATUS.md updated to say the old figure measured nothing).


JumpLink:
## First CI evidence

**`Detect runtime-triplet drift` — pass.** The invariant runs on the bare `ubuntu-latest` Node runner and reports its own boundary rather than overclaiming:

```
prebuild-artifact audit: 27 committed prebuild directories across 11 packages verified STRUCTURALLY …
  functional load (env-free `dlopen`…): 8 libraries on this host's own target `linux-x64`;
  16 directories are for OTHER targets and CANNOT be loaded here — a cross-arch prebuild is
  verifiable only from its file headers, and this audit does not pretend otherwise.
  3 host-target load(s) skipped (see notes).
  · @gjsify/http-soup-bridge: `linux-x64/…` not load-tested — this host lacks a system dependency
    it links against (libsoup-3.0.so.0: cannot open shared object file). Structure was verified;
    the functional load was not.
  · @gjsify/napi: … (libgjs.so.0)
  · @gjsify/webrtc-native: … (libgstwebrtc-1.0.so.0)
```

8 loaded / 3 skipped-for-host / 16 structural-only — the runner's missing libsoup, libgjs and GStreamer are reported as facts about the runner, never as broken artifacts. (Locally on Fedora, where those libraries exist, all 11 load.)

## The `--platform=` fix works, and it exposes a second real problem

`Build prebuilds (s390x) [QEMU]` is now **red**, and this is the fix working rather than a regression. The generated Dockerfile now carries the platform:

```
1 | FROM --platform=linux/s390x fedora:43
3 | >>> RUN chmod +x /root/run-on-arch-install.sh && /root/run-on-arch-install.sh
ERROR: … did not complete successfully: exit code: 139
```

and the emulated container dies before installing anything:

```
/root/run-on-arch-install.sh: line 33: 12 Segmentation fault (core dumped)
  dnf install -y git tar xz findutils curl meson vala gcc pkgconf glib2-devel …
```

So `linux-s390x` has **no working build path at all** — Fedora's `dnf` segfaults under `qemu-user` on emulated s390x. The leg has been "passing" only because it was never emulating; there has never been an s390x build, and the artifact previously committed as one was an x86-64 binary.

That is a support-matrix decision rather than a scripting one, so this PR does not make it: either the s390x leg needs a base image whose package manager survives emulation (or cross-compilation via `cross` instead of qemu-user), or `linux-s390x` should come out of `gjsify.platforms` on the eight packages that declare it. Whichever way it goes, `audit-runtimes --check` now forces the declaration and the reality to agree.

`ppc64` and `riscv64` are still running at the time of writing — themselves evidence the fix took effect, since those legs used to finish in about three minutes. I will follow up with their outcome.

JumpLink:
## `ppc64` too — emulation reached, build still impossible

`Build prebuilds (ppc64) [QEMU]` ran for **7m50s** (against the ~3m00s once recorded as its "measured" emulated cost) and got much further than s390x — real `ppc64le` packages, 574 of them:

```
FROM --platform=linux/ppc64le fedora:43
…
[552/574] Installing pipewire-pulseaudio-0:1.4.11-1.fc43.ppc64le
[574/574] Removing libgomp-0:15.2.1-7.fc43…
Transaction failed: Rpm transaction failed.
ERROR: … exit code: 1
```

So the dependency install itself dies in RPM's scriptlet phase under `qemu-user`. Same conclusion as s390x, different symptom: **`linux-ppc64` has no working build path today either.** `riscv64` (Ubuntu/apt rather than Fedora/dnf) is still running.

### Two things worth deciding, neither of them scripting

1. **The emulated targets need a build that works, or they need to come out of `gjsify.platforms`.** Under real emulation, Fedora's package manager does not survive `qemu-user` on either s390x (segfault) or ppc64 (RPM transaction failure). Options are a base image whose package manager does survive, or dropping qemu-user for cross-compilation (`cross`), or narrowing the declared matrix. This PR deliberately does not choose — it makes the choice unavoidable instead of optional, and `audit-runtimes --check` now keeps the declaration and the reality in step whichever way it goes.

2. **`arch: none` collapses `run-on-arch-action`'s GHCR cache key.** The action derives its cached image name from `arch` + `distro`, so all three legs now share `run-on-arch-…-none-none` and the first run reported `failed to configure registry cache importer: …-none-none:latest: not found`. Harmless while the legs cannot build at all, but it must be resolved alongside (1) — otherwise three different base images contend for one cache tag. The `arch` input cannot simply be restored: the action only accepts its own arch names (`ppc64le`, not the `ppc64` this repo uses as the prebuild-directory token), and it is the input the action ignores for platform selection in the first place.

The invariant itself is unaffected by either: `Detect runtime-triplet drift` passes, the committed tree is consistent, and the three targets are recorded as declared-but-uncommitted with their reason until a build exists that can produce them.

JumpLink:
## The emulated legs are fixed — two stacked defects, both verified locally

### 1. Wrong architecture (the original finding)

`uraimo/run-on-arch-action` ignores its `arch` input whenever a custom `base_image` is supplied. The job passed `arch: ${{ matrix.qemu_arch }}` next to a bare `base_image: fedora:43`, so all three legs compiled **x86-64** and staged it as ppc64/s390x/riscv64.

### 2. An unmaintained emulator (found by fixing #1)

Setting the platform explicitly made the legs emulate — and fail. The action *also* resets binfmt to `multiarch/qemu-user-static` (**qemu 7.2**, effectively unmaintained) before it builds, and under that emulator Fedora's package manager does not survive:

* `s390x` — `dnf` segfaults at startup (job dead in 28 s)
* `ppc64` — 574 packages install, then `Transaction failed: Rpm transaction failed.` in wireplumber's `%posttrans` scriptlet (7 m 50 s)

Neither is an architecture problem. **Both build cleanly under a current qemu**, which I could establish without CI because Fedora ships `qemu-user-static` — an emulated leg reproduces with `podman run --platform`. Under **qemu 10.2.2**:

```
# s390x — @gjsify/terminal-native, full Vala + typelib build, exit 0
$ file build/libgjsifyterminal.so
build/libgjsifyterminal.so: ELF 64-bit MSB shared object, IBM S/390, version 1 (SYSV) …
```

a genuine **big-endian s390x** library — the first that has ever existed in this repo. And the ppc64le side:

```
# the exact transaction that fails on CI's qemu 7.2
INSTALL_OK        (525 packages, real 5m58s)
```

## What changed

* **The emulator is pinned like any other toolchain component** — `docker/setup-qemu-action@v3` + `tonistiigi/binfmt:qemu-v10.2.3` — and `run-on-arch-action` is gone. Once binfmt is right, a plain `docker run --platform` in the target-arch image is all the job needs, and dropping the action removes its other two liabilities at the same time.
* **Cache-key collapse resolved by construction** (constraint 2). The action derived its GHCR image name from `arch`+`distro`, so with the documented `arch: none` all three legs shared one `run-on-arch-…-none-none` tag. There is no input to override it. With the action gone nothing pushes to GHCR, so the job also sheds `packages: write`.
* **Weak/recommended dependencies OFF.** The Fedora list pulled 574 packages, mostly weak deps of gtk4-devel/gstreamer — pipewire, wireplumber, fonts, bat, 7zip. None is needed to compile a Vala library, each costs emulated install time, and the failing RPM scriptlet was wireplumber's: the build was dying inside a package it does not use. 574 → 383 requested.
* **The build moved to `.github/prebuild-toolchain/emulated-build.sh`** so reproduction is one command rather than a copy-paste out of YAML, plus a fail-fast `g-ir-compiler` probe (meson resolves it by name and the riscv64 image is a different distro) and a `file` on every staged artifact — the same machine-vs-directory fact the audit holds against the committed tree, asserted at the source where a mismatch names the leg that produced it. Both workflow triggers now include the script.
* `parseCiPlatforms()` attribution verified unchanged — `--platforms` still shows all three emulated columns, because it reads the per-package **upload** steps, which stay in the workflow.

## A parser bug the first real s390x artifact found

A GI typelib is written in the byte order of the machine that compiled it and carries no endianness flag. `readTypelibSharedLibraries` read the s390x one little-endian, got an out-of-range offset, and reported *"this namespace records no library"* — silently skipping the staged-leaf check on the one architecture we ship that is big-endian. The header's `size` field holds the blob's own length, so the byte order in which it equals the file size is the file's: a self-validating probe rather than a guess. Covered by a test that byte-swaps the committed little-endian typelib, so the suite needs no second binary fixture.

This is the second time a check written for this PR was wrong in a way only a real artifact could show, and both times the artifact was reachable locally. Worth keeping: `podman run --platform linux/<arch>` is cheap next to a 20-minute CI round trip per hypothesis.

JumpLink:
## Machine-checked artifacts, and the real cost

### `file` per target — genuine cross-architecture output

Built under emulation with the pinned qemu (10.2.2 locally), then identified on the host:

```
# s390x — @gjsify/terminal-native (Vala + typelib), exit 0
libgjsifyterminal.so:       ELF 64-bit MSB shared object, IBM S/390

# ppc64le — @gjsify/lightningcss-native, BOTH libraries (Vala + Rust cdylib)
libgjsifylightningcss.so:   ELF 64-bit LSB shared object, 64-bit PowerPC, OpenPOWER ELF V2 ABI
libgjsify_lightningcss.so:  ELF 64-bit LSB shared object, 64-bit PowerPC, OpenPOWER ELF V2 ABI
```

`MSB` on the s390x one is the point: genuinely big-endian, which is also what exposed the typelib-endianness bug in my own checker. Both lightningcss libraries are staged, so the #832 sibling trap is covered on this path too.

### CI: past both old failure points

| leg | qemu 7.2 (before) | qemu 10.2.3 (now) |
|---|---|---|
| s390x | failed at **28 s** — `dnf` segfault | **1945 s** and running |
| ppc64 | failed at **981 s** — RPM transaction | **1944 s** and running |
| riscv64 | cancelled at 1471 s | **1945 s** and running |

Both architectures that used to die are now well past the point where they died, doing real emulated compilation.

### Per-leg cost, measured (ppc64le, local, qemu 10.2.2)

| step | wall-clock |
|---|---|
| dependency install (weak deps off) | **5 m 59 s** (525 pkgs) |
| rustup | 1 m 22 s |
| `meson setup` | 0 m 15 s |
| **`meson compile` — lightningcss Rust cdylib** | **29 m 10 s** (user 72 m) |
| the seven Vala/GI packages | minutes |

CI elapsed (~32 min and counting on all three) matches that shape.

## Recommendation on #831

**The architectures are not the expense — one package is.** `@gjsify/lightningcss-native`'s Rust cdylib is ~29 of every ~37 emulated minutes. Everything else is the Vala/GI half, which is both the cheap part and the part the architecture actually changes (ABI, endianness, typelib layout).

So I would **not** revisit #831's "every Linux leg on every PR" decision by dropping architectures. Drop the *package* from the emulated legs instead:

- **On `pull_request`**: build the seven Vala/GI bridges on ppc64/s390x/riscv64 (~8-10 min per leg, parallel with the 8-15 min native legs — no critical-path change, which is what #831 assumed and is now true rather than accidental).
- **On `main` + the nightly sweep**: add `lightningcss-native` back, so the committed exotic-arch prebuild for it stays real.

The justification for exempting exactly that package is that its crate graph is portable and every PR already compiles it on x64 **and** arm64 — a ppc64 compile is unlikely to catch anything those two do not, whereas the Vala/GI bridges are precisely where a big-endian or ABI difference shows up. If you would rather not split the matrix, the fallback is to move all three emulated legs to `main` + nightly and accept post-merge discovery; I would take the split.

I have **not** implemented that split — it changes what a PR verifies, so it is your call. Everything above is measurement.

PR #843, opened 2026-07-28
Build each native prebuild only when it changed

JumpLink:
`prebuilds.yml` built every package it owns on every leg on every run. On a
native runner that is 7-8 min; under real emulation it is the whole cost.

Measured on the first fully-emulated run (#838, run 30310941686):

| leg | total | of which `@gjsify/lightningcss-native` |
| --- | --- | --- |
| ppc64 [QEMU] | 41m38s | 32m02s (77%) |
| s390x [QEMU] | 38m54s | 32m18s (83%) |
| riscv64 [QEMU] | 96m47s | 70m28s (73%) |
| x64 (native) | 8m24s | — |
| arm64 (native) | 7m06s | — |

`@gjsify/rolldown-native` is already excluded from the emulated matrix for
exactly that reason. This continues that line per-package instead of
hard-coding a second name.

## The rule

Uniform across events — pull request, push to main, dispatch:

> a package is built when something it depends on changed; otherwise its
> build **and** its upload are skipped.

A new `changes` job decides once; every leg and `commit-prebuilds` read that
one decision, so "a package was not built" and "no artifact is downloaded for
it" are the same fact rather than two guesses.

## What counts as changed

1. **The package's own sources** — *derived* from this workflow's own
   `on: paths:` filter rather than restated, plus its `meson.build` and
   `package.json`.
2. **Its `refs/` submodule pin** — the union of `package.json#gjsify.refsLockstep`
   and the Cargo `path = ".../refs/<x>/..."` deps of `src/rust/Cargo.toml`
   (the same derivation `scripts/check-refs-pin.mjs` uses).
   `@gjsify/oxfmt-native` declares no lockstep (task #59, deliberately not
   fixed here) — the union is what stops that absence from quietly meaning
   "this package never rebuilds".
3. **Shared inputs** — this workflow, `.github/prebuild-toolchain/**` (where
   the base images live), `scripts/stage-prebuild.mjs`,
   `scripts/check-refs-pin.mjs`, `scripts/check-prebuild-loader-path.mjs` —
   rebuild **all** packages.

## Why not `gjsify affected`

It was evaluated first, and it does not fit — for two independent reasons:

* **It cannot see a submodule pin.** `refs/**` is the *first* entry of its
  `IGNORE` table (dropped before every other rule) and it ignores every
  workflow file except `main.yml`, including this one. Two of the three input
  classes above are structurally invisible to it.
* **Its graph is the wrong graph.** It walks npm production-dependency edges
  to answer *whose tests re-run*. A prebuild's inputs are Vala/C/Rust sources,
  `meson.build` and a pinned submodule. It would be wrong in both directions:
  an edit to a package `lightningcss-native` depends on would rebuild a
  32-minute cdylib whose sources did not change, while a `refs/rolldown` bump
  would leave `rolldown-native` out of the closure entirely.

So: path matching — but the path table is **derived** from this workflow
(package set from its own `path: packages/…/prebuilds/…` steps, per-package
globs from its own `paths:` filter), not a second hand-written copy of it.

## Two silent-staleness bugs found on the way, fixed here

* `refs/oxc` and `refs/rolldown` were **not in the trigger list**, so bumping
  a pin changed a binary without even starting this workflow.
* The three shared scripts were likewise absent.

`changed-packages.mjs` now **fails** if a package's sources or a linked
submodule are missing from the trigger list, so the gate and the trigger
cannot drift apart again.

## Fail-open, everywhere

Gates read `!contains(needs.changes.outputs.skip, '"<key>"')` on the raw
string. An unset output — job failed, output renamed, key missing — is the
empty string, `contains` is false, and the package **builds**. Gating on a
build list would have turned each of those into a silent skip. The classifier
step also writes an empty skip list and warns if the script exits non-zero,
and an unresolvable diff base (or a dispatch) builds everything.

## `commit-prebuilds`

`actions/download-artifact` **fails** on a missing artifact —
`src/download-artifact.ts` throws ``Artifact '<name>' not found`` into
`core.setFailed`, and there is no `if-no-artifact-found` input (no counterpart
to upload's `if-no-files-found`). So each download carries the same gate its
build did.

What that leaves: nothing was downloaded for a skipped package, its
`prebuilds/` directory is exactly what `actions/checkout` wrote, `git add`
stages nothing, and the existing `git diff --cached --quiet` short-circuit
commits nothing. **The committed prebuild is untouched — not deleted, not
emptied.**

A new **`Refuse to delete a committed prebuild`** step asserts it rather than
assuming it: `git add <dir>` has staged deletions since git 2.0, so anything
that removed a file under `prebuilds/` would otherwise be committed as an
unshipped binary. (An empty directory cannot be committed at all — git does
not track directories — so the reachable failure is the deletion, and that is
what is checked.)

Corollary, and it is load-bearing: **no build leg may be skipped as a whole
job**, only step by step. A skipped `needs` skips `commit-prebuilds` (there is
no `always()` there, deliberately — it is gate 3 of the PR guard), so a leg
that skipped itself would suppress committing everything the *other* legs
produced. The emulated legs therefore still run their ~20 s shell when they
have nothing to build.

## Honesty

`prebuilds-summary` now prints the per-package table with the reason, from the
`changes` job's own `report` output — the same source the gates read. A run
that builds two of ten packages says so, and names them, so it cannot read
like a full verification.

Also stated rather than glossed, in `changed-packages.mjs`' header: this
cannot see the **crates.io** half of a Rust bridge's graph (no `*-native`
package commits a `Cargo.lock`). Before this gate every run rebuilt
`lightningcss-native`, so registry drift landed incidentally; now it lands
when the package or a shared input changes, or on a `workflow_dispatch`, which
builds everything by design.

## Verification

* `scripts/audit-runtimes.mjs --check --strict` passes, and `--platforms`
  renders **byte-identical** to before — `parseCiPlatforms()` attributes per
  step block, and a step gains an `if:` line rather than losing its per-package
  upload step, so the declared-vs-built invariant is unaffected.
* `commit-prebuilds`' three steps were driven verbatim against a scratch repo
  for five cases (skipped/all-skipped/identical-bytes/file-removed/directory-wiped);
  the guard fires on the last two and `HEAD` is untouched in every one.
* Wall-clock demonstrations are in the linked verification PRs below.


JumpLink:
## Verification — measured, not asserted

Three demonstrations, each on its own branch based on **this** one so the
workflow change itself is not in the diff.

### 1. A diff touching only a Vala bridge — #844, run [30334507660](https://github.com/gjsify/gjsify/actions/runs/30334507660)

`build=["terminal-native"]`, nine packages skipped including `lightningcss-native`.

| leg | before ([30310941686](https://github.com/gjsify/gjsify/actions/runs/30310941686)) | after | Δ |
| --- | --- | --- | --- |
| x64 (native) | 8m24s | **51s** | 9.9× |
| arm64 (native) | 7m06s | **1m03s** | 6.8× |
| ppc64 [QEMU] | 41m38s | **5m30s** | 7.6× |
| s390x [QEMU] | 38m54s | **4m48s** | 8.1× |
| riscv64 [QEMU] | 96m47s | **22m23s** | 4.3× |

`changes` costs a flat 20 s. On the x64 leg every build/collect/upload group but
`terminal-native`'s reported `skipped`, along with the rustup install and both
submodule inits; inside the emulated container the log reads

```
SKIP rustup — no Rust bridge is being built on this leg
SKIP packages/framework/webgl — unchanged (PREBUILD_SKIP=[…])
…
packages/node/terminal-native prebuilds for linux-riscv64:
```

Where the time goes now: the **dependency install is the whole leg** —
s390x 4m05s of 4m48s, riscv64 21m02s of 22m23s (94%). The next lever there is a
prebuilt per-arch base image, not more build gating.

### 2. A diff touching `lightningcss-native/src` — #845, run [30334533299](https://github.com/gjsify/gjsify/actions/runs/30334533299)

`build=["lightningcss-native"]`. The x64 leg ran exactly these steps and nothing else:

```
Install prerequisites
Install Rust toolchain (rustup, stable)     ← gated on the Rust bridges
Build   @gjsify/lightningcss-native (Rust cdylib + Vala bridge)
Collect @gjsify/lightningcss-native prebuilds
Upload  @gjsify/lightningcss-native prebuilds artifact
```

x64 **2m59s**, arm64 **3m03s** (vs 8m24s / 7m06s for the full build). The
emulated legs pay full price for it, as they must — that is the negative
control.

### 3. A shared-input change rebuilds everything — this PR, run [30334467073](https://github.com/gjsify/gjsify/actions/runs/30334467073)

`skip=[]`, `reason=shared input changed (.github/workflows/prebuilds.yml, .github/prebuild-toolchain/changed-packages.mjs)`.
The x64 leg ran **40 steps with zero skips**; x64 8m40s / arm64 7m50s, i.e. the
pre-change baseline.

### 4. `commit-prebuilds` cannot damage a committed prebuild

It is main-only, so it was driven **verbatim** — its `Stage prebuilds`,
`Refuse to delete a committed prebuild` and `Commit prebuilds` bodies copied out
of this file — against a scratch repo laid out like the real one:

| case | staging | guard | `HEAD` afterwards |
| --- | --- | --- | --- |
| two packages skipped, one rebuilt with new bytes | 1 file | pass | skipped packages byte-identical; rebuilt one updated |
| every package skipped (no download ran) | nothing | pass | untouched, "No prebuild changes to commit." |
| rebuilt, bytes identical | nothing | pass | untouched |
| **a committed prebuild file removed** | deletion staged | **FAIL** | untouched — job stopped before the commit |
| **a skipped package's whole target dir wiped** | 2 deletions staged | **FAIL** | untouched |

The reason the guard is not redundant: `git add <dir>` has staged deletions
since git 2.0, so without it either hostile case would have been committed as an
unshipped binary.

`actions/download-artifact`'s behaviour was checked rather than assumed —
`src/download-artifact.ts` does `throw new Error("Artifact '<name>' not found")`
into the top-level `core.setFailed`, and the action has no
`if-no-artifact-found` input.

### Also verified

- `scripts/audit-runtimes.mjs --check --strict` passes and `--platforms` renders
  **byte-identical** to before, so `parseCiPlatforms()`' declared-vs-built
  invariant is untouched (it attributes per step block; a step gains an `if:`
  line and keeps its per-package upload step).
- `tests/e2e/prebuild-change-gate/` — 13 cases, 0.9 s, no built workspace
  required: per-package mapping, both `refs/` derivations, all five shared
  inputs, the package.json asymmetry, fail-open on an unresolvable base, the
  quoted-key contract, every workflow gate naming a key the classifier emits,
  every artifact step gated on its own package, and the emulated script's
  `should_build` helper driven out of the real file.

### One thing this PR does not fix, reported rather than absorbed

`.github/prebuild-toolchain/**` matches neither `IGNORE` nor `GLOBAL_TRIGGERS`
in `packages/infra/cli/src/commands/affected.ts`, so it lands in `unmatched` and
forces a **full `main.yml` run** — including this PR's own. That gap arrived with
`#838` (the directory's first file), not here. The fix is one `IGNORE` entry
`/^\.github\/prebuild-toolchain\//` next to the existing per-workflow carve-outs,
plus its spec case and a rebuild of the two committed CLI bundles — a change in
a different subsystem, so it is left as a follow-up rather than folded in.

JumpLink:
Adding the negative-control row now that it has landed (#845, run [30334533299](https://github.com/gjsify/gjsify/actions/runs/30334533299)): with `lightningcss-native` as the only change, **ppc64 took 37m56s** against 41m38s for the full build. That is the point — the gate buys nothing when the expensive package genuinely changed, and it should not. What comes off is only the six Vala bridges (~3.5 min emulated) plus the two skipped Rust bridges.

JumpLink:
## Merged main (through #847 / #848 / `1ec2f6742`)

`main` did **not** touch `prebuilds.yml` or `.github/prebuild-toolchain/`, so the
only conflict was root `package.json#scripts.test:e2e` — both sides appended a
suite. Resolved by taking main's line and re-inserting `prebuild-change-gate`
next to main's new `toolchain-bundle-napi-free`.

### #847 needed a real fix, not just a merge

The registry refactor left `check-refs-pin.mjs` and
`check-prebuild-loader-path.mjs` as **thin CLI entry points** (51 and 55 lines)
over `scripts/manifest-conformance/rules/*`. This PR's shared-input list named
only those wrappers — so after #847, editing a *rule* would change what every
build **verifies** while nothing rebuilt. That is precisely the silent-staleness
shape the gate exists to prevent, so `scripts/manifest-conformance/**` is now a
shared input in both `on: paths:` lists and in `SHARED_SCRIPTS`.

The `refs/` derivation itself is unaffected — #847's exported
`linkedRefsSubmodules` is byte-identical to the six lines `changed-packages.mjs`
carries (same regex, same `src/rust/Cargo.toml` path). It stays reimplemented
rather than imported *on purpose*: the `changes` job runs with nothing but
`actions/checkout` and the runner's node, and gating must not acquire a module
graph that can fail to load there (`refs-pin.mjs` pulls
`packages/infra/manifest-conformance/lib/index.mjs`, which re-exports ~10
modules). The duplication is now held to the original by a test that imports the
real rule and asserts both agree for every native package.

### #848: nothing to do, and nothing resurrected

The classifier never read `gjsify.platformsUncommitted`. Confirmed the removal
survived the merge (`grep -c` → 0 on every bridge).

### Re-verification surfaced a real hole in the gate's own self-check

Checking that the "gate and trigger must agree" assertion still fires, it
**didn't**. The `paths:` parser UNIONED the push and pull_request lists, so
deleting `refs/oxc` from *one* of them left the union intact — assertion green,
while a pin bump could no longer start the workflow on that event.
`prebuilds.yml` states in a comment that the two lists must be identical; nothing
enforced it.

Fixed: the parser returns **one array per list** and the self-check holds each
separately. Same deletion now yields

```
::error::prebuilds.yml: `on: paths:` list #2 differs from list #1 — the two MUST be
identical (a path that can change a prebuild on main has to be able to prove itself
on a PR). Only in #2: refs/oxc.
::error::oxfmt-native: links refs/oxc by Cargo path dependency …, but `refs/oxc` is
missing from prebuilds.yml `on: paths:` list(s) #1 — bumping that submodule pin
changes the binary and would not even run this workflow on that event.
```

with exit 1. Locked in as an e2e case that drives the real script against a
mutated copy of the workflow.

### Re-verification results

| check | result |
| --- | --- |
| `node scripts/audit-runtimes.mjs --check` (restructured) | **pass**, 51 committed prebuild dirs across 11 packages verified |
| `--platforms` matrix, gated vs ungated `prebuilds.yml` | **byte-identical** — `platforms-ci` attribution unaffected |
| package-set derivation | still exactly the 10 the workflow owns; self-check green |
| `tests/e2e/prebuild-change-gate/` | **15/15**, 1.3 s (was 13) |
| gate scenarios (Vala-only / lightningcss / shared input) | unchanged: 1, 1, 10 packages |
| new shared input `scripts/manifest-conformance/rules/refs-pin.mjs` | rebuilds all 10 |
| `commit-prebuilds` guard simulation (real step bodies, 10-pkg pathspec) | unchanged: 0 committed prebuilds destroyed in any case; guard fires on both hostile ones |
| repo-wide `gjsify lint` | 0 errors |

### Wall clock

**Carried over, not re-measured** — the merge changes no build step, only which
steps run. The gate decisions were re-verified above and are identical, and both
`main`-side changes are outside the build legs (#847 = scripts, #848 =
manifests).

| leg | before | after (Vala-only) | Δ |
| --- | --- | --- | --- |
| x64 (native) | 8m24s | 51s | 9.9× |
| arm64 (native) | 7m06s | 1m03s | 6.8× |
| ppc64 [QEMU] | 41m38s | 5m30s | 7.6× |
| s390x [QEMU] | 38m54s | 4m48s | 8.1× |
| riscv64 [QEMU] | 96m47s | 22m23s | 4.3× |

Re-measured earlier in this PR: the lightningcss-only control (ppc64 **37m56s**
vs 41m38s full) and the shared-input case (x64 8m40s / arm64 7m50s = baseline).

### Note on the merge commit

Committed with `SKIP_GJSIFY_HOOKS=1`. The merge carries main's **already
rebuilt** `dist/cli.gjs.mjs` byte-for-byte (`sha256 80de1b64…` on both sides) and
this branch touches **no** bundle input — nothing under
`packages/infra/{cli,resolve-npm,rolldown-plugin-gjsify}`. The hook's rebuild is
therefore provably a no-op, and this worktree has no built `lib/`, so attempting
it produces a *wrong* artifact instead of the correct one already committed.
`verify-committed-bundles.mjs` in CI is the real gate and rebuilds from source.

JumpLink:
## Merged main (through #841 / #850) + fixed the `E2E 4/4` EACCES

### The failure — cause, not symptom

`emit()`'s `--format=github-actions` branch appends to `$GITHUB_OUTPUT`. When the
e2e suite spawned the classifier it **inherited the runner's real output file**
(`/__w/_temp/_runner_file_commands/set_output_…`), which the container's test
user cannot write. Locally the variable is unset, so the branch is never taken —
the case could only ever fail in CI, and did.

One correction for the record: the note said two cases were affected. The job log
shows **exactly one** — `emits a skip list…`. `FAILS when the two 'on: paths:'
lists diverge` **passed**; it spawns with `--format=json`, so it never reaches
that branch. I routed it through the new helper anyway, because a case asserting
*"the classifier FAILED"* must not be satisfiable by an unrelated EACCES.

### Test side — owned output file, and now actually asserted

Every spawn goes through a `runScript()` helper that points `GITHUB_OUTPUT` at a
file the suite owns (per-call, cleaned in `after()`). That is strictly better
than neutralising the variable, because the file becomes **assertable** — the
github-actions format was previously verified only through stdout, so a dropped
append would have passed.

The case (renamed `PUBLISHES the decision to $GITHUB_OUTPUT, not just stdout`)
now checks the published file:

- all four keys present (`skip`/`build`/`reason`/`report`), nothing extra;
- `build` = `["terminal-native"]`, `skip` has the other nine;
- `report` parses and covers **all ten** packages, each with a `key` + a `why`
  (this is what `prebuilds-summary` renders);
- **every value is single-line** — a `key=value` file has no multi-line form
  without a heredoc delimiter, so an embedded newline would silently truncate
  what GitHub records;
- the keys are **quoted**, which is what makes the workflow's
  `!contains(needs.changes.outputs.skip, '"<key>"')` gates work and what stops
  `"http2-native"` matching `"http2-native-x"`;
- stdout still mirrors the file, so a failed publish stays diagnosable.

Cases that never needed the format were already on `--format=json` and stay
there.

### `emit()` now fails loudly — the decision you asked about

**Decided: fail.** Silently continuing publishes nothing, every gate then reads
`''`, `contains` is false, and all ten packages build. That is *safe* — but it is
a dead gate wearing a green tick, and nothing would ever report it. A hard
failure is still fail-open (the `changes` job's wrapper catches a non-zero exit,
writes the all-build fallback and `::warning::`s), it just refuses to be quiet.

Two details that make it diagnosable rather than a mystery:

- **stdout is written BEFORE the append**, so the log shows the decision that
  could not be published;
- the error names the path and the variable, and tells a non-CI caller to unset
  `GITHUB_OUTPUT` or use `--format=json`.

New case drives it with `GITHUB_OUTPUT` pointed at a **directory** — portable
across platforms and CI users, unlike `chmod` (root ignores mode bits).

**Proof the fix addresses the cause:** re-running the whole suite with an
unwritable `GITHUB_OUTPUT` inherited from the environment — the exact CI
condition — now passes 17/17. Before the fix, that same command failed the case.

### Merge

`main` did not touch `prebuilds.yml` or `.github/prebuild-toolchain/`. The only
conflict was root `package.json#test:e2e`, where #841 added
`napi-transparent-app-gjs` and this branch adds `prebuild-change-gate` — both
kept, plus the `test:e2e:prebuild-change-gate` alias.

**Checked explicitly, because it was this branch's own finding:**
`scripts/manifest-conformance/**` survived in **both** `on: paths:` lists (30
entries each, identical) and in `SHARED_SCRIPTS`.

### #852 (`Cargo.lock`) — already covered, now asserted

All three Rust bridges are listed in the trigger as `<pkg>/src/**`, so the
derived globs already match `src/rust/Cargo.lock`:

```
packages/infra/lightningcss-native/src/rust/Cargo.lock  → BUILD lightningcss-native
packages/infra/oxfmt-native/src/rust/Cargo.lock         → BUILD oxfmt-native
packages/infra/rolldown-native/src/rust/Cargo.lock      → BUILD rolldown-native
```

No change needed, but there is now a test case — narrowing any of those three
trigger entries would silently reopen exactly the registry-drift hole the
lockfile is being committed to close.

### Verification

| check | result |
| --- | --- |
| `tests/e2e/prebuild-change-gate/` | **17/17** (was 15) |
| same suite with an unwritable inherited `GITHUB_OUTPUT` | **17/17** — the CI condition, reproduced and fixed |
| `node scripts/audit-runtimes.mjs --check` | pass |
| classifier self-check on the merged tree | pass, 10 packages |
| repo-wide `gjsify lint` | 0 errors |

### Merge commit note

`SKIP_GJSIFY_HOOKS=1` again: the merge carries main's already-rebuilt
`dist/cli.gjs.mjs` byte-for-byte (`sha256 6d491169…`) and this branch touches no
bundle input, so the hook's rebuild is a no-op that this unbuilt worktree would
only corrupt. `verify-committed-bundles.mjs` in CI is the real gate.

JumpLink:
Follow-up: `Lint commit messages` failed on the merge commit, not on the code. commitlint's default ignore covers `Merge <x> into <y>` but not `Merge main; fix …`, so the subject was linted and reported `type-empty` + `subject-empty`. Amended the merge commit to a conventional subject (`fix(e2e): isolate the gate suite's GITHUB_OUTPUT`, parents unchanged, tree unchanged — `dist/cli.gjs.mjs` still `sha256 6d491169…`) and force-pushed. Commitlint is green on run [30344663514](https://github.com/gjsify/gjsify/actions/runs/30344663514); the E2E shard carrying `prebuild-change-gate` is in flight on run [30344661221](https://github.com/gjsify/gjsify/actions/runs/30344661221).

PR #845, opened 2026-07-28
[verification, do not merge] lightningcss-native only

JumpLink:
Verification run for #843. Base is #843's branch, so the workflow change itself is NOT in the diff — the only changed file is `packages/infra/lightningcss-native/src/rust/src/lib.rs`.

Expected: `lightningcss-native` builds on every leg (the expensive one, on purpose); the other nine packages are skipped.

Close after the wall-clock is recorded. Do not merge.

### DIFF
--- a/packages/infra/lightningcss-native/src/rust/src/lib.rs
+++ b/packages/infra/lightningcss-native/src/rust/src/lib.rs
@@ -1,5 +1,8 @@
 //! gjsify_lightningcss — single-call Rust→C FFI for lightningcss.
 //!
+//! (VERIFICATION-ONLY comment for PR #843's per-package prebuild gate — this
+//! branch touches nothing but this file. Do not merge.)
+//!
 //! Adapted from refs/lightningcss/c/src/lib.rs (lightningcss_c_bindings,
 //! Devon Govett, MIT). The upstream C bindings expose a 3-step pipeline
 //! (parse → transform → to_css) plus separate Targets/PseudoClasses/etc


JumpLink:
## Result — lightningcss-native only

Classifier decision (`Detect affected prebuild packages`, 25 s):

```
build=["lightningcss-native"]
skip =["webgl","oxfmt-native","rolldown-native","http-soup-bridge","http2-native",
       "sab-native","terminal-native","tls-native","webrtc-native"]
```

The x64 leg ran exactly these steps and nothing else — note the Rust toolchain
install turning itself back **on**, because a Rust bridge is in the set:

```
Install prerequisites
Install Rust toolchain (rustup, stable)
Build   @gjsify/lightningcss-native (Rust cdylib + Vala bridge)
Collect @gjsify/lightningcss-native prebuilds
Upload  @gjsify/lightningcss-native prebuilds artifact
```

| leg | full build ([30310941686](https://github.com/gjsify/gjsify/actions/runs/30310941686)) | this run |
| --- | --- | --- |
| x64 (native) | 8m24s | **2m59s** |
| arm64 (native) | 7m06s | **3m03s** |
| ppc64 [QEMU] | 41m38s | **37m56s** |

This is the negative control and it behaves like one: when the expensive package
IS the change, the emulated leg still pays for it — only the six Vala bridges
(~3.5 min emulated) and the two skipped Rust bridges come off. The gate buys
nothing here, and should not.

s390x / riscv64 were still compiling the same cdylib when this was closed; they
add no information beyond the ppc64 row.

Closing — verification only, never for merge. See #843.

PR #855, opened 2026-07-28
fix(ci): stop selective builds matching nothing

JumpLink:
## The bug

`gjsify affected --format=github-actions` emitted its CI filter **pre-quoted**:

```
include-args=--include '@gjsify/fs' --include '@gjsify/stream'
```

`main.yml` expands that value **unquoted** inside a `su testuser -c "… sh -c '…'"` nesting, and the result of a parameter expansion is never re-scanned for quoting. So the apostrophes stayed *inside* each glob, every `--include` pattern matched zero workspaces, and `gjsify foreach` printed one line and exited 0:

```
gjsify foreach: no workspaces match (script="build",
  include=["'@gjsify/integration-lightningcss'","'@gjsify/lightningcss-native'", …])
```

**On every selective PR run, CI built nothing.** PR #852's job `90231787853` shows the Build step at **0.65 s**; every later step — guards, tests, checks — then inspected whatever the build cache had restored. `main` stayed green only because a push to main takes the `FULL` branch (`gjsify run build`, no include args), which never hits this.

### 1. The transport — the boundary is at the producer

`include-args` is now a space-separated list of **already-shell-safe tokens** (`--include @gjsify/fs --include @gjsify/stream`), whose only supported consumption is an unquoted `$INCLUDE_ARGS` expansion, where word splitting is exactly the intent.

The boundary belongs at the producer because **this one output has five consumers**: the build / browser-bundle / check / test steps splice it into a command line, and thirteen `contains(needs.changes.outputs.include-args, …)` expressions substring-match it. Any format needing per-consumer unquoting would have to be repeated inside the very `su … -c "… sh -c '…'"` nesting that produced the bug — pre-quoting for an unknown number of nesting levels cannot work, while emitting tokens that need no quoting at any level can. main.yml's own comment already documented the contract as `--include @gjsify/x …`; the implementation had drifted from it.

The `contains()` gates are unaffected: they matched the quoted spelling too — which is precisely why they kept firing while the build did nothing. Pinned by a test.

Shell-safety is **machine-checked, not assumed**: `assertShellSafeWorkspaceName` rejects any name outside `[A-Za-z0-9@._/-]`. npm names cannot contain those characters, so it never fires in practice; if it ever does the command exits non-zero, the classify step's `continue-on-error` leaves the outputs unset, and CI falls back to a **full** run — failing towards "run everything".

### 2. The silence — this is the real defect

A build that did nothing looked exactly like a build that succeeded, which is why a one-character quoting bug survived on every selective run. `gjsify foreach` now **fails** when an `--include` pattern matches zero workspaces, and says why:

```
gjsify foreach: --include matched no workspace (2 of 2 pattern(s)):
  "'@gjsify/querystring'"
    → matches as "@gjsify/querystring": the QUOTES are part of the value. A shell expansion
      is not re-scanned for quoting, so a pre-quoted filter spliced through `sh -c` arrives with
      the quotes attached. Pass the bare name and let word splitting do the work.
  …
  238 workspace(s) discovered. Refusing to run: a filter that selects nothing would
  otherwise exit 0 and look exactly like a successful run over the whole set.
```

**Scope is "a PATTERN matched nothing", not "the selection is empty"** — the distinction is load-bearing. A final selection of zero is legitimate and stays exit-0 in three supported cases: `--shard N/4` over a small closure (the sharded `test` job depends on this), `--exclude` removing what the include kept, and script mode dropping workspaces that don't declare the script. All three narrow a *real* set to empty. A pattern naming something the tree does not contain is different in kind.

**Why the CLI and not a workflow assertion:** every `--include` in this repo matches non-zero — `@gjsify/example-*` (75), `@gjsify/example-node-cli-*` (26), `@gjsify/example-node-net-*` (8), `@gjsify/integration-*` (35), plus the classifier's exact names — so no legitimate empty include exists. Putting it in the CLI protects every caller (the root `test:examples` / `check:examples` scripts, all four workflow steps, any consumer repo) and is the only layer that can tell the two conditions apart. There is deliberately **no opt-out flag**: an escape hatch would re-arm the exact silence it closes. The one consumer that legitimately matches nothing — `tests/browser/scripts/build-bundles.mjs`, where a closure with no browser-capable package produces zero bundles — is a separate script and is unaffected.

## Verification

`main.yml`'s Build-step nesting reproduced verbatim (only `su testuser -c` → `sh -c`, which changes the uid, not shell parsing), driven with a real 38-workspace closure:

| | producer | `foreach` | result |
|---|---|---|---|
| before | pre-fix (committed bundle) | pre-fix | `no workspaces match` — **exit 0 in 1.1 s**, nothing built |
| after | fixed | fixed | builds the intended packages, names each one |
| zero-match | pre-fix (quoted) value | fixed | **exit 1**, names every bad pattern and the quoting cause |

Tests:

* `affected-classifier.spec.ts` — asserts the emitted value carries no quoting **and word-splits in a real POSIX shell into exactly the intended argv**. A string comparison cannot tell a token from a token-with-quotes-attached, which is the whole bug; the shell is the oracle. Plus: a global run still emits an empty value (the full-run signal), and the `contains()` gates still match.
* `tests/e2e/foreach-command` — zero match fails; quotes-in-value fails and names the cause; one bad pattern beside a good one fails; a matching glob still succeeds; narrowing to empty stays exit-0.

## Follow-on for the maintainer

With selective builds actually building, checks that have been passing may start failing — they will finally run against a genuinely built tree instead of a cache-restored one. That is the point, not a regression. The blast radius is bounded by the fact that **every push to main already runs the full path and main is green**, so the fixed selective path (a strict subset of the same work over the same tree) can only surface PR-specific regressions that were being masked — which is exactly what it is for.


JumpLink:
## Real-run evidence

This PR touches `packages/infra/cli/**` and `.github/workflows/main.yml`, both GLOBAL_TRIGGERS, so its own Build step takes the **FULL** branch and structurally cannot exercise the selective path. I measured that on a throwaway draft PR (#856, now closed + branch deleted) based on this branch, whose only change was a one-line touch in `packages/node/querystring`.

| | run | `include-args` | Build step | what `foreach` did |
|---|---|---|---|---|
| before | #852 job `90231787853` | `--include '@gjsify/integration-lightningcss' …` (quoted) | **0.65 s** | `no workspaces match` |
| after | #856 job `90251012576` | `--include @gjsify/browser-node-polyfills --include @gjsify/cli …` (unquoted) | **6 s** | `start @gjsify/cli` / `start @gjsify/querystring` |

```
[gjsify foreach] start @gjsify/cli (0/2 done, 1 in flight)
[gjsify foreach] start @gjsify/querystring (0/2 done, 2 in flight)
```

(38-workspace closure; `--no-private` drops the 35 integration packages and the two dep-only umbrella metas declare no `build`, so 2 remain — correct.)

## A pre-existing failure this makes visible — NOT caused by this PR

Both the before and the after run fail the **next** step, `Verify declared package outputs exist`, on the same package:

```
@gjsify/xmlhttprequest (packages/web/xmlhttprequest) declares 3 path(s) that do not exist:
    types → lib/types/index.d.ts
    exports["."]["types"] → ./lib/types/index.d.ts
    exports["./register"]["types"] → ./lib/types/register.d.ts
```

It reproduces on #852's **pre-fix** run (Build 0.65 s, then the identical error), so it is independent of this change. It is a build-cache **completeness** problem specific to the selective path: `@gjsify/xmlhttprequest` is outside the closure, so it is never rebuilt, and the restored `build-v3-…` archive does not carry its `lib/types/**`. The FULL path is unaffected — this PR's own build regenerates everything and the step passes (`10:47:12 → 10:47:13`, success).

Worth a separate issue: today **every** selective PR reds at that step, which is a second reason nobody could see the build doing nothing.

PR #858, opened 2026-07-28
fix: run the committed-bundle verification on a cold tree

JumpLink:
## The defect

`scripts/verify-committed-bundles.mjs` could not run on a freshly cloned,
freshly installed checkout. Every recipe died at its first step:

```
gjsify build: no usable bundler engine under GJS — `@gjsify/rolldown-native` is not loadable …
ERROR: @gjsify/cli: `gjsify workspace @gjsify/cli build --with-dependencies` failed (exit 1).
```

Under GJS the only bundler engine is `@gjsify/rolldown-native`, whose JS facade
(`packages/infra/{rolldown,lightningcss}-native/lib/`) is a **build output**
that `gjsify install` does not produce. `scripts/bootstrap-native-facades.mjs`
produces it and is idempotent — nothing called it from the verify path. In CI
the gap is invisible, because `gjsify run build` → `build:infra` runs the
bootstrap long before the verify step.

**Why this blocks releases rather than being cosmetic:** since #835 the
release-it `after:bump` hook runs
`node scripts/verify-committed-bundles.mjs --rebuild`. Moving releases into CI
— the direction this project is going, after a local release shipped three
regressed bundles — means running that on a cold tree *by definition*.

## What changed

1. **`verify-committed-bundles.mjs` bootstraps the tree it needs**
   (`ensureBuildableWorkspace()`), rather than documenting a precondition that
   gets forgotten exactly when it matters. Two entry points, because the
   bootstrap has a precondition of its own — it drives the CLI's Node entry
   (`packages/infra/cli/lib/index.js`, the one emitter that needs no bundler),
   itself a build output:
   * Node entry present (warm; the CI ordering) → run the bootstrap directly.
     It is mtime-idempotent: a warm run pays a few stat sweeps.
   * Node entry absent (cold clone) → run root `build:infra`, the documented
     tsc chain that produces that entry *and* ends in the same bootstrap.
     `deploy-docs.yml` already runs `build:infra` on an install-only tree for
     exactly this reason, so the path is proven; re-spelling its prefix here
     would be a second copy of an existing recipe.

   It runs **inside the `try`**, so a preflight failure still unwinds through
   the artifact restore, and it names the exact command it ran on failure.

2. **The engine's own failure message no longer guesses.** It used to assert
   the rarest of three causes as "most likely" — *"the CLI bundle was invoked
   DIRECTLY … bypasses the `gjsify` launcher"* — which on a cold tree sends the
   reader after an env-var problem that does not exist (the launcher *was* used
   correctly). `diagnoseNativeEngine()` now reads the answer off disk
   (`existsSync`-only, so it cannot throw or recurse into the failure it
   explains) and each branch names the exact command: **not installed** /
   **facade not built** (`gjsify run build:infra`,
   `node scripts/bootstrap-native-facades.mjs`) / **typelib-lookup**, which is
   the launcher case the old text always claimed.

3. **The install hard-exit message stops asserting a wedge** — same defect
   class, and the evidence contradicts the claim (see below). It now states
   what the timer actually observed, names both shapes it cannot distinguish,
   and points at the one action that does distinguish them: re-running, which
   is safe and resumes. No behaviour change; **no timeout change**.

## Cold-tree verification

Scratch clone at `/home/jumplink/gjsify-cold-verify`, installed with
`gjsify install --immutable`, then `git clean -xfd -e node_modules` back to
cold before every run (probes printed in each log). Exit codes captured
directly, never through a pipe.

| tree | commit | mode | result |
|---|---|---|---|
| cold | `c700af4c2` (base) | compare | **exit 1** — `no usable bundler engine under GJS`, 2 s |
| cold | `c700af4c2` (base) | `--rebuild` | **exit 1** — identical failure, 2 s |
| cold | this branch | compare | **exit 0** — all 111 artifacts reproduce byte-identically, 6 m 22 s |
| cold | this branch | `--rebuild` | **exit 0** — all artifacts produced, 7 m 12 s; `git status` clean afterwards |
| warm | this branch | compare | **exit 0** — takes the cheap branch (`bootstrap … up to date — skipping`), 4 m 47 s |

Before (base commit, cold):

```
[verify-bundles] @gjsify/cli: gjsify workspace @gjsify/cli build --with-dependencies
gjsify build: no usable bundler engine under GJS — `@gjsify/rolldown-native` is not loadable, …
Most likely cause: the CLI bundle was invoked DIRECTLY (`gjs -m …/cli.gjs.mjs` …
ERROR: @gjsify/cli: `gjsify workspace @gjsify/cli build --with-dependencies` failed (exit 1).
EXIT=1
```

After (this branch, cold, both modes):

```
[verify-bundles] preflight (cold tree — no packages/infra/cli/lib/index.js): gjsify run build:infra
…
  ✓ packages/infra/cli/dist/cli.gjs.mjs (6570173 B) reproduces from source
  ✓ packages/infra/cli/dist/affected.gjs.mjs (263871 B) reproduces from source
  ✓ packages/infra/tsc/dist/tsc.gjs.mjs (3581066 B) reproduces from source
  ✓ packages/infra/tsc/lib/: 108/108 reproduce from source
All committed bundles reproduce byte-identically from the source at HEAD.
EXIT=0
```

The compare run is a stronger result than it looks: the bundle it reproduced
byte-for-byte was built in a *different directory* by a different checkout, so
it also re-confirms the "bundles bake no build-host state" property the whole
check rests on.

## The install stall — findings

Reproduced on two genuinely fresh trees (fresh clone, no `node_modules`;
tarball cache warm; linux-x64, 20 cores).

**It is not a deadlock — it is ~4× more work than npm would do, and the budget
is what hits.**

* A cold install materialises **1597 packages / 4.78 GB**. **183 packages /
  3.36 GB — 70% of the bytes — declare an `os`/`cpu` that EXCLUDES the host**,
  so npm/yarn/pnpm would never place them: six `@anthropic-ai/claude-agent-sdk-*`
  siblings at ~230 MB each, six `@pagefind/*` at ~55 MB, plus the
  `@rolldown`/`@oxlint`/`@oxfmt`/`@img/sharp`/`@deltachat` binding sets.
* Those six 230 MB tarballs took **~22–45 s each** to gunzip + parse + write
  under GJS. That is the entire "9/1597 after N minutes" shape agents keep
  reporting — the counter *is* moving, it is just crawling through the biggest
  packages first (the queue is sorted by depth, then path, and `@anthropic-ai`
  sorts near the front).
* Run 1 reached **1372/1597 in 34 min** while still making steady progress. A
  re-run finished the remainder in **120 s** off the `isAlreadyExtracted` skip
  path — which is exactly why "a second run completed".
* **Contention is real but secondary.** With two installs on the box, the
  second sat at 7–9/1597 for minutes while the first — past the big packages —
  ran at ~1.5 pkg/s. Same absolute per-package cost, 20 idle cores; they
  contend on memory bandwidth and page cache, not CPU. Reports of "9/1597 in
  35 min under load" and "17/1597 in 5 min" are both this shape.
* **The hard-exit backstop can fire with nothing wedged.** Extraction has
  multi-second *synchronous* stretches (tar parse + `writeFileSync` per entry)
  and 16 workers run concurrently, so the clean abort path can easily miss its
  10 s grace window while the install is healthy.
* **Could not determine**, honestly: whether a genuine dropped-Gio-event wedge
  *also* exists. Nothing in these runs required that hypothesis, and
  `tests/e2e/install-concurrent` already records that the historical hang was
  never reproduced deterministically. (Both of my runs were ultimately killed
  by an external `SIGTERM` from the agent harness, not by the install — exit
  143, both processes at the same instant — so neither run is evidence of a
  hang.)

**Not fixed here, tracked in STATUS.md → Toolchain hygiene follow-ups.** The
real fix is to honour `os`/`cpu` like every other package manager, and it is a
two-part change because `--immutable` materialises straight from
`gjsify-lock.json`: record `os`/`cpu`/`libc` on lock entries at resolve time
(npm's `package-lock.json` carries them for exactly this reason) and filter at
materialisation. That is a lockfile-format change plus a full lockfile
regeneration, so it wants its own PR and e2e. The napi-rs entry-replacement in
`napi-node-addon.ts` already selects its sibling *by host triple*, so it is
unaffected.

**Explicitly not done:** raising `--timeout`. A budget that exists to bound a
hang must not be tuned to accommodate one.


JumpLink:
Rebased onto `c185bbaf4`. One conflict, in `packages/infra/cli/dist/cli.gjs.mjs` — a derived artifact, so it was resolved by taking the new base's bytes and **rebuilding from source at that commit** (`build` → `build:gjs-bundle` → `build:affected-bundle`) rather than by merging minified text. The rebase does not run the pre-commit hook, so this is the step that would otherwise be silently skipped.

Local green on the rebased head:

* `node scripts/verify-committed-bundles.mjs` → **exit 0**, all four groups reproduce byte-identically (`cli.gjs.mjs` 6 572 148 B, `affected.gjs.mjs` 245 617 B, `tsc.gjs.mjs` 3 581 066 B, 108/108 `lib*.d.ts`). This also proves the two intermediate commits are self-consistent: HEAD touches no CLI source, so a byte-match at HEAD is a byte-match at the install-message commit too.
* `gjsify workspace @gjsify/cli test` → **exit 0**, 1229 tests.
* oxfmt/oxlint clean on the changed files.

Note on #855: the local evidence in the PR body was never produced by a CI selective build — every run was a cold `build:infra` + the verify recipes in a scratch clone — so it is unaffected by "selective builds built nothing".

---

### Re #862: should the warm-cache probe and `ensureBuildableWorkspace()` share a definition?

**No — and the reason is worth writing down, so I added it to AGENTS.md next to the verify bullet.**

They ask different questions and differ in *kind*:

|  | `main.yml` warm-cache probe (#862) | `ensureBuildableWorkspace()` |
|---|---|---|
| question | is this restored cache a **complete baseline** a *selective* build may trust? | can this workspace **run a bundler at all**? |
| scope | the whole tree, every declared output | two tools: `packages/infra/cli/lib/index.js`, the two `*-native` facades |
| kind | read-only; **decides** a build path | **repairs** — runs the bootstrap |
| on failure | fall back to a full build | fall back to `build:infra` |

The relation is **strict containment**, and I verified it rather than assuming it: every path my preflight needs is a *declared output* that `scripts/verify-package-outputs.mjs` already asserts — `@gjsify/cli`'s `main`/`module`/`bin` → `lib/index.js` (checked by moving it aside: the probe fails naming all three declarations), and `@gjsify/rolldown-native`'s `main`/`exports` → `lib/esm/index.js` + `lib/types/index.d.ts`. So **probe warm ⇒ my preflight no-ops** (it takes the mtime-idempotent bootstrap branch and skips everything in ~0.1 s). No overlapping work, no contradiction: in CI the probe runs first and either way the tree is complete before the verify step.

The converse does not hold, and must not: the tree that motivated #862 — `@gjsify/xmlhttprequest` with `lib/esm` but no `lib/types` — is *cold* for the probe yet perfectly *buildable*. Unifying the two would break in both directions:

* the probe's predicate as my precondition ⇒ the release hook runs a full `build:infra` because some unrelated package is missing a `lib/types` that cannot affect a bundle rebuild;
* my two-file predicate as the CI probe ⇒ reinstates exactly the "a few paths exist ⇒ assume the tree is complete" fallacy #862 removed.

What they *do* share, and should keep sharing, is the **policy**: an incomplete tree means "do the fuller thing", never "error".

PR #863, opened 2026-07-28
fix(ci): evict the poisoned build cache (v3 → v4)

JumpLink:
The missing half of #862 — **not** a shortcut around it.

#862 fixed the **consumer**: an incomplete archive now forces a full build instead of a silently-selective one. It cannot fix the **producer**, because `actions/cache` saves only when the PRIMARY key MISSED.

## An entry under a live key is immortal

Once a bad archive exists under a key runs still compute, every later run exact-hits it, restores it and skips the save — the tree gets repaired in-job forever while the archive is never replaced. **Bumping the key is the only eviction mechanism the action offers.**

Measured on #862's own build job, *after* the probe landed:

```
Cache hit for: build-v3-Linux-fedora44-7b9e4e8e…-1cb5eb2cd…
Cache Size: ~4 MB (4573502 B)
[ci] build cache cold or INCOMPLETE — forcing full build/check
[ci] full build
Post Run ./.github/actions/gjsify-setup … 2s   ← nothing written
```

That 4 MB archive holds `@gjsify/xmlhttprequest`'s `lib/esm` but not its `lib/types`. main's own key — and every PR touching no `src/**` or manifest — restores it and pays a full build **forever**. Runs whose key *differs* do self-heal (miss → full build → save a complete archive that later prefix lookups prefer), so only the exact live key is stuck. A wider probe cannot reach it; a version bump can.

I declined this in #862 for a reason that still holds — *it fixes the instance, not the class* — which is exactly why it lands **with** the probe rather than instead of it. The probe makes any future incomplete archive harmless; this removes the one that already exists.

## The restore-keys trap

A surviving `build-v3-` prefix tier would restore the very archive being evicted and the bump would achieve **nothing**. So v4 has **no v3/v2 fallback**:

```yaml
key: build-v4-…-<manifests>-<src>
restore-keys: |
  build-v4-…-<manifests>-
  build-v4-…-
```

The v2→v3 bump kept an old tier deliberately (*"so the first v3 run isn't a cold full build"*) — a sensible nicety when that bump was a schema change and the old archives were fine. Here the old archives **are** the thing being thrown away, so the one-time cold build is the price of eviction, and it is the point rather than a side effect.

## Cost — stated up front

Measured on #862's build job, which already full-builds today, so these **are** the post-bump numbers for main's key:

| step | duration |
|---|---|
| Build | 6m07s |
| Verify committed bundles | 5m04s |
| Build examples | 6m07s |
| Check examples | 3m33s |
| **build job total** | **22m35s** |
| **whole run** | **32m45s** |

**One Fedora leg** today (`fedora-matrix` is a single entry — not two), and **downstream jobs are unaffected**: they `needs: build` and exact-hit the archive the build job saves in the same run.

So the bump costs **one extra full build per distinct key, once** — roughly **+10-15 min wall clock on first contact** (vs a warm selective build job of ~10-12 min). For main's current key it is actually **cheaper than today**: that run already full-builds and merely cannot save, whereas after the bump it saves, and the key can go selective again.

## Also

A sentence in AGENTS.md's caches bullet recording the general fact this repo just learned the hard way — an entry under a live key is immortal, bumping is the only retirement, and the older `restore-keys` tiers must be dropped in the same edit.


JumpLink:
Closing this — **my premise was wrong**, and the change makes CI more fragile rather than less.

## The 4 MB "corpse" was not a corpse

`gh cache list` shows **every** build-output archive this repo has ever saved is ~4.5 MB:

```
14:17:33  4572010  build-v3-…7b9e4e8e…
13:16:19  4572570  build-v3-…7b9e4e8e…
13:08:51  4573502  build-v3-…7b9e4e8e…
12:30:49  4573316  build-v3-…7b9e4e8e…
11:20:12  4578061  build-v3-…073606a9…
10:42:28  4576920  build-v3-…bffadf7e…
09:31:25  4573754  build-v3-…81e9f7ba…
09:09:09  4571116  build-v3-…8c1619b2…
```

Across many different key hashes and hours, all ~4.57 MB. That is the **steady-state size** of this cache, not the signature of a poisoned entry — the committed multi-MB bundles are excluded by `!` and most `dist/` content is only produced by `foreach test`, so a zstd'd `lib/` tree really is a few MB. I read one number as damage without checking the distribution. The v4 archive that did get created is 3.3 MB — if anything *smaller*.

So "evict the poison" had nothing to evict, and the eviction argument I made — and that you reasonably accepted — does not survive the data.

## It also actively broke the run

Four concurrent `pull_request` runs existed for the same commit `792f95926`. They all compute the identical brand-new v4 key, so they raced to reserve it:

```
Failed to save: Unable to reserve cache with key build-v4-…,
another job may be creating this cache.
```

Three runs went green; the loser's build job could not save — and **because I removed every fallback `restore-keys` tier, its downstream jobs restored nothing at all**. Check, all four Test shards, all four E2E shards, Cross-runtime, Examples and Browser failed together, each at its first real step (`Check` died in 25 s). Under v3 the loser would have prefix-restored *something* and survived.

The fallback tiers I deleted as "the trap" are in fact the safety net that makes a reserve race survivable. Removing them converts a benign, transient collision into a total-red run.

## What stands

#862's completeness probe is the correct and sufficient fix. Whatever leaves a restored tree short of `@gjsify/xmlhttprequest`'s `lib/types`, the probe detects it and forces a full build, so the job is *correct*. That is real and merged.

What remains genuinely open is the cause: why does a restored archive lack `lib/types` when the full build immediately preceding the save demonstrably produces it (`Verify declared package outputs exist` passes in the same job)? That is worth its own investigation, and the answer may well be cheap — but it is not a cache-key problem, and bumping the key would not have touched it.

Cost of the mistake: one wasted CI round. Nothing merged.

PR #893, opened 2026-08-01
feat(cli): skip foreign-platform optional deps

JumpLink:
Two commits, 15 files:

- **`feat(cli)`** — the install backend skips optional dependencies declared for other platforms, instead of materialising every platform package. Ships an e2e suite (`tests/e2e/install-platform-filter/`) and a cli-reference doc update.
- **`fix(fs)`** — removes the unquoted `GLib.spawn_command_line_sync(\`ln ${a} ${b}\`)` shell-outs in `@gjsify/fs`. That is the **standing counter-example in AGENTS.md** ("fix at next touch, never copy"): a path containing a space or a shell metacharacter breaks or injects.

## Before merging, please compare with a second attempt

An independent branch `perf/install-skip-foreign-platform-packages` (2026-07-28, 2 commits) targets the same feature and touches **eight of the same files**, including the same `tests/e2e/install-platform-filter/run.mjs`. Neither branch contains the other — they are separate takes, not a rebase.

This one is the newer (2026-07-31) and the smaller diff. I have not judged which is better; that needs the author's context. The other branch is pushed and preserved, so nothing is lost either way.

---

*Provenance: found in an agent worktree while reclaiming disk. Verified clean and not in `main`. Opened so the work is visible rather than stranded — not my authorship, please review on its merits.*


JumpLink:
**Assessment after comparing all three takes — none of them is superfluous, and none subsumes the others.** So I have not closed anything; that would lose work.

| branch | core filter | libc axis | lockfile metadata | fs shell-out fix |
|---|---|---|---|---|
| `wip/platform-filter-worktree-state` | 548 lines | 11 files | — | — |
| `perf/install-skip-foreign-platform-packages` | 349 | 6 | **yes** | — |
| `feat/install-platform-filter` (this PR) | 189 | 7 | — | **yes** |

The third one is new to this list: it was uncommitted in a working tree and I snapshotted it to `wip/platform-filter-worktree-state` so a reboot could not lose it. It is the furthest along on the filter itself and on the libc axis, but it carries no lockfile platform metadata — which only `perf/` has — and it does not contain the fs fix.

Consolidating three partial takes into one is real work, not a choice between them. It is also the ADR-0017 precondition, so it deserves doing deliberately rather than by picking whichever branch is nearest.

**What could be resolved cleanly right now:** the `fix(fs)` commit here has nothing to do with the platform filter. It is extracted, verified and open as #899. Once that lands, this PR is purely the filter and the three-way consolidation is the only remaining question.

JumpLink:
Superseded by #897 — and worth recording *why*, because the difference is not cosmetic.

**This branch's libc check cannot fire.** It reads `v.libc` from the packument and gates
on `target.libc && env.libc ? checkList(...) : true` (`platform-check.ts`), but it never
escalates to the full document. The registry does not serve `libc` in the abbreviated
("corgi") body — measured against the live registry:

```
lightningcss-linux-x64-musl@1.33.0
  abbreviated → {"os":["linux"],"cpu":["x64"]}
  full        → {"os":["linux"],"cpu":["x64"],"libc":["musl"]}
```

So `target.libc` is always `undefined` and the branch is structurally present but
practically dead. `perf/install-skip-foreign-platform-packages` diagnosed exactly that
and solved it; #897 keeps that insight.

For the record, the numbers this work is worth, measured with npm's own `checkPlatform`
over all 1597 lockfile entries on `linux-x64`: **183 packages / 3361 MB** excluded by
`os`/`cpu`, **9 more / 308 MB** by `libc` alone.

#897 also had to go further than either branch: our OWN packages fall through the
"escalate only where it can matter" predicate. `@gjsify/sab-native` and
`@gjsify/lightningcss-native` declare `libc: ["glibc"]` with **no** `os`/`cpu`, so a
trigger keyed on those fields never fires for them and a musl host would be handed a
glibc-only prebuild — failing at `dlopen` instead of at install time, on precisely the
platform the libc axis exists for. #897 therefore reads one authoritative document per
package, which measured at 1.26x the abbreviated one: ~10% more metadata bytes and ~11%
**fewer** round-trips than the two-document shape.

No action needed here beyond closing it whenever convenient.

JumpLink:
**Superseded by #897 — closing.** This was my branch (`pf/gates`), so nothing of anyone else's is being discarded, and both halves of it are accounted for.

**The `fix(fs)` half landed separately as #899.** It never belonged here: the unquoted `GLib.spawn_command_line_sync(\`ln …\`)` shell-outs have nothing to do with the platform filter, and bundling them meant an independently-verifiable fix waited on an unresolved feature. Merged with the three defects it actually carried (space-broken argv, metacharacter injection, unchecked exit status) and five tests covering exactly those.

**The filter half is fully covered by #897.** When I compared these earlier, none of the three takes subsumed the others and I said so. #897 has since moved past that:

| | #893 (this) | #897 |
|---|---|---|
| `platform-check.ts` | 137 lines | **463** |
| `platform-check.spec.ts` | 178 | **340** |
| `tests/e2e/install-platform-filter/run.mjs` | 348 | **447** |
| libc axis | partial | full, plus the prebuild-libc conformance rule |

It is a rewrite rather than an extension — none of my exported names survive verbatim (`detectLibcFamily` → `detectLibc`, `platformMatches` → `checkPlatform`/`declaresPlatform`) — but every capability is present and the surface is considerably larger. #897 is also out of draft with no failing checks, while this branch is 15 commits behind and would need the same work done twice.

The third take, `wip/platform-filter-worktree-state`, was a rescue snapshot of uncommitted work; it is also covered by #897 and has already been deleted.

This is the ADR-0017 precondition, so it matters that it lands *once*, well — not that it lands from my branch.

PR #894, opened 2026-08-01
build(oxfmt-native): declare the refs/oxc lockstep

JumpLink:
AGENTS.md documents this pin as active:

> **version skew vs the npm engine** — … Declared as `gjsify.refsLockstep: {…}`, checked against the installed npm version (**same for `oxfmt-native` ↔ `refs/oxc` @ `oxfmt_v<X>`**)

But `packages/infra/oxfmt-native/package.json` never carried the declaration. `scripts/manifest-conformance/rules/refs-pin.mjs` reads `gjsify.refsLockstep` (line 81) and gets an empty table for this package — so it checks **nothing**, silently.

That is worse than having no rule: the docs assert the skew between the committed `refs/oxc` prebuild source and the npm `oxfmt` devDep is caught, while nothing catches it. The identical guarantee for `@gjsify/rolldown-native` is real; this one was a claim with nothing behind it. AGENTS.md itself names the cost of exactly this skew — the rolldown-native prebuild that had drifted behind its pin and emitted two chunks where npm emitted one.

## Scope

**Declaration only.** The `refs/oxc` submodule pointer is deliberately *not* bumped here: per AGENTS.md a refs bump invalidates the committed `Cargo.lock` and requires its regeneration plus a prebuild rebuild in the same commit. That belongs in its own change by someone who can run the rebuild.

## Verification

The shape mirrors the working `@gjsify/rolldown-native` declaration exactly, with the tag template following oxfmt's own scheme (`oxfmt_v{version}`) as AGENTS.md specifies:

| package | refs | npm | tag |
|---|---|---|---|
| rolldown-native | `refs/rolldown` | `rolldown` | `v{version}` |
| oxfmt-native *(this PR)* | `refs/oxc` | `oxfmt` | `oxfmt_v{version}` |

I did **not** execute `refs-pin` end-to-end — it needs initialised submodules, which is also why `audit-runtimes --check` registers but does not select it. Verified by shape comparison against the working declaration and by reading the consuming rule.

---

*Provenance: recovered from uncommitted work in a stale agent worktree that was being removed to reclaim disk. I preserved the diff rather than discard it, and split out the part that stands on its own. Not my authorship of the idea — please review on its merits.*


JumpLink:
**Reworked** — the declaration alone would have been a promise with nothing behind it, so this now also makes it hold.

The pin was already broken when this PR was written: `refs/oxc` sat at `crates_v0.140.0-3-g582ea7547f` (14 Jul) while the npm `oxfmt` devDep is exactly `0.57.0` (tag dated 29 Jun). Nothing noticed, precisely because `gjsify.refsLockstep` was absent — `refs-pin.mjs` read an empty table and checked nothing while AGENTS.md asserted the guarantee was live. So the rule found real drift on its own first run.

`Cargo.lock` makes the size concrete: regenerating against the new pin **downgrades `oxfmt 0.59.0 → 0.57.0`**. The committed prebuild was built from 0.59.0 sources while `gjsify format` on Node runs 0.57.0 — two builds of the same formatter, two release lines apart. That is the "engines disagree on some inputs only" failure the rolldown-native incident documents.

```
refs/oxc  582ea7547f (crates_v0.140.0-3)  ->  5306f24d9e (oxfmt_v0.57.0)
```

I verified the target commit **is** the tag rather than trusting `git describe`, which reports `apps_v1.72.0` for it — oxc puts `apps_v1.72.0`, `oxfmt_v0.57.0` and `oxlint_v1.72.0` on one commit, so the described name is not the pin.

`Cargo.lock` is regenerated in the same commit, as a `refs/` bump requires (it is a `rust_sources` input and CI builds with `--locked`).

**One consequence to expect:** the committed `prebuilds/linux-x64` binary is now older than its pin. Self-correcting — `prebuilds.yml` carries `refs/oxc` in its paths filter, so this triggers a rebuild and `commit-prebuilds` updates the artifact on main after merge.

Verified: `node scripts/check-refs-pin.mjs packages/infra/oxfmt-native` exits 0 (it exited 1 before, naming the skew), and `audit-runtimes --check --strict` is green.

JumpLink:
**Corrected direction** — the previous push resolved the drift by pinning `refs/oxc` *back* to match an older npm `oxfmt`. That was the wrong way round: it would have frozen the build two releases behind instead of closing the gap. Both engines now go **forward** to the newest release that exists on npm *and* as an upstream tag.

```
refs/oxc          582ea7547f (crates_v0.140.0-3)  ->  65fe65d842 (oxfmt_v0.61.0)
npm oxfmt devDep  ^0.57.0                         ->  ^0.61.0
```

**Cost measured before choosing**, because AGENTS.md warns that pre-1.0 oxfmt churns formatting across minors and this crosses four of them:

| oxfmt | files with format differences |
|---|---|
| 0.57.0 (was pinned) | 388 |
| 0.61.0 (new) | 390 |

A delta of **exactly two files** — `webgl2-rendering-context.ts` and `composeVisitors.js` — and none fixed. The 388 are pre-existing and unrelated: CI gates `gjsify lint`, never `format --check`, so formatting drifts unchecked. Not reformatting here (two files out of 388 is noise); the missing gate is the actual issue and belongs in its own change.

Everything else from the earlier rework still applies: the pin was already broken when this PR was written, and nothing noticed because the declaration was absent — `refs-pin` read an empty table and verified nothing while AGENTS.md asserted the guarantee was live. `Cargo.lock` is regenerated in the same commit (oxc crates 0.140.0 → 0.142.0, all updates, no downgrades). The target commit was verified by SHA rather than `git describe`, which calls it `apps_v1.76.0` — oxc puts three tags on it.

AGENTS.md also stops restating the current version: that copy said `oxfmt_v0.54.0 ↔ oxfmt@0.54.0`, three releases stale, describing a check that was not wired up. It now points at the declaration and the rule that holds it.

`check-refs-pin.mjs` exits 0 (1 before, naming the skew); `audit-runtimes --check --strict` green.

PR #897, opened 2026-08-01
feat: platform-aware installs + the prebuild libc axis (WIP)

JumpLink:
> **DRAFT / WIP — nothing here has been type-checked, linted or tested yet.**
> Opened to preserve the work, not to be reviewed as finished. The two commits
> are honest about that in their own messages.

Implements the consumer side and the vocabulary groundwork for
[ADR 0017 — Distribution of platform-specific native builds](docs/adr/0017-native-package-distribution.md).

## What is here

**1. `feat(cli): filter installs by os/cpu/libc`** — the consumer side ADR 0017
step 4 requires to land and be proven *before* the split.

Measured on this workspace, using npm's own `checkPlatform` over all 1597
lockfile entries on `linux-x64`:

| | packages | bytes |
|---|---|---|
| installable on this host | 1405 | — |
| excluded by `os`/`cpu` | 183 | **3361 MB** |
| excluded by `libc` alone | 9 | **308 MB** |

Semantics are ported from npm's `npm-install-checks`, not invented: `checkList`
with its `any` / `!`-negation rules, the three-field check, and the
"declares `libc` but the host libc is unknown" case.

**`libc` requires the FULL packument.** The abbreviated document the registry
serves for `application/vnd.npm.install-v1+json` carries `os` and `cpu` but
never `libc` — verified against the live registry:

```
lightningcss-linux-x64-musl@1.33.0
  abbreviated → {"os":["linux"],"cpu":["x64"]}
  full        → {"os":["linux"],"cpu":["x64"],"libc":["musl"]}
```

npm pays for this with a blanket `fullMetadata: true` in arborist's
`#fetchManifest`, and after trying to be cleverer, so do we — one authoritative
document per package.

The first draft escalated only where the abbreviated entry declared `os` or `cpu`
(~190 of 1597 packages). **That predicate has a hole, and this repo's own packages
fall through it:** `@gjsify/sab-native` and `@gjsify/lightningcss-native` declare
`libc: ["glibc"]` with *no* `os`/`cpu`, so the trigger never fired for them and a
musl host would have been handed a glibc-only prebuild — failing at `dlopen`
instead of at install time, on exactly the platform the libc axis was added for.
The abbreviated body omits `libc` entirely, so it can never *prove* the absence of
a restriction, and every cheaper predicate has some version of that hole.

The cost turned out not to argue for cleverness either. Measured over 15
representative dependencies, largest packages dominating: the full document is
**1.26x** the abbreviated one, so reading it everywhere costs ~10% more metadata
bytes and ~11% **fewer** round-trips than the two-document shape it replaces.

Foreign-platform packages stay **recorded in the lockfile** and are skipped only
at download/extract, so a lockfile written on one host still describes every
platform (npm marks such nodes `inert` for the same reason).

**2. `feat(prebuilds): add the libc axis to targets)`** — the target vocabulary
becomes `<os>-<arch>[-musl]`, and the `libc` manifest field is verified against
the binary instead of hand-maintained. `binary.mjs` gains hand-written ELF
readers for `DT_NEEDED` and `.gnu.version_r`, keeping the property that a Linux
host can verify a darwin artifact without `readelf`/`objdump`. A new portable
`prebuild-libc` rule fails a declaration that disagrees with the ELF.

Why the axis is not cosmetic, measured across the committed `linux-x64`
prebuilds:

| prebuild | `libc.so.6` | max `GLIBC_` |
|---|---|---|
| `tls-native`, `webrtc-native` | **no** | — |
| `terminal-native` | yes | 2.2.5 |
| `webgl`, `http2-native`, `http-soup-bridge` | yes | 2.14 |
| `sab-native` | yes | 2.28 |
| `oxfmt-native`, `rolldown-native`, `napi` | yes | 2.34 |
| `lightningcss-native` | yes | **2.39** |

Two bridges link no libc at all and run on either; `lightningcss-native` alone
pins the floor to Ubuntu 24.04 / Debian 13. `gjsify.glibcRequires` turns that
floor into a gate rather than an accident.

`darwin-x64` is declared with a `platformsUncommitted` reason — every macOS job
in this repo runs on `macos-latest` (arm64), so the target is locally buildable
and awaits a self-hosted runner.

## Relationship to the existing branches

Two earlier branches attempt the installer half. This supersedes both, and the
reason is worth recording:

- `feat/install-platform-filter` (31 Jul, the branch ADR 0017 names) reads
  `v.libc` and checks it — but never escalates to the full packument, so
  `target.libc` is always `undefined` and **the libc branch is structurally
  present and practically dead code**.
- `perf/install-skip-foreign-platform-packages` (28 Jul) diagnosed exactly that
  and solved it with a `backfillLibcFields()` pass. That insight is what this PR
  keeps.

The deliberate divergence from npm on an *undetectable* host libc — npm excludes,
we install — is kept from `feat/install-platform-filter`, because with our guarded
`imports.gi` degradation an unloadable extra is cheaper than a wrongly skipped
dependency. It only becomes load-bearing once `target.libc` is real, which is
what this PR makes it.

## Not done yet

- [x] Type-check (0 errors, both packages), unit specs (1504 cases), e2e install suites (30) — all green
- [x] `gjsify.glibcRequires` is claimed by `prebuild-libc`; `audit-runtimes --check` exits 0
- [x] New e2e fixtures registered in the root `package.json` `test:e2e` list — `e2e-shard.mjs` parses that list rather than globbing, so an unregistered fixture never ran
- [x] repo-wide `gjsify lint`: **0 errors**, 56 warnings, **none introduced** — each of the five in changed files verified present on `main` and absent from the diff
- [x] Committed `dist/cli.gjs.mjs` rebuilt (6591888 B)
- [ ] The split itself — all 11 bridges into ~51 generated per-target packages
      (ADR 0017 decision), plus the `gjsify onboard` bootstrap sweep it needs
- [x] CI: glibc-floor gate added to `commit-prebuilds` (it commits with `[skip ci]`,
      so `audit-runtimes.yml` never saw the commit that can change the floor), then
      `fedora:43` → `fedora:44`
- [x] Alpine/musl leg for postmarketOS — `build-prebuilds-musl`, dispatch-only,
      gating (not compile-only). Scoped by measurement to the **two** packages that
      actually fail on musl; six others load and run there unchanged
- [ ] Promotion of that leg: it must run green once before any `-musl` target may
      be declared, and both packages' `libc` must become `["glibc","musl"]` or a
      shipped musl prebuild can never reach a musl host

---

## Rebased onto main (2026-08-01)

Two commits were DROPPED: the Windows junction fix and the packument-cache
filename fix shipped separately as #898, so they are already in `main`. Backup of
the pre-rebase branch: `backup/897-pre-rebase`.

That rebase produced one conflict worth recording, because it was a genuine
COMBINATION rather than a choice — `main` and this branch improved
`install-packument-cache.ts` for different reasons and both are right:

- from `main` (#898): the separator moved off the Win32-reserved `|` to `,`,
  `encodePart` additionally escapes `*`, and the key got ONE spelling (the shard
  hashed the raw values while the filename used the encoded ones — which is how a
  reserved character survived review of a function whose other half looked
  correct).
- from this branch: the document SHAPE became part of the key. Without it an
  escalated full read is answered from the corgi document, which carries no
  `libc` — so a musl-only package comes back unconstrained, is judged compatible
  on a glibc host and gets installed. A wrong answer shaped exactly like a cache
  hit, with no failure until something `dlopen`s the binary.

Resolved as `<shape>,<encoded-registry>,<encoded-name>` at layout **v3**: either
change alone makes older entries unreachable rather than merely stale, so a new
root retires them instead of leaving ambiguous orphans.

### Verification on the rebased branch

| | |
|---|---|
| `tsc -p packages/infra/cli` | 0 errors |
| unit suite | 1578 pass |
| e2e (`install-platform-filter` + 4 prebuild suites) | 108 pass |
| `audit-runtimes --check` | exit 0 |
| repo-wide `gjsify lint` | 0 errors |
| `changed-packages.mjs --all` | OK |
| workflow YAML (3 files) | parses |


JumpLink:
## Empirical correction from a podman probe — `prebuild-libc`'s predicate is wrong

A container probe on `alpine:3.24.1` (x86_64 and aarch64 under qemu) measured what the
committed artifacts actually do on musl. It **contradicts the premise this PR's
`prebuild-libc` rule was built on**, so the rule and the `libc` declarations it drives
need correcting before this leaves draft.

### `libc.so.6` in `DT_NEEDED` is a provenance marker, not a loadability predicate

musl's dynamic linker treats a `DT_NEEDED` of `libc.so.6` as a request for **itself** —
`load_library()` refuses to reload the implementation under any reserved name (`c`,
`pthread`, `rt`, `m`, `dl`, `util`, `xnet`). No `libc.so.6` file exists on the image at all
(`find / -name libc.so.6` is empty; only `/lib/libc.musl-x86_64.so.1`), and four
glibc-built bridges that name it load anyway.

The three failure modes that are real, at `RTLD_NOW`:

| failure mode | error | affected |
|---|---|---|
| glibc loader in `DT_NEEDED` (`ld-linux-*` is not a musl reserved name) | `Error loading shared library ld-linux-x86-64.so.2` | all three **x64** Rust cdylibs |
| a glibc-only **symbol** | `symbol not found` | `sab-native` x64 (`fcntl64`), arm64 (`__cmsg_nxthdr`); `lightningcss-native` arm64 (`gnu_get_libc_version`) |
| distro **soname** difference (not libc at all) | `Error loading shared library libmozjs-140.so.0` | `napi` — Fedora has `.0`, Alpine ships `libmozjs-140.so` |

### The consequence for this PR

The `libc: ["glibc"]` declarations added here are **too aggressive**. Declaring it makes the
install filter *skip* the package on Alpine — but `webgl`, `http2-native`,
`http-soup-bridge` and `terminal-native` demonstrably load and run there. That would remove
working functionality on postmarketOS, which is precisely the platform the axis was added
for. It is also the exact failure this PR's own kept deviation from npm argues against ("a
wrong skip breaks a working install").

Recommendation: keep the `libc.so.6` read as *provenance* (it answers "was this linked
against glibc", which is what the field means to an installer), but do not let it stand as
the loadability predicate, and **omit `libc` where the artifact loads on either**. Asserting
loadability requires a real `dlopen` on a musl host — a CI-leg job, not a portable rule.

### The load matrix, and the good news for the phone

**aarch64 — the postmarketOS case — is markedly better than x64:**

| package | glibc **arm64** build on musl/aarch64 |
|---|---|
| `webgl`, `http-soup-bridge`, `http2-native`, `terminal-native`, `tls-native`, `webrtc-native` | **OK** |
| `oxfmt-native`, `rolldown-native` (both halves) | **OK** |
| `lightningcss-native` | FAIL `gnu_get_libc_version` |
| `sab-native` | FAIL `__cmsg_nxthdr` |

**11 of 13 committed `linux-arm64` libraries already load on Alpine aarch64 — only two
packages strictly need a musl build for the phone.** The arm64 Rust cdylibs load because,
unlike their x64 siblings, they carry no `ld-linux-aarch64.so.1` in `DT_NEEDED`.
`tls-native`'s arm64 artifact also passed a real functional call on musl/aarch64.

### Caveat, stated plainly

`RTLD_NOW` means every relocation bound, not that behaviour is correct. **musl ignores ELF
symbol versioning entirely**, so a reference to `GLIBC_2.39`-versioned `foo` binds to musl's
unversioned `foo` with no compatibility check. The probe independently reproduced this PR's
glibc-floor numbers, `lightningcss-native` at 2.39 included.

### Toolchain answer for the musl CI leg

`alpine:3.24` carries the whole toolchain from `main`/`community` — no third-party repo, no
`edge`, no source builds. All seven Vala bridges build with **zero `meson.build` patches**,
and `glib-networking` uses the GnuTLS backend. Rust bridges build with Alpine's own
`apk rust`/`cargo` **without** `--target` (the `*-unknown-linux-musl` targets cannot emit a
cdylib). For the workflow: mount with `--security-opt label=disable`, **never `:Z`** on the
work tree — two concurrent `:Z` mounts relabel it to private MCS categories and lock each
other out.

PR #908, opened 2026-08-01
ci: catch what silently reaches main — resurrected prose and unlinted PR titles

JumpLink:
Twice in one day a doc change was written against a copy of a file that predated a rewrite already on `main`, and put the old wording back. Git merged it **without a conflict** — the merge base has neither version, so *"replace these paragraphs with those"* is an ordinary edit. #886 restored five paragraphs #885 had consolidated; #892 carried the entire pre-#885 `AGENTS.md`. Nothing could see either: prose has no test, no type and no conformance rule.

### The signature is not the obvious one

My first attempt looked for *"removes lines the base added"*. **It does not fire on either incident** — both had the new wording in their own merge base and replaced it anyway, so nothing was removed that the base gained later.

What distinguishes them is the other direction: **the change ADDS lines the base branch had DELETED.** Content coming back from the dead is what writing against a stale copy produces, and what an ordinary edit does not.

### It covers one half of the class, and the other half needs nothing

| shape | merge behaviour | covered by |
|---|---|---|
| branch holds the new version and replaces it (#886) | merges **silently** | this check |
| branch predates the rewrite (#892) | git **conflicts** | git itself |

#892 never added those lines, so there is nothing to detect — and the merge conflicted, which is how I found it. Only the #886 shape is invisible.

### Warns, never gates

A deliberate revert has the identical signature, and a hard gate on an ambiguous signal is one people learn to route around. `continue-on-error: true`; a `Revert-Of:` trailer marks intent explicitly; `--strict` exists for a caller that wants the gate.

### Two things keep it from becoming noise

- **A churn filter.** A line that still exists on the base in a slightly different form was not resurrected. Without it, the `AGENTS.md` header carrying the workspace version — deleted and rewritten by *every release* — fired on every branch spanning a release. That is precisely how a check earns being ignored.
- **A shallow clone reports `SKIPPED`, not "nothing found".** A check that cannot run must not read as a clean bill of health. The step deepens 200 commits first, and only for this step, so the rest of the job keeps its fast checkout.

### Verification — against the real commits, not fixtures

| case | result |
|---|---|
| the #886 incident | fires, **6 lines**, exit 1 under `--strict` |
| the commit that fixed it | silent |
| this branch | silent, exit 0 |
| no merge base available | `SKIPPED`, exit 0 |

### DIFF
--- a/.github/workflows/audit-runtimes.yml
+++ b/.github/workflows/audit-runtimes.yml
@@ -109,6 +109,12 @@ jobs:
       #
       # Needs history, and this checkout is shallow — deepen just for this step
       # rather than making every other check in this job pay for a full clone.
+      # The PR-title linter inherits its type list from the action default
+      # instead of carrying a second copy of `type-enum`. That is only sound
+      # while the two agree — this is what says so out loud.
+      - name: Check the PR-title type list still matches commitlint
+        run: node scripts/check-pr-title-types.mjs
+
       - name: Check for resurrected prose (advisory)
         continue-on-error: true
         env:
--- a/.github/workflows/audit-runtimes.yml
+++ b/.github/workflows/audit-runtimes.yml
@@ -94,3 +94,25 @@ jobs:
       # exactly what silently reintroduces a per-run `dnf install`.
       - name: Check CI container images cover their jobs
         run: node scripts/check-ci-image-packages.mjs
+
+      # Prose that a change resurrects after the base branch deleted it. Git
+      # merges that WITHOUT a conflict — the merge base has neither wording, so
+      # "replace these paragraphs with those" is an ordinary edit — which is how
+      # #886 put back five paragraphs #885 had consolidated, and how #892 carried
+      # the whole pre-#885 AGENTS.md. Nothing else can see it: prose has no test,
+      # no type, no conformance rule.
+      #
+      # WARNS by design and never gates: a deliberate revert has the identical
+      # signature, and a hard gate on an ambiguous signal is one people learn to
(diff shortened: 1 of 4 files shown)


JumpLink:
**Second commit added: the PR-title lint.** Same theme as the doc-revert check — something that reaches `main` while nothing looks at it — so it belongs in one PR rather than two.

`commitlint.yml` lints the PR's **commits**. With squash merges what actually lands is the PR **title**. #849 merged as *"Run the transparent addon matrix in CI (#849)"* with no conventional prefix and is therefore **absent from the v0.26.1 CHANGELOG entirely** — the work shipped with no release note. This workflow already triggers on `edited`, so the intent was there; only the check was missing.

### No `types:` input, on purpose

Our `type-enum` is exactly the action's default — verified as a set, the same 11. Passing an explicit list would put a **second copy** in a workflow file, and a second copy is the thing that drifts. `check-pr-title-types.mjs` holds the assumption instead:

| situation | result |
|---|---|
| lists diverge, workflow silent | **exit 1**, names the divergent type, says to make the list explicit |
| lists diverge, workflow declares its own | exit 0 |
| lists agree, workflow declares one anyway | **exit 1** — a dead copy |

All three exercised by hand. The hardcoded default inside the script is not a duplicate of our config; it *is* the claim being checked.

I considered hand-rolling a title regex and rejected it: `config-conventional` also enforces `subject-case`, `subject-full-stop` and `header-max-length`, so a partial reimplementation would accept titles the commit linter rejects — precisely the drift the rule exists to prevent.

PR #913, opened 2026-08-01
fix(cli): restore the reproducible cli.gjs.mjs

JumpLink:
**main has been red since #906** (`6257afe46`) on `Verify committed bundles rebuild from source`, so every PR opened since inherits a failing build — including #911, which is how this surfaced.

## What is actually wrong

The committed `packages/infra/cli/dist/cli.gjs.mjs` does not reproduce from the source at HEAD:

```
committed: 6594685 B · rebuilt: 6594347 B · first difference at byte 5395787
```

At that offset:

| | |
|---|---|
| committed | `…__name$1(a$5,'a')});function e$1(n,a=!1){…` |
| CI rebuild | `…__name$1(a$5,'a')});function e$2(n,a=!1){…` |

`e$1` vs `e$2` — a **minifier name-suffix counter**, i.e. the module-order-derived `$N` drift `release-cut.yml` documents at length. 338 bytes of accumulated suffix differences, no semantic delta. This is *not* a bundle built from older source.

## The fix is the documented one

The verify step uploads its own rebuild as the `rebuilt-bundles` artifact precisely for the case where a machine cannot reproduce the bytes. This takes them from main's own failing run (`30709249982`, at `6257afe46`).

Valid for the current HEAD: the only commits since are #905 and #912, touching `.github/workflows/**` and `status/open-todos.md` — neither is inlined into the bundle (checked with `git diff --name-only`, not assumed).

## Verified rather than trusted

Swapping a committed bundle blind is the exact mistake this gate exists to catch, so, under stock `gjs`:

- `--version` → `0.26.1`, matching `packages/infra/cli/package.json`
- `--help` → exit 0
- single-`@gjsify/web-streams`-copy assertion still holds

## What this does not answer

Why #906's tree produced a different module order. `globToEntryPoints` sorting each glob's expansion is supposed to have closed that race; a local build cache or a stale `lib/esm` on the authoring machine are the documented suspects. Unblocking main takes priority — but if this recurs, the sorting fix does not cover whatever this is, and that deserves its own investigation.


JumpLink:
### CI's determinism is now measured, not assumed

The PR body claims the committed file is the outlier. That is worth more than an assertion, so I compared the `rebuilt-bundles` artifacts from **two independent CI runs**:

| run | commit | bytes |
|---|---|---|
| `30709249982` (main) | `6257afe46` | 6594347 |
| `30710590470` (#911) | `ci/derive-baked-image-exemptions` | 6594347 |

`cmp` reports them **byte-identical** — two separate runs, two different commits, two container instances, same output. Together with the committed file's 6594685 B, that isolates the divergence to whatever produced the committed bytes; it is not run-to-run noise, and CI is not the unstable side.

It also means the bytes this PR installs are not a one-off snapshot of a lucky run — the next verify run has already produced them twice.

PR #920, opened 2026-08-01
revert(node-gi): unconditional GTK bundle install

JumpLink:
Reverts #910 (`da9ca4001`), which broke `macOS build + conformance (Node+Bun+Deno / macos-latest arm64)` on main.

## Evidence

Same job, same runner image, parent vs. merge commit:

| commit | `npm install` in `packages/node-gi/node-gi` | `boxed-out` / `async-error` |
|---|---|---|
| `b607be808` (parent) | `added 1 package in 20s` | ✓ pass (node, bun, deno) |
| `da9ca4001` (#910 merge) | `added 2 packages in 36s` | ✗ fail |

The second package is `@gjsify/gtk-runtime-darwin-arm64` — **31.5 MB unpacked, 90 files**, a real payload.

## Mechanism

The edge works exactly as designed; the problem is where it lands. This job builds the addon against **Homebrew GTK 4.22.4**, but the newly installed bundle now satisfies candidate 4 of `resolveGtkRuntimeBundle()` (`require.resolve('@gjsify/gtk-runtime-darwin-arm64')`). node-gi therefore re-execs onto the **bundle's** typelibs while its native code is linked against Homebrew's. Method lookups land on the wrong entries:

```
no method 'match' on GObject-weak-notifies      (GLib.Regex.match)
no method 'get_identifier' on GIRepository
no method 'get_year' on GObject-closure-array
```

The runner then produced no output for 29 minutes and the job hit its timeout — the `cancelled` conclusion is that timeout, not a supersede.

This collapses a separation the repo already models deliberately: a brew-GTK conformance job **and** a separate `macOS batteries-included conformance (no brew GTK)` job. A bundle that installs unconditionally makes the first one test the second one's configuration, against a mismatched addon.

## Scope

**No published consumer is affected.** `0.26.1` shipped before this edge existed, so the change only ever reached CI. Nothing needs a patch release.

## What was actually wrong

Not the ADR-0003 rule-5 amendment, the conformance rule, or the pin sync — those stand and were verified green. The gap is a precedence question the PR never had to answer, and explicitly flagged under *"Not proven here: Nothing ran on Windows or macOS"*:

**when a host already has a GTK the addon was built against, does the bundle win?** Today it does, unconditionally. It should not.

Re-landing needs that answered — plausible shapes: only activate the bundle when no usable system GI is present; or honour an explicitly-set `GI_TYPELIB_PATH` as an opt-out; or have the brew-GTK legs install with `--omit=optional` if the override is the intended semantics after all — plus a real macOS run proving it, which is the check #910 was missing.


JumpLink:
Independent confirmation from #921, with a control that isolates the mechanism further than a parent-vs-merge comparison can.

#921 matrixes this job over both macOS architectures. On the **same commit, same workflow, same steps**, the only difference between the two legs is whether a `@gjsify/gtk-runtime-<target>` package exists to be installed:

| leg | bundle in `optionalDependencies` | result |
|---|---|---|
| `darwin-arm64` | `@gjsify/gtk-runtime-darwin-arm64` installs | hangs in `Conformance subset on Node`, killed at `timeout-minutes: 30` |
| `darwin-x64` | no such package → falls back to Homebrew GTK | **green end to end** — node-gyp build → conformance on Node, Bun and Deno → stage → prebuild load → upload |

That removes the remaining alternative explanations in one shot: not the runner image, not the conformance suite, not the addon, not the commit. The arch that cannot install the bundle takes the pre-#910 path and passes; the arch that can, hangs. Exactly the precedence bug described above.

It also gives the re-landing work a ready-made regression test: once the bundle only wins where no usable system GI is present, both legs should stay green — and the x64 leg keeps proving the Homebrew path independently.

I had opened #922 for this before finding this PR; closing it as a duplicate and pointing it here.

PR #923, opened 2026-08-01
feat(prebuilds): split into per-target packages

JumpLink:
ADR 0017, the esbuild model: each native bridge's committed `prebuilds/<target>/`
moves into its own platform-gated npm package `@gjsify/<name>-<os>-<arch>`,
referenced from the bridge as an `optionalDependency`. A `linux-x64` consumer
stops downloading ~66 MB it cannot load.

- **156 renames** — pure file moves, all `(100%)` similarity
- **60 new packages** — `os`/`cpu` gated, `files: ["prebuilds"]`, no JavaScript
- **11 bridges** rewired
- **derived, not hand-listed** — `scripts/generate-platform-packages.mjs` emits
  the set, and the `platform-packages` conformance rule IS that function
  (imported, not reimplemented), so "the generator and the audit agree" is a
  fact rather than a hope

## Verified

Measured on this tree, not predicted:

| | |
|---|---|
| committed prebuild bytes | **98.3 MB** total; a `linux-x64` consumer now downloads **31.6 MB** — 68 % never fetched again, matching ADR 0017's estimate |
| `gjsify pack` on a platform package | tarball holds exactly `README.md`, `package.json`, `prebuilds/linux-x64/{*.gir,*.typelib,*.so}` |
| `gjsify pack` on its bridge | **zero** `prebuilds` entries — the byte saving is real, not just declared |
| `node scripts/audit-runtimes.mjs --check --strict` | **OK** (the gate every PR runs) |
| `node scripts/generate-platform-packages.mjs --check` | **OK** — 60 packages, 9 artifacts deferred |
| `gjsify install --immutable` | **exit 0**, `gjsify-lock.json` unchanged |
| `detectNativePackages()` after that install | finds all 10 host-target platform packages, and no foreign-platform ones — **ADR 0017's "no new runtime code path" claim, measured** |
| `tests/e2e/{terminal-native,prebuild-declaration-invariant,platform-exemption-clearing,ci-runner-arch}` | **67/67**, incl. `native_loaded is true` — the typelib resolves at the new path |
| `gjsify lint` / `gjsify format --check` | 0 errors (57 pre-existing warnings, unchanged) / OK |
| platform matrix | byte-compared before/after the CI-attribution fix — unchanged for every bridge |
| committed GJS bundles | untouched: no `cli/src`, `resolve-npm/lib`, `rolldown-plugin-gjsify/src` or `tsc` source in the diff, and CI's `Verify committed bundles rebuild from source` passes |

## What the five open items from the last review became

1. **Rebase** — done, onto `b963a3ac3`. Five conflicts, all resolved; the 156
   renames survived rename detection, including the four `oxfmt-native`
   binaries main rebuilt in between.
2. **Versions** — the 51 manifests said `0.26.0`; regenerated at `0.26.1`.
   *Regenerating exposed a real defect*: the plan judged "is there an artifact
   for this target" by the **parent's** `shipped` set while `artifactDir` had
   always looked in the child too, so a half-migrated tree — a `--write`
   interrupted between the `git mv` and the manifest patch, or exactly the
   rebase I had just done — printed `0 committed target(s) into `, an
   instruction to re-run the command that had refused to do anything. Both call
   sites ask the same question now.
3. **`gjsify onboard`** — **DONE.** Swept from this branch: 60 published +
   trusted at 0.26.1, 131 already done, 0 failed. All 60 were covered, the two
   napi ones included — `mergePublishables()` carves `packages/{node-gi,napi}/*`
   in for exactly that reason. Verified against the registry, not the sweep's
   own output: `@gjsify/webgl-linux-x64` (144K) ships typelib + `.so`,
   `@gjsify/tls-native-darwin-x64` (4.0K) is README + manifest only — the
   deferred case behaving as designed — and the published metadata is per-target
   (`libc: ["glibc"]` on riscv64 and ppc64 ONLY, the two whose ELF records the
   glibc loader).

4. **Tier direction** — no rule 5 needed. A child inherits its parent's
   `gjsify.tier`, so every new edge is same-tier and the existing
   dependency-direction check passes unchanged.
5. **`node-gi` `win32-x64`** — no decision needed: node-gi builds with node-gyp
   at install time, so `prebuildOwnership()` returns `install-time` and the
   generator skips it on the same signal `prebuild-artifacts` already exempts it
   with. Nothing to package.

## What pointing the workflow at the new directories found

The recovery commit changed no CI. Doing so surfaced five defects, each one
silent:

- **`commit-prebuilds` would have committed binaries into a path no tarball
  ships.** All 58 download paths now name the platform package.
  `changed-packages.mjs` derives its package table from those same lines, so it
  maps `<bridge>-<target>` back to the bridge.
- **`@gjsify/napi`'s macOS prebuild would have stopped shipping.** That artifact
  is not committed — `release.yml` builds it and stages it into the tarball at
  pack time — and the split takes `prebuilds` out of the bridge's `files`, so
  the staging would pack nothing, with no error anywhere. This is why a
  **declared-but-deferred target now gets a package too**, inheriting the
  parent's `gjsify.platformsUncommitted`. It also restores the tripwire the
  split would otherwise remove, and claims the npm name at declaration time.
  51 packages → 60.
- **`napi.yml` had three `cp … prebuilds/linux-x64/` with no `mkdir -p`** — that
  directory was committed before the split; now the step just fails.
- **Two rules read a DOWNLOAD as production.** Every download step is called
  "Download `<pkg>` `<arch>` prebuilds" and `publish-napi`'s is "Confirm both
  prebuilds are STAGED" — all production verbs. Harmless while those steps named
  the bridge (the build legs had already credited the same targets); once they
  name the children, both consuming jobs are plain `ubuntu-latest`, so all 60
  platform packages were credited with `linux-x64` and failed in both directions
  at once. Keyed on `uses: actions/download-artifact` at the JOB level now.
- **The platform matrix would have claimed `✓` "artifact committed" for nine
  targets that have none.** Its exempt cell was gated on `gjsify.prebuilds`,
  which a split bridge no longer has. That table is rendered into the website's
  Platform Support page, so it is the "declared looks like delivered" lie the
  cell was written to prevent, reintroduced through the back door. It asks
  `prebuildOwnership()` now.

Plus two facts that had to move off the bridge with the bytes:

- **`libc` on a split parent is an install-time regression.** npm, yarn and pnpm
  all honour it, so `@gjsify/lightningcss-native`'s inherited `["glibc"]` would
  refuse to install its pure-TypeScript half on musl hosts where it runs fine.
  Per-target packages let the field say something true for the first time: only
  `@gjsify/tls-native-linux-riscv64` needs it, not all five of that bridge's
  Linux targets.
- **The generator had its own copy of the glibc-loader predicate.** It matched
  `ld-linux-*` but not ppc64le's `ld64.so.2` or s390x's `ld64.so.1`, so `--write`
  emitted manifests `--check` then rejected. It calls `measurePrebuildLibc()` —
  the `prebuild-libc` rule's own reader — so the tool that writes the field and
  the tool that grades it cannot disagree.

## Still open before this can ship

**The seven `darwin-x64` exemptions** need a human edit the first time
`commit-prebuilds` runs green: the artifact lands beside a live exemption and
`prebuild-artifacts` refuses it, which is what the exemption's own reason text
has always promised would be needed. **Pre-existing** — the same deadlock exists
on `main`, where the artifact lands in `<bridge>/prebuilds/darwin-x64/` and trips
the identical check — and the split neither fixes nor worsens it; it only moves
where the directory appears. Tracked in `status/open-todos.md`.

The npm bootstrap that used to be item 1 here is done (above).

## Unrelated reds, both diagnosed

**`main`** — `commit-prebuilds` has been failing since 2026-08-01, for a reason
with nothing to do with this PR: #897 bumped `prebuilds.yml`'s base image
`fedora:43` → `fedora:44` in a hygiene sweep, and glibc 2.43 re-versions
`acosf`/`asinf`/`atan2f`, which lightningcss's colour conversion calls — so the
measured floor jumped 2.39 → 2.43 with no source change. Fixed separately in
**#927**.

**This PR's one red check** was `Bun + Deno conformance` on `arrays`, and it is a
flake in the teardown class `scripts/cross-runtime.mjs` documents (#47/#58):
every assertion passes, then `free(): invalid pointer` at exit. Established
rather than assumed:

- **Nondeterministic on one commit** — attempt 1 failed on *bun*, attempt 2 on
  *deno*, attempt 3 **passed**. A consequence of a diff does not move runtimes.
- **This PR's code cannot execute in that job.** It runs `npm install` +
  node-gyp *inside* `packages/node-gi/node-gi`, whose only dependency is
  `node-addon-api` — no workspace, no `gjsify install`, no platform package. The
  two node-gi test files this PR touches are not in `cross-runtime.mjs`'s
  curated 38-file list, so they never run there.
- **The environment did not change.** The two CI image digests either side of
  the first failure are package-identical (797 packages, empty `diff`, same
  `glibc-2.43-6`), and both runs used bun 1.3.14 / deno 2.9.4.
- A control re-run of the last `main` node-gi run passed, as did the third
  attempt here.

Worth flagging on its own: the bun side of that class was measured at *0 crashes
in ~7000 runs* and is deliberately left un-carved-out, with the instruction that
a bun hard-crash be re-confirmed with a gdb backtrace before anything is
inferred. This is the first one. Recorded in #927.


JumpLink:
## Coverage check against current `main` — what the follow-up has to close

Measured, not assumed. Two separate findings.

### 1. Against today's *committed* prebuilds the split is complete

The set of `(package, target)` pairs with a committed `prebuilds/<target>/` directory:

| | pairs |
|---|---|
| `main` today | 51 |
| this branch covers | 51 |

`diff` of the two sorted lists is **empty** — identical. Targets: `linux-{x64,arm64,ppc64,riscv64,s390x}` + `darwin-arm64`. So nothing that exists right now is missing a package.

### 2. But `main` already *declares* 7 targets this branch has no package for

These came in with the macOS and Windows ports, after this branch's base (`f61464b44`):

| package | declared target | artifact status |
|---|---|---|
| `@gjsify/http2-native` | `darwin-x64` | `platformsUncommitted` — CI builds it, lands on the next `commit-prebuilds` |
| `@gjsify/http-soup-bridge` | `darwin-x64` | same |
| `@gjsify/lightningcss-native` | `darwin-x64` | same |
| `@gjsify/oxfmt-native` | `darwin-x64` | same |
| `@gjsify/terminal-native` | `darwin-x64` | same |
| `@gjsify/tls-native` | `darwin-x64` | same |
| `@gjsify/node-gi` | `win32-x64` | no `platformsUncommitted` entry |

The six `darwin-x64` entries carry an explicit note: *"the artifact lands here on the first main-branch run of `commit-prebuilds` after this declaration merges, and this entry becomes a failure the moment it does."* So this is a **moving target** — the pairs go 51 → 57+ on their own, without anyone touching this branch.

### Why that is mostly good news

The split is **derived, not hand-listed**: `scripts/generate-platform-packages.mjs` produces the package set and `platform-packages.mjs` checks it. Re-running the generator after the rebase should emit the additional packages, the same way the 51st cost what the 2nd did.

### Suggested order for whoever picks this up

1. Rebase onto `main` (currently ~25 commits ahead: #897, #916, #919, #920).
2. Re-run `scripts/generate-platform-packages.mjs`; expect new `-darwin-x64` packages, and `node-gi`'s `win32-x64` to need a decision.
3. Fix the version skew (new manifests say `0.26.0`, `main` is `0.26.1`).
4. Re-check the tier direction — #910 added ADR 0003 rule 5 for the parent→artifact edge and was reverted in #920.
5. `gjsify onboard` before the release that ships this. Count is no longer 51 but 51 + however many targets have landed by then; every one is an unpublished name, and an unbootstrapped one stalls the serialized OIDC publish loop for all alphabetically-later packages.

PR #927, opened 2026-08-02
fix(ci): stop the prebuild image from moving the glibc floor

JumpLink:
`main`'s `Build Native Prebuilds` has been red since 2026-08-01 20:05 — three
consecutive `commit-prebuilds` runs, so no new prebuild has landed since.

## What broke

`prebuilds.yml`'s base image decides which glibc every published binary is
**linked against**. #897 bumped it `fedora:43` → `fedora:44` as part of a hygiene
sweep across the workflows' images. That sweep is right for every *other*
`fedora:<n>` in the repo — those pick a machine to run tests on, and newer is
strictly better. This one picks an ABI promise.

Measured, on this machine:

| image | glibc | resulting floor |
|---|---|---|
| `fedora:43` | 2.42 | **2.39** — `pidfd_spawnp`/`pidfd_getpid`, from Rust std. The value every package declares today. |
| `fedora:44` | 2.43 | **2.43** — glibc 2.43 RE-VERSIONS `acosf`, `asinf` and `atan2f`, which lightningcss's colour conversion calls. |

From the artifact the failing run actually uploaded:

```
$ readelf --dyn-syms -W libgjsify_lightningcss.so | grep GLIBC_2.43
  32: … UND atan2f@GLIBC_2.43 (11)
  33: … UND asinf@GLIBC_2.43  (11)
  37: … UND acosf@GLIBC_2.43  (11)
```

The linker binds the newest *default* version available, so merely building
there rewrites the requirement — no source change, no review, nothing to
attribute it to. The old symbol versions still exist in 2.43 and behave the way
lightningcss needs, so the new ones buy nothing. What they cost is every consumer
below 2.43: at 2.39 this toolchain runs on Ubuntu 24.04 and Debian 13; at 2.43,
on essentially nothing shipping today.

## The fix

Revert the five image references (native container + the two emulated
`base_image`s + two comments), and put the reason **at the `container:` line**.
The failure mode is that the next person doing image hygiene sees a
stale-looking tag; a commit message will not be in front of them.

No declaration changes: the committed artifacts already measure 2.39, which is
what every package declares. Raising the declaration to 2.43 was the other way
out and is rejected on the numbers above.

## The gate did its job

Nothing about `prebuild-libc` needs changing. `commit-prebuilds` measured the
freshly downloaded artifacts, refused them and named both numbers — exactly what
that step's own comment predicted a base-image bump would do:

> a base-image bump (fedora:43 → 44 → …) is exactly the change that moves that
> number without touching one line of source

It fired **late**, though: `commit-prebuilds` is main-only, so the PR that caused
it was green. That is the shape AGENTS.md § "PR coverage parity" calls dishonest.
Closing it means running the same measurement in the build legs, which do run on
`pull_request` — recorded in `status/open-todos.md` with both prerequisites
(`nodejs` in the legs' `dnf` line; the legs must stage where the artifact is
published from, which ADR 0017's split changes, so the two want to move
together).

Also recorded there: underneath all of this the floor is **observed, never
chosen**. Even pinned to `fedora:43` it moves the day that image's glibc
re-versions something else. Choosing it means building the three Rust bridges
against a declared baseline (old-glibc container or `cargo-zigbuild --target
<triple>.<glibc>`) — a policy decision about how old a distro we support, and its
own change.

## Verified

- `node scripts/audit-runtimes.mjs --check --strict` → OK
- `changed-packages.mjs` still parses the workflow and gates correctly (its
  package table is derived from these very files)
- both glibc numbers measured under `podman run --platform linux/amd64`

### DIFF
--- a/.github/workflows/prebuilds.yml
+++ b/.github/workflows/prebuilds.yml
@@ -56,7 +56,7 @@ name: Build Native Prebuilds
 # while checking out the PR's head code — precisely the combination that must
 # not exist in a workflow that contains a job which pushes commits. Plain
 # `pull_request` runs with a token GitHub forces to read-only for fork PRs, and
-# the build legs need nothing more: they pull the PUBLIC `fedora:44` image, so
+# the build legs need nothing more: they pull the PUBLIC `fedora:43` image, so
 # unlike main.yml's private `ghcr.io/gjsify/ci-fedora` they also work on a fork
 # PR. Least privilege backs that up rather than relying on it — see
 # `permissions` below.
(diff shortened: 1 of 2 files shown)


JumpLink:
## Verified against the artifact this PR's own CI built

Not asserted — the `Build prebuilds (x64)` leg of this branch ran on `fedora:43`,
and I measured what it uploaded:

```
$ readelf --dyn-syms -W libgjsify_lightningcss.so | grep -oE 'GLIBC_2\.[0-9]+' | sort -uV | tail -4
GLIBC_2.33
GLIBC_2.34
GLIBC_2.35
GLIBC_2.39
$ … | grep -c 'GLIBC_2.43'
0
```

Zero `GLIBC_2.43` symbols, and the same numbers through the gate's own reader
rather than my `readelf`:

```
gate reader → flavour=glibc musl=incompatible glibcRequires=2.39 unreadable=0
declared    → 2.39
MATCH — prebuild-libc passes on this artifact
```

For contrast, the same package from the last **failing** `main` run
(`fedora:44`, run 30722880760):

```
32: … UND atan2f@GLIBC_2.43
33: … UND asinf@GLIBC_2.43
37: … UND acosf@GLIBC_2.43
```

Three re-versioned float-math symbols, nothing else — which is why the floor
moved 2.39 → 2.43 with no source change, and why reverting the image is enough to
put it back.

Note this PR's build legs rebuild **every** package (it edits `prebuilds.yml`, so
the change classifier fans out), which is what makes the fix provable on the PR
instead of only after merge.

JumpLink:
### All four flagged targets restored — measured, from this PR's own artifacts

The failing `main` run named four `@gjsify/lightningcss-native` targets. Every
one of them, rebuilt on `fedora:43` by this branch's build legs and read with the
gate's own `measurePrebuildLibc()`:

| target | measured | declared | libc | verdict |
|---|---|---|---|---|
| `linux-x64` | 2.39 | 2.39 | glibc-only | MATCH |
| `linux-arm64` | 2.39 | 2.39 | — | MATCH |
| `linux-ppc64` | 2.39 | 2.39 | glibc-only | MATCH |
| `linux-s390x` | 2.39 | 2.39 | glibc-only | MATCH |

Zero `GLIBC_2.43` symbols in any of them. `linux-riscv64` is not in the table
because it was never affected — that leg builds on `ubuntu:26.04`, not Fedora, so
the image revert does not touch it.

That closes the claim: the floor is back at the declared 2.39 on every target the
bump moved, so `commit-prebuilds` has nothing left to refuse.

PR #939, opened 2026-08-02
fix(webgl): report the drawing buffer in device pixels

JumpLink:
## What

`HTMLCanvasElement.width`/`height` on the `Gtk.GLArea`-backed canvas returned the GTK **allocation**. A GLArea's framebuffer is `allocation × scale-factor` and `gl.viewport` takes raw **device** pixels, so every consumer that sizes its viewport from `canvas.width` drew into the bottom-left `1/scale²` of the widget:

* Three.js — `renderer.setSize(canvas.width, canvas.height, false)`
* Excalibur — `ExcaliburGraphicsContextWebGL`
* anything reading `gl.drawingBufferWidth`, which is a re-export of `canvas.width`

## Why it was never caught

Scale-factor 1 makes the bug arithmetically invisible — that is every desktop, every CI runner and both port-test VMs. It surfaced on the first HiDPI host the project has run on: a OnePlus 6T on postmarketOS (GNOME Mobile, 1080×2340, scale 3), where three showcases each rendered a 120×218 corner of a 360×655-logical widget.

| before | after |
|---|---|
| teapot in the bottom-left ninth | teapot fills the widget |

## The fix

* `canvas.width`/`height` → the **drawing buffer**, in device pixels (also what the browser contract means by those attributes).
* `clientWidth`/`clientHeight`/`offsetWidth`/`offsetHeight` → CSS layout size, **logical** pixels.
* `installGlobals()` publishes `devicePixelRatio` as an **accessor** over the widget's live `get_scale_factor()`, replacing the widget-less `1` that `@gjsify/dom-elements` seeds — so the identity `clientWidth × devicePixelRatio === canvas.width` holds on both kinds of host, which is what an unmodified browser consumer relies on.
* An unrealized widget answers `0` for its scale factor; clamped to 1 so the buffer is never reported as 0×0.

`Canvas2DBridge` needs no change and gets none: GTK pre-scales a `Gtk.DrawingArea`'s Cairo context. That asymmetry is now written down in AGENTS.md, along with the corrected "no HiDPI scaling in GTK GL context" comment that caused this.

## Tests

`html-canvas-element.spec.ts` (6 cases, +17 assertions, `@gjsify/webgl` GJS suite 875 → 892 green). It drives a **stub** GLArea on purpose: a real one reports scale 1 on any host that can run the suite — precisely the case that never failed — so only a stub can vary the number that matters. Covers scale 3, scale 2, scale 1 (no-op), the unrealized `0`, the CSS×ratio identity, and that width/height writes stay ignored.

## Showcase fixes carried along

* **teapot**: the resize branch compared `renderer.domElement.width` with `canvas.width` — the same object, so it was never true and the camera aspect stayed frozen at the first allocation. Now tracked in `prevW`/`prevH`, the shape the pixel showcase already used.
* **pixel**: dropped the `canvas.width = w` sync writes — silent no-ops against a GTK-owned buffer — and the comment that claimed they were what made the resize check work.

## Verification

Built and run on the device against the real `@gjsify/webgl` linux-arm64 prebuild (glibc build loading on musl), screenshotted through `org.gjsify.Devtools.Screenshot`: teapot, pixel post-processing and Excalibur all fill the widget.


JumpLink:
Regression check on a scale-factor-1 host (this workstation, Fedora GNOME/Wayland, 1100×700 window): the teapot fills the GL pane exactly as before the change — the arithmetic is identity at scale 1, and that is now measured end-to-end, not only unit-tested.

Re-verified on the device after the change as well: teapot, pixel post-processing and Excalibur all fill the widget.

PR #974, opened 2026-08-04
ci: order the publish sweep, verify the closure

JumpLink:
## The measured failure

A publish sweep has no transaction, so its **order** is a correctness property. Three facts compose:

1. **The sweep is fail-fast.** `packages/infra/cli/src/commands/foreach.ts` defines `failFast` at :654, called at :673 (stall abort) and :705 (a child rejecting); the sequential driver `runSequential` (:576) rethrows out of `runOne`, is caught at :475 and exits 1. Either way one `gjsify publish` failure aborts the remaining sweep.
2. **Pack time pins exactly.** `resolveWorkspaceProtocol` rewrites `workspace:*` to the target's exact version. Verified on the shipped 0.27.0 tarball: all 7 `@gjsify/webgl` siblings pinned to literal `0.27.0`.
3. **npm skips an unresolvable `optionalDependency` in silence.** A platform mismatch is not an error, and neither is a version that was never published.

**Measured publish order at 0.27.0, from the registry:**

| package | published |
|---|---|
| `@gjsify/webgl` | 19:38:22.572Z |
| `@gjsify/webgl-darwin-arm64` | 19:38:26.291Z |
| `@gjsify/webgl-darwin-x64` | 19:38:29.551Z |

The **parent** goes out before its platform **children** — on all eleven split bridges. An abort between those points leaves `@gjsify/webgl@<v>` live pinning `@gjsify/webgl-darwin-x64@<v>` that does not exist: the install *succeeds*, and a macOS consumer gets the fixed JS bridge with **no typelib and no dylib**. Strictly worse than the previous version's error state, which at least failed loudly.

Nothing re-checked it: root `npm:publish` ended at the `foreach`, and `verify-package-outputs.mjs` runs `--scope examples`.

At 0.27.1 all 7 siblings landed, so the tree on npm is intact today. **This is an unguarded hazard, not a live incident.**

## Cause

The policy already existed — in one job only. `publish-napi` publishes its two platform packages before the bridge and states the reason in a comment ("PLATFORM PACKAGES FIRST … an optional dependency that does not resolve is skipped in silence", release.yml:1319). The workspace sweep never inherited it, and no gate would have noticed.

## Prevention — the sweep is now dependency-ordered

`npm:publish` gains `--topological`. The load-bearing detail: `buildDependencyGraph`'s `includeOptional` defaults to `true`, so a platform child *is* a topological dependency of its bridge, and **any prefix of an aborted sweep is a resolvable tree**.

Measured with the real CLI over this tree (`foreach -v --no-private --exclude '@girs/*' --exec -- true`, 187 workspaces):

| order | `optionalDependencies` edges in the sweep | parent published first |
|---|---|---|
| current (discovery) | 60 | **60** |
| `--topological` | 60 | **0** |

**The composition — fail-fast × order — measured end to end with the real CLI** (three-package fixture, a `pub` script that fails in one workspace):

| sweep order | abort at | left "published" | tree state |
|---|---|---|---|
| discovery (today) | `child-a` | `[bridge]` | **broken** — the bridge is live pinning two children that do not exist |
| `--topological` | `bridge` | `[child-a, child-b]` | **resolvable** — nothing published points at anything missing |

That is the hazard and the fix in one table: fail-fast is not the bug, unordered fail-fast is.

The production graph is acyclic over the 187 selected packages, and `gjsify run build` already sorts topologically over the same graph — so this adds no new failure mode.

## Detection — `verify-release-closure`

New `scripts/verify-published-closure.mjs`, run by a new final job that `needs:` all five publishing jobs (so it also covers `publish-napi`, whose two edges are **literal** exact pins rather than `workspace:*`) and is gated `!cancelled()` — when a publish job *fails* is exactly when the state of npm is worth reporting.

Plain Node on a bare checkout: no install, no build, so the committed CLI bundle cannot stale it (the circularity `verify-committed-bundles.mjs` exists to break) and it still runs when the release that would have installed the tree is what failed. The package set comes from `createContext`, the same oracle every conformance rule reads.

### The exact assertions

It asserts **counts**, not the absence of findings:

1. `live > 0` — at least one candidate package is on the registry at the released version. Otherwise the release published nothing, there was no closure, and **the check verified nothing**.
2. `pinnedEdges > 0` — the *tree* declares at least one release-pinned intra-repo edge. A manifest property, not a registry one: this is the assertion that outlives a refactor of the enumeration. If it ever stops seeing the ~60 platform-sibling edges, every future release would otherwise pass on an empty set.
3. On a **complete** release, `examined > 0`. Implied by 1+2, asserted anyway so a change to either derivation cannot quietly produce an empty examination.
4. Every examined edge resolves: for each live parent `P@<v>` and each `dependencies`/`optionalDependencies` entry pinned to `<v>`, the target must be present at `<v>`. Failure names every offender as `UNRESOLVABLE: <parent>@<v> → <block>.<target>@<v> is NOT on the registry`.

A **probe that was not answered** (5xx, timeout, DNS) is `unknown`, not `absent`. The three states stay distinct because collapsing them was a real bug in the first push of this PR: `violations()` asked `!isLive(target)`, so one transient registry 5xx on a platform package would have printed `UNRESOLVABLE … is NOT on the registry` about a release that was complete — replacing a missing signal with a wrong one, precisely the substitution this guard exists to prevent. A violation now requires a confirmed 404; an unanswered probe fails the job on its own terms ("not evidence of anything; re-run"), and the liveness-derived conclusions are only stated when every probe was answered.

An **unclassifiable** spec is a failure, never a skip — an unclassified edge is coverage lost without a signal. (Zero today: every intra-repo edge is `workspace:*|^|~` or napi's literal `0.27.1`.)

Exact-version presence is a complete answer with no semver engine: the released version is the maximum of the train, so `<v>`, `=<v>`, `^<v>`, `~<v>` each admit exactly one existing version.

### What it deliberately does not fail on

An **incomplete** release in which no published package has a pinned dependency. That is precisely what prevention promises, it can only co-occur with an already-red publish job, and a second misleading red is how a job gets ignored. The log says `NO edge was examined` outright, so it can never read as a verified closure.

Absence is re-queried over a few rounds before it becomes a verdict — npm's CDN can serve a packument minted before a publish, and a false red at release time costs a manual re-run of a workflow that did its job.

## Detection is not a substitute for prevention

`--tolerate-republish` makes a **re-run of `release.yml` the recovery path**: it no-ops for every package that already landed and publishes the rest. But npm has no transaction to roll back, so detection alone would only **shorten the window** in which a consumer can install a bridge with no binary — the ordering is what closes it. That is why both halves are in one change.

## Verification

- Against the **live registry**: 194/194 packages at 0.27.1, 351 release-pinned edges (62 `optionalDependencies` + 289 `dependencies`), all resolve, exit 0.
- `tests/e2e/published-closure` — 12 cases against a local fake registry: complete release; dangling optional child (names it, and *only* it); resolvable prefix stays green; nothing published fails; a tree with no declared edge fails; an unclassified spec fails; a 503 probe is unknown rather than missing; propagation retry passes on a later round and re-queries only the absent name; private packages out of scope — plus a pin on the sweep's `--topological` flag.
- `tests/e2e/foreach-command` gains the `optionalDependencies`-are-topological-edges case. That one graph default is all that keeps the publish order correct, so it gets its own assertion.
- `node scripts/audit-runtimes.mjs --check --strict` green; the two other `release.yml`-parsing gates (`tests/e2e/ci-runner-arch`, `scripts/check-ci-image-packages.mjs`) green; `oxfmt --check` + `oxlint` clean on every touched file.

## A defect CI found in the guard itself

Worth recording, because it is the same class as the hazard: **E2E 4/4 went red on four fixtures that had each verified their closure perfectly.**

The suite spawned the script with the job env inherited, so `GITHUB_STEP_SUMMARY` pointed at the runner's summary file — which under the `ci-fedora` container belongs to a different user. `appendFileSync` threw `EACCES`, and because that write happens *after* the closure is computed and *before* `process.exit`, a run whose verdict was "every one of 3 edges resolves" exited 1 anyway. **A reporting side-channel decided the verdict.**

Two bugs, two fixes: the script now warns and carries on when the summary is unwritable (`problems` is the verdict; the table is a courtesy), and the suite strips `GITHUB_STEP_SUMMARY` from the child unless a case sets it deliberately — a test must not append to the enclosing job's summary, and inheriting it is exactly what hid the bug locally. Two cases pin both directions: an unwritable path keeps exit 0 *and says the summary is missing*; a usable path must still actually contain the table, so tolerating a failed write cannot decay into never writing.

Everything else on that run was green, including **E2E 1/4**, the shard carrying the `foreach-command` ordering case — so the prevention property is machine-verified in CI, not just locally.

## What the new gate would have caught

A sweep that aborted after a bridge and before its children. Today that produces an npm tree which installs cleanly and does not work, with nothing looking at it afterwards. With this change the order makes that state unreachable for the workspace sweep, and if it ever arises by another route the release goes red naming the missing packages, so the re-run is deliberate rather than a guess.

## Out of scope

`release.yml` is also touched by a sibling PR adding `--windowing` to the two gtk-runtime builder invocations. This PR only **appends** a job at the end of the file and changes no existing job or line.

Manifest-declared edges only: `@gjsify/gtk-runtime-*` is resolved by name at install time with no `optionalDependencies` entry, so there is no edge to check and its own publish jobs stay the guard.


JumpLink:
## Both findings fixed — `785145791`, `675ef7551`

Both were the batch's own failure class recurring inside the fix for it, so both are closed here rather than deferred. Nothing outside `scripts/verify-published-closure.mjs`, `tests/e2e/published-closure/run.mjs` and one AGENTS.md line changed — **`release.yml` is untouched by this push.**

### Finding 1 — the summary now carries the finding

Reproduced first, exactly as reported: the prefix fixture with `GITHUB_STEP_SUMMARY` set rendered `Pinned edges examined | 0` / `Unresolvable edges | 0`, no alert, exit 0, while the honest sentence went to stdout only.

The root cause was structural: the sentence was a `console.log`, and a print statement can only reach one channel. Noteworthy-but-not-fatal states are now **values** (`notes`), routed to all three:

```
## Published dependency closure

> [!WARNING]
> NO edge was examined: none of the 3 release-pinned edge(s) this tree declares has a published parent (1 package(s) absent at 1.2.3). Nothing on npm points at anything missing, so the closure is intact — but this run verified no edge, and an unverified closure is not a verified one. The publish job that stopped short is the finding: re-run the release (`gjsify publish --tolerate-republish` no-ops for what already landed).

| Fact | Value |
|---|---|
| Registry | `https://registry.npmjs.org` |
| Version | `1.2.3` |
| Candidate packages | 4 |
| Live at this version | 3 |
| Pinned edges examined | 0 |
| Unresolvable edges | 0 |
| Undecidable edges | 0 |
| Verdict | **NOTHING VERIFIED** — no edge was examined; see the warning above |
```

Three deliberate choices there:

- **The alert is above the table.** A reader who stops at the first numbers has already been told. `problems` moved up too, keeping `> [!CAUTION]` (the syntax this file already used — it is the only alert usage in the repo, so there was no other convention to match; `[!WARNING]` marks what the gate declined to fail but still insists on being read).
- **A `Verdict` row in words.** Every other row is a count, and a count needs a reader who already knows which ones are supposed to be zero.
- **`::warning title=Release closure::…` on stdout as well**, matching `cli-cross-platform.yml`'s existing use. It is the one channel that survives an unwritable summary file — the defect this PR already fixed once — and it renders *next to* the green check instead of behind it. `fail()` likewise now emits `::error::` under Actions, matching `verify-package-outputs.mjs` / `verify-committed-bundles.mjs`.

The e2e harness now strips `GITHUB_ACTIONS` from the child alongside `GITHUB_STEP_SUMMARY`, for the reason one level up: every negative case is a pass when it fails, and inheriting the flag would decorate the enclosing job with a dozen red annotations about fixtures. One case sets it deliberately and asserts the annotation.

### Finding 2 — an undecidable edge is a failure, with the target named

Reproduced both shapes first. Each needed a legitimate pinned edge alongside the bad one — otherwise assertion 2 fires and *masks* the finding, which is exactly why the drop stayed invisible:

| fixture | before | after |
|---|---|---|
| literal `"1.2.3"` pin, child manifest at `1.2.2` | exit 0, *"Every one of 1 release-pinned dependency edge(s) across 2 published package(s) resolves"* | **exit 1** |
| published parent pinning a `private` sibling | exit 0, same sentence | **exit 1** |

The enumeration now drops **exactly one shape**, and the header says so: a target that is not a package of this repository at all — genuinely the registry's problem. Every in-repo target is decided; `decide()` returns `release` or `undecidable` **with a reason**:

```
UNDECIDABLE EDGE: @fix/napi@1.2.3 → optionalDependencies.@fix/napi-darwin-arm64 = "1.2.3" — it pins 1.2.3,
but that package's own manifest is at 1.2.2 — this release does not publish it at 1.2.3, so the pin cannot
resolve. Put the two manifests on the same version.

UNDECIDABLE EDGE: @fix/app@1.2.3 → dependencies.@fix/secret = "workspace:^" — the target is a `private`
package — no release ever publishes it, so a pin on it from a PUBLISHED package is dead on arrival (npm would
resolve the NAME to whatever unrelated package owns it, or to nothing). Unmark the target or drop the
dependency.
```

Each is also listed under a `### Undecidable dependency edges` section in the summary — the aggregate caution cannot carry which target is at fault.

Both new reasons are decided from the **manifests**, with no probe involved, so they fail identically on a complete release, a partial one and a dry tree — including the case where nothing was published, where a probe-based check would never have examined the edge at all. `classify(spec)` became `decide(spec, target)` because deciding needs the target: `publishedSpec()` now models what `resolveWorkspaceProtocol` actually writes at pack time (`*`/`^`/`~` against the **target's own** version), which is precisely the mechanism that makes a `private` or off-train target produce a dead pin.

The third reason — an in-repo edge that packs to something other than the release version — is a failure too, per the header's existing rule that an intra-repo spec pinned to an older range does not exist here and an unrecognised shape fails loudly. A new case pins the counterweight: an edge into a package outside this repo (`left-pad`) must stay green, so widening the enumeration did not turn every external dep into a finding.

### Both nits closed

- **`:313`'s third assertion** now carries `pinnedEdges.length > 0`. Measured before/after on a no-declared-edge fixture: two errors including *"those three facts cannot all be true — the examination filter is broken"* → one. Zero declared and zero examined is the one arrangement in which all three *can* be true, and the case now asserts `ERROR:` appears exactly once.
- **`run.mjs:400`'s regex.** Each sweep is split at its `--exec`: group 1 is `foreach`'s own argv, group 2 the execed command, and the flag is required in group 1 — only for sweeps whose execed command is `gjsify publish`, so an unrelated preparatory `foreach` can neither satisfy the assertion nor spuriously fail it. A shape where no publish sweep can be located **fails**: an assertion that cannot find its subject has verified nothing.

### How I established the gates still fail on the bad input

Not by inspection — by mutation. Each mutant was applied to the merged code and the suite re-run; the point is not only that something went red, but that **only** the intended case did:

| mutant | red cases |
|---|---|
| restore `if (!candidateByName.has(to)) continue;` | *exactly* the 2 new shape cases (14 pass / 2 fail) |
| route the note to stdout only + drop the `Verdict` row | *exactly* the 2 summary cases (14 / 2) |
| remove the `pinnedEdges.length > 0` guard | *exactly* the no-declared-edge case (15 / 1) |
| move `--topological` past `--exec --` in `package.json` | *exactly* the order case (15 / 1) — **green before this PR's change** |
| remove `--topological` from `package.json` | *exactly* the order case (15 / 1) — no regression on what already worked |

Nine spellings of the sweep were checked against the tightened matcher offline as well (`-t` and the `-vt` cluster pass; flag removed / moved / only on a preparatory foreach / no locatable sweep all fail).

### Verification

- **Live registry, after the change: 194/194 packages at 0.27.1, 351 release-pinned edges (62 optional + 289 dependencies), 0 undecidable, exit 0.** The widened enumeration finds the *same 351* edges on the real tree (`--json`: `{"total":351,"pinned":351}`), so the two new failure modes are armed, not noisy — same measurement as before the fix.
- `tests/e2e/published-closure` **16/16** (was 12; +4: the summary-carries-the-finding case, the two new shapes, the external-dep counterweight; 3 existing cases gained assertions).
- `node scripts/audit-runtimes.mjs --check --strict` green; `scripts/check-ci-image-packages.mjs` green.
- `oxfmt --check` and `oxlint` clean on both changed files. No TypeScript changed, so `gjsify tsc` has no subject here.
- AGENTS.md's closure-gate paragraph amended to state both rules, so the doc and the gate cannot disagree: the one legitimate drop, and *anything the script finds noteworthy reaches the artifact a human opens*.

PR #976, opened 2026-08-04
fix(ci): stop rebasing over prebuild binaries

JumpLink:
## The prebuild refresh channel on `main` is dead, deterministically

`prebuilds.yml` run [30864276535](https://github.com/gjsify/gjsify/actions/runs/30864276535) (push, head `f7b9c46d9`) failed in **Commit prebuilds to repo**. The local bot commit `ba5ca42` was *"26 files changed, 26848 insertions(+)"* and created ten new `.gir` files. Then `git pull --rebase origin main`:

```
warning: Cannot merge binary files: packages/node/tls-native-darwin-x64/prebuilds/darwin-x64/libgjsifytls.dylib (HEAD vs. ba5ca42 ...)
CONFLICT (content): Merge conflict in ...libgjsifytls.dylib
Rebasing (1/1)
error: could not apply ba5ca42... chore: update native prebuilds [skip ci]
##[error]Process completed with exit code 1
```

Nothing was pushed.

**Cause:** the job COMMITS first and REBASES second. git has no merge driver for a binary file, so a replay of the prebuild commit onto a moved `main` conflicts on the first dylib. Not a race between runs — the next run conflicts identically. **Consequence: no 0.27.x could receive a rebuilt darwin dylib at all.**

### The "fresh build id" claim, measured rather than assumed

Diffed two consecutive bot commits of unchanged sources, `7f4e81291 → a03206649`:

* All **16** committed darwin dylibs changed. Every one at an **identical size**. **No** linux `.so` changed — the ELF legs reproduce exactly; only Mach-O does not.
* On **darwin-x64**, *every* differing byte is build identity and nothing else. For `libgjsifytls.dylib`, 25 of 44264 bytes:
  * **16 bytes** at offset 1160 — the `LC_UUID` payload, confirmed against the parsed load commands (`LC_UUID size=24 off=1152 payloadOffset=1160`), `a62ed19e…` → `777060d8…`. ld64 derives it per link.
  * **9 bytes** spread over three `N_OSO` debug-map stabs' `n_value` — each intermediate object file's **mtime**, `0x6a7083a6` → `0x6a7122e2` (2026-08-03T12:03:50Z → 23:23:14Z), for `meson-generated_src_vala_tls.c.o` and its two siblings.
* On **darwin-arm64** there is one extra block, inside `LC_CODE_SIGNATURE` (`dataoff=64704 datasize=640`) — the ad-hoc signature, whose hashes cover the header those bytes live in.
* **Zero** bytes of `__TEXT`, `__DATA_CONST`, the string table, the chained fixups or the exports trie differ, in any of the sixteen.

So this is a build id and not real content — which is the *worst* case for a rebase, because the artifacts are interchangeable and git still cannot merge them. Reproduce with `git show <sha>:<path>` on both sides and a Mach-O load-command walk.

## What changed

**1. Sync before staging.** New order: download → **snapshot the artifacts out of the worktree** → `fetch` + `reset --hard origin/main` → re-apply the snapshot → clear satisfied markers → stage → gate → commit → push. No rebase exists, so there is nothing for git to merge. The snapshot (`tar`, into `$RUNNER_TEMP`) is required because `reset --hard` restores tracked files and would otherwise discard the very binaries the job downloaded; being outside the repository, it also cannot be staged by accident.

**2. The remaining race, handled honestly.** `main` can still move between the sync and the push. The push retries up to five times and **each attempt re-syncs and RE-STAGES** — retrying the push alone would push a tree built on a stale base, the old rebase defect one layer out. Every attempt is logged (`push attempt N/5 (tree …)`, plus a `sync … attempt N` group). A push that succeeds while `origin/main` has since moved on is *not* treated as an error: it falls into the same re-stage, which answers the only question that matters — are our bytes in `origin/main`? — by having nothing left to stage. This also closes the *"Two `commit-prebuilds` jobs can race"* item in `status/open-todos.md`, where a rejected push discarded every binary the run had downloaded.

**3. `Gate the tree being pushed` is preserved, and no longer trusted to step order.** How I verified: parsed the job with `yaml` and listed its steps — the tail is `Snapshot the downloaded artifacts` → `Sync to origin/main and stage the artifacts` → `Gate on the measured glibc floor + libc flavour` → `Commit prebuilds` → `Gate the tree being pushed` → `Push`, so the gate still sits after staging and after the commit, and before the push. Its `if:` is unchanged (`steps.commit.outputs.changed == 'true'`, same condition as the push).

But position is only the readable form of the invariant, and a retry *does* change the tree after the gate ran. So it is now enforced: the gate script **stamps the tree hash it audited** (`git rev-parse 'HEAD^{tree}' > "$PREBUILD_GATED_TREE"`), the push refuses any tree that does not match it, and the retry **re-gates** before each further push. **No stamp file, no push** — that assertion, not the `if:`, is what makes the gate unskippable. A future `git pull --rebase` slipped between gate and push could not go unnoticed again.

Also moved into the staging script rather than left as steps: the *"Refuse to delete a committed prebuild"* guard and the *"nothing unstaged"* assertion. As steps they ran once, against a tree the retry then replaced; in the script they hold on every re-stage.

**4. The `.gir` blind spot** — same PR because it lives in the same job and would otherwise self-conflict.

Ten of the sixty committed prebuild directories have `.typelib` + `.so` and no `.gir`: `webgl-linux-{x64,arm64,ppc64,s390x,riscv64}` and `webrtc-native-linux-{…}`. Nothing caught it — `prebuild-artifacts` asked for "a library plus a typelib", `prebuild-libc` reads ELF, and the only `.gir` assertion in the tree is a shell loop in `prebuilds.yml`'s **macOS** job, which is precisely why the ten that got away were all linux.

**Runtime-relevant or not — established first, because the severity depends on it: NOT runtime.** girepository loads the `.typelib`; no path in `detectNativePackages()` / `buildNativeEnv()` opens a `.gir`, and `imports.gi.Gwebgl` works on all ten today. What a missing `.gir` breaks is everything *downstream of the committed artifact*: six bridges' `build:gir-types` scripts point `ts-for-gir --girDirectories` straight at `../<bridge>-linux-x64/prebuilds/linux-x64` (tls-native, http-soup-bridge, http2-native, terminal-native, sab-native, lightningcss-native), and the GIR-parsing `shared-library`-leaf assertion reads it. Not a coincidence that `webgl` and `webrtc-native` are the two bridges with *no* `build:gir-types` script: they are the two whose committed `.gir` was never there to point one at.

It is a **hard failure anyway, on provenance**: `stage-prebuild.mjs` matches artifacts by extension and `.gir` is in its list, so every directory the shared stager wrote has one — a directory without one was written by something else.

**Where it lives, and why:** in `prebuild-artifacts` (`packages/infra/manifest-conformance/lib/rules/prebuild-artifacts.mjs`), the `portable` rule that already OPENS every committed directory and already runs on the staged tree inside `commit-prebuilds` (`audit-runtimes --check` as the fast fail, `--check --strict` in the gate) as well as on every PR via `audit-runtimes.yml`. It is also the rule `status/open-todos.md` names as the place that did not check this. Checked per DIRECTORY, not per package — the shape that went wrong was one bridge shipping two different file sets, which a package-level "has a gir somewhere" would pass. Ordered after the library and typelib checks, so an empty directory still reports the one thing a consumer needs.

**The ten `.gir` files are NOT hand-committed** — that would repair the symptom and leave the channel broken. They are deferred with a mandatory reason in `scripts/manifest-conformance/prebuild-gir-gaps.mjs` (a caller option injected via `ctx.options`, exactly like the unchecked-field ledger, because the manifests it would otherwise live in are generated from derived fields only), printed on every run, and they must arrive **through the fixed channel** — that is the proof it works.

Both exits are automatic, because a self-retiring exemption with no clearing step is a trap on a `[skip ci]` push: `prebuild-artifacts` **fails** an entry whose `.gir` has landed, and `scripts/clear-satisfied-gir-gaps.mjs` (run beside its `platformsUncommitted` sibling, before staging) deletes that entry in the **same commit** that adds the file. Without the second half, the run that finally lands a `.gir` would push a self-contradictory tree and redden `main` for every open PR — verbatim what `platformsUncommitted` did on 2026-08-01.

## What the new gates would have caught

Each was verified by reintroducing the exact defect and watching it fail:

| Gate | Reintroduced defect | Result |
|---|---|---|
| no `git pull`/`git rebase` in `commit-prebuilds` | put `git pull --rebase origin main` back in `Commit prebuilds` | ✗ *"step "Commit prebuilds" runs git pull/rebase. This job commits BINARIES…"* |
| sync precedes the first `git add`, snapshot restored in between | moved a `git add` above the reset | ✗ *"stages before it syncs…"* |
| push compares `HEAD^{tree}` to the stamp | replaced the tree read with the stamp value | ✗ *"the push step must compare HEAD's tree against the stamp, or the stamp certifies nothing"* |
| `.gir` in every committed directory | emptied the ledger | ✗ `audit-runtimes --check` exit 1, **exactly 10** findings, e.g. `webgl-linux-x64 … holds no '.gir' (present: Gwebgl-0.1.typelib, libgwebgl.so)` |
| ledger cannot rot | — | rule fails an unused entry; an entry with no reason fails; an entry cannot excuse a missing library or typelib |

Everything here asserts a positive fact rather than exiting 0: the snapshot asserts a non-empty file set, the sync asserts `HEAD == origin/main`, the restore asserts every snapshot file is back on disk, the gate stamps what it read, the push asserts the stamp matches and reports the ref it landed on, and the clearing script re-imports the ledger and asserts the surviving key set.

## Validation

* `node scripts/audit-runtimes.mjs --check --strict` → exit 0, 60 directories verified, 10 `.gir` deferrals printed with their reason
* `node scripts/audit-test-scripts.mjs` → exit 0
* the gate's five e2e suites → **137 tests, 137 pass** (was 110 before the additions)
* `oxfmt --check .` → clean, 2588 files; `oxlint` repo-wide → **0 errors**, 57 warnings (the documented pre-existing count, unchanged)
* no TypeScript changed, and no source the committed GJS bundles inline, so no bundle rebuild is involved
* snapshot/restore shell exercised locally against the real tree: 187 files / 132 MB captured, 187 restored, assertion passes
* clearing script exercised in a sandbox: given 2 of 10 directories a `.gir`, it cleared exactly those two, left an entry whose file had not arrived, and printed only the ledger path on stdout; emptying it fully leaves an `oxfmt --check`-clean, `oxlint`-clean file

## One defect this PR introduced and fixed, worth reading

The first push carried `PREBUILD_SNAPSHOT: ${{ runner.temp }}/…` in `jobs.commit-prebuilds.env`. The `runner` context does not exist outside steps, so **GitHub refused to load the entire workflow** — [run 30890615702](https://github.com/gjsify/gjsify/actions/runs/30890615702), *"This run likely failed because of a workflow file issue"*, raised as a `push`-event run on a topic branch that `prebuilds.yml`'s `push` trigger does not even cover (a file GitHub cannot parse never gets its triggers evaluated). It is not a PR check: `gh pr checks` showed only passes. `python3 -c 'yaml.safe_load(...)'` parses the file fine, so local YAML validation says nothing about it.

That is the second instance of the class `status/open-todos.md` records as *"A workflow GitHub refuses to load reports the PR as GREEN"*, and the first that is a **context** error rather than a shape error — recorded there, because it constrains what the eventual general gate has to evaluate (expression scope, not just outline shape). Both paths are now published once from `$RUNNER_TEMP` in a step, and `tests/e2e/prebuild-change-gate` fails a `runner`/`steps` reference in any job-level `env:` block: a point fix for the one mistake in the one file, since a file that cannot be parsed cannot check itself.

## What a PR run cannot show

`commit-prebuilds` is `if: github.ref == 'refs/heads/main' && github.event_name != 'pull_request'`, so the restructured commit path does not execute on this PR — by design, and unchanged. What CI proves here is the build legs plus every static assertion about the job (no rebase, sync-before-stage, the stamp, the bounded re-staging retry, and the `.gir` rule). The commit path itself is exercised by the first `main` run after merge, which this PR's own diff triggers: `.github/workflows/prebuilds.yml`, `.github/prebuild-toolchain/**` and `scripts/manifest-conformance/**` are all in the `on: push: paths:` list, and a shared-input change rebuilds every package — which is also the run that should land the ten deferred `.gir` files and clear their ledger entries.

## Two more defects, found by rehearsing the step rather than reading it

Both in the snapshot step, both fixed here, both worth recording because they are the same class as the bug this PR is about.

**The pathspec matched nothing.** `git status --porcelain -- 'packages/*/*/prebuilds/'` reports a **clean tree on a dirty one** — git's default wildmatch runs without `WM_PATHNAME`, so `*` crosses `/` and the pattern has to match the whole path, which a file three levels deeper does not. The snapshot would have come out empty on every run, `Commit prebuilds` would have reported *"No prebuild changes to commit"*, and the job would have been **green with every downloaded binary discarded**. `:(glob)` switches on pathname-aware matching. Guarded by **two oracles**, because the failure is silent by nature: the scoped count is checked against an unscoped `git status` filtered for `/prebuilds/` (identical sets — every tracked `/prebuilds/` path here lives under `packages/<pillar>/<pkg>/`, 187 = 187 on main). With the bad pathspec the counts are 0 and 2 and the guard fires.

**Restoring all sixty directories could revert someone else's commit.** If `main` retires a target or renames a library while the run is in flight, re-applying every directory over the synced tree resurrects the deleted file — and `git add` stages a resurrection as an *addition*, so the deletion guard cannot see it. The snapshot is now scoped to what this run changed. Rehearsed end to end against a scratch repository: a rewritten artifact comes back with the new bytes, a brand-new target directory survives, and an artifact retired concurrently stays retired. It also takes the archive from ~132 MB to a few.

The restore assertion no longer treats zero as failure — an empty snapshot is legitimate, since the ELF legs reproduce byte for byte and a linux-only run can rebuild everything identically. What may never happen is a **shortfall**, which is what it now asserts; a separate positive check covers the other question (that the downloads put files in the tree at all).


JumpLink:
## The blocker was real, and the rehearsal reproduced it before the fix

Two reviewers found the same thing and one rehearsed it end to end. I reproduced both halves on the PR head before touching anything, because a fix for a "the gate would have failed" claim is worth nothing unless the gate is observed failing:

| Known-bad state, on the PR head as pushed | `node --test tests/e2e/prebuild-declaration-invariant/run.mjs` |
|---|---|
| ledger emptied (what `clear-satisfied-gir-gaps.mjs` produces) | **fail 1** — `AssertionError: the ledger needs at least two entries for this test to distinguish anything` |
| ledger file deleted (the documented human cleanup) | **0 tests ran** — `ERR_MODULE_NOT_FOUND` at load, because `audit-runtimes.mjs` imports the ledger at top level |

So: confirmed, and the second case is worse than reported — the whole suite fails to *load*, not one test.

## 1. The fixture is hermetic, and the fix is not a weaker bound

The suite's subject is now a **synthetic** ledger written by the test in the real one's load-bearing shape. Every state the real ledger passes through has its own test: entries present · partially cleared · last entry gone · **EMPTY** · **file ABSENT**. No count is asserted in either direction — a lower bound on this ledger is what failed on the very run it exists to permit.

Two reads of the real file survive, and both hold for ten entries, one, none, *or deleted*:

* **its shape is one the clearer can edit** — the clearer now exports `entryKeyOnLine()`, the grammar of the one line an entry occupies, so the assertion asks about the same grammar the surgery uses instead of paraphrasing the regex;
* **every static importer of the ledger is NAMED in the ledger's own "delete this file by hand" instruction** — the mechanism behind the ENOENT case above, rather than trust in it. A second importer appearing without that sentence being updated fails the suite, so the cleanup commit cannot break a script nobody was thinking about.

Anti-loosening is asserted, not claimed: with the ledger **drained**, a directory holding a library and a typelib and no `.gir` is asserted to fail again, and the clearer is asserted to only ever *remove* — it never learns about a gap that has no entry.

## 2. A defect the hermetic test found on the way

`clearSatisfiedGirGaps()` read the ledger with a bare `import('file://…')` and cache-busted only the *verify* read. A second call in one process therefore saw the first call's module, re-derived the entry it had already deleted, found no line for it, and threw

```
could not find a one-line entry for … Restore it, or teach this script the new shape
```

— blaming the ledger for being malformed while the ledger was correct and the script's view of it was stale. `sync-and-stage.sh` documents this function as idempotent ("the push step calls this again per attempt"); today's shell caller is saved only by accident, one process per invocation. Both reads now go through one `readLedger()` with a **monotonic counter**, not a timestamp — two calls in the same millisecond would hit the cache again, which is the bug rather than a smaller version of it.

## 3. The second half: what makes a PERMANENT deferral visible

The reviewers are right that "the files never arrive → the ledger never empties → everything keeps passing" was untouched. The entry's reason makes exactly one promise — *the next `prebuilds.yml` run that rebuilds this target lands the file* — and it can be false in two ways. **Both are now gates, and I chose gates over a stale-entry age deliberately**: an age reddens `main` on a date with no diff, for every open PR, which is the incident `gate-pushed-tree.sh` was written to stop. Each of these fires on a *fact*, in the job that owns it.

**(a) The build stops emitting the `.gir`** → `scripts/stage-prebuild.mjs` refuses to stage a `.typelib` with no `.gir` beside it. Not a taste rule: `g-ir-compiler` takes the `.gir` as **input**, and every meson build here runs it on a path inside the same build directory, so a `.typelib` in `build/` is *proof* a `.gir` was in `build/`. Deliberately **not** ledger-aware — an entry claims the next rebuild lands the file, so the rebuild is where the claim gets checked, not excused. Measured safe for every leg that runs today: all 60 committed directories hold a `.typelib`, and the 50 the stager produced hold a `.gir` too; the ten that do not are exactly the pre-stager `cp` lists it replaced. One e2e fixture described a build that cannot happen (typelib, no gir) and became honest rather than the check indulgent.

**(b) Nothing ever rebuilds the target** → new `auditGirGapArrival()` in the repo-scoped `platforms-ci` rule fails an entry whose package `prebuilds.yml` does not build *for the deferred target*. It reuses the coverage map that rule already parses, so it costs a comparison, and it catches the near miss: a `linux-x64` leg does not justify deferring `linux-riscv64`. **Advisory when the coverage map is empty** — then it is this parser that could not read the workflow, and turning a parser gap into a red `main` is the failure mode, not the signal.

Verified on this tree: all ten entries are credited by `prebuilds.yml` with exactly the target whose directory is short a file, so `audit-runtimes --check` now prints *which leg will end each deferral* rather than only its reason:

```
platform audit: OK. 74 native package(s) declare `gjsify.platforms`; …
  · @gjsify/webgl-linux-ppc64: `prebuilds.yml` builds linux-ppc64 — a run restages it.
  · @gjsify/webrtc-native-linux-riscv64: `prebuilds.yml` builds linux-riscv64 — a run restages it.
```

Producibility, stated because a deferral for a file that can never exist is the third way to be permanent: both bridges' `meson.build` ask for the GIR (`vala_gir:`), `webgl`'s two darwin directories ship theirs, and the committed `.typelib` is itself the proof — `g-ir-compiler` could not have produced it otherwise.

## 4. Rehearsed against the real gate script, not argued

`.github/prebuild-toolchain/gate-pushed-tree.sh` run in a throwaway worktree, one commit per state, replaying the merge-run sequence (land the `.gir` files → `clear-satisfied-gir-gaps.mjs` → commit → gate):

| State | Ledger | `gate-pushed-tree.sh` |
|---|---|---|
| 1 — as merged | 10 entries | **exit 0** (153 tests) |
| 2 — webgl's five land, clearer runs | 5 entries | **exit 0** (153 tests) |
| 3 — the rest land, clearer empties it | `PREBUILD_GIR_GAPS = {}` | **exit 0** (153 tests) ← *this is the state that used to fail* |
| 4a — file deleted, import left behind | absent | **exit 1**, `ERR_MODULE_NOT_FOUND` — correct and loud |
| 4 — file deleted **and** its import removed (the documented cleanup) | absent | **exit 0** (153 tests) |

The emptied ledger the clearer produced is `oxfmt --check`-clean, and the five `.gir` files cleared exactly their own five entries and left the other five describing reality.

**And the gate still fails on every known-bad input** — same script, same worktree:

| Known-bad input | Result |
|---|---|
| committed directory missing its `.gir`, ledger gone (no excuse available) | **exit 1** — `` holds no `.gir` `` |
| ledger entry whose `.gir` HAS landed (stale entry) | **exit 1** — `nothing in this tree matched it` |
| ledger entry with no reason | **exit 1** — `with no reason` |
| deferral for a target no `prebuilds.yml` leg builds (`@gjsify/napi-darwin-arm64`, built by `napi.yml` only) | **exit 1** — ``no job in `prebuilds.yml` builds this package for darwin-arm64`` … `PERMANENT` |
| stager given a `.typelib` with no `.gir` | **exit 1**, and **nothing written** — no half-staged directory for a later step to upload |

The last one's complement passes as soon as the `.gir` is present, so the check is "the set is complete", not "always demand a `.gir`" — a library with no typelib at all still stages (asserted).

## Validation

* five gate suites: **153 tests, 153 pass** (was 137)
* `node scripts/audit-runtimes.mjs --check --strict` → **exit 0**; `node scripts/audit-test-scripts.mjs` → **exit 0**
* `oxfmt --check .` → clean, 2588 files; `oxlint` → **0 errors**, 57 warnings (the documented pre-existing count, unchanged)
* no TypeScript changed and no source the committed GJS bundles inline, so `gjsify tsc` is not applicable and no bundle rebuild is involved

## Recorded, and follow-ups

`status/open-todos.md` gains the class rather than the instance: **this is the SECOND time a gate's fixture read state the gated job writes** — the first (`platform-exemption-clearing`, f5d250b32) is the incident `gate-pushed-tree.sh`'s own header cites as its reason for existing, and the second appeared *inside the fix for it*. Twice is a class, so the entry names the property that would catch a third and says why the obvious rule is wrong: several gated suites read committed artifacts **on purpose** (`prebuild-loader-path` asserts exact glibc floors of bytes this job replaces), so it cannot be "fixtures may not read the tree". First cut that would have caught both: fail a gate-listed suite that reads a path the same job's clearing scripts print on stdout — a set the job already computes. Not built here; it is a gate on the gate list and deserves its own change.

Two notes for the reviewers:

* Only the blocker text reached me — if there were non-blocker findings, they were not relayed, so please repeat them and I will fold them in.
* The PR description still says "137 tests" and does not mention the two arrival gates; the numbers above supersede it.

PR #977, opened 2026-08-04
fix: make the GTK runtime bundles self-contained

JumpLink:
## The measured failure

Measured on the **published 0.27.1 tarballs** (`npm pack @gjsify/gtk-runtime-*@0.27.1`), not on the repo:

| tarball | files | unpacked | native | typelibs | `windowing` | `dataBytes` | license files |
|---|---|---|---|---|---|---|---|
| `gtk-runtime-darwin-x64` | 88 | 29,920,533 B | 45 dylibs | 38 | `false` | `0` | **none** |
| `gtk-runtime-darwin-arm64` | 88 | 31,459,350 B | 45 dylibs | 38 | `false` | `0` | **none** |
| `gtk-runtime-win32-x64` | 80 | 36,350,027 B | 37 DLLs | 37 | `false` | `0` | **none** |

Three defects, all the same class — a missing signal read as a pass:

**1. Typelibs whose backing library is not in the bundle.** Parsing every shipped typelib's `shared_library` header field and comparing it against the bundle's own `lib/`/`bin/`:

```
darwin-x64 / darwin-arm64 — 6 of 38 unbacked
  Adw-1            wants libadwaita-1.0.dylib          MISSING
  AppStream-1.0    wants libappstream.5.dylib          MISSING
  GIRepository-2.0 wants libgirepository-1.0.1.dylib   MISSING
  HarfBuzz-0.0     wants libharfbuzz-gobject.0.dylib   MISSING
  Rsvg-2.0         wants librsvg-2.2.dylib             MISSING
  Xmlb-2.0         wants libxmlb.2.dylib               MISSING
win32-x64 — 3 of 37 unbacked
  Adw-1            wants adwaita-1-0.dll               MISSING
  GtkSource-5      wants gtksourceview-5-0.dll         MISSING
  Rsvg-2.0         wants rsvg-2-2.dll                  MISSING
```

GI resolves a namespace's symbols with `g_module_open(<shared_library>)`, so such a namespace **resolves**, advertises its classes, and then dies in the constructor with `Failed to load shared library '…'` — the type is not a constructible GObject. That is worse than an absent typelib, which fails cleanly at `require`.

**2. No runtime data at all.** No `share/`: no `gschemas.compiled` (a hard `Gio.Settings` startup blocker), no icon theme, no GtkSource language-specs. GSettings-backed APIs and icon lookups then misbehave *silently*.

**3. No license text of any kind**, while each tarball carries 37–45 relocated LGPL/MPL/GPL libraries — and on darwin *modified* ones (`install_name_tool` rewrites every install name to `@loader_path`, then an ad-hoc `codesign`).

## The cause

`release.yml` invoked both builders **without `--windowing`**, so the published artifact was the display-free *conformance* variant; and both builders copied the **source prefix's entire typelib directory**, which covers every installed formula rather than the bundle's own closure. Nothing compared the two sets, so the mismatch was invisible.

## The gate

`packages/node-gi/scripts/typelib-backers.mjs`, shared by both builders (the win32 one imports it across `packages/node-gi/`):

- **Every typelib in the OUTPUT bundle must have its backing library in that same bundle**, else the build fails naming the offending pairs.
- **The mapping is derived from the typelib itself** — the `shared_library` field of girepository's header, comma-split (`GLib-2.0` names *two* backers, and a check that stopped at the first would pass a bundle missing the second), empty for the header-only namespaces that legitimately have none (`xlib`, `freetype2`, `Vulkan`, `DBus`, …). Never a hand-written leaf table: the same namespace is `libadwaita-1.0.dylib` on brew, `adwaita-1-0.dll` on gvsbuild, `libadwaita-1.so.0` on Linux.
- **A mis-parse throws instead of degrading.** An unreadable header must not look like "header-only, no backer needed" — that would make the whole gate pass vacuously.
- **The drop decision is dependency-aware**, because the naive filter *regresses* the bundle: measured, `Pango-1.0` **depends on** `HarfBuzz-0.0`, whose backer `libharfbuzz-gobject` was in the win32 closure only by accident of a seed pattern and never in the darwin one. `gi_repository_require` loads a typelib's dependencies first, so dropping HarfBuzz would have broken Pango — and with it Gdk, Gsk, Gtk. An unbacked typelib something kept depends on is therefore a **hard failure naming the missing library**; the repair is the seed, which is why `libharfbuzz-gobject` is now a base darwin seed.
- **It asserts positive facts, not exit 0**: it re-reads *both* sets off the finished bundle (not the plan that produced it), requires ≥ 1 library-backed typelib, and requires every namespace the packages promise (`REQUIRED_NAMESPACES`, plus `Adw` + `GtkSource` under `--windowing`). So a filter that quietly removed everything, or an empty typelib dir, cannot pass green. `manifest.json` records the outcome under `typelibSymmetry`, including each dropped namespace and the library it wanted.

**What it would have caught:** the 0.27.1 publish. Run against the three published tarballs the gate reports 6/6/3 unbacked pairs plus, with the windowing floor, `required namespace Adw is absent`.

## The content decision the gate forces

- the **display-free conformance** bundle (`node-gi.yml`) drops the typelibs it cannot back — its closure is exactly what the display-free conformance loads, nothing more;
- the **published** bundle is the `--windowing` **superset**, which is what that flag exists for: libadwaita + libgtksourceview + compiled GSettings schemas + Adwaita/hicolor icon themes + GtkSource data. `release.yml` now passes `--windowing` on all three legs, and its brew install gains `gtksourceview5` + `adwaita-icon-theme` (mirroring `node-gi.yml`'s windowing job) so nothing is seeded that the prefix cannot provide. Both release verify steps assert the *artifact*: `Adw-1.typelib` beside a libadwaita image, `gschemas.compiled`, `share/icons`, the notice, and a manifest saying `windowing: true` with non-zero `dataBytes`.

## License compliance

`packages/node-gi/scripts/bundle-licenses.mjs` collects the terms **from the source prefix at build time**, so the set cannot drift from the closure, and ships them as `gtk/licenses/<component>/` + `gtk/THIRD-PARTY-NOTICES.md` (recorded in the manifest under `licenses`).

- **darwin — per-binary attribution.** Every bundled dylib realpaths through `…/Cellar/<formula>/<version>/` and Homebrew keeps the formula *inside* the keg (`.brew/<formula>.rb`), so both the attribution and the license expression come from the prefix. An unattributable dylib **fails the build**. The notice states the modification the LGPL asks about: `install_name_tool` + ad-hoc `codesign`.
- **win32 — prefix-wide attribution, stated as such.** A flat gvsbuild `bin/` cannot prove a per-DLL mapping, and the DLLs' `VERSIONINFO` cannot either (measured: present in glib/gtk/zlib, absent in adwaita/harfbuzz/libpng), so the mapping is *not invented*: the whole documented corpus ships — measured **37 texts from 31 projects, 454 KiB** in `GTK4_Gvsbuild_2026.6.0_x64` — and the notice says plainly that the per-binary mapping is not recoverable.
- **Zero recovered texts is a build failure**, not a quiet omission.
- Each package also gains its own MIT `LICENSE`; the tarballs had none.

## Manette — decided, not asserted

**Not added as a seed.** Three reasons, in order of decisiveness:

1. **There is nothing to seed from.** homebrew-core has no `libmanette` formula (`formulae.brew.sh/api/formula/libmanette.json` → HTTP 404) and `GTK4_Gvsbuild_2026.6.0_x64.zip` contains **zero** manette/evdev entries. The pattern would match nothing, and with the new symmetry rule a Manette typelib could not ship anyway — adding the pattern would be a no-op that *looks* like support.
2. **Even with a dylib it would be the same lie one layer down.** libmanette's backend reads Linux `/dev/input/event*` through evdev/udev; a `Manette.Monitor` on macOS/Windows would satisfy every symmetry check and enumerate nothing forever.
3. **Scope.** The principled rule has two halves. "The bundle ships the closure of exactly the typelibs it ships" is unconditionally right and is what this PR implements. "The typelib set follows what the workspace imports via `gi://`" cannot be read as *all* such imports: the workspace also imports `gi://Gst` ×17, `gi://GstWebRTC` ×11, `gi://WebKit` ×4, `Soup`, `Gda`, `JavaScriptCore`, `X`, `GstApp`, `GstSdp` — WebKitGTK alone dwarfs the bundle, and `gi://X` is meaningless on win32. Half 2 has to be scoped to what the bundle *claims* (its description + the `--windowing` superset); the rest needs a per-namespace availability contract, not an ever-growing bundle. Manette is the smallest of ten instances, and picking it alone would be arbitrary.

The real defect is in the consumer and is now recorded with its evidence in `status/open-todos.md`: `packages/web/gamepad/src/gamepad-manager.ts:58` catches the failed `await import('gi://Manette')`, sets `_initialized = true` and returns, so `navigator.getGamepads()` answers `[]` with no diagnostic — indistinguishable from "no controller plugged in", on a tier-1 package. That path has never run on macOS or Windows. The fix belongs there (make unavailability observable; a platform backend if gamepads are actually wanted), not in the bundle's seed list.

## Sizes, before → after

**Corrected 2026-08-04.** The earlier version of this table claimed darwin-arm64 "~37 MiB, 35 typelibs", which this PR's own CI refutes: 37 MiB was the pre-dereference measurement (0.2 MiB of links where the icon theme is 19 MiB of files), and 35 was derived from the published tarball's headers rather than measured. Every number below is read off a CI log — the darwin-x64 row for the first time, because until this push nothing built that bundle.

| platform | before (published 0.27.1, unpacked) | after (measured, run 30905012513) |
|---|---|---|
| win32-x64 | 34.7 MiB (36,350,027 B), 37 typelibs (3 unloadable), no data, no licenses | **77.8 MiB** — 41 DLLs 69.9 + 37 typelibs (25 backed + 12 header-only, all load) 2.2 + data 5.4 + 37 license texts 0.4 |
| darwin-arm64 | 30.0 MiB (31,459,350 B), 38 typelibs (6 unloadable), no data, no licenses | **56.5 MiB** — 52 dylibs 34.3 + 37 typelibs (25 + 12, all load) 2.3 + data 19.3 + 65 license texts 0.6 |
| darwin-x64 | 28.5 MiB (29,920,533 B), 38 typelibs (6 unloadable), no data, no licenses | **54.7 MiB** — 52 dylibs 32.5 + 37 typelibs (25 + 12, all load) 2.3 + data 19.3 + 65 license texts 0.6 |

Per-leg detail, all from run **30905012513** (the run of this branch's current head):

- **`Windows FULL-windowing GTK runtime bundle`** — 77.8 MiB by the manifest, **77.9 MiB** by `Get-ChildItem` on disk. Data gate: `schemas (1 file), icons (820), gtksource (6), pixbuf-loaders (2)`. The display-free variant in the same run: `34 typelibs (22 backed + 12 header-only)`, total 34.8 MiB, dropping exactly `Adw-1 → adwaita-1-0.dll`, `GtkSource-5 → gtksourceview-5-0.dll`, `Rsvg-2.0 → rsvg-2-2.dll`.
- **`macOS windowing GTK runtime bundle / darwin-arm64`** — `totalBytes: 59222513` (56.5 MiB), `dataBytes: 20247017`, `du -sh` **64M**. Data gate: `schemas (1), icons (863), gtksource (6)`.
- **`macOS windowing GTK runtime bundle / darwin-x64`** — `totalBytes: 57345325` (54.7 MiB), same `dataBytes: 20247017`, `du -sh` **63M**. Identical typelib set and closure size (52 dylibs); only the Mach-O bytes differ (32.5 vs 34.3 MiB).

Both darwin legs drop the same two typelibs, both correctly: `GIRepository-2.0 → libgirepository-1.0.1.dylib` and `Rsvg-2.0 → librsvg-2.2.dylib`.

The `du -sh` figures exceed `totalBytes` by ~7 MiB because ~960 of the files are small (icons, license texts) and occupy a 4 KiB block each — block overhead, not a manifest/reality disagreement. Before the `lstat` fix the arm64 manifest reported 19.4 MiB of data that was not on disk at all.

The win32 jump is ANGLE (`libGLESv2`/`libEGL`) arriving with the windowing seeds; the darwin jump is the icon theme (19 of the 26 MiB). Size is not a constraint: no `package.json` in the workspace depends on `gtk-runtime-*` — verified by grep — they are opt-in installs.

## The declared data must BE there, and the shipped shape must be exercised

Three follow-ups from review, all the same class as the defect above — a missing signal read as a pass — fixed here rather than deferred.

**1. A missing declared data set now FAILS the publish path.** Both builders' data steps were warn-and-continue, so a bundle could publish with `share/gtksourceview-5` (or the schemas, or the icon theme) absent and nothing failed — which is exactly how `"dataBytes": 0` shipped. `verifyWindowingData()` (shared `bundle-data.mjs`, called by both builders after the typelib symmetry check) re-reads the FINISHED bundle and requires every set whose namespace that bundle ships: `Gio` → a non-empty `gschemas.compiled`, `Gtk` → `share/icons/*/index.theme` **and** files under it, `GtkSource` → a non-empty `share/gtksourceview-5`, plus win32's `loaders.cache` + a loader. Derived from the shipped typelibs rather than a second list, so it is the data-side twin of the typelib rule and cannot be dodged by declaring less (the `--windowing` namespace floor forces those namespaces to be present). It asserts BYTES, not directories, and an empty applied-set list is itself a failure.

*Why the display-free variant keeps no relaxed copy of it:* that bundle declares no data at all — no `share/` tree, `windowing: false`, the loader keys the whole windowing env on the `gschemas.compiled` marker and wires none of it, and the backer filter has already dropped `Adw`/`GtkSource`. The same rule holds there; it has no subject. The gdk-pixbuf loaders are win32-only for the same honest reason: darwin does not bundle them yet (tracked), so it does not claim them.

*Rehearsed against the bad input*, on Linux: the gate passes on a complete copy of this host's real GTK prefix (`schemas 1, icons 1674, gtksource 6`), FAILS with each set removed in turn, and FAILS the published 0.27.1 shape (`lib/` + `girepository-1.0/`, no `share/`) with 4 problems on darwin and 6 on win32.

Found while writing it, measured: GtkSourceView 5 compiles its built-in language-specs/styles/snippets into a **GResource inside the library** (198 resources via `gresource list`), so the highlighting travels with the dylib; `share/gtksourceview-5` carries the RNG/DTD schemas that validate USER `.lang`/`.snippets` files + `BuilderBlocks.ttf`. The old comments claimed the opposite, and the copy took only `language-specs` + `styles`, silently dropping `snippets/` and `fonts/` — now the whole tree.

**2. The darwin-x64 windowing bundle is now exercised.** `release.yml` publishes `--windowing` on both darwin legs, but `macos-gtk-windowing-runtime` → `macos-gtk-windowing` was arm64-only: the bundle an Intel Mac installs was built by the release and tested by nothing. The display-free x64 leg does not cover it — different artifact (no libadwaita, no libgtksourceview, no data tree). Both jobs are now a two-arch matrix on the runner labels `macos-gtk-runtime` already uses, and the reason is in the workflow comment.

**3. The conformance now runs against the shape that ships.** `macos-gtk-batteries-included` (and `windows-batteries-included`) consume the DISPLAY-FREE bundle, so no conformance had ever run on the published superset. Both windowing proof jobs now run `check-batteries.mjs` + the full `cross-runtime.mjs node` subset on the published bundle before the GUI proof. **Both shapes run, and the authority is stated in each job:** the windowing pair is authoritative for "what a consumer installs works"; the display-free pair is the minimal-closure check (evidence the backer-drop filter left a closure the conformance still passes on, and that bundle is what node-gi's own prebuilds stage).

Also tightened, same class: the windowing uploads were `if-no-files-found: warn` (an empty artifact downloads fine and fails four steps later as a dlopen error), and both windowing consumers now assert the bundle's own manifest stamp — `platform` matches the runner AND `windowing: true, dataBytes > 0` — now that two arches × two variants share one artifact namespace. `release.yml` additionally requires `windowingData.verified` in the manifest, so a bundle built by an older builder cannot publish.

And the AGENTS.md paragraph that instructed `dereference: true` is corrected: the code deliberately does not use that flag (it governs only the top-level path and left 859 links in place), so the doc now names `copyTreeDereferenced()` and records the flag as the trap a test pins. Resolved in favour of the code, because for a bundle that ships to another machine "real files" is the requirement — `npm pack` archives a symlink as a symlink, and a link into a Homebrew Cellar is worthless in a tarball.

## Tested

Both modules are pure and platform-agnostic on purpose: the builders only ever run on macOS/Windows runners, so until now their logic was exercised on no machine a developer has. `packages/node-gi/node-gi/test/gtk-runtime-bundle-gates.test.mjs` (**28 tests**, runs in `node-gi.yml`'s `build-test` job) drives them on Linux from a **synthesised** typelib — which pins the header offsets by assertion instead of by comment — and from the host's own ~200-file typelib corpus, asserting among others: the comma-split backer list, the mis-parse throw, the `Pango → HarfBuzz` refusal-to-drop, the required-namespace floor, the vacuous-empty-bundle rejection, win32-case-insensitive vs darwin-case-sensitive matching, keg attribution + the unattributed-binary failure, that the notice names every bundled binary, the `cpSync dereference: true` trap, and — for the data gate — one missing set at a time, the empty-tree / zero-byte shapes, the namespace derivation, the win32-only loaders and the vacuous no-subject case.

Also fixed on the way: `dirSize` followed symlinks, so the manifest counted every icon-theme alias at its target's size — the arm64 `--windowing` manifest claimed 19.4 MiB of runtime data inside a bundle `du -sh` measured at 37 MiB. `lstat` now: a size the manifest reports must be the size on disk. And `files` no longer ships the win32 builder, which imports the shared rules from `packages/node-gi/scripts/` and would otherwise be a script with dangling imports in the tarball (`manifest.builder` already names the recipe).


JumpLink:
Review fixes pushed — `08deb8f51`, `b8b02151d`, `553d93e5b`. All five findings; the first two were this batch's own failure class recurring inside the fix, so they are closed with a gate rather than a note.

### 1. `dereference` — code and doc now agree, resolved in favour of the CODE

The doc said "copy it with `dereference: true`"; the code deliberately does not use that flag (`45b2bdfb9` removed it) because it governs only the stat of the path handed to `cpSync` and leaves every NESTED link a link — measured on the macOS runner: the first attempt passed the flag and the gate still found **859 links** under `share/icons/Adwaita`, each into `/opt/homebrew/Cellar/adwaita-icon-theme/50.0/…`.

**Real files is the correct requirement** for a bundle that ships to another machine: `npm pack` archives a symlink as a symlink, so a link into a Homebrew Cellar path is worthless in a tarball — the consumer gets a dangling pointer, and the build host is the only place it ever resolved. So the code stays (`copyTreeDereferenced()`, a walk that `statSync`s each entry and `copyFileSync`s the bytes) and AGENTS.md now names it, records `cpSync`'s flag as the trap, and keeps the pinning test. Two stale `dirSize` comments that also described the pre-dereference code were corrected in the same pass.

### 2. Missing declared data now FAILS on the publish path — and the failure is rehearsed

`verifyWindowingData()` in the shared `bundle-data.mjs`, called by BOTH builders after the typelib symmetry check, re-reads the FINISHED bundle. A data set is required **iff the bundle ships the namespace it belongs to** (namespaces taken from the typelib set just read off disk, not from the copy plan):

| set | required when the bundle ships | what must be there |
|---|---|---|
| `schemas` | `Gio` | a non-empty `share/glib-2.0/schemas/gschemas.compiled` |
| `icons` | `Gtk` | `share/icons/*/index.theme` **and** ≥1 file under `share/icons` |
| `gtksource` | `GtkSource` | ≥1 non-empty file under `share/gtksourceview-5` |
| `pixbuf-loaders` | `GdkPixbuf` (win32 only) | `loaders.cache` + ≥1 loader |

Derived rather than a second hand-kept list, so it is the data-side twin of "a bundle ships the closure of exactly the typelibs it ships", and it cannot be dodged by declaring less: the `--windowing` namespace floor already forces `Gio`/`Gtk`/`Adw`/`GtkSource`. It asserts BYTES, not directories (a `test -d` is satisfied by an empty dir and by a link into the build host's Cellar), and an **empty applied-set list is itself a failure**, so it cannot pass vacuously.

**Why the display-free variant differs, and why it is not a relaxed copy of the gate:** that bundle *declares no data*. No `share/` tree, `windowing: false`, node-gi's loader keys the entire windowing env on the `gschemas.compiled` marker and wires none of it, and the backer filter has already dropped `Adw`/`GtkSource` (libadwaita/libgtksourceview are not in that closure). The same rule holds there — it has no subject. The gdk-pixbuf loaders are win32-only for the same honest reason: the darwin builder does not bundle them yet (tracked in `status/open-todos.md`, and that entry now says so explicitly), so it does not claim them. Assert what you ship.

**How I established the gate still fails on the bad input** — three rehearsals, two of them on real prefixes:

1. **Unit, on Linux** — 8 new cases (28 total, green): one missing set at a time, the empty-tree and zero-byte-file shapes, an icon theme with an index but no icons, the namespace derivation (display-free shape → `gtksource` skipped, the others still applied), the win32-only loaders, and the vacuous "no set applied" case.
2. **Against this host's REAL GTK prefix** — built a bundle-shaped tree from `/usr/share` the way the builders do (`schemas 1, icons 1674, gtksource 6`): PASSES complete, and FAILS with each of the three sets removed in turn.
3. **Against the published 0.27.1 shape** (`lib/` + `girepository-1.0/`, no `share/` at all — the bytes that actually shipped): FAILS with **4 problems** on darwin and **6** on win32.

And the positive direction is now CI-measured on both real build prefixes — run 30905012513: `windowing data verified — schemas (1 file(s)), icons (820), gtksource (6), pixbuf-loaders (2)` on gvsbuild, `schemas (1), icons (863), gtksource (6)` on both Homebrew legs.

Found while writing it, measured with `gresource list`: **GtkSourceView 5 compiles its built-in language-specs/styles/snippets into a GResource inside the library** (198 resources), so the highlighting travels with the dylib. `share/gtksourceview-5` carries the RNG/DTD schemas that validate USER `.lang`/`.snippets` files plus `BuilderBlocks.ttf`. The old comments claimed the opposite ("the editor's built-in highlighting will be unavailable"), and the copy took only `language-specs` + `styles`, silently dropping `snippets/` and `fonts/` — now the whole tree, which is why the count went 4 → 6 files on both platforms.

### 3. x64 windowing coverage — the gap is CLOSED, not explained away

x64 *can* be covered, so it is: both `macos-gtk-windowing-runtime` and `macos-gtk-windowing` are now a two-arch matrix on the same runner labels `macos-gtk-runtime` already uses, artifact names carry the arch, and the WORKFLOW COMMENT states the reason (the release publishes the superset for darwin-x64, so the bundle an Intel Mac installs must be exercised; the display-free x64 leg does not cover it because it consumes a different artifact). The `status/open-todos.md` entry that tracked this is deleted. **Verified, run 30905012513:** `macOS windowing GTK runtime bundle (build + relocate / darwin-x64)` pass 1m41s, `macOS windowing proof (no brew GTK / darwin-x64)` pass 41s — including the GUI render leg, so no scoping caveat is needed. The one KNOWN gap is unchanged and shared by both arches: no gdk-pixbuf loaders on darwin.

### 4. The conformance now consumes the shipped shape — both run, authority stated

`macos-gtk-batteries-included` and `windows-batteries-included` consume the DISPLAY-FREE bundle, so no conformance had ever run against the published superset. Both windowing proof jobs now run `check-batteries.mjs` + the full `cross-runtime.mjs node` subset on the PUBLISHED bundle before the GUI proof (verified running, not skipped: 4/4 legs green — darwin-arm64, darwin-x64, win32-x64).

**Which is authoritative** — written into both job comments, not just here: the **windowing pair is authoritative** for "what a consumer installs works". The display-free pair stays as the MINIMAL-CLOSURE check: evidence that the backer-drop filter left a closure the conformance still passes on, and that bundle is what node-gi's own `prebuilds/` stage.

Also tightened, same class: the windowing uploads were `if-no-files-found: warn` (an empty artifact downloads fine and fails four steps later as a dlopen error), and both windowing consumers now assert the bundle's own manifest stamp — `platform` matches the runner AND `windowing: true, dataBytes > 0` — now that two arches × two variants share one artifact namespace. `release.yml`'s two verify steps assert each data set on bytes and additionally require `windowingData.verified` in the manifest, so a bundle built by an older builder (or with the gate bypassed) cannot publish. That `node -e` inside `pwsh` pattern is no longer unrehearsed either: the win32 windowing job now runs the same shape on every PR.

### 5. Size table corrected from this PR's own CI

The body claimed darwin-arm64 "~37 MiB, 35 typelibs". Both halves were wrong: 37 MiB was the PRE-dereference measurement (0.2 MiB of links where the icon theme is 19 MiB of files), and 35 was derived from the published tarball's headers rather than measured. Re-derived from run **30905012513**, which is also the first run that ever built the darwin-x64 windowing bundle:

| platform | before (published 0.27.1, unpacked) | after (measured) |
|---|---|---|
| win32-x64 | 34.7 MiB (36,350,027 B), 37 typelibs (3 unloadable) | **77.8 MiB** — 41 DLLs 69.9 + 37 typelibs (25 backed + 12 header-only) 2.2 + data 5.4 + 37 license texts 0.4 (`Get-ChildItem`: 77.9 MiB) |
| darwin-arm64 | 30.0 MiB (31,459,350 B), 38 typelibs (6 unloadable) | **56.5 MiB** — 52 dylibs 34.3 + 37 typelibs (25 + 12) 2.3 + data 19.3 + 65 license texts 0.6 (`totalBytes: 59222513`, `du -sh` 64M) |
| darwin-x64 | 28.5 MiB (29,920,533 B), 38 typelibs (6 unloadable) | **54.7 MiB** — 52 dylibs 32.5 + 37 typelibs (25 + 12) 2.3 + data 19.3 + 65 license texts 0.6 (`totalBytes: 57345325`, `du -sh` 63M) |

37 typelibs, not 35: both darwin legs drop exactly two, `GIRepository-2.0 → libgirepository-1.0.1.dylib` and `Rsvg-2.0 → librsvg-2.2.dylib`. `du -sh` exceeds `totalBytes` by ~7 MiB because ~960 of the files are small and take a 4 KiB block each — block overhead, not a manifest/reality disagreement (before the `lstat` fix the manifest reported 19.4 MiB of data that was not on disk at all).

### Validation

`node --test test/gtk-runtime-bundle-gates.test.mjs` → **28 pass / 0 fail**. `oxfmt --check packages/node-gi/` clean; `oxlint` clean on every changed file (the three warnings it reports in `packages/node-gi/` are pre-existing, in files this push does not touch). No TypeScript changed, so `gjsify tsc` has no new subject. Both workflows re-parsed after editing. CI on `553d93e5b`: every node-gi leg green, including the four new ones.

PR #978, opened 2026-08-04
feat(devtools): reach the control plane without a bus

JumpLink:
## What

`@gjsify/devtools` is now reachable **without a session bus**, so macOS and Windows need one env var (or nothing at all) instead of an external daemon.

The capture never needed fixing: `captureWidgetPng` is pure in-process GTK4/GSK (`get_native().get_renderer()` → `Gtk.WidgetPaintable` → `render_texture` → `save_to_png_bytes`) with no grim, no portal, no GStreamer and no OS branch — CI-proven on darwin-arm64 by the `macos-gtk-windowing` job. The **transport** was the only Linux-shaped part, and on a bus-less host `installDevtools` constructed a `DevtoolsService` and never exported it: one stderr line, then every call failing from the other side.

- `startDevtoolsPeerServer(service, objectPath, address?)` — a `Gio.DBusServer` exporting the SAME `org.gjsify.Devtools` interface on each incoming peer connection through the unchanged `DevtoolsService.export()`. Standard GDBus peer machinery (generated GUID, no auth observer needed), no bespoke protocol.
- `DevtoolsService` keys its exported objects **by connection** — a peer server hands out one connection per client with no shared registry, so the old single slot served the FIRST client and left every later one seeing `UnknownMethod` for all 26 methods.
- Client side: `DbusDevtoolsClient` accepts an address, sends **no destination bus name** on a peer connection, and answers `nameHasOwner`/`listInstances` from the connection instead of the daemon. `gjsify debug --address` (or `gjsify.devtools.address`) threads it through.
- `packages/node-gi/node-gi` `test:dbus` gets `--config-file=test/session.conf` (a plain `unix:tmpdir=/tmp` bus) so the suite is runnable on macOS at all. Linux unchanged (verified locally).
- Docs: a `## No session bus? (macOS, Windows)` section in both READMEs with the busless recipe, the `dbus-run-session` trap, the 94 s cold-start caveat and "poll `GetStatus`, do not sleep".

## The precedence I chose, and why

App side (`chooseDevtoolsTransport`, a pure function so the table is spec-checked rather than prose):

| explicit address | app holds a bus connection | session bus answers | transport |
|---|---|---|---|
| set   | any | any | peer server at that address |
| unset | yes | any | the session bus — **Linux byte-unchanged** |
| unset | no  | no  | peer server at an **auto-picked** socket |
| unset | no  | yes | none: the call-site diagnostic |

**Row 3 auto-picks rather than failing loudly**, because `GJSIFY_DEVTOOLS=1` is an explicit request for a control plane and "no control plane, one stderr line" is exactly the silent absence this work removes. It cannot regress Linux: a host whose bus answers never reaches that row. The auto-picked address is logged *and* published to `<runtime-dir>/gjsify-devtools/<app-id>[.<instance>].address` (deleted on `GApplication::shutdown`) — the bus-less analogue of `DBUS_SESSION_BUS_ADDRESS`, without which the bridge would have nothing to dial and `gjsify debug` would still need hand-wiring on macOS.

**Row 4 deliberately does NOT fall back.** The bus works, so the honest diagnosis is a too-early `installDevtools` call; standing a socket up there would paper over a call-site bug behind a working-looking address.

Bridge side mirrors it, with the published address file ranked **above** the session bus: the file exists only while an app of exactly this id is listening on a peer socket, whereas "a session bus exists" says nothing about the app. With nothing in peer mode no file is ever written, so on Linux that row is a `stat` that misses.

The diagnostic now distinguishes three states instead of blaming the call site unconditionally: `DBUS_SESSION_BUS_ADDRESS` unset (no bus here) · set but nothing answers (the usual macOS launchd state, named) · bus answers but the app is unregistered (call from `startup`). Probing whether a bus *answers* is what makes the middle row possible; the env check runs first so the probe cannot autolaunch a daemon.

## What the Linux test asserts

`packages/framework/devtools/src/peer-transport.spec.ts` runs on plain `gjs` with **no display and no bus daemon** — the shape of a macOS host, and of CI, so the busless path is covered on every affected PR rather than only on a Mac someone owns. It asserts positive facts, never "nothing blew up":

- a real `GetStatus` round trip over an auto-picked socket, returning **this app's** live `appId` + `instance` (not a default-constructed shape);
- a **second** peer connection served too (`connectionCount === 2`), the per-connection-export regression;
- the async `Screenshot` method marshalling its `(ay)` reply over the peer wire (the gjs 1.86.0 Promise-return guard, now on the peer transport);
- an explicit `unix:path=` honoured, socket present on disk;
- **`installDevtools` itself** exporting over an explicit address on an application with no bus connection at all, with the published address file read back and compared;
- both precedence tables (4 + 6 rows) and all three diagnostic wordings — so "the no-bus message is the no-bus one" is machine-checked, including that it does *not* mention the startup handler.

74 specs green in `@gjsify/devtools`, 32 in `@gjsify/devtools-mcp`.

## The Windows answer

Supported, by address family rather than by platform branch — and measured, not assumed:

- `nonce-tcp:host=127.0.0.1` **works** under gjs 1.88.1 (server + client round trip); it is the auto-picked default when `GLib.DIR_SEPARATOR === 92`, since GLib implements no unix-socket transport on win32.
- Server flags follow the family: `unix:` gets `AUTHENTICATION_REQUIRE_SAME_USER`, `nonce-tcp:` **must** get `NONE` — measured, that flag rejects every TCP client (`Gio.IOErrorEnum: Unexpected lack of content when trying to read a line`) because TCP carries no peer credentials to compare. There the nonce file's permissions are the access control.
- Platform detection is `GLib.DIR_SEPARATOR` — GLib's own introspected compile-time constant — so no `uname` probe and no new dependency in a GJS-only package.

Honest limit: **nothing exercises this on Windows yet.** There is no libgjs prebuild for win32 (`@gjsify/napi`'s blocked port documents the same wall), so the shipped Windows GUI path is node-gi plus the DBus-free storybook probe. The code is address-agnostic and has no unix-only parsing, but treat win32 as groundwork until a Windows leg loads it.

## Recorded, not fixed (`status/open-todos.md`)

- **`Screenshot`'s `scope` in-arg is dead** — `ScreenshotAsync(_params, invocation)` (`devtools-service.ts:151`) ignores it and always captures `app.get_active_window()` via `_captureActiveWindowPng()` (`:171`), even though `_resolveRootWidget` (`:289`) and `captureWidgetPng` already support a widget path. Asking for `toplevel:1` returns the active window: a wrong answer, not an error. Verified; left out because it changes what an existing argument *means* and wants its own per-scope regression test.
- **`SetProperty` / `EmitSignal` are classified with nothing behind them** — `methods.ts:32` and `:35` declare them `mutating`, but there is no XML fragment in `devtools-iface.ts`, no handler on the service and no MCP tool. Verified; both are widget-state mutators, a separate surface from the transport.

## Notes

- `CHANGELOG.md` deliberately untouched (other PRs in flight there).
- `packages/infra/cli/dist/cli.gjs.mjs` was rebuilt by the pre-commit hook because `commands/debug.ts` changed; it boots under stock `gjs` and reports `0.27.1`.


JumpLink:
## Both blockers fixed in `8f9c774c5` — and rehearsed against the old code first

Thanks — both were real, and both were reproducible exactly as described. Neither gate got looser: every new test **fails on the pre-fix tree**, and I ran them there before writing the fix.

### BLOCKER 1 — `installDevtools` is TOTAL again

Measured first, gjs 1.88.1: a second `Gio.DBusServer` on the same `unix:path=` throws `G_IO_ERROR_ADDRESS_IN_USE` (code 33), and so does binding over the socket **inode** a killed process leaves (`stop()` unlinks it, an abnormal exit does not; `connect()` to that inode gets `G_IO_ERROR_CONNECTION_REFUSED`, and after `unlink` the bind succeeds).

- `installDevtools` now wraps ONE fallible inner function (`exportDevtools`) in a single try/catch: it logs `[gjsify-devtools] devtools stayed OFF, the app is unaffected — <actionable sentence>`, calls `service.unexport()` (idempotent cleanup) and returns `null`, exactly like the disabled path. Its JSDoc states the contract; keeping all fallible work in one function is what makes totality structural rather than a promise.
- `startDevtoolsPeerServer` now throws a typed `DevtoolsPeerServerError` (`address`, `reason`, `cause`) whose message names the address and the way out, instead of GDBus' localised `Fehler beim Binden an Adresse (GUnixSocketAddress): Die Adresse wird bereits verwendet`.
- **Stale socket: reclaim, deliberately, on proof.** For a `unix:path=` address that already exists: `unix::mode` must say `S_IFSOCK` *and* a real connect must be refused → unlink + log `reclaimed the stale socket …`. A **live** socket is never touched (the message tells you to stop it or use `GJSIFY_DEVTOOLS_INSTANCE`), and a **non-socket** at that path is never deleted — `GJSIFY_DEVTOOLS_ADDRESS=unix:path=<typo>` must not be a delete primitive.
  *Why reclaim rather than pick a fresh address:* an explicit address is a contract — it is set on the app **and** on `gjsify debug`, so moving elsewhere would leave the bridge dialling the path it was told to dial and finding nothing, i.e. trading a visible failure for an invisible one. Where there is no contract (the auto-picked address) "pick a fresh one" is already what happens: `unix:tmpdir=` generates a unique name per server, and `unix:abstract=`/TCP cannot go stale at all.

**Rehearsal (pre-fix tree, same tests):** `5 of 80 tests failed`, with the swallow visible in the log —

```
(gjs:280723): Gjs-CRITICAL **: JS ERROR: Gio.IOErrorEnum: Fehler beim Binden an Adresse …
    startDevtoolsPeerServer@…  installDevtools@…
❌ keeps the app running when the pinned address is already taken
```

The collision test calls `installDevtools` **from inside a `Gio.SimpleAction::activate` handler** and asserts the statement after it ran (the stand-in for `present()` / `onStartup`), because that is the actual blast radius — I confirmed separately that GJS logs a handler throw and skips the rest of that handler. Post-fix it asserts positive facts: install returns `null`, no orphan address file was published, **and the first app still answers `GetStatus` over the same socket** (proof the collision did not steal a live socket).

### BLOCKER 2 — the address file's claim is now VERIFIED, and retracted with the socket

- `connectToDevtools` (the bridge's resolution step) is **dial-then-fallback**: a published address that does not answer is deleted, and resolution continues down the precedence → session bus → the three-ways-in diagnostic. The file keeps its rank, but it is now checked rather than trusted.
- An **explicit** `--address` / `GJSIFY_DEVTOOLS_ADDRESS` still does **not** fall back (pinned on both sides ⇒ a dead one must be reported, not routed around) — but it now fails with a message naming that address instead of a raw GIO error.
- `DevtoolsService.unexport()` removes the published file together with the socket (`attachPeerServer(peer, addressFile)`, `service.addressFile`), so `uninstallDevtools` no longer leaves a claim behind; the `shutdown` handler stays as the app-exit path.
- **On SIGINT I deliberately did not install a handler**, and said so in the code: it is unix-only (this transport exists for macOS/Windows), it would change an app's termination semantics from inside an opt-in diagnostic, and it still could not cover `SIGKILL`. So the claim is verified where it is *consumed*.

**Rehearsal (pre-fix client, same test):**

```
REHEARSAL MESSAGE: Gio.IOErrorEnum: Verbindung ist gescheitert: Datei oder Verzeichnis nicht gefunden
REHEARSAL FILE STILL THERE: true
❌ deletes a STALE published address and reports the three ways in
```

Post-fix the same test (end to end through `new DbusDevtoolsClient`, real GDBus dial, `DBUS_SESSION_BUS_ADDRESS` unset for determinism) gets `no way to reach <app> … <dead address>, is dead … GJSIFY_DEVTOOLS=1 … --address` and the file is **gone**. Three more cases: a **live** file is dialled and **kept** (the positive control — self-healing must not make the file worthless, and the session bus must not be consulted), a stale file **degrades to the session bus** (the Linux-regression row), and a pinned dead address fails without consulting the bus.

*Harness note, stated in the spec:* a peer server and its client cannot be dialled **synchronously** in one process (GDBus services the new connection on the main context, which a sync dial blocks — it deadlocks; I hit it). So where a spec needs a live connection it dials async and passes that **real, open** `Gio.DBusConnection` through the new `deps` seam; the dead-address cases use the real dial untouched, since those fail immediately. What is stubbed is the synchronous call, never the connection.

### The seventh reviewer's three questions — checked, not assumed

1. **Session-bus precedence unchanged (Linux).** App side untouched: the reclaim/probe code only runs on a `unix:path=` peer address, and row 2 of `chooseDevtoolsTransport` is unchanged. Bridge side: with nothing in peer mode there is no address file, so that row is still a `stat` that misses — and a *stale* file now degrades **to the session bus**, which is asserted by a test that did not exist before. Both suites also pass with `DBUS_SESSION_BUS_ADDRESS` **and** `XDG_RUNTIME_DIR` unset (the CI/macOS shape): `@gjsify/devtools` 96 specs, `@gjsify/devtools-mcp` 52 specs, green either way.
2. **`nonce-tcp:` with `DBusServerFlags.NONE` on a shared machine.** Measured on gjs 1.88.1, and now **spec-checked** instead of asserted in prose: GDBus creates the nonce file with **mode 0600**, binds **127.0.0.1 only** (`ss -ltn` → `LISTEN 127.0.0.1:38013`, not `0.0.0.0`), and a client that opens the port **without** sending the 16 nonce bytes gets no reply at all — the server waits and the connection times out. So another local user can reach the socket and still not speak the protocol; the effective control is "can read a 0600 file", i.e. same-uid, which is what `AUTHENTICATION_REQUIRE_SAME_USER` would have given. The new spec asserts the 0600, so a GLib change or a wrong flag surfaces as a failing test rather than as an open control plane.
3. **`node-gi` `test:dbus` on macOS *and* Linux — honest answer: it RUNS on Linux, and is only made RUNNABLE on macOS.** Verified locally on Fedora that `dbus-run-session --config-file=packages/node-gi/node-gi/test/session.conf -- …` starts and round-trips (`ADDR=unix:path=/tmp/dbus-xHMx4Ezvc8,guid=…`, `GetId` answers), with the same `<listen>unix:tmpdir=/tmp</listen>` Fedora's own `/usr/share/dbus-1/session.conf` uses — the only deliberate differences are the omitted `<standard_session_servicedirs/>` (no activation, and the suite owns both ends) and a permissive policy for a throwaway private bus. In CI it is exercised only by the Linux `Build & test (Node / Fedora 44)` job; **`node-gi.yml`'s macOS legs never run `test:dbus`** (they run `test:conformance`/`test:bun`/`test:deno` and install no dbus-daemon), so nothing verifies that leg on a Mac and this PR does not claim otherwise — the config removes a hand-run blocker, it does not add coverage.

### Validation

`gjsify tsc --noEmit` clean in both packages · `node_modules/.bin/oxfmt` + `node_modules/.bin/oxlint` clean · `@gjsify/devtools` **96** specs and `@gjsify/devtools-mcp` **52** specs green, each run twice (with a session bus, and with `DBUS_SESSION_BUS_ADDRESS`/`XDG_RUNTIME_DIR` unset). `packages/infra/cli/src/` is untouched, so the committed `cli.gjs.mjs` needs no rebuild. Docs updated where the behaviour is now different: both READMEs (a new "Restarting, and two apps at once" section, the verified-not-trusted address file, the measured nonce-tcp facts) and the devtools block in `AGENTS.md`.

PR #979, opened 2026-08-04
fix(gamepad): make a missing backend observable

JumpLink:
## The defect

`packages/web/gamepad/src/gamepad-manager.ts` wrapped `await import('gi://Manette')` in a `try`/`catch` whose catch set `_initialized = true` and returned, and the README's third line called it "**degrades gracefully when libmanette is unavailable**".

Two distinct failures hid behind that one catch, and a third hid behind the code *after* it. Measured on this branch's parent by modelling the pre-fix `_init()` verbatim (only the import expression replaced by an injectable one, so all three host situations can be produced on a machine that *has* libmanette):

```
[absent] getGamepads() -> [null,null,null,null]   ← no diagnostic
[broken] getGamepads() -> [null,null,null,null]   ← no diagnostic
[nodegi] getGamepads() -> [null,null,null,null]   ← no diagnostic
    + Gjs-WARNING: Unhandled promise rejection … _init@…:7
```

* **absent** — no `Manette` typelib. Expected on macOS/Windows.
* **broken** — typelib present, `Failed to load shared library 'libmanette-0.2.so.0'`. A fault, reported as "no gamepads".
* **nodegi** — the `--app node` shape (below). Import resolves, `new Monitor()` throws into a promise nobody awaits.

## What the spec actually says

`navigator.getGamepads()`, **§ Extensions to the `Navigator` Interface** of [w3c.github.io/gamepad](https://w3c.github.io/gamepad/):

> The `getGamepads()` method steps are:
> 1. Let |doc| be the current global object's associated `Document`.
> 2. If |doc| is `null` or |doc| is not fully active, then **return an empty list**.
> 3. If |doc| is not allowed to use the `"gamepad"` permission, then **throw a `"SecurityError"` `DOMException`** and abort these steps.
> 4. If this.`[[hasGamepadGesture]]` is `false`, then **return an empty list**.
> 5. Let |now| be the current high resolution time…
> 6. Let |gamepads| be an empty list.
> 7. For each |gamepad| of this.`[[gamepads]]`: …
> 8. Return |gamepads|.

So: the steps **only ever return a list**. The single `throw` is the permission-policy `SecurityError`; every other negative case — no `Document`, no user gesture, nothing connected — is spelled *return an empty list*. **There is no state, exception or error condition for "this platform has no gamepad subsystem."** The spec is silent because a browser does not need one.

And browsers behave accordingly, structurally rather than incidentally. WebKit compiles an `EmptyGamepadProvider` (`Source/WebCore/platform/gamepad/EmptyGamepadProvider.cpp`) on every port with no gamepad backend:

```cpp
const Vector<WeakPtr<PlatformGamepad>>& EmptyGamepadProvider::platformGamepads()
{
    static NeverDestroyed<Vector<WeakPtr<PlatformGamepad>>> emptyGamepads;
    return emptyGamepads;
}
```

`startMonitoringGamepads()` is `{}` and `playEffect()` completes with `false`. That is exactly the surface we present with no backend.

**Therefore `getGamepads()` does not throw here, and the earlier `(C)`-stub proposal is dropped.** The `(C)` convention (`packages/node/tls/src/browser.ts`) is for a slot with **no meaningful behaviour**; "no gamepads are connected" *is* real, correct behaviour. AGENTS.md already carves out the same exception on the browser axis — "Exports whose answer IS platform-independent stay REAL (`net.isIP*`, … `hasOcspSupport()`→`false`); a blanket thrower would replace one lie with another". A throw would break `navigator.getGamepads().length`, the canonical poll. There is a comment on `getGamepads()` saying all of this so nobody "fixes" it later.

## How the catch is split, and why

`src/backend.ts` is now the one place the package decides whether a backend exists — deliberately not inside `GamepadManager`, because the answer must be obtainable without constructing a monitor.

**The discriminator.** GI reports a missing typelib as `GI_REPOSITORY_ERROR_TYPELIB_NOT_FOUND`, but neither loader hands the structured error to JS. Measured on gjs 1.88: the thrown object is a plain `Error` whose own properties are `["fileName","lineNumber","columnNumber","message"]` — no `domain`, no `code`. `@gjsify/node-gi`'s `requireNamespace()` (`src/repo.cc`) likewise re-wraps `error->message` into a `Napi::Error` string. The message is the only signal **both** loaders preserve, so the test is one substring — GI's own wording, not GJS's `Requiring Manette, version none: …` framing nor node-gi's `Failed to require Manette: …` framing, which is what lets one test serve both targets:

```ts
const TYPELIB_ABSENT = "Typelib file for namespace 'Manette'";
```

It names the namespace on purpose: `… for namespace 'GObject' …` (a missing *dependency*) is a broken install, not an absent backend, and stays loud. Measured wordings it must span: `'Manette' (any version) not found` and `'Manette', version '0.2' not found`.

The classification **defaults to fault**; only that wording is quiet. If upstream rewords it we get a loud fault on a host that merely has no libmanette — noisy and fixable. The other direction would restore a silent wrong answer, which is the defect being removed.

**Warning vs. rethrow — argued against the repo rule.** AGENTS.md § Code anti-patterns: *"A kept catch must STATE ITS REASON"* and *"the worst variant swallows a `ReferenceError` from a missing binding into a silent wrong answer"*. Both apply:

* The catch is **kept** because the operation inside it has exactly one real, per-host failure mode — GI resolving a typelib that may not exist — and the code says so at the site.
* It no longer **converts an error into a silent wrong answer**: each branch produces a distinguishable value (`status: 'absent' | 'failed'`, the original error retained) *and* a message.
* **Not a rethrow.** `_ensureInit()` starts `_init()` without awaiting it, because `getGamepads()` is synchronous per the polling contract. A rethrow therefore becomes an unhandled rejection — unattributable on GJS (see the model output above: it names `_init@…:7` and nothing about libmanette) and, under Node's default `--unhandled-rejections=throw`, a **process kill**. A half-installed typelib must not be able to take down the host app, and an unhandled rejection is not more visible than a named `console.error`. So: `console.warn` once for absence, `console.error` with the original error for a fault — **two levels for two situations**, which is what makes the split visible rather than merely present.

Separately, `_ensureInit()` now handles the rejection that everything *past* the probe (`new Monitor()`, `iterate()`, `connect()`) could produce. Pre-fix that was a permanent silent "no gamepads" plus an unhandled rejection.

## The capability export

```ts
export async function hasGamepadBackend(): Promise<boolean>
```

Barrel-exported from `src/index.ts`. **Mirrored on `isSecureRandomSource(source)`** (`packages/web/webcrypto/src/random.ts:102`) — the near-exact analogue named in the brief: Web Crypto says "returns random bytes" and has no word for "these are not cryptographic", so the package added a boolean predicate *next to* the conformant surface rather than changing it. `hasNativeSab()` and `hasOcspSupport()` supplied the `has*` naming and the "the bridge is usable, not every call will succeed" semantics, which the doc comment states.

A **boolean**, not a richer shape: all three exemplars are booleans, and the two facts a caller can act on ("is there a backend" / "what do I install") are the return value and the stderr line. `GamepadBackendStatus` is exported as a *type* only, for the internal three-way result the manager consumes.

`Promise<boolean>`, not `boolean`, because `gi://` is only reachable through a dynamic `import()` on both targets — `hasNativeSab()` can be synchronous because `imports.gi` is, and `globalThis.imports?.gi?.Manette` is not an option here (it resolves on GJS only, so node-gi parity would be broken by construction — AGENTS.md line 296). It needs no monitor and no connected device: the probe resolves the namespace and reads `Monitor`, nothing more. `loadGamepadBackend()` caches the probe, so the capability export and the manager cannot disagree, and one probe per process is also what makes the diagnostic fire once.

> On the warn-once shape: `random.ts`'s `let warnedInsecure = false` guard exists because `fillRandomBytes()` re-runs its chain on **every call**. This probe runs at most once per process by construction, so a second flag could never be the thing that suppressed a duplicate — AGENTS.md forbids exactly that kind of redundant guard ("hides real bugs as silent no-calls"). The `_resetGamepadBackendCache()` testability export from `random.ts:210` **is** mirrored, and clearing the cache is what lets a test observe the message again. The code says this where the cache is declared.

## node-gi parity — the import guard was a no-op on the node target

The package declares `node: "partial"`, and `--app node` rewrites `gi://Manette` via `giNodeShimSource()` (`packages/infra/rolldown-plugin-gjsify/src/plugins/gjs-gi-node.ts`) to a virtual module whose default export is a **lazy Proxy**; `requireGi('Manette')` runs on the first **property access**. So on Node the import *always* resolves and the pre-fix `try`/`catch` caught nothing — the failure landed on `new this._ManetteModule.Monitor()`, outside it, as an unhandled rejection (scenario `nodegi` above).

The probe therefore **resolves `Monitor` inside its own guard**. Free on GJS (the namespace object is already materialised); on Node it is what moves the fault back where it can be classified. Verified on an emitted `--app node` bundle — the read is inside the `try`:

```js
try{if(r=(await t()).default,typeof r?.Monitor!=`function`){…}}catch(e){let t=classifyLoadError(e);…}
```

and run under `node` with `@gjsify/node-gi` absent:

```
[@gjsify/gamepad] The gi://Manette gamepad backend failed to load — a fault on this host,
NOT a platform without libmanette. … Error: Cannot find module '@gjsify/node-gi/gi'
```

A missing *bridge* is a setup fault carrying its own fix, not an OS without libmanette — that run is also why the fault message no longer claims "the typelib is present", which would have been a lie in exactly this case.

## Tests

`src/backend.spec.ts` (9 cases) + `src/register.spec.ts` (4 cases), both wired into `src/test.mts` through `run()`. `104 completed` under gjs 1.88.1. An injected `importer` is the seam — same role as `fillRandomBytes({ webcrypto })` — because every runner this package is tested on *has* libmanette, so the absent and failed paths are otherwise unreachable. Every assertion is a positive fact about injected, known state; none passes vacuously.

| assertion | pins |
|---|---|
| module loads → `status: 'manette'`, `module` is the injected object, **no** warn, **no** error | the happy path emits nothing |
| absent wording → `'absent'`, original error retained, `hasGamepadBackend()` **false**, exactly **1 warn** naming `libmanette` + `hasGamepadBackend()`, **0 errors** | absence is quiet-but-said, and never a fault |
| `Failed to load shared library 'libmanette-0.2.so.0'` → `'failed'`, **1 error** containing `NOT a platform without libmanette` **and** the original text, **0 warns** | a fault is distinguishable from "no controllers" — the core defect |
| `Typelib file for namespace 'GObject'` → `'failed'` | a missing *dependency* is not absence |
| `Version 0.2 … already loaded, cannot load version 9.9` → `'failed'` | a version conflict is not absence |
| a Proxy that throws on first member access → `'absent'`, 1 warn | node-gi parity; **fails pre-fix**, where nothing touched a member inside the guard |
| namespace without `Monitor` → `'failed'`, error names it | ABI skew, instead of a `TypeError` at first poll |
| two `loadGamepadBackend()` + one `hasGamepadBackend()` → importer called **once**, **1 warn** | warn-once *is* the cached probe |
| manager on a primed-absent probe → 4 slots, all `null`, no throw; `dispose()` + re-poll adds **no** second warn | the W3C shape survives, and re-init stays quiet |
| `/register`: `navigator.getGamepads` is a function; `GamepadEvent` registered; **`navigator.getGamepads()` returns a 4-slot array and does not throw**; `hasGamepadBackend` reachable from the package root and returns a boolean | the anti-throw guard, plus the barrel export, through the published `/register` + root entries |

`/register` coverage is in `register.spec.ts` per the convention, importing `@gjsify/gamepad/register` and `@gjsify/gamepad` by package specifier (the `@gjsify/webrtc` / `@gjsify/iframe` shape).

**Confirmed failing on pre-fix behaviour**, two ways:

1. Structurally — with `src/backend.ts` removed and `gamepad-manager.ts` + `index.ts` restored from `origin/main`, the test bundle does not build: `[UNRESOLVED_IMPORT] Could not resolve './backend.js' in src/backend.spec.ts`. Pre-fix there is nothing to ask and nothing that emits a diagnostic.
2. Behaviourally — the model run at the top of this description: all three host situations returned `[null,null,null,null]` with **zero** console output, the broken and absent cases byte-identical, and the node-gi case additionally leaking an unhandled rejection.

## Deliberately not done

* **`gjsify.runtimes` untouched.** This is an OS-axis gap inside a working `gjs: "polyfill"` slot; AGENTS.md: *"The two axes are blind to each other BY DESIGN … Never let one axis answer the other's question."* `gjs: "none"` would kill Linux, where the package works. No `gjsify.platforms` either — that field is about prebuilds, and this package has none. `node scripts/audit-runtimes.mjs --check` exits 0.
* **No libmanette seeding into the GTK-runtime bundles.** Already recorded and unchanged: nothing to seed from, and a `Manette.Monitor` on macOS would enumerate nothing while satisfying every symmetry check.

## Recorded in `status/open-todos.md`, not implemented

Every claim verified first-hand.

**A darwin gamepad backend is the only route to macOS support, and it is a separate project.** `GameController.framework` alone is insufficient, and both reference implementations say so by shipping two paths. WebKit's `Source/WebCore/platform/gamepad/` has `cocoa/GameControllerGamepadProvider.mm` **and** `mac/HIDGamepadProvider.mm` (+ per-device `Dualshock3HIDGamepad`, `StadiaHIDGamepad`, `LogitechGamepad`, `GenericHIDGamepad`), combined by `mac/MultiGamepadProvider.mm`, which calls `HIDGamepadProvider::ignoreGameControllerFrameworkDevices()` and gates GCF on `GameControllerGamepadProvider::willHandleVendorAndProduct()` — a hardcoded allow-list. Verbatim from the source: *"we use GameController framework for some controllers, but it's much too aggressive in handling devices it shouldn't. So we check Vendor/Product against an explicit allow-list to determine if we should let GCF handle the device. (We have the opposite check in HIDGamepadProvider, as well)"*. SDL ships both too: `src/joystick/apple/SDL_mfijoystick.m` and `src/joystick/darwin/SDL_iokitjoystick.c`. Second piece: `src/button-mapping.ts` maps raw evdev codes (`BTN_SOUTH: 304` … `BTN_DPAD_RIGHT: 547`), which nothing on macOS produces — GCF gives named `GCControllerButtonInput`s, IOKit gives HID usage pages — so a darwin backend needs a **second source vocabulary** onto the same `W3CButton`/`W3CAxis` targets, not another row in the existing table.

**libmanette is not portable and upstream has never considered it.** Checked out `gitlab.gnome.org/GNOME/libmanette` at tag `0.2.13` and `main`: `dependency('libevdev', version: '>= 1.4.5')` and `dependency('hidapi-hidraw')`, neither with a `required:` argument; the only "Dependencies" toggle in `meson_options.txt` is `gudev`; **no `host_machine` conditional in any `meson.build`** in either revision; **`hid_enumerate()` is never called anywhere** — `manette-hid-backend.c` does `hid_open_path(self->filename)` and `filename` comes from the gudev walk in `manette-monitor.c` (`g_udev_client_new({"input","hidraw"})`, `g_udev_device_get_device_file()`, `DEV_DIRECTORY "/hidraw"`), so hidapi being cross-platform buys nothing. Tracker: **51 issues and 155 merge requests** (GitLab API `x-total`, oldest MR 2017-12-03) scanned for macos/darwin/osx/portab*/windows/win32/freebsd/cross-platform in title and description — **zero relevant hits in nine years** (one keyword match, MR !104, is a comment about SDL button mappings). And the hard dependency is unavailable: homebrew-core's `libevdev` carries `depends_on :linux` with `arm64_linux`/`x86_64_linux` bottles only, MacPorts has no `libevdev` port (ports API exact-name query → `{"count":0}`), nixpkgs declares `platforms = lib.platforms.linux ++ lib.platforms.freebsd`.

**A forced migration is coming: `Manette-1`.** libmanette `main` is `version: '1.0.alpha'` with `libmanette_api_version = '1'`. `ManetteEvent` is **deleted** (`src/manette-event.c` + `manette-event-private.h` removed in `3255105` "Remove ManetteEvent", part of MR **!126** "Bump API version and revamp API", merged 2025-04-01) — every `event.get_button()`/`get_absolute()`/`get_hat()` call goes with it. Signals are renamed and re-typed: `button-pressed` / `button-released` / `absolute-axis-changed` (+ `unmapped-*`), and **there is no `hat-axis` signal any more** — the d-pad is four buttons. Typed `ManetteButton`/`ManetteAxis` enums (`src/manette-inputs.h`, `MANETTE_BUTTON_DPAD_UP` … `MANETTE_BUTTON_TOUCHPAD`, `MANETTE_AXIS_LEFT_X` … `MANETTE_AXIS_RIGHT_TRIGGER`) replace the raw kernel codes, retiring `button-mapping.ts`'s `LinuxButton` table rather than extending it. `@girs/manette-0.2` does not bind that namespace, so this needs a `ts-for-gir` run first. Independent of the macOS work and of this PR.

## Validation

`gjsify tsc --noEmit` clean · `node_modules/.bin/oxfmt` clean · `node_modules/.bin/oxlint` clean · `gjsify workspace @gjsify/gamepad test` → **104 completed** under gjs 1.88.1 · `node scripts/audit-runtimes.mjs --check` exit 0 · `--app node` bundle built and executed.


JumpLink:
Fixed in 7898958. Every finding below is measured; where a measurement contradicted the PR's own text, the text lost.

## The two GJS wordings, measured first

Everything in B4 turns on these, so they came first — gjs 1.88.1, Fedora 44.

**Absent typelib.** `Manette-0.2.typelib` hidden from `/usr/lib64/girepository-1.0` (a copy of the directory minus that one file, bind-mounted over it with `bwrap`), then `await import('gi://Manette')`:

```
Error: Requiring Manette, version none: Typelib file for namespace 'Manette' (any version) not found
own properties: ["fileName","lineNumber","columnNumber","message"]   // no domain, no code
```

So `ABSENT_MESSAGE` was right, and it is now a *measured* fixture rather than a plausible one.

**Present typelib, shared library will not load.** A byte-patched copy of the real typelib (the `shared-library` string rewritten to an equal-length name that does not exist), placed first on `GI_TYPELIB_PATH`:

```
(gjs:…): GLib-GIRepository-WARNING **: Failed to load shared library 'libmanette-9.9.so.9'
         referenced by the typelib: … Kann die Shared-Object-Datei nicht öffnen: …
Error: Unsupported type void, deriving from fundamental void
```

Two things follow, and both were wrong in the PR:

1. **The `Failed to load shared library …` text never reaches JS.** It is a `GLib-GIRepository-WARNING` on stderr, in the process **locale** (note the German above), attached to no error object. `backend.spec.ts:129` asserted against the C format template — a string GJS does not produce. That fixture is gone.
2. **What JS actually gets names neither libmanette nor the library.** The ESM `gi://` loader materialises the namespace eagerly, so the first GType whose `get_type` symbol is unreachable throws. `Unsupported type void, deriving from fundamental void` is the measured message and is now the fixture; the test asserts that the emitted `console.error` contains it, and a comment records that this case is *precisely* why the classifier defaults to fault.

**The discriminator no longer compares a string to a copy of itself.** `_diagnoseGiLoadError(error, namespace)` takes the namespace, and one case asks the **live** loader for a namespace that cannot exist, then classifies whatever it threw:

```ts
try { await import(ABSENT_SPECIFIER) } catch (error) { liveError = error }
expect(_diagnoseGiLoadError(liveError, ABSENT_NAMESPACE).status).toBe('absent');
// …and the scoping is real, not decoration:
expect(_diagnoseGiLoadError(liveError).status).toBe('failed');   // scoped to 'Manette'
```

An upstream rewording now fails *that* test instead of silently reclassifying every backend-less host as a fault.

## B1 — capability export on the non-GJS entry

Reproduced verbatim on the PR head before touching anything, from README.md:52-58:

```
{"error":"rolldown: … MISSING_EXPORT: \"hasGamepadBackend\" is not exported by
 \"packages/web/gamepad/globals.mjs\"."}
```

`globals.mjs` now exports it **real**, per the rule and the exemplar you named — answered from the runtime's own surface (`typeof globalThis.navigator?.getGamepads === 'function'`): `true` in a browser, `false` on NativeScript, which ships no Gamepad API. Both targets now build, and the emitted body is the whole answer:

```
$ gjsify build tmp/readme-example.ts --app browser      → ok
$ gjsify build tmp/readme-example.ts --app nativescript  → ok
$ cat out.browser.mjs
globalThis.Gamepad,…;async function hasGamepadBackend(){return typeof globalThis.navigator?.getGamepads==`function`}await hasGamepadBackend()||console.log(`no gamepad subsystem on this host`);
$ node out.browser.mjs
no gamepad subsystem on this host
```

**Should `probeGlobalsExports` have caught it?** No, structurally — it only validates the `export … from '<spec>'` **sources** a `globals.mjs` names, so a file that re-exports nothing (every hand-written `export const X = globalThis.X`) passes it vacuously. The missing check is the `native` mirror of `platform-entry-parity`: `native` routes the package **root** to `@gjsify/<X>/globals` exactly as `polyfill` + a declared subpath routes it to `src/<target>.ts`, so the same invariant applies. Added as check 5 in `auditReachability`, `globals-entry-parity`.

It **reports**, it does not gate, and that is a measurement rather than a preference — the check's own first run:

```
reachability notes — 38 `native` slot package(s) whose globals.mjs is NARROWER than
the root entry (314 export(s); globals-entry-parity, reported not enforced):
  · @gjsify/assert (node) — 19: AssertionError, deepEqual, deepStrictEqual, …
  · @gjsify/util (node) — 34: TextDecoder, TextEncoder, aborted, callbackify, …
  · @gjsify/webcrypto (browser, nativescript, node) — 4: Crypto, CryptoKey, SubtleCrypto, crypto
  · @gjsify/gamepad (browser, nativescript) — 11: GamepadManager, MANETTE_TO_W3C_AXIS, …
  …
```

Making it fatal is a rewrite of 314 export names across nearly every `native` slot — AGENTS.md exception (c) territory (Plan + confirm + split PRs), and this PR was told to *reduce* scope. Two shapes hide in those 314 and only one is a re-export away: names the runtime provides (`@gjsify/assert`'s `strictEqual` from `node:assert`) versus names it does not (gamepad's Manette→W3C mapping tables). **No `globals.mjs` in the tree imports its own package body**, so the second shape needs a platform entry — a slot decision, not a line in `globals.mjs`. That is why gamepad's row still shows 11 and why closing them here would have meant inventing a mechanism. Ledgered in `status/open-todos.md` under the ADR-0014 follow-ups.

## B2 — the two targets where `gi://` is stubbed by design

Two changes, one structural and one classificatory.

**Structural, and it fixes something worse than the wrong label.** `register.ts` overwrote `navigator.getGamepads` unconditionally, so on a browser build it replaced a *working native* implementation with the libmanette manager. It now guards — the same guard the `GamepadEvent` registration two lines below always had — so on browser/NativeScript the native one stays, the manager is never constructed, nothing probes, and there is nothing to be silent about.

**Classificatory,** for a bundle that reaches `backend.ts` anyway (an explicit `GamepadManager`). The by-design stub is recognised and produces `status: 'absent'` with a `null` diagnostic — completely silent, because there is nothing to install. The three shapes are distinguishable, measured:

| resolved `gi://Manette` default | own property names | prototype |
|---|---|---|
| `gjsImportsEmptyPlugin` stub `{}` | 0 | `Object.prototype` |
| real GJS namespace | **16, eagerly** (`Device`, `Monitor`, `MAJOR_VERSION`, …) | `[object GIRepositoryNamespace]` |
| node-gi lazy shim | 0 | `null` (`Object.create(null)` target) |

So an ABI skew (namespace loaded, no `Monitor`) still has own properties and stays a fault, and node-gi's proxy is separated by its `null` prototype — plus it is classified from the error its `get` trap throws, before the stub check is reached. All three are asserted.

## B3 — node-gi parity

A missing bridge is "no backend here", not a broken host: ADR 0005 forbids hard-depending on `@gjsify/node-gi`, the `node` slot is `partial`, and `giNodeShimSource()`'s lazy `require` exists precisely to keep that configuration harmless. It now warns, with its **own** advice, because the libmanette text would be the wrong fix:

```
[@gjsify/gamepad] No gamepad backend in this process — @gjsify/node-gi is not installed,
and it is how a --app node bundle reaches gi://Manette. … (npm install @gjsify/node-gi,
plus libmanette + its typelib). Gate on hasGamepadBackend().
```

Guarded so a genuine node-gi crash stays a fault: the error must both name the bridge **and** have a module-resolution shape (`code` `MODULE_NOT_FOUND` / `ERR_MODULE_NOT_FOUND`, or a `Cannot find module` message).

## B5 — the test bundle could not be Manette-free

Root cause found, and it is not the `/register` import: `--globals auto` reads `globalThis.GamepadEvent` and `globalThis.navigator` (Pattern A in `detect-free-globals.ts`, plus the `navigator.getGamepads` method marker) as free globals and injects the register set. On the PR head:

```
[gjsify] note: --globals auto injected GTK/GNOME-backed register(s) — this bundle now
requires gi://Gdk, gi://GdkPixbuf, gi://Manette, gi://Pango, gi://PangoCairo at load.
Triggered by: GamepadEvent, navigator.
```

On `main` the identifiers are *imported* bindings, so nothing was injected — the PR introduced it, exactly as you said. The spec now reads the host through a typed local view (`const host = globalThis as unknown as HostGlobals`), which removes the trigger at the source rather than papering over it with `--exclude-globals`, keeps the `/register` convention, and retires the file-level `no-explicit-any` disable. The note is gone, and the bundle now contains **no static imports at all** (`grep -c '^import\|;import' test.gjs.mjs` → 0); all three `gi://Manette` occurrences are dynamic — two `()=>import(`gi://Manette`)` from the probe, one `await import(`gi://Manette`)` from the live-loader assertion.

Verified where it counts, on a host that genuinely has no Manette:

```
$ bwrap --ro-bind / / --ro-bind <girepository-1.0-minus-Manette> /usr/lib64/girepository-1.0 \
        gjs -m test.gjs.mjs
✔ [Gjs 1.88.1] 138 completed
$ gjs -m test.gjs.mjs          # same bundle, libmanette present
✔ [Gjs 1.88.1] 138 completed
```

with `No gamepad backend on this host` on stderr only in the first case. `register.spec.ts`'s capability assertion holds on both because it asks the live loader what the host has and requires the export to agree — that also replaces the vacuous `typeof … === 'boolean'`.

## R1 — the 4-slot array: made conformant, deviation removed

`Navigator.[[gamepads]]` "is initially the empty list"; index selection is *first `null` slot, else append*; `getGamepads()` step 6 starts from an empty list and step 7 copies `[[gamepads]]` into it. A host where nothing ever connected therefore answers `[]` — Firefox and WebKit do, and WebKit's `EmptyGamepadProvider::platformGamepads()` returning a static empty vector is the same fact one layer down. The four pre-filled nulls are **Chrome's** shape, and hard-coding them made `navigator.getGamepads().length` report four ports that do not exist.

Under "the spec is the authority" I could not see an argument for hardening that, and AGENTS.md's root-cause rule forbids booking it as a known limitation, so: `_slots` starts empty and grows by the spec's index selection, disconnect nulls the index (keeping it allocated, per the disconnect steps), `dispose()` returns to the empty list, and the artificial `MAX_GAMEPADS = 4` cap is gone — the algorithm has no upper bound, and a cap silently drops the 5th controller. The claim is now removed from all three docs instead of asserted in them, and the two pre-existing 4-slot assertions became one conformance assertion each.

The manager-level test that "passed even if `GamepadManager` never consults the probe" was replaced rather than patched: the probe is primed with a fake namespace whose monitor hands out **one device**, and the test asserts a 1-element list with `id === 'Injected Pad'`. Remove the wiring and the list stays empty.

## The remaining should-fixes

- **S1** — `status/status.json:182`'s "graceful degradation without libmanette" is gone; the note now states the three failure classes and the empty-list contract.
- **S2** — the warning moved from the **query** to the **use**: `probeGamepadBackend()` prints nothing and *carries* the text, `GamepadManager._init()` emits it once via `reportGamepadBackendOnce()`. That is `isSecureRandomSource()` (pure) vs. `fillRandomBytes()` (warns), and it needed a real `warnedInsecure`-style flag, because each manager init and each re-init after `dispose()` reads the same cached probe. Two tests pin it: the query is asserted silent, and the manager's line is asserted present — and since the probe no longer prints, that line can only come from `_init()`.
- **S3** — the false comment is fixed rather than kept: the shared cache means the two cannot disagree *about the probe*, "which is all the cache buys", and the header now says a monitor can still fail afterwards and points at the report for it.
- **S4** — that report exists: `reportGamepadMonitorFault()`, for `new Monitor()` / the device walk / `connect()` failing on a host whose backend loaded fine (a sandbox without udev or `/dev/input`; flatpak `--device=input`). Asserted to say "could not be started" and **not** "failed to load".
- **R2** — the AGENTS.md row is one line plus a pointer: 1225 → 379 chars (neighbours are 51–306).
- **R3** — the WebKit quote is verbatim and scoped. It begins **"On macOS 10.15"** and sits inside `#if HAVE(MULTIGAMEPADPROVIDER_SUPPORT) && !HAVE(GCCONTROLLER_HID_DEVICE_CHECK)`, so it is the fallback for builds without the newer HID-device check, not a standing verdict on GCF. The entry now says so, and says the two-provider conclusion rests on `MultiGamepadProvider.mm` driving both providers (`HIDGamepadProvider::singleton().ignoreGameControllerFrameworkDevices()`) and on the per-device HID classes, not on the comment.

Minors: the orphaned `capturingConsole` JSDoc is attached to its own function; `{@link hasGamepadBackend}` is plain prose naming the root export; and `loadGamepadBackend({ importer })` after the probe now **throws** instead of discarding the importer — the old `cached ??= probe(options)` let a suite that forgot `_resetGamepadBackendCache()` assert against the previous test's injected module, so a test pins the throw too.

## Validation

`gjsify tsc --noEmit` clean · `oxfmt` clean · `oxlint` clean (the one `fmtSlot` warning in `audit-runtimes.mjs` pre-exists on this branch — verified by linting the unmodified file) · `node scripts/audit-runtimes.mjs --check --strict` **exit 0** with the new report · `gjsify workspace @gjsify/gamepad test` → **138 completed**, and 138 again under a Manette-less `bwrap` · README example built on `--app browser` and `--app nativescript` and executed.

JumpLink:
Two follow-up commits, no behaviour change:

* `a3310ffe` — `backend.spec.ts` reads the probe result through a local const + an explicit not-null assertion instead of `as unknown as GamepadBackend`; a cast there would also have swallowed a genuine type change.
* `322c0cd6` — records why the `gi://Manette` specifier stays a LITERAL while `GI_NAMESPACE` drives the wordings: every plugin that claims `gi://*` (`gjsGiNodePlugin`, `gjsImportsEmptyPlugin`, the `--app gjs` externals predicate) matches the resolved specifier at build time, so a template literal would leave the import unclaimed on all four targets.

Re-validated: `gjsify tsc --noEmit` / `oxfmt` / `oxlint` clean, and the suite is **139 completed** both with libmanette present and under the Manette-less `bwrap`.

JumpLink:
One more, and it is the piece of R1 I had left as recall rather than measurement (`4da0b78a`).

"Firefox and WebKit both answer `[]`" was asserted, not checked. Measured now — one machine, `about:blank`, no controller attached, via the Playwright browsers this repo already uses:

```
firefox  → {"length":0,"isArray":true,"json":"[]"}
chromium → {"length":4,"isArray":true,"json":"[null,null,null,null]"}
```

So the four pre-filled `null` slots this PR had certified as "the W3C shape" are demonstrably **Chrome's** shape and nobody else's. WebKit agrees with Firefox in source rather than by observation: `NavigatorGamepad::gamepads()` does `if (m_gamepads.isEmpty()) return m_gamepads;` (`Source/WebCore/Modules/gamepad/NavigatorGamepad.cpp`), which is the same fact as `EmptyGamepadProvider` one layer up. The numbers are now in the `getGamepads()` doc, in its spec case, and in the open-todos entry, so the next reader gets the evidence and not my memory.

JumpLink:
**Correction to my own B1 numbers** (`ebc15a14`) — the new check was wrong about 17 packages, and I would rather say so than have it found.

`globals-entry-parity` compared root value exports against `globals.mjs` via `collectValueExports`, which cannot follow a **bare** `export * from '…'`. So `@gjsify/util`'s `globals.mjs` — literally `export * from 'node:util'`, i.e. the runtime's entire surface — was read as missing all 34 root names. `tests/e2e/runtimes-routing` disproves that by importing `format`/`inspect` through exactly that file and passing. Same for `assert`, `stream`, `buffer`, `process` and 13 others.

Those are now **skipped, with the skip counted and printed next to the findings** so it cannot become a silent carve-out:

```
reachability notes — 23 `native` slot package(s) whose globals.mjs is NARROWER than
the root entry (152 export(s); globals-entry-parity, reported not enforced):
  · @gjsify/webcrypto (browser, nativescript, node) — 4: Crypto, CryptoKey, SubtleCrypto, crypto
  · @gjsify/timers (browser, node) — 2: Immediate, Timeout
  · @gjsify/gamepad (browser, nativescript) — 11: GamepadManager, MANETTE_TO_W3C_AXIS, …
  …
  (17 further package(s) NOT compared: their globals.mjs star-re-exports a runtime
   module — @gjsify/assert, @gjsify/async_hooks, @gjsify/buffer, … — which surfaces
   the whole runtime surface and is not statically enumerable.)
```

So the honest figure is **23 of 40 comparable packages / 152 names**, not 38 / 314; my earlier comment's block is superseded by this one. I spot-checked two survivors against their sources rather than trusting the tool twice: `@gjsify/timers` does export `Timeout`/`Immediate` from the root only, and `@gjsify/webcrypto`'s `globals.mjs` really carries just `subtle` plus the default.

The skip has a **residual blind spot, and it is written down rather than hidden**: a `globals.mjs` that stars a runtime module *and* has a root export that module does not carry is skipped too, so a real gap there stays invisible. Closing it means asking the runtime for the star target's export set — runtime **evaluation**, which `audit-runtimes` deliberately does not do (it must not crash on a browser-only re-export), so that is its own decision and not a quiet widening of this check. Noted in the function and in `status/open-todos.md`.

The conclusion is unchanged — reported not fatal, and gamepad's 11 remaining names need a platform entry rather than a `globals.mjs` line — but the class is a third of the size I first reported.

JumpLink:
**B3 / S2 / R1 verified end-to-end on a real emitted bundle**, not only through the injected `importer` — because "it classifies correctly in a unit test" is not the claim; the claim is what a plain-Node consumer of this `partial` package experiences.

A three-line consumer (`hasGamepadBackend()`, then `new GamepadManager().getGamepads()`), built with `gjsify build --app node` and run under Node **with `@gjsify/node-gi` absent from `node_modules`**:

```
$ node out.node.mjs
query says: false
getGamepads(): []
[@gjsify/gamepad] No gamepad backend in this process — @gjsify/node-gi is not installed,
and it is how a --app node bundle reaches gi://Manette. getGamepads() reports no
controllers until it is added (npm install @gjsify/node-gi, plus libmanette + its
typelib). Gate on hasGamepadBackend().
$ echo $?
0
```

Four things at once, all of them the point of the review:

* **B3** — the missing bridge is a **warn** with its own install advice, not an `error` with a stack trace calling the host broken.
* **S2** — the two `console.log` lines are the program's own; `hasGamepadBackend()` itself printed **nothing**. The diagnostic arrives afterwards, from the `getGamepads()` that actually wanted a monitor.
* **R1** — `getGamepads()` answers `[]`, not `[null,null,null,null]`.
* and **exit 0** under Node's default `--unhandled-rejections=throw`: the un-awaited init still cannot take the host process down.

JumpLink:
**B2 verified in real browsers**, since the finding was about what a web page sees.

A `--app browser` bundle whose entry is `import '@gjsify/gamepad/register'` plus one `navigator.getGamepads()` call, loaded as a module in headless Firefox and Chromium with every `console` event and `pageerror` captured:

```
firefox  probe = {"length":0,"json":"[]",                 "impl":"function getGamepads() { [native code] }"}
firefox  console output = []            gjsify lines = 0
chromium probe = {"length":4,"json":"[null,null,null,null]","impl":"function getGamepads() { [native code] }"}
chromium console output = []            gjsify lines = 0
```

Three claims, all measured:

* **`[native code]`** — `/register` no longer overwrites the platform's `getGamepads`. That was the worse half of B2: the old unconditional assignment replaced a *working* browser implementation with a libmanette manager that can never report a controller, because a `--app browser` build maps `gi://Manette` to an empty module by design.
* **zero console output** — no `console.error` asserting a host fault that does not exist. Not "a quieter message": nothing at all, because on this target there is nothing to install and nothing to fix.
* each browser returns **its own** shape (Firefox `[]`, Chromium four nulls), i.e. the page gets the platform's real answer, which is what `runtimes.browser: "native"` promises in the first place.

The fault path is not weakened by any of this — the by-design empty stub is separated from a broken namespace by measured object shape (0 own props + `Object.prototype` vs. 16 own props eagerly), and both directions are asserted in `backend.spec.ts`.

PR #980, opened 2026-08-04
fix: gate GTK bundles from a script, not node -e

JumpLink:
v0.28.0 published `@gjsify/cli`, `@gjsify/node-gi` and `@gjsify/napi` and left all three `@gjsify/gtk-runtime-*` packages at **0.27.1** — the one version whose defects the gate that blocked them had just been written to prevent shipping.

**Neither bundle was at fault.** Both were correct, measured from the manifests the failing steps had already printed:

| | darwin-arm64 | win32-x64 |
|---|---|---|
| `windowing` | `true` | `true` |
| `dataBytes` | 20 247 017 | 5 628 218 |
| backed typelibs | 25 | 25 |
| license texts | 65 | 37 |

Every assertion would have passed.

## What actually failed

Both verify steps ended in

```js
… data sets ${verified.map((v) => v.id).join('+')}`);
```

inside a single-quoted `node -e '…'`. The shell closes the body at the quote before `+`, so node got `join(+)` and died with `SyntaxError: Unexpected token ')'` before evaluating anything. The quote is visually inert — the JS reads correctly, only the shell layer between the YAML and node disagrees — and it survived review twice because the block existed twice, once under bash and once under pwsh.

## The fix

The manifest half becomes **one script both legs call** (`packages/node-gi/scripts/verify-bundle-manifest.mjs`): no shell-quoting layer, one copy instead of two that must agree across two shells, testable without cutting a release. `--expect-host-target <os>` keeps the property the inline version had and tightens it — OS *and* arch compared against the **runner**, not the matrix value, which is the class that let the emulated prebuild legs stage x86-64 into `prebuilds/linux-ppc64/` with every downstream check green.

## Why it reached a release at all

`release.yml` triggers on `release` / `workflow_dispatch` only, so **no pull request has ever executed either block** — the PR-coverage-parity hole AGENTS.md already documents for `deploy-docs` and `build-ci-image`. Two things close it:

- **`scripts/check-workflow-inline-scripts.mjs`**, in `audit-runtimes.yml` (behind a required check, pure Node, no install): refuses a multi-line inline body containing the quote that opened it. Scoped to the multi-line form on purpose — a first draft that counted quotes on single-line invocations produced 23 findings of which **21 were false**, including the *correct* `cli-cross-platform.yml:193` block, and a check with false positives gets switched off.
- **`tests/e2e/release-bundle-gate/`** — the only place either script runs before a release does. 12 tests, asserting the failure paths incl. the exact 0.27.1 manifest shape and the exact body that broke, because a gate that cannot fail is what 0.27.1 shipped behind.

## Verification

- guard against the broken tree → exactly the 2 real findings, nothing else; against the fixed tree → 12 workflows clean
- `node --test tests/e2e/release-bundle-gate/run.mjs` → 12/12
- gate accepts the measured 0.28.0 manifest, rejects the 0.27.1 shape on all 4 counts, rejects an empty declared data set, a foreign arch, a foreign OS, a missing manifest
- `oxlint` clean, `oxfmt` applied, both workflows parse as YAML

## Follow-up

The bundles' `license: MIT` still contradicts their LGPL payload now that the texts ship — recorded in `status/open-todos.md` rather than fixed here, because the right spelling is a judgement and wants the same derived-not-hand-edited treatment as the notices.

**After merge the three bundle publish legs need re-running so 0.28.0 bundles exist on npm.**

### DIFF
--- a/.github/workflows/audit-runtimes.yml
+++ b/.github/workflows/audit-runtimes.yml
@@ -115,6 +115,16 @@ jobs:
       - name: Check the PR-title type list still matches commitlint
         run: node scripts/check-pr-title-types.mjs
 
+      # A multi-line inline `node -e '…'` whose body contains the quote that opened
+      # it: the shell truncates the program and the step dies before asserting
+      # anything, so the gate reports nothing about what it was gating. Belongs here
+      # for the same reason as the rest of this job, plus one more — the workflows
+      # most in need of it (`release.yml`, `cli-cross-platform.yml`) never run on a
+      # pull request, so this is the only place their inline scripts get read before
+      # a release. Incident + scope in the script header.
+      - name: Check inline workflow scripts are not self-truncating
+        run: node scripts/check-workflow-inline-scripts.mjs
+
       - name: Check for resurrected prose (advisory)
         continue-on-error: true
         env:
@@ -197,3 +207,13 @@ jobs:
 
       - name: Check the PR-title type list still matches commitlint
         run: node scripts/check-pr-title-types.mjs
+
+      # A multi-line inline `node -e '…'` whose body contains the quote that opened
+      # it: the shell truncates the program and the step dies before asserting
+      # anything, so the gate reports nothing about what it was gating. Belongs here
+      # for the same reason as the rest of this job, plus one more — the workflows
+      # most in need of it (`release.yml`, `cli-cross-platform.yml`) never run on a
+      # pull request, so this is the only place their inline scripts get read before
+      # a release. Incident + scope in the script header.
+      - name: Check inline workflow scripts are not self-truncating
+        run: node scripts/check-workflow-inline-scripts.mjs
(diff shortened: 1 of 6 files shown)


JumpLink:
**Correction — the coverage claim in this PR's body is false, and the test it names never ran.**

I wrote that `tests/e2e/release-bundle-gate/` "IS the pre-release coverage: it is the only place either script runs before a release does". It is in **no npm script**, so it runs nowhere. Verified: `package.json#scripts.test:e2e` lists 100 suite paths, 112 exist on disk, and this suite is one of the 12 absent ones.

That is the same defect class this PR was fixing — a gate that cannot fail — reintroduced by the fix, one layer up. The scripts themselves are sound (they were exercised by hand and the guard demonstrably found the two real findings), but nothing re-runs them, so the next edit to either is unguarded.

`scripts/e2e-shard.mjs` parses that script rather than globbing `tests/e2e/*`, deliberately: *"Some suite dirs are intentionally NOT in test:e2e, e.g. terminal-native, which needs its own `gjsify run build` to produce a probe; a glob would wrongly run them."* Sound reason — but only terminal-native's exclusion is recorded anywhere, so the other eleven are indistinguishable from oversights, and mine was one.

Being fixed in a follow-up: the suite gets added to `test:e2e`, and the omission list becomes an explicit suite → reason ledger with a check behind it, so "not run" has to be a stated decision rather than a silent gap. Found by Pascal, not by me.

PR #984, opened 2026-08-05
ci: make every e2e suite run or say why not

JumpLink:
`package.json#scripts.test:e2e` named **100** suite paths while **112** existed on disk. Twelve ran nowhere — no shard, no workflow, no local `run test:e2e`.

```
css-as-string-bare-import   gjs-cli-config-load      pack-gitignored-files
devtools-export             gjs-less-host            prebuild-declaration-invariant
flatpak-sources             homepage-run-variations  release-bundle-gate
library-src-leak-guard      terminal-native          workspace-unknown-name
```

**Eleven were oversights** — I ran each one; all eleven pass unchanged. Two show what the gap cost:

- **`gjs-less-host`** covers the Node-free host path, which is exactly where a released `gjsify lint` regression was found this week.
- **`release-bundle-gate`** was written (#980) to cover the release gate that let v0.28.0 publish `cli`/`node-gi`/`napi` while its three GTK bundles stayed a version behind. That PR's body claimed the suite *was* the pre-release coverage. It was in no script, so it was coverage of nothing. Corrected on #980.

## Why the list exists, and what was missing

`e2e-shard.mjs` parses the script rather than globbing `tests/e2e/*`, deliberately and correctly — a suite *can* need setup the shared batch does not do, and `terminal-native` genuinely does: it drives a probe binary that exists only after its prebuild is staged. What was missing is the other half, **a record of which omissions are meant.**

`scripts/e2e-unlisted-suites.mjs` is that record. `scripts/check-e2e-suite-coverage.mjs` holds it in three directions:

| | |
|---|---|
| a suite neither listed nor ledgered | **fail** — the silent gap |
| a ledger entry that IS listed | **fail** — a deferral cannot outlive its cause |
| a ledger entry whose directory is gone | **fail** — stale deferral |

Same shape as `PREBUILD_GIR_GAPS` and `scripts/manifest-conformance/unchecked-fields.mjs`, for the same reason: an honest "not covered" must be available, a silent one must not. Every deferral prints on every run, and the reason must say what the suite *needs* — not that it is slow, and not that it fails. A failing suite is a finding, not an exemption.

Wired into `audit-runtimes.yml` beside the other repo-scoped guards — plain Node over the repo's own files, no install, no build, behind a required check.

## One suite gate fixed along the way

`devtools-export` checked for `gjs`, `dbus-run-session`, `gdbus`, the CLI bundle, the rolldown typelib and `node_modules` — every condition about the **toolchain**, none about whether the package under test was **built**. On a tree with `node_modules` but no `packages/framework/devtools/lib` it ran and died inside rolldown with `gjsify-unresolved-workspace-import: NotFound("@gjsify/devtools")`, i.e. an environment fact reported as a failed assertion about DBus.

## Verification

- check passes on the fixed tree: `110/111 e2e suite(s) run, 1 deliberately unlisted, 0 unaccounted for`
- fails when a suite is dropped from the script (exit 1, names the suite)
- fails when a ledgered suite is also listed (exit 1, self-retiring)
- all four shards now enumerate 111 suites (29 / 28 / 27 / 27), and `release-bundle-gate` lands in shard 2
- `oxlint` clean, `oxfmt` applied

Found by Pascal, not by the tooling.


JumpLink:
**The red `E2E 4/4` was one of the newly-listed suites, and it is exactly the information this PR was built to surface.**

`devtools-export` had never run in CI. Listing it gave it a first run, which failed — and the measurement points at the environment rather than at `@gjsify/devtools`:

```
APP_ON_BUS=yes            the fixture DID own org.example.reprotest
INSTALL_RETURNED=null     installDevtools returned null
EXPORT_LOG=no             the "exported org.gjsify.Devtools" line never printed
GETSTATUS  -> GDBus.Error:…ServiceUnknown: The name org.example.reprotest
                          was not provided by any .service files
```

with `dbus-daemon` activating `org.freedesktop.portal.Desktop` on the fixture's own request in between, and `xdg-desktop-portal` then failing on `Document portal fuse mount point unknown`. So the app **owned its name, lost it, and devtools never installed.** The same suite passes on a normal desktop session.

**Ledgered, with the cause stated as an open question rather than an exemption.** This ledger's own rule is that a reason must say what the suite *needs* — "not that it is slow, and not that it fails" — so the entry records the measured environmental dependency *and* says plainly that why the name goes away is not yet understood. It also names the real repair, which is not a ledger entry: a precondition in the suite's own SKIP gate (it already carries nine) so it skips where an Adwaita GApplication cannot complete startup and keeps running where it can. Candidates to separate first — early app exit, portal churn, or `app.get_dbus_connection()` genuinely absent at `startup` — are in `status/open-todos.md`, along with the note that the driver already captures the app's own log and that log is the next thing to read.

The other ten stay listed. Now **109/111 run, 2 deliberately unlisted, 0 unaccounted for.**

Worth saying plainly: a suite that has never run is not a suite that works. Ten of the eleven did work; the eleventh did not, and there was no way to know which was which before they ran. That asymmetry is the whole argument for the check.

PR #1026, opened 2026-08-06
fix(cli): resolve native prebuilds in-process

JumpLink:
> ### Status — DRAFT
>
> **Implementation is complete and fully verified locally** (every command in the
> Verification section below was run and is green, including a negative control
> proving the new e2e is a real gate).
>
> **What is outstanding is CI evidence, and only that.**
>
> GitHub Actions is degraded for this repo right now, and it is not a
> configuration problem on this branch:
>
> - The `pull_request` trigger has produced **zero** runs for this PR — not on
>   `opened`, not on `reopened` (I closed/reopened to re-fire it), not on
>   `synchronize` (second push). No workflow in `.github/workflows/` gates on
>   draft state, so draft status is not the cause.
> - `workflow_dispatch` on `main.yml` (the documented on-demand full sweep) DOES
>   create runs, so the API accepts them — they then sit `queued`.
> - Repo-wide, `GJS` runs have been `queued` since **19:46 UTC on `main`** and
>   20:58 UTC on another branch. Runners are not picking work up.
>
> A dispatched full sweep for the current head SHA is queued:
> https://github.com/gjsify/gjsify/actions/runs/31128417929
>
> **`gh pr checks` reports "no checks reported on the branch". That is not a
> green — it is the absence of checks**, which is precisely the false green
> `docs/governance.md` § required-checks was written about (#910 merged with 12
> checks still running).
>
> I have deliberately **not** merged this. `docs/governance.md` makes the
> three-required-check set non-negotiable specifically so that `main` cannot take
> a commit with no verdict behind it, and merging with zero CI would be the exact
> incident that rule exists to prevent. Mark ready + merge once a real run for the
> head SHA reports green — or explicitly decide to override, which should be the
> repo owner's call and not mine.

---

## The problem

`gjsify build` under GJS needs `@gjsify/rolldown-native`'s prebuild, and it only
resolved when the process had been started through the **`gjsify` launcher**,
which exports `GI_TYPELIB_PATH` + the host library-path variable before `exec`.
Invoked any other way — `gjs -m …/dist/cli.gjs.mjs build …`, which several e2e
suites and any embedding consumer do — the build died with *"no usable bundler
engine under GJS"* on a tree where the engine was installed, built and loadable.

The CLI's own diagnostic asserted it was unfixable, and so did
`packages/infra/cli/AGENTS.md` and ~30 per-target package READMEs:

> The typelib lookup happens inside the GJS runtime, so those must be set BEFORE
> the process starts — **the CLI cannot repair it from the inside.**

## The measurement

gjs 1.88.1 / GLib 2.88.3, importing `gi://GjsifyRolldown` from a bare `gjs -m`:

| environment | result |
|---|---|
| neither variable | `Typelib file for namespace 'GjsifyRolldown' (any version) not found` |
| `GI_TYPELIB_PATH` only | `Failed to load shared library 'libgjsifyrolldown.so' referenced by the typelib` → surfaces as `Unsupported type void, deriving from fundamental void` |
| both | loads |

**There are two lookups, not one**, and that is what the old answer collapsed.

Half of it is true forever — `LD_LIBRARY_PATH` cannot change in-process, ld.so
freezes its search path at startup — but it is the wrong frame, because
**girepository delegates neither lookup to the loader's environment**. It keeps
its own typelib path *and* its own library path, consults both before falling
back to the system loader, and both are writable at runtime:

- `gi_repository_prepend_search_path()` — typelib (GLib 2.80)
- `gi_repository_prepend_library_path()` — library (GLib 2.80), documented as
  *"If the library is not found in the directories configured in this way,
  loading will fall back to the system library path (i.e. `LD_LIBRARY_PATH` and
  `DT_RPATH`…)"*

The second is the one nobody went looking for, because `LD_LIBRARY_PATH` was
assumed to *be* the mechanism. It is only girepository's fallback — and the
middle row above is the proof, since girepository's library path starts empty
and no environment variable seeds it.

## The decisive proof

`gjs -m packages/infra/cli/dist/cli.gjs.mjs build … --app node`, invoked
directly, **no launcher**, with both variables *deleted* from the environment:

**BEFORE** — `EXIT=1`
```
gjsify build: no usable bundler engine under GJS — `@gjsify/rolldown-native` is not loadable…
GI_TYPELIB_PATH / LD_LIBRARY_PATH do NOT contain …/rolldown-native/prebuilds/linux-x64, so the
`gjsify` launcher was bypassed … the CLI cannot repair it from the inside.
```

**AFTER** — `EXIT=0`
```
$ cat /tmp/rpath-proof/out.js
const hello=e=>`hi ${e}`;export{hello};
```

Pinned by `tests/e2e/launcher-free-build/`. **Negative control run**: with only
the three source files reverted and the bundle rebuilt, the suite goes red on
exactly that assertion — it is a real gate, not a vacuous one.

## No artifact changes — the rpath was the expected fix and is not one

I set out to add an `$ORIGIN`/`@loader_path` rpath. It is **already there**, and
it does a *different* job:

```
$ readelf -d …/rolldown-native-linux-x64/prebuilds/linux-x64/libgjsifyrolldown.so
 (NEEDED)   Shared library: [libgjsify_rolldown.so]
 (SONAME)   Library soname: [libgjsifyrolldown.so]
 (RUNPATH)  Library runpath: [$ORIGIN]
```

Every bridge's `meson.build` has set it since it was written, and it is already
machine-checked by `scripts/check-prebuild-loader-path.mjs` +
`tests/e2e/prebuild-loader-path/`. It lets the Vala library find the Rust cdylib
shipped beside it. It **structurally cannot** help gjs find
`libgjsifyrolldown.so` itself: the typelib names it by bare SONAME, and
`dlopen()` of a bare name searches the RPATH/RUNPATH of the **calling** object —
libgirepository, whose `$ORIGIN` is `/usr/lib64`. An rpath on the callee is never
consulted to find the callee.

So **nothing is rebuilt, re-staged or re-published**, which also means no
`prebuilds.yml` run and no release coupling.

## Scope: how wide, and why

The brief was "start with rolldown-native, decide deliberately about the rest".
The mechanism turned out **not to be per-bridge**, which dissolves the question:
`activateNativePrebuilds()` activates every directory `detectNativePackages()`
resolves — the *same* function whose output the launcher turns into environment
variables.

> The in-process activation is the launcher's environment, applied to the process
> that is already running. Same detection, same set, two mechanisms for two
> different processes.

That covers all three engines the CLI loads in-process today —
`rolldown-native`, `lightningcss-native` (via the `css-as-string` plugin) and
`oxfmt-native` — with **no list of covered bridges to drift**, and a bridge added
later is covered without touching the file. Two call sites, both already
memoized: `tryLoadNative()` and `tryLoadNativeOxfmt()`.

**The launcher is not removed.** `gjsify run <bundle>` hands prebuilds to a
*child* process, which only an environment can do, and a Node-less host still
needs a runnable bin. This removes the launcher as a *precondition for the CLI's
own engines*, nothing more.

## What it deletes

- **`diagnoseNativeEngine()`'s bypassed-launcher branch** (~30 lines). Doubly
  wrong by now: the cause no longer exists, and the path it told readers to
  export — `<facade>/prebuilds/<target>` — **has held nothing since ADR 0017**
  moved every artifact into a per-target sibling package. Verified: that
  directory does not exist in an installed tree.
- **The class** "works through the launcher, fails when invoked directly" —
  never fixed, only worked around each time (most recently by retargeting an e2e
  assertion from `gjsify build` to `gjsify copy`).
- **The e2e constraint** that `runGjs`-style suites must avoid an inner
  `gjsify build`.

A third e2e test was written and **deleted** rather than kept: it asserted the
old advice was absent, and it passed against an unfixed tree, because a build
failing for any other reason never reaches that diagnostic. A check whose input
set is empty passes while checking nothing (`docs/governance.md` § simplicity).
The reason it is gone is recorded in the suite so it is not re-added.

## What it does NOT solve

- `LD_LIBRARY_PATH` is still immutable in-process. Unchanged, and why the
  launcher stays.
- Only the CLI's own process. A user bundle under a bare `gjs -m` still needs
  the launcher or its own `prepend_*` call.
- Nothing about macOS/Windows is *proven* — the API is platform-independent and
  `detectNativePackages()` already resolves per host, but measurements are
  linux-x64.
- Not a system-library fix; the measured `libjson-glib` branch (#994,
  ts-for-gir#437) is untouched.
- GLib < 2.86 gains nothing (`dup_default()` is the process-global accessor).
  Handled by a capability probe, so those hosts keep today's behaviour exactly —
  the change is strictly additive.

## Governance

- **ADR 0021** (`docs/adr/0021-launcher-free-prebuild-resolution.md`) + index
  row; `check-adr-index.mjs` green.
- `packages/infra/cli/AGENTS.md` corrected — the launcher paragraph now says what
  the launcher is still *for* instead of what it is *required* for, keeping the
  postmarketOS incident that justifies the rule and adding "do not re-derive
  'the CLI cannot repair it from inside' from `LD_LIBRARY_PATH`".

## Verification (all local, all green)

```
gjsify workspace @gjsify/cli run check          ✔
gjsify workspace @gjsify/cli run test           ✔ 1947 tests
node --test tests/e2e/launcher-free-build       ✔ 2/2  (red without the fix)
node --test tests/e2e/run-command               ✔ 15/15
node --test tests/e2e/global-install-engine     ✔ 3/3
node --test tests/e2e/oxc                       ✔ 10/10   (touched oxc-resolve)
node --test tests/e2e/css-as-string-bare-import ✔ 2/2     (lightningcss path)
node scripts/audit-runtimes.mjs --check --strict ✔
node scripts/check-adr-index.mjs                ✔ 21 ADRs
node scripts/verify-committed-bundles.mjs       ✔ all reproduce byte-identically
gjsify format && gjsify lint                    ✔
```

`affected.gjs.mjs` rebuilds byte-identical (this change is not in its graph).


JumpLink:
Independent local verification, from a **separate worktree** (fresh checkout of `main` + `gjsify install --immutable`, not the authoring tree) — recorded here because CI still cannot produce a run.

**Respecting the DRAFT condition:** not marking ready, not merging. The head SHA still has no run.

| check | result |
|---|---|
| `audit-runtimes --check --strict` + 4 further audit scripts | 5/5 PASS |
| `@gjsify/cli` suite | **1947 completed** (+3 vs main = the new `gi-search-path` specs) |
| `tests/e2e/launcher-free-build` | **2/2 pass** |

The e2e passes the decisive one — *builds with `GI_TYPELIB_PATH` and `LD_LIBRARY_PATH` deleted from the env* — plus the girepository API pin.

### One trap worth recording for whoever runs this next

The suite **failed for me on the first attempt**, and it was not this PR. `run.mjs` boots `packages/infra/cli/dist/cli.gjs.mjs` directly and builds nothing itself, so it measured a `dist/` left over from a `main` build. The giveaway was that the assertion diff contained *verbatim the old diagnostic this PR deletes* — "the CLI cannot repair it from the inside".

After `gjsify workspace @gjsify/cli build --with-dependencies && … build:gjs-bundle`, both tests pass. In CI the ordering makes this invisible (the build job produces the bundle first), so it only bites locally — same shape as the repo's own "it is never the source you edited that runs" rule.

### Non-blocking follow-up

15 per-target `README.md` files still document `GI_TYPELIB_PATH`/the launcher as a requirement. The PR updates `packages/infra/cli/AGENTS.md` and ADR 0021, so the claim this PR disproves survives in those 15 copies.

JumpLink:
Correction to my previous comment — the "non-blocking follow-up" in it is wrong, and I would rather retract it than leave someone chasing it.

I claimed 15 per-target READMEs still document the launcher as a requirement. I had only grepped for `GI_TYPELIB_PATH`, not read what they say. They actually say:

> `@gjsify/cli` finds the artifacts at run time by scanning `node_modules` for packages declaring `gjsify.prebuilds`, and puts this directory on `GI_TYPELIB_PATH` plus the host loader's own search path.

That is *run time*, by the CLI — which is precisely what this PR implements, and it is generated text (`scripts/generate-platform-packages.mjs`), so all copies agree. **No follow-up is needed there.**

The verification in my earlier comment stands unchanged: audit 5/5, `@gjsify/cli` 1947 completed, `launcher-free-build` 2/2 against a freshly built bundle — as does the stale-`dist/` trap, which is the part actually worth keeping.

PR #1031, opened 2026-08-06
fix: make the macOS and Windows suites green

JumpLink:
Both platform legs have been red since they landed, and neither had ever been
green: #1021 and #1022 merged during the 2026-08-06 GitHub Actions degradation
with **no checks reported on their PRs**. Every finding below is therefore
*pre-existing* — none of the nine PRs that merged that day is involved,
including #1026 and #1029, the two plausible suspects.

**Result: macOS arm64 28/28, macOS x64 28/28, Windows 25/25 — zero steps
skipped, zero failed.**

Each fix uncovered the next, because no leg had ever run far enough to see it.

| | leg | defect |
|---|---|---|
| 1 | both | `build:infra` ran **before** `install`, so `@gjsify/vite-plugin-blueprint` (`"types": ["node"]`) could not typecheck — `TS2688`. The whole macOS failure, both arches, ~30 s in. |
| 2 | Windows | `where sh && exit 1 \|\| echo "good"` — `ECHO` does not reset `ERRORLEVEL` in cmd.exe, so the step failed *because* the check passed. |
| 3 | both | After `install`, the tree's own `node_modules/.bin/gjsify` exists and dispatches to build outputs `build:infra` is what produces. |
| 4 | `main` | `gjsify format --check` is a whole-repo gate; eight files from #1021 were unformatted, so the `check` job was red too. |
| 5 | both | `build:infra` left the CLI's own four runtime deps at `build:types` — no `lib/esm` — so the parallel sweep was *creating* them while another package's build imported them. |
| 6 | Windows | Every `--include`/`--exclude` in the root scripts was **single-quoted**, which cmd.exe does not strip. All exclusions were silent no-ops: `run build` selected **181** packages vs 160 and died building GTK showcases on a toolkit-less runner. |
| 7 | both | `@gjsify/node` is not a package (it is `@gjsify/node-globals`), and `@gjsify/utils` is GJS-only with no `build:test:node` — the Windows leg died on it as its FIRST suite step. |
| 8 | macOS | `os.machine()`'s known-values list was a Linux `uname -m` vocabulary; Apple Silicon reports `arm64`. |
| 9 | both | `fs.promises.watch` specs never cancelled their deferred write, so it fired after teardown and the ENOENT was blamed on the next spec. |
| 10 | Windows | `makeTmp()` needs `realpathSync.**native**`. An 8.3 short `tmpdir()` makes libuv's `uv__relative_path` assert and **abort** (exit 3221226505), taking all 589 specs with it. Plain `realpathSync` does not canonicalise short names — measured, the first attempt aborted identically. |
| 11 | macOS | `@gjsify/net` client sockets: five had no reset tolerance and two used a bare `reject`, so a teardown RST was blamed on the next spec. |
| 12 | macOS | `computeNativeEnvForBundle` is not the empty case on darwin — `buildNativeEnv()` puts `systemGiLibraryDirs()` on `DYLD_FALLBACK_LIBRARY_PATH` unconditionally. |
| 13 | Windows | `build:test:node` did not run `prebuild:test:fixtures`, so `test:node` ran against a worker fixture that did not exist. |
| 14 | macOS | The `prebuild-artifacts` rule **blamed the artifact for its own name**: dyld names both the library it opened and the one it could not find, so the substring test always matched the probed leaf. glibc names only the missing one — which is why Linux never saw it. All three darwin prebuilds were reported as staging defects for Homebrew libraries none of them stages. |
| 15 | Windows | `systemGiLibraryDirs` used host-flavoured `join`, so on win32 it built `\usr\local\lib\…` and never matched the darwin prefix it was asked about. Fixed in **both** mirror copies. |
| 16 | Windows | `dlx-cache`'s forced-POSIX refresh row asserts a rename-over-existing-symlink guarantee win32 does not make. |

Only 2, 6, 7, 8, 10, 15 and 16 are genuinely platform-specific. 1, 3, 5, 9, 11
and 14 all reproduce or are explainable on Linux — CI never saw them because its
bootstrap runs the published *GJS* bundle (where `detectPackageManager()`
already returns `gjsify`) and it restores a complete `lib/` cache. Exactly the
shape ADR 0018 § 5 predicts.

## The core fix (`fix(cli): let a bootstrap CLI drive nested calls`)

Two halves, because one alone is measurably not enough:

- `ensureGjsifyShimOnPath()` was a no-op under Node. It now also serves a
  **bootstrap CLI** — entry outside the workspace, i.e. npx or a global install
  — writing npm's three-sibling shim trio so `%COMSPEC%` can run it. Scoped that
  way deliberately: when the invocation already came through the tree's own
  `.bin/gjsify`, today's dispatch stays byte-for-byte unchanged.
- `detectPackageManager()` returns `gjsify` for that CLI. Without it the repair
  is undone one level down — npm re-prepends `node_modules/.bin` for every
  lifecycle script. Measured: with only the first half, `build:infra` cleared
  its first clause and then failed inside `npm run build` at the same message.

`needsSelfShim()` takes the runtime and both paths as parameters rather than
reading them ambiently, so the branch only Windows can *hit* is unit-tested from
a Linux host.

## Nothing was weakened

No `it.skip`, no deleted check, no relaxed assertion.

- Defect 16 is the one tolerated failure, and it uses the sanctioned
  `it.failing(…, { when })`: it **runs** on Windows, is reported as
  `✗ … (expected failure — …)` with `1 expected failure` in the summary, and
  fails the day win32 honours it. The Windows log shows 11 such rows total.
- Defect 12 replaced `toBe('')` with a **sharper** claim (the fallback variable
  carries the repair; the override and `GI_TYPELIB_PATH` stay untouched).
- Defect 14 kept the rule's teeth: a genuine unresolved `libgjsify*` sibling
  still fails; only the self-blaming substring match was corrected.
- Defect 8 added a value the OS genuinely returns to a list that lacked it.
- The `style:` commit is pure oxfmt output.

Defect 5 is **narrowed, not closed**, and the ledger says so: a concurrent
rewrite is still a window, just one where every file already exists. The
structural fix belongs in the `foreach` scheduler and is filed.

## The bridge, and how it retires

Until a CLI carrying `usingSelfShim()` is `latest` on npm, both legs set
`GJSIFY_SHIM_DIR` and `npm_config_user_agent=gjsify/bootstrap` — two levers the
published 0.30.0 already honours. Both carry a named **retirement trigger** plus
a `status/open-todos.md` entry, because an inert workaround left in place hides
whether the core fix works.

`status/open-todos.md`'s "`@gjsify/child_process` on Windows — 86 of 145" entry
is deleted: the leg now runs it green at 226 specs. The incident survives in
`windows-suites.yml`'s header and the residual claim in the package's own
`gjsify.osNotes.win32`.

## Measured vs inferred

**On real runners (this PR):** everything in the table above, and the final
green run — macOS 28/28 on both arches, Windows 25/25.

**Locally (Linux/x64/glibc), cold worktree with the build cache cleared:** the
`TS2688` failure; the `Cannot find module …/cli/lib/index.js` bootstrap failure
and its exit-0 after the fix; the same with the *published* 0.30.0 plus the two
bridge variables; a cold full build of all 160 packages; defect 6's mechanism
(`--exclude "@girs/*"` selects 179, `--exclude "'@girs/*'"` selects 187);
`audit-runtimes --check`, lint, `format --check`, and
`verify-committed-bundles` all green.

**Inferred, never measured:** nothing material remains — defect 2 and defect 10
were both confirmed on the Windows runner after being diagnosed from logs.


JumpLink:
Heads-up: this branch now conflicts with `main`, and it is my doing — I merged #1030 a moment ago (`dae079a09`), which touches the same two files:

- `.github/workflows/macos-suites.yml`
- `.github/workflows/windows-suites.yml`

Not rebasing it for you, since you are still committing here (last commit 8 minutes ago). Say the word and I will.

**Two things I verified about this branch while checking main:**

1. `gjsify format --check` on `pr-1031` → *"All matched files use the correct format"*, and it covers **all 8** files that were unformatted on main.
2. After #1030 landed, main is down to **2** remaining format offenders — `scripts/report-symlink-capability.mjs` and `tests/e2e/run-command/run.mjs` — and this branch is what fixes them. #1030 fixed the other six as a side effect of rewriting the fs specs.

For the record: those 8 files reached main because I ran the audit scripts, actionlint, the package suites and the e2e sweep locally while Actions was down, but **not** `gjsify format --check` (`main.yml:808`). That is on me — it costs 150 ms over 2640 files. It is in my checklist now.

JumpLink:
Small overlap to flag, so you do not redo it: I took the two remaining format offenders into #1032 (`scripts/report-symlink-capability.mjs`, `tests/e2e/run-command/run.mjs`).

Reason: #1032 inherits `main`, so `Check Fedora 44` reds on it for those two files and it could not go green without them — while this PR is still a draft. The formatter is deterministic, so both branches produce identical bytes and the overlap should resolve to nothing; if git still flags it, take either side.

Everything else in this PR stays yours — the macOS/Windows bootstrap repair and the `gjsify-shim`/`run.ts` work have no counterpart in #1032.

JumpLink:
Rebase attempted and **deliberately aborted** — this needs your judgement, not mine.

Your branch is based on `de7f55254`; `main` has since taken #1030 (`dae079a09`) and #1032 (`bad2d1aed`). Five conflicts. Two I could resolve on evidence, three I could not:

**Resolved cleanly (had I continued):**
- `macos-suites.yml` / `windows-suites.yml` → **your** version. It is a strict superset of what #1030 landed: same install-before-build order, plus the `GJSIFY_SHIM_DIR` / `npm_config_user_agent` bridge that keeps the nested `gjsify` calls off the dead workspace shim. Nothing in #1030's variant is missing from yours.
- `fix(os): accept darwin's arm64` → **skippable**. #1032 landed the identical list entry (I hit the same failure independently). Only the comment differs; the assertion is the same.

**Where I stopped — two parallel solutions to the same problem:**
- `packages/node/fs/src/watch.spec.ts` — #1030 cancels the deferred writes through a `pendingWrites` set with `scheduleWrite()`/`cleanup()` owning timer *and* directory together; your `fix(fs): cancel the deferred write in the watch specs` does it with `finally` + `clearTimeout`. Both correct, mutually exclusive.
- `packages/infra/cli/src/run-node-resolve.spec.ts` — same shape.

Picking one would mean guessing which intent survives, and your later commits may build on yours. Since both approaches came from you, you are the one who can say. Say the word and I will redo it with your call — or push a rebase yourself and I will take it from there.

Sorry for the collision: #1030 and #1032 were merged while this was a draft, and I did not know you were fixing the same specs in parallel.

JumpLink:
The one red check here is **not** yours, and not a blocker (it is not a required check) — but it is the same class you are fixing, in a third place #1019 missed:

`better-sqlite3 consumer gate (GJS / Fedora 44)`:

```
TRANSPARENT GATE bufferutil: FAIL — GJSIFY_GATE_ENGINE=gjs but the committed CLI bundle
is missing: /__w/gjsify/gjsify/packages/infra/cli/dist/cli.gjs.mjs
```

`packages/napi/napi/test/transparent-gate.mjs:106` requires that file, and says why in its docblock:

> Using the COMMITTED bundle is deliberate: it carries the plugin code of the commit under test (the pre-commit hook rebuilds it; `verify-committed-bundles` proves it reproduces from that source), **so no in-repo CLI build is needed** — which is what let this run inside the existing consumer job instead of needing its own workspace build.

ADR 0002 removed exactly that premise. So this gate now dies on all four transparent addons (`bufferutil`, `utf-8-validate`, `sqlite3`, `argon2`) whenever `napi.yml` triggers, which your CLI-source changes do.

It cannot fall back to the Node driver on a missing bundle — that would silently stop measuring the GJS path, which is the whole point of the gate. It needs a real build in `napi.yml`, the same shape you gave the platform legs.

**I am not touching it until this PR lands**, to avoid a second collision in the workflows. Once you merge, I will fix it on top of the new `main` and verify it there. Flagging it now so it is not a surprise.

PR #1033, opened 2026-08-07
ci(napi): build the CLI bundle the transparent matrix drives

JumpLink:
The third place ADR 0002 left behind, after the macOS and Windows legs #1031 just repaired.

## What is broken

`packages/napi/napi/test/transparent-gate.mjs` requires `packages/infra/cli/dist/cli.gjs.mjs` by absolute path. #1019 untracked it, so the better-sqlite3 consumer job dies on **all four** transparent addons the moment `napi.yml` triggers. This is red on `main` too (run 31142398622, job 92754830935).

```
TRANSPARENT GATE bufferutil:      FAIL — GJSIFY_GATE_ENGINE=gjs but the committed CLI bundle is missing
TRANSPARENT GATE utf-8-validate:  FAIL — …
TRANSPARENT GATE sqlite3:         FAIL — …
TRANSPARENT GATE argon2:          FAIL — …
```

The gate's own docblock names the premise that was removed:

> Using the COMMITTED bundle is deliberate: it carries the plugin code of the commit under test … **so no in-repo CLI build is needed** — which is what let this run inside the existing consumer job instead of needing its own workspace build.

That matrix is the only coverage `napiNodeAddonPlugin`'s acquisition path has. It ran in no workflow until #840, which is why six defects in it shipped at once, two of them visible only under `@gjsify/rolldown-native` on gjs.

## Three easier repairs, all wrong

- **Point at the published CLI.** Its neighbouring comment rules it out in its own words: that CLI *"carries the LAST RELEASE's plugin, so a PR changing `napiNodeAddonPlugin` would test nothing at all"*.
- **Fall back to the Node driver** when the file is absent. Worse than the red: the GJS path stops being exercised while the job reports green — the exact false green this matrix exists to catch.
- **Alias the one unresolvable dep to `@gjsify/empty`.** This goes green immediately and is the sharpest trap, because it is *nearly* harmless — the matrix never invokes `gjsify create`. But it makes the artifact under test structurally different from the bundle that ships, and the gate's whole contract is "the bundle from this commit". The same objection retires a narrower build-only entry: a bundle shape no user runs is a bundle whose future divergence from the real one is invisible to the gate meant to catch it. (It also would not work as drafted — the gate passes `build` as a positional subcommand, which the cited `affected-entry.ts` parser ignores.)

## So it builds

One step, in the job that is already warm. The published CLI is the **driver**; the **entry** is this checkout's `src/index.ts`. That distinction is the whole design: what matters is whose source is bundled, not who runs rolldown.

`packages/infra/cli/src/actions/build.ts:5` statically imports `@gjsify/rolldown-plugin-gjsify`, whose `src/app/gjs.ts:19,269` is where `napiNodeAddonPlugin` lives and gets installed for every `--app gjs` build. That specifier resolves through the workspace symlink into `packages/infra/rolldown-plugin-gjsify/lib/` — built earlier in this job, or restored from `napi-libs-v2`, which is exact-only with **no `restore-keys`** over `packages/*/*/src/**` (all 43 of that package's sources are `.ts`, so the hash covers the whole input). So the emitted bundle carries **this commit's** plugin, and the matrix then runs `gjs -m dist/cli.gjs.mjs build …` on `@gjsify/rolldown-native`. Only the transform pass is last release's.

**`@gjsify/create-app` is built first.** It is the one workspace dep of this entry that neither existing fill step produces: no `build:gjsify` (the polyfill foreach skips it) and not in `@gjsify/rolldown-plugin-gjsify`'s dep closure (`--with-dependencies` never reaches it), while `src/commands/create.ts` imports it statically and its `exports` name a gitignored `lib/`. Two independent walks of the `--app gjs` closure (~330 files, 32 `@gjsify/*` specifiers, 27 packages) agree it is the **only** one; the rest are covered by `build:gjsify`, by `--with-dependencies`, or — for `@gjsify/resolve-npm` — by having a committed `lib/`.

**Both commands share one unguarded step, on purpose.** The guarded steps define what a `napi-libs-v2` entry carries, and every live entry was saved by a run that never built create-app — this PR's own run took an exact hit (`Cache hit for: napi-libs-v2-104a5967…`) and skipped both of them. A guard here would be skipped on precisely that hit and red the matrix: the v1 → v2 incident, repeated. Folding the two commands into one step is what stops that guard being re-added by someone reading only a step name — and it keeps create-app **out** of the fill-step set the cache comment's invariant is about, so that invariant stays literally true and no key bump is needed.

## Honest correction of the earlier commits

Both previous attempts on this branch carried a wrong diagnosis, and the second failure is a direct consequence of the first.

`8ef235413` ran the CLI package's own `build` + `build:gjs-bundle` and died in `gjsify tsc`. The commit message and the in-workflow comment blamed `@types/node`, "measured here, not guessed". The log it cites says otherwise: 63 errors in 26 files, of which 34 are TS7016 `Could not find a declaration file for module` against `@gjsify/{workspace,npm-registry,tar,semver}` — this job fills `packages/*/*/lib` transpile-only, so there are no `.d.ts` — and the remainder cascade off the resulting `any`s. Zero errors mention `@types/node`, and zero touch any `node:` builtin, which is itself proof it resolved.

The same log also contained `TS2307: Cannot find module '@gjsify/create-app'`. `0a8e5f3d9` read past it, which is why the second attempt failed on the identical missing artifact by a different route.

`0a8e5f3d9` also rewrote the `GJSIFY_CLI_ENTRY` comment to say the new step "builds `packages/infra/cli/lib/` too". It does not — no step in this job does. `main`'s wording was correct and is restored.

## Also corrected

Not scope creep — these are statements this change or ADR 0002 made untrue:

- **`transparent-gate.mjs`'s docblock and `die()`** still assert the committed-bundle premise, including two guarantees that no longer exist ("the pre-commit hook rebuilds it", "`verify-committed-bundles` proves it reproduces"). That stale docblock is *why* this failure was invisible. It now names the producing step and records that the missing Node fallback is deliberate.
- **`dist/cli.gjs.mjs` described as "committed"** in the shim-build comment, in both jobs that carry that duplicated block. That comment also gains the provenance note above, so the no-`restore-keys` invariant is written down where a future edit would break it.

## Verification, stated honestly

**Not reproduced end to end.** This is a container job on `ghcr.io/gjsify/ci-fedora:44`; a locally built workspace has exactly the artifacts CI lacks, which is what made the previous attempt look proven when it was not. That trap is the reason for the checks below — each was chosen because it holds *independently* of the local tree being warm.

Run locally against this branch:

- `actionlint 1.7.12` (the pinned CI version, sha256-verified) over all workflows — clean.
- `scripts/check-workflow-inline-scripts.mjs` — 14 workflows clean; `scripts/check-ci-image-packages.mjs` — clean.
- `gjsify format --check` — 2641 files clean. `node --check transparent-gate.mjs` — parses.
- **`gjsify workspace @gjsify/create-app build` from a wiped `lib/`** — the one command here with no CI precedent.
- **`gjsify tsc --listFiles` on `@gjsify/create-app`** — the closure is 135 files and the only `packages/` entries are its own four sources. No workspace sibling, so the missing-`.d.ts` wall that killed `8ef235413` cannot reach it. This is the check that does not depend on the local tree being warm.
- `packages/infra/tsc/lib/lib*.d.ts` are **still tracked** (108 files), so `gjsify tsc`'s default libs are present in any checkout, including this job. Independently corroborated by run 31142741432, where `gjsify tsc` ran in this exact job and emitted ordinary type diagnostics rather than a missing-lib error.

The proof is the next run of `better-sqlite3 consumer gate (GJS / Fedora 44)`. It fails closed: the matrix cannot pass with the bundle missing and cannot fall back to Node.

## Not in this PR

`Build & gates & conformance (GJS / Fedora 44)` → `test:gate:tsfn-teardown` failed once here with `FAIL draining: no warning`. Not this PR: the diff is confined to the consumer job's steps, that job shares no cache with it, the same job passed on this branch's previous commit and on `main`'s tip `04773c9c1` (this PR's base — identical tree), the macOS leg of the very same run passed the same assertion, and only 1 of that job's 3 Fedora repetitions warned.

It is a fixture bug: `tsfn-teardown-case.js` waits on `stats()[0]`, a **global** delivered-callback counter, where the `draining` shape needs per-thread claim *attribution* — N deliveries from fewer than N threads satisfies it, leaving one claim unattributed, which teardown then correctly warns about. Its own PR, against `packages/napi/napi/test/{tsfn-addon/tsfn.c,tsfn-teardown-case.js}`; the runtime is not at fault, and the warning it emitted was accurate.

Note that `napi.yml` carries no required status check, so `--auto` will not wait for it. Confirm this run green by hand before merging.

### DIFF
--- a/.github/workflows/napi.yml
+++ b/.github/workflows/napi.yml
@@ -375,17 +375,26 @@ jobs:
       # went green. The bundle has to come from THIS commit's source, which
       # means building it.
       #
-      # Two steps because they answer different questions: `build` produces the
-      # CLI's own `lib/` (the bundle's entry), `build:gjs-bundle` produces the
-      # artifact. The polyfill libs and the bundler plugin they both need were
-      # built further up, and the published CLI on PATH is what drives this
-      # build — the same bootstrap `.github/actions/gjsify-setup` performs.
+      # The package's own `build:gjs-bundle` cannot be used here, and the reason
+      # is the same one that made this job attractive in the first place. It is
+      # `node lib/index.js build …`, so it needs the CLI's `lib/`, which comes
+      # from `build` = `gjsify tsc …` — and this job deliberately installs only
+      # the slice the bundler needs, so that typecheck dies on TS7016/TS7006
+      # against `@types/node` it never installed. Measured here, not guessed.
+      #
+      # So drive the SAME bundle out of the published CLI instead. That is not
+      # the substitution the comment below rules out: what matters is whose
+      # SOURCE is bundled, not who runs rolldown. The entry is this checkout's
+      # `src/index.ts`, so the artifact carries THIS commit's
+      # `napiNodeAddonPlugin` — which is the whole property the matrix needs.
+      # Using the published CLI as the DRIVER would be the broken variant, since
+      # it would execute last release's plugin.
       - name: Build the CLI's GJS bundle (untracked since ADR 0002)
+        working-directory: packages/infra/cli
         run: |
-          export PATH="$PWD/node_modules/.bin:$PATH"
-          gjsify workspace @gjsify/cli build
-          gjsify workspace @gjsify/cli build:gjs-bundle
-          ls -l packages/infra/cli/dist/cli.gjs.mjs
+          node /tmp/gjsify-node-cli/node_modules/@gjsify/cli/lib/index.js \
+            build src/index.ts --app gjs --outfile dist/cli.gjs.mjs --shebang
+          ls -l dist/cli.gjs.mjs
 
       # Driven by the dist/cli.gjs.mjs BUILT ABOVE, NOT the published CLI the
       # consumer gate uses — and that distinction is the whole point.
--- a/.github/workflows/napi.yml
+++ b/.github/workflows/napi.yml
@@ -316,7 +316,7 @@ jobs:
           # These jobs never use the tsc bundle, so skip building it.
           gjsify foreach build:gjsify --exclude '@gjsify/tsc'
 
-      # The committed dist/cli.gjs.mjs INLINES the build pipeline's CODE but not
+      # The CLI's `dist/cli.gjs.mjs` INLINES the build pipeline's CODE but not
       # its runtime DATA files: `resolveShim()` (rolldown-plugin-gjsify's
       # app/gjs.ts) hands rolldown FILE paths under
       # packages/infra/rolldown-plugin-gjsify/lib/shims/ (unicorn-magic,
@@ -324,14 +324,24 @@ jobs:
       # `./shims/<name>` subpath export at BUILD time and inlined into the
       # OUTPUT bundle. That `lib/` is a plain-`tsc` build output the polyfill
       # step above never produces (the package has no `build:gjsify` script), so
-      # without this step every `--app gjs` build driven by the committed bundle
-      # dies at config time with `cannot locate the "unicorn-magic" shim` — the
-      # first red of the GJS-engine transparent matrix. `--with-dependencies`
+      # without this step every `--app gjs` build driven by that bundle dies at
+      # config time with `cannot locate the "unicorn-magic" shim` — the first
+      # red of the GJS-engine transparent matrix. `--with-dependencies`
       # tsc-builds the small workspace-dep closure the plugin's own `tsc` needs
       # declarations from (console, deepkit, pnp, blueprint); deps without a
       # `build` script are skipped. Runs in BOTH jobs that save the napi-libs
       # cache so an entry always carries this output regardless of which job's
       # post-job save wins the per-key race (see the cache step above).
+      #
+      # This `lib/` is ALSO what makes the transparent matrix measure THIS
+      # commit's plugin: `napiNodeAddonPlugin` lives in THIS package, not in the
+      # CLI, and the bundle step below reaches it by resolving
+      # `@gjsify/rolldown-plugin-gjsify` through the workspace symlink. That
+      # survives a cache HIT only because the key above is exact-only with NO
+      # `restore-keys` over `packages/*/*/src/**` — all 43 of this package's
+      # sources are `.ts`, so the hash covers the whole input. Adding a
+      # `restore-keys:` fallback there would let this matrix report PASS while
+      # driving a plugin built from a DIFFERENT commit. Do not add one.
       - name: Build @gjsify/rolldown-plugin-gjsify lib (bundler shims)
         if: steps.napi-libs-cache.outputs.cache-hit != 'true'
         run: |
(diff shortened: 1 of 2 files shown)


JumpLink:
**Converting to draft — two attempts, both wrong, and the second failure explains why the premise was wrong.**

### Attempt 1 — `build` + `build:gjs-bundle`
`build` is `gjsify tsc …`. This job installs only the slice the bundler needs, so the typecheck died on TS7016/TS7006 against `@types/node` it never installed. I should have read the script before adding the step.

### Attempt 2 — drive rolldown from the published CLI on this checkout's `src/index.ts`
Locally clean (6,714,918-byte bundle, no `tsc`). In the job:

```
UnresolvedWorkspaceImportError: gjsify build --app gjs: cannot resolve the
workspace package `@gjsify/create-app`
```

Bundling the CLI pulls its whole workspace-dependency closure, and their `lib/` outputs do not exist here — the job builds only the polyfill libs and `@gjsify/rolldown-plugin-gjsify`. My local run passed *because my worktree had a full `gjsify install` + build*, which is exactly the difference that matters.

**That is the point I kept walking past:** the gate's docblock says *"so no in-repo CLI build is needed — which is what let this run inside the existing consumer job instead of needing its own workspace build"*. The cheap in-job build was never available. ADR 0002 did not just remove a file; it removed the reason this gate could live in a warm, narrow job.

### What actually remains

1. **Build the closure here** — `gjsify workspace @gjsify/cli build --with-dependencies` + `build:gjs-bundle`. Correct, and it turns a narrow job into a workspace build. Needs a measurement of what that costs before anyone commits to it.
2. **Produce the bundle in an earlier job of this same workflow and pass it as an artifact.** `napi.yml` already has `Build & gates & conformance`; if it can produce the bundle, `upload-artifact`/`download-artifact` carries it — the shape `main.yml` uses for `bootstrap-bundles-fedora*`. Probably the right answer, but it is a workflow restructure, not a step.
3. **Move the matrix into `main.yml`**, where the bundle already exists as an artifact, and drop it from `napi.yml`.

Not guessing a third time at 07:45 after a night of merges. Branch and both failing runs are kept for whoever picks this up.

**Unrelated red on this PR:** `Build & gates & conformance` failed at `TSFN TEARDOWN GATE: FAIL (1)` — a timing gate (deadline 2000 ms / fast bound 500 ms) in a job my diff does not touch, and green on `main` at `04773c9c`. Looks flaky; worth a re-run rather than a fix.

JumpLink:
**Picked this up. Closing out the three routes from the handoff above.**

Your read of the failure was right, and it was one package, not a class — I enumerated the entry's whole `--app gjs` closure before touching anything, precisely because "fix one, reveal the next" was the risk you flagged. Two independent walks (~330 files, 32 `@gjsify/*` specifiers, 27 packages) agree `@gjsify/create-app` is the *only* hole.

- **Route 1 (build the closure here)** — taken, but far narrower than "turn a narrow job into a workspace build". Not `@gjsify/cli --with-dependencies`; just `@gjsify/create-app`, whose own tsc closure is empty (dep: `yargs`, `types: ["node"]`, no `@gjsify/*` imports). **Measured, as you asked:** 3.5 s from a wiped `lib/`, and `tsc --listFiles` reports 135 files of which the only `packages/` entries are its own four sources.
- **Route 2 (artifact from an earlier job)** — dead, and not for the reason it looks. `download-artifact` is same-run-scoped, *and* `affected.ts`'s IGNORE list carves out `^packages/napi/`, `^packages/node-gi/` and `napi.yml` itself, so main.yml's producer `build` job is **skipped for three of napi.yml's four trigger paths**. Verified on this PR: GJS run 31143104852 has a skipped build job and zero artifacts. Route 3 inherits the same problem.

**One correction to the record.** The `@types/node` diagnosis in `8ef235413` does not survive its own log. Job 92755831272: 63 errors in 26 files, 34 × TS7016 `Could not find a declaration file for module` against `@gjsify/{workspace,npm-registry,tar,semver}` — this job fills the workspace libs *transpile-only*, so there are no `.d.ts`. Zero errors mention `@types/node`, and zero touch a `node:` builtin, which is itself proof it resolved. That same log also contained `TS2307: Cannot find module '@gjsify/create-app'` — the second attempt's failure was already visible in the first attempt's output.

Also reverted a regression from `0a8e5f3d9`: it rewrote the `GJSIFY_CLI_ENTRY` comment to claim the new step builds `packages/infra/cli/lib/`. No step in this job does.

**On `tsfn-teardown`** — your flake call holds, and it has a cause. `tsfn-teardown-case.js` gates readiness on `stats()[0]`, a **global** delivered-callback counter, where the `draining` shape needs per-thread claim *attribution*: N deliveries from fewer than N threads satisfies the predicate, leaving one claim unattributed, which teardown then correctly warns about. Fixture bug, not the runtime — the warning was accurate. Left for its own PR.

Note `napi.yml` carries no required status check, so `--auto` will not wait for it.

PR #1035, opened 2026-08-07
WIP fix(fs): mode, wx and the write cursor — needs redesign

JumpLink:
> **Status: draft, do not merge.** The `@gjsify/cli` half was split out and is ready in #1037. What remains here — the `@gjsify/fs` work — needs a deliberate redesign rather than another patch, and the analysis below is the specification for it.

## What this was trying to fix

Three things `open()`/`mkdir()` promise that `@gjsify/fs` does not, all surfaced by one consumer streaming mail attachments to disk:

1. **`writeSync` does not advance the file position.** `_readSync` tracks `_readPos` for `position: null`; the write side has no mirror, so every `writeSync(fd, chunk)` restarts at offset 0 and a chunked writer keeps only its **last** chunk.
2. **`'wx'` does not fail on an existing file.** fopen(3) has no exclusive mode, so `resolveIOMode` maps `wx`→`w` and IOChannel **truncates** instead of throwing `EEXIST` — defeating the one guarantee the flag exists to give.
3. **`mode` is parsed and dropped.** Neither `GLib.IOChannel.new_file()` nor `Gio.File.make_directory()` takes a mode, so `openSync(p,'w',0o600)` yields a 0644 file and `mkdirSync(p,{mode:0o700})` a 0755 directory. For a caller writing private data that is a disclosure — mail attachments landed world-readable.

All three are real and worth fixing. `mkdirSyncRecursive` even carries a `NOTE ... Tracked for follow-up` conceding (3).

## Why it is not merging as-is

Two max-effort reviews found **30 confirmed defects**, most of them regressions introduced by the attempted fixes. The pattern is the point: this touches **fd-offset semantics** and **permission semantics** in a module everything builds on, and every patch rippled into paths the patch did not consider.

The sharpest examples:

- **`mkdtempSync` passes `mode: 0o777`.** Inert while `mode` was ignored; the moment `mode` works it hands out a **world-readable** temp directory from the one API whose whole purpose is a private scratch space. Fixing (3) makes a dormant wrong constant live.
- **The post-create chmod strips inherited setgid.** The kernel inherits `S_ISGID` from the parent — the standard shared-workspace setup — so a mode computed from a plain `mode` silently breaks group access for everyone else in that tree.
- **`file_test(EXISTS)` follows symlinks.** A dangling symlink reads as "free", so `wx` creates and writes **through** it, letting whoever planted the link choose where the caller's bytes land — the precise guarantee the check was added to restore.
- **A positional write on an append fd now overwrites.** Narrowing the append branch to `!usePos` means `writeSync(fd, buf, 0, n, 0)` on an `'a'` fd destroys the head of the log. gjsify's own `WriteStream` always passes an explicit position starting at 0, so an append stream could only ever clobber. **This is worse than the bug it replaced.**
- **One shared cursor is right, but was only half-wired.** Async `read()` never consults or advances it (so `while ((await fh.read(b,0,4,null)).bytesRead)` never terminates); `writeFile`/`readFile`/`appendFile` drive the IOChannel's own separate offset; and the read-modify-write straddles the `_serialize` boundary, so `Promise.all([fh.write(a), fh.write(b)])` loses one.

## What a real fix needs to decide first

- **What is the fd offset, and who owns it?** One cursor shared by reads and writes is correct, but it has to be the *only* one — `_readSync`, `_writeSync`, async `read`/`write`, `readFile`/`writeFile`/`appendFile` and the streams all have to go through it, and the read-modify-write has to sit inside `_serialize`.
- **When is `mode` applied?** At creation (Node's contract, survives a crash, but the handle's own positional I/O re-opens the path and a 0444 file then fails to write), or at close (works, but is lost if the process dies and can revert an intervening `chmod`). Neither is free; pick one deliberately.
- **Positional I/O should use the fd, not re-open by path.** Most of the mode trouble and the close-time chmod-by-path hazard dissolve if it does.
- **Where does the umask come from?** Deriving it from a probe works, but `GLib.Dir.make_tmp` is `mkdtemp(3)` and forces 0700 regardless — it must be an ordinary `make_directory`. A probe failure must not cache "umask 0" for the process.
- **Special bits.** Preserve inherited setuid/setgid/sticky, *and* honour them when explicitly requested (`mkdirSync(d, {mode: 0o2770})` is the normal way to make a setgid directory).

## Meanwhile

The consumer works today with four local shims, each carrying a `// fixed upstream in gjsify:` note: an explicit write position, an `existsSync` pre-check before `'wx'`, a `chmod` after open/mkdir, and `minify: false`. Nothing is blocked on this PR — #1037 removes the need for the fourth.

Full finding lists for both review rounds are in the branch history (`ad8869bc4` and its parent).


JumpLink:
Two rounds of work against this specification. Nothing pushed here — it lives on a local `fix/fs-semantics-redesign` (17 commits off `c5a8b4dd5`). Recording what it established, including three places where **this PR body is wrong**.

## Corrections to the specification above

- **"Two max-effort reviews found 30 confirmed defects."** Reconstructing the list yields **41**. The round-2 list is *not* in the branch history as the body claims: `039c73dbe` and `296eb53d2` have **empty commit bodies**, and `gh api …/pulls/1035/{comments,reviews}` is empty too. The missing ~14 had to be recovered by reading the code at `ad8869bc4`. Anyone relying on "the branch history" for that list will not find it.

- **The `Special bits` bullet is wrong for `mkdir`.** It says to "honour them when explicitly requested (`mkdirSync(d, {mode: 0o2770})` is the normal way to make a setgid directory)". Measured against Node v24.15.0: `mkdirSync(d,{mode:0o2770})` → **0750**. `vfs_mkdir` masks to `S_IRWXUGO|S_ISVTX`, so mkdir keeps *sticky* and drops setuid/setgid, while `open(2)` keeps them (`openSync(p,'w',0o4755)` → **04755**). The asymmetry is correct kernel behaviour; the code should follow it and this bullet should be corrected rather than implemented.

- **`mkdtempSync`'s 0o777 is not the only dormant-constant of its shape.** `fs.constants` freezes the platform `O_*` table at module-eval time, defeating the reason `openFlagValues()` was made a function (`posix-flags.ts` documents the hazard, `index.ts` then spreads the call into a module-level `export const`). Darwin-only, so inert on Linux and unmeasured — exactly the profile 0o777 had.

## What the two rounds produced

Three designs were built independently against the five decisions above and judged twice. The judges **split**: on correctness, fd-first ranked first (it *deletes* the umask probe, the post-create chmod and the existence check rather than fixing them, taking whole defect classes with them); on blast radius, minimal-surface ranked first, because it ships a seam (`IoCarrier`) making fd-first a later drop-in rather than a second rewrite.

fd-first was implemented. Result:

| | |
|---|---|
| Round 1 | suite **778/778 green** — and three adversarial lenses then demonstrated **22 defects**, each with a running failing sequence |
| Round 2 | 15 unique defects closed (21 of 24 new tests verified failing pre-fix), GJS suite 778 → **841** |
| Round 2 verify | **35 remaining, 12 of them NEW regressions from round 2's own fixes** |

Several of the new regressions are in `classifyOpenFailure()` — the function round 2 rewrote to close defect #15. It now emits `GLib-GIO-CRITICAL` (so **`G_DEBUG=fatal-criticals` aborts**, the very condition the seek fix was made to satisfy), reports a permission denial as `EEXIST`, and returns `ENOENT` for a name that exists. Also new: `normalizeMode(mode, 0o666)` treats an *absent* mode as a request for 0666 on `chmod`; `lchmodSync` now throws, breaking the portable guard; a `WriteStream` on a closed descriptor never emits `'close'` or `'error'`; and `filehandle.createReadStream()` is **7.6× slower**.

**Read that alongside the judges' split.** Three attempts now (the two in this branch's history, plus round 2 here) have each closed defects and opened new ones in the same module. That is the pattern the blast-radius judge predicted for a near-total rewrite of a tier-1 package, and it is the strongest argument in the record for the smaller-surface design plus a seam.

## Worth keeping regardless of what happens to the design

- A genuine **core bug found and fixed in `@gjsify/stream`**: `Writable_._maybeFinish()` emitted `'finish'` then `'close'` without ever destroying, so `_destroy()` — the only route to `WriteStream.close()` — never ran and every `createWriteStream` leaked a descriptor. Node's `autoDestroy` implemented; every other `Writable` subclass's suite re-run.
- **`#41` (the `wx` TOCTOU) is genuinely closable** — real `O_CREAT|O_EXCL` in one syscall, with `existsSync(target) === false` asserted after a refused `wx` on a dangling symlink. Both prior rounds gave up on it.
- Platform facts measured rather than assumed, two of which falsified design claims: `GioUnix.{Input,Output}Stream` is **not** `GSeekable`; `IOChannel.unix_new(fd)` **does** drive the same description's offset; `seek_position()` returns `G_IO_STATUS_NORMAL`, not a position.

## Remaining work, split by kind

The 35 are not one pile:

1. **12 new regressions** from round 2 — must go, whatever the design decision.
2. **~8 incompletely-closed** items from the original list (notably `ftruncate` on a 0444 file, where GLib offers no fd→truncate route at all, and the `ENXIO` third of `#15`).
3. **~15 pre-existing defects** unrelated to this work, surfaced on the way: `filehandle.readableWebStream()` returns a source-less `ReadableStream` that never closes; `fs.mkdtemp` (callback) is not exported at all; the descriptor write path ignores `encoding`; `copyFile` carries setuid/setgid onto the copy; `statSync` fabricates a `Stats` for an unreachable path. **These deserve their own issues, not this PR.**

Continuing on 1 and 2 with an explicit stop criterion (two consecutive rounds with no new regression) rather than as an open-ended sweep.

JumpLink:
## Reconciled against `main` — nothing is left

Merged `main` in and resolved the eight conflicts. The diff is now **+0/−0 over 0 files**: this branch's tree is byte-identical to `main`. That is the check that answers the question, rather than my reading of it.

### Where each half went

| | |
|---|---|
| `@gjsify/cli` NUL escape | merged as **#1037**, unchanged |
| `@gjsify/fs` mode / `wx` / write cursor | reimplemented as the fd-first redesign in **#1039** |
| that ledger made portable | **#1059**, after it reddened the macOS and Windows legs |

### Every assertion this branch made, and where it lives now

The names all changed, so here is the mapping rather than an assurance:

| this branch (`creation-mode.spec.ts`) | `main` |
|---|---|
| creates a file with the requested mode | honours an explicit mode on the created file |
| applies the umask, as open(2) does | masks the default 0o666 with the live umask |
| does NOT change the mode of an existing file | ignores mode when the file already exists |
| honours the mode on an exclusive create | creates a free path and honours the mode |
| creates a directory with the requested mode | honours { mode } |
| applies the mode to EVERY level of a recursive create | applies the mode to EVERY level it creates recursively |
| applies the umask, as mkdir(2) does | defaults to 0o777 masked by the umask |
| does NOT change the mode of a directory that already existed | does not re-apply the mode to directories that already exist |
| parses a STRING mode as octal, as Node does | parses a string mode as OCTAL |
| PRESERVES a setgid bit the kernel inherited | preserves a setgid INHERITED from the parent |
| keeps a restrictive mode usable for the handle that created it | the creating handle can write a mode-0444 file |
| mkdtempSync creates a PRIVATE directory | creates the directory 0700, whatever the umask |
| refuses wx on a DANGLING symlink instead of writing through it | refuses a DANGLING symlink and does not create its target |

| this branch (`fd-ops.spec.ts`) | `main` |
|---|---|
| writeSync with position null ADVANCES the file position | O-1 |
| writeSync with an explicit position leaves the cursor alone | O-2 |
| writeSync in append mode always writes at the end | keeps append semantics for "as" and numeric O_RDWR\|O_APPEND |
| openSync with wx throws EEXIST on an existing file | refuses an existing file WITHOUT truncating it |
| openSync with wx succeeds when the file is absent | applies to wx+, ax and ax+ alike |

The sixteenth, **"does not chmod at all when no mode was passed"**, has no counterpart and needs none: it guarded the post-create-`chmod` design. `main` hands the mode to `GLib.open(path, spec.posix, mode)` — a real `open(2)` — so there is no post-create chmod left to get wrong. The hazard is gone by construction rather than by a test, which is the stronger outcome and the one the PR body asked for ("positional I/O should use the fd, not re-open by path").

### What I removed rather than carried

- `creation-mode.ts` / `.spec.ts` — the superseded design. Keeping the spec was not even viable: `main`'s ledger has a rule that `test.mts` imports every spec that exports one, and `test.mts` here comes from `main`.
- `file-handle-open.ts` — replaced by `fd-io.ts` + `posix-flags.ts`.
- the five extra `fd-ops.spec.ts` tests — duplicates of ledger rules, and a second copy is a second truth that drifts.

### Also worth knowing

The three consumer shims the PR body mentions (explicit write position, `existsSync` pre-check before `'wx'`, `chmod` after open/mkdir) are all obsolete on `main` and can be removed from Postbote; the fourth (`minify: false`) went with #1037.

**Recommend closing** — superseded, with nothing outstanding. Not closing it myself since it is not my call.

JumpLink:
Superseded. The scope named here — the write cursor, `mode` and `wx` — landed in `6321ad22d7` ("fix(fs): fd-first redesign — cursor, mode and wx", #1039), which also pinned ~50 observable rules of `open(2)`/`mkdir(2)`/`mkdtemp(3)` that this branch had no coverage for.

What is left of the fs backlog is #1046 — four divergences from Node measured on both legs (setuid bits surviving `copyFileSync`/`cp`, no callback `mkdtemp`, a sourceless `readableWebStream`, and nine fd/dir callbacks that accept a missing callback). That is where the remaining work belongs; nothing in this branch is needed for it.

PR #1038, opened 2026-08-07
fix(cli): make gjsify tsc fail when it fails

JumpLink:
Found while diagnosing why the **v0.31.0 release-cut dry run failed**.

## What breaks

`gjsify tsc` ends all six of its exit paths with a bare `process.exit()`. Under GJS that is **idle-scheduled** — `exitProcess` queues `quitMainLoop()` + `imports.system.exit()` — so from a `child.on('close'|'error')` handler it can lose the race with the module's own shutdown and the process reports **0** whatever was asked for.

## Why the release-cut died two steps past the fault

`@gjsify/create-app`'s build is an `&&` chain:

```
node scripts/process-template.mjs && gjsify tsc && node scripts/set-bin-mode.mjs
```

In [the failed run](https://github.com/gjsify/gjsify/actions/runs/31177657929) `gjsify tsc` printed `spawn … ENOENT` five times **and `set-bin-mode.mjs` still ran** — which in an `&&` chain is only possible if tsc exited 0. It then died on a `lib/index.js` that tsc had never written, so the error named `set-bin-mode.mjs` while the fault was one link earlier.

This path is reachable in CI by construction: `ghcr.io/gjsify/ci-fedora:44` ships **no node** (verified by running the image), and `tsc.ts`'s own comment calls the Node fallback "the NORMAL path" on a cold tree, since `@gjsify/tsc`'s bundle is a build output that `build:infra` produces only later.

Related, and why it went unnoticed until now: the cold-tree bootstrap step in `release-cut.yml` arrived with #1019 (ADR 0002) on 2026-08-06, while v0.30.0 was cut on 2026-08-05. **No release has ever gone through it.**

## The fix

The repo already had the rule and two copies of it — `utils/spawn.ts` documents the main-loop machinery, `gjsExit()` exists for precisely this ("GJS has no atexit hook and would exit 0 at natural shutdown, so a non-zero `process.exitCode` alone does not stick"), and both `install.ts::forceExit` and `oxc-resolve.ts::setOxcExitCode` implement it locally. `tsc.ts` had none of it.

Hoisted into `utils/force-exit.ts` and used at all six sites, so the next async exit path gets it right by importing rather than by remembering. `install.ts` delegates instead of carrying copy three; `setOxcExitCode` is left alone (different contract — it deliberately does not exit on Node).

## The test gap this closes

`tests/e2e/tsc-node-fallback` already asserted a non-zero exit for a type error — and passes. It only ever exercised `child.on('close')`: the child RAN, so the async spawn had armed the main loop and the deferred exit was delivered. **Nothing covered a spawn that FAILS**, where there is no child and no armed loop. Added that case, with a PATH that resolves `gjs` but not `node`.

## What is proven, and what is not

**Proven:**
- tsc exited 0 in CI while reporting failure — the `&&` chain reaching `set-bin-mode.mjs` cannot happen otherwise.
- `ci-fedora:44` has no node (`command -v node` → empty in the image).
- `isGjs()` is NOT the culprit: measured inside that image, `imports.gi` is present and `isGjs()` returns `true` (gjs 1.88.1, same as the host).
- The existing e2e's coverage boundary, read from its own assertions.

**Not proven:** I could not reproduce the silent exit locally. Runs against the unmodified v0.30.0 bundle with `node` off PATH and no `@gjsify/tsc` bundle present still compiled and exited 0 *successfully*, so some path I have not yet identified served the compile. The fix rests on the documented GJS exit semantics and the CI evidence, not on a local before/after.

**The verification that settles it** is a re-run of the release-cut dry run once this lands: either it gets past `@gjsify/create-app`, or it now fails AT tsc with a diagnosis instead of at `chmod`. Both outcomes are progress; the second is the one this PR promises.

Not fixed here, and worth its own change: `@gjsify/child_process`'s spawn error does not carry `err.code`, so `tsc.ts`'s dedicated ENOENT branch ("`node` is not on PATH") never fires and the vaguer `else` message is printed instead. That is why the CI log named the bootstrap bundle rather than the missing interpreter.

### DIFF
--- a/packages/infra/cli/src/commands/install.ts
+++ b/packages/infra/cli/src/commands/install.ts
@@ -32,7 +32,7 @@ import { chmodSync, existsSync, lstatSync, mkdirSync, readFileSync, rmSync, syml
 import { dirname, join, relative, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { spawnSync } from 'node:child_process';
-import { gjsExit } from '@gjsify/rolldown-plugin-gjsify/runtime';
+import { forceExit } from '../utils/force-exit.js';
 import { discoverWorkspaces } from '@gjsify/workspace';
 import type { Command } from '../types/index.js';
 import { buildInstallCommand, detectPackageManager, runMinimalChecks } from '../utils/check-system-deps.js';
@@ -406,21 +406,6 @@ function isAbortedFromOverallTimeout(err: unknown): boolean {
     // the overall budget but typically the symptom the overall timer reports.
     if (name === 'RegistryTimeoutError') return true;
     return false;
-}
-
-/**
- * Force a non-zero process exit that is honored on BOTH runtimes.
- *
- * Node exits at natural shutdown once `process.exitCode` is set, but GJS has
- * no atexit hook, so `imports.system.exit` is called directly there —
- * SpiderMonkey raises its uncatchable exit exception from any context,
- * including the `setTimeout` continuation the backstop timer runs in. Mirrors
- * the entry wrapper (`index.ts`) and `setOxcExitCode` (`oxc-resolve.ts`).
- */
-function forceExit(code: number): void {
-    process.exitCode = code;
-    if (gjsExit(code)) return;
-    process.exit(code);
 }
 
 const CLI_PACKAGE_NAME = '@gjsify/cli';
--- a/packages/infra/cli/src/commands/tsc.ts
+++ b/packages/infra/cli/src/commands/tsc.ts
@@ -24,6 +24,7 @@ import { dirname, join } from 'node:path';
 import { pathToFileURL } from 'node:url';
 import type { Command } from '../types/index.js';
 import { detectNativePackages, buildNativeEnv } from '../utils/detect-native-packages.js';
+import { forceExit } from '../utils/force-exit.js';
 import { nodeBinary } from '../utils/run-node.js';
 import { findWorkspaceRoot } from '../utils/workspace-root.js';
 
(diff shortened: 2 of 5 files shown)


JumpLink:
**The re-run settled it, and it overturned the diagnosis rather than confirming it.**

The first commit's own new e2e case went RED on this PR — and the assertion printed an **empty** captured output. No stdout, no stderr, exit 0.

That rules out the theory the first commit was built on. If the spawn's `'error'` handler had run, it would have printed its diagnosis; it printed nothing. So under the GJS bundle the async `'error'` event is **not delivered** when the program is missing: the handler never runs, `runNodeTsc()` returns normally, and `gjsify tsc` exits 0 having compiled nothing. `forceExit` in that handler was fixing a path that is never taken.

`db4fe081d` moves the decision **before** the spawn — resolve `nodeBinary()` against PATH (`existsSync` for a qualified path) and fail there, on the caller's own stack, where the exit is reliable. `isOnPath` already existed in `check-system-deps.ts` and is exported rather than re-implemented; its comment explains why it walks PATH instead of shelling out to `which`(1) — the Fedora containers CI runs in do not ship it.

`forceExit` stays. The `'close'` path still needs it to propagate a compiler's non-zero exit code, and on Node the `'error'` handler is correct — the event IS delivered there.

Worth stating plainly: **the test was more valuable than the fix it shipped with.** It reproduced in CI what I could not reproduce locally, and it falsified my explanation instead of rubber-stamping it. The `errno`-fidelity gap noted at the bottom of the description is now also better understood — `err.code` is not merely lost, the error object may not arrive at all.

PR #1040, opened 2026-08-07
ci(release): bootstrap a cold tree under Node

JumpLink:
**v0.31.0 cannot be cut without this.** Two dry runs have died in the same step, and #1038 — a real CLI fix, merged — provably could not help.

## The trap

```yaml
gjs -m "$GJSIFY_BOOTSTRAP" run build:infra
```

`$GJSIFY_BOOTSTRAP` is downloaded from `releases/latest/download/cli.gjs.mjs`. It is **always the previous release's CLI**. So this step cannot be fixed from the CLI source: a bug that reaches here is only repairable by a release, and this step is what a release runs.

That is not theory. #1038 fixed the exact defect this step trips over, landed on `main`, and the very next dry run ([31188521136](https://github.com/gjsify/gjsify/actions/runs/31188521136)) failed **byte-identically** to the one before it — because the bootstrap is still 0.30.0 and will be until a release ships, which is the thing being blocked.

## What actually breaks

`build:infra` reaches `@gjsify/create-app`, whose build is an `&&` chain:

```
node scripts/process-template.mjs && gjsify tsc && node scripts/set-bin-mode.mjs
```

`gjsify tsc` prints `spawn <bootstrap-bundle> ENOENT` five times — it resolved the compiler to the bootstrap bundle rather than `node` — writes no `lib/index.js`, **and exits 0**. `set-bin-mode.mjs` therefore runs and dies on the missing file, so the error names a step past the fault. (The exit-0 half is what #1038 fixes; it will matter for the release *after* this one, when the bootstrap becomes 0.31.0.)

Worth stating: this step arrived with #1019 (ADR 0002) on 2026-08-06 and v0.30.0 was cut on 2026-08-05, so **no release has ever gone through it**. Its first real run was its first failure.

## The change

Drive the same build from a published `@gjsify/cli` **run under Node**. None of the failing surface exists there: `isNode()` is genuinely true so `nodeBinary()` resolves the real interpreter, and `process.exit()` is immediate rather than idle-scheduled through a GLib main loop.

This is not a new mechanism — `napi.yml` already does exactly this, for exactly this reason, and this reuses its `install-published-cli.sh` rather than re-implementing the retry window it documents (a release bumps every `package.json` before npm serves that version).

`Setup Node.js` (26.x) already runs two steps earlier, so nothing new is asked of the image — which ships no `node` of its own, and that is precisely why the GJS-hosted fallback had nothing to spawn.

The warm-tree branch is untouched: a tree that already has `packages/infra/cli/lib/index.js` still skips this entirely.

## Verification

A dry run on this branch is the test, and it is a real one — the two prior dry runs both failed here, so a pass is not a vacuous green. I will report the result on this PR before merging.

### DIFF
--- a/.github/workflows/release-cut.yml
+++ b/.github/workflows/release-cut.yml
@@ -139,6 +139,21 @@ jobs:
       # mechanism napi.yml already uses for the same reason, down to the shared
       # `install-published-cli.sh` (which retries through the window where a release
       # has bumped every package.json but npm does not serve the version yet).
+      #
+      # TEMPORARY, AND MEANT TO BE REVERTED. The project's direction is to run as
+      # much as possible ON gjsify rather than on Node — CI included — because
+      # running the toolchain on itself is what makes its defects observable at
+      # all. This one step is the exception only because it cannot repair itself.
+      # Revert it to `gjs -m "$GJSIFY_BOOTSTRAP" run build:infra` once BOTH hold:
+      #   1. a RELEASED @gjsify/cli carries the fixes this step trips over
+      #      (gjsify#1038: `gjsify tsc` exiting 0 when it can spawn no compiler),
+      #      so `releases/latest` is no longer the broken one; and
+      #   2. a NON-BLOCKING job exercises the GJS-hosted bootstrap, so the next
+      #      regression there is a signal rather than a deadlocked release.
+      # Without (2) a revert only re-arms the trap for the next CLI bug.
+      # Note: every step AFTER this one already runs the repo's own CLI
+      # (`PATH="$PWD/node_modules/.bin:$PATH" gjsify …`), so what is on Node here
+      # is the cold start, not the release.
       - name: Bootstrap the build toolchain (cold trees only)
         run: |
           if [ -f packages/infra/cli/lib/index.js ]; then
--- a/.github/workflows/release-cut.yml
+++ b/.github/workflows/release-cut.yml
@@ -149,6 +149,13 @@ jobs:
             cd /tmp/gjsify-node-cli && npm init -y >/dev/null 2>&1
             "${GITHUB_WORKSPACE}/.github/scripts/install-published-cli.sh" "${CLI_VERSION}"
             cd - >/dev/null
+            # Point the workspace shim at the published CLI too. `build:infra`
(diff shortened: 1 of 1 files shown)


JumpLink:
**Dry run on this branch: [green](https://github.com/gjsify/gjsify/actions/runs/31191764111).** Not a vacuous one — four consecutive dry runs died before it, each on a different real cause.

```
success   6. Bootstrap the build toolchain (cold trees only)
success   7. Build packages
success   8. Verify the committed bundles reproduce
success  10. release-it
skipped  11-14.   (exactly what dry_run suppresses)
```

## The four blockers, in the order they surfaced

| # | Cause | Whose |
|---|---|---|
| 1 | The GJS-hosted CLI cannot build itself on a cold tree | pre-existing — this step arrived with #1019 the day *after* v0.30.0 was cut, so no release had ever run it |
| 2 | Fixing the CLI (#1038) changed nothing: `$GJSIFY_BOOTSTRAP` is **always the previous release's** CLI | structural |
| 3 | `node_modules/.bin/gjsify` still pointed at the unbuilt local CLI, so `build:infra`'s child scripts died on MODULE_NOT_FOUND | mine — I took napi.yml's recipe without its `ln -sf` |
| 4 | …and then left the shim pointed at the *published* CLI, so `verify-committed-bundles` rebuilt `affected.gjs.mjs` 157 bytes short and called it STALE | mine |

Blocker 4 is the one worth dwelling on. Left in, the release would have **built, verified and cut with 0.30.0** rather than with the CLI the commit ships — silently, and in a repo whose whole point is running on its own toolchain. `Verify the committed bundles reproduce` caught it, before a bump, a tag or a release record existed, exactly where its comment says it sits. The check was right the whole time; it was being handed the wrong compiler.

The same rebuild driven by the freshly built CLI reproduces byte-identically — measured locally: `All committed bundles reproduce byte-identically from the source at HEAD`, with `git status` clean after a full three-step rebuild.

## What this is not

A retreat from GJS. The Node CLI's reach ends the moment `packages/infra/cli/lib/index.js` exists — every later step (`Build packages`, the bundle verification, `release-it`) runs the repo's own CLI, as before. The header now carries the two conditions for reverting the cold start to GJS as well, so the exception does not quietly become the design.

Issue #1052, opened 2026-08-07
"the three targets screenshot-compare 1:1" is documented but not implemented


JumpLink:
`@gjsify/adwaita-storybook` and `@gjsify/storybook-nativescript` both describe their purpose as rendering the same stories "so the three targets screenshot-compare 1:1", and `packages/nativescript-bridge/AGENTS.md` repeats it.

No such harness exists. `tests/browser/specs/` holds a single `unit.spec.ts`, and a search across `tests/`, `scripts/` and `.github/workflows/` finds nothing that compares renderer output. The claim describes an intention.

That mattered concretely: the avatar divergence (single-word names rendering `"AD"` on web and `"A"` on NativeScript, and every accented name getting a different colour on the two) was exactly what a parity harness is supposed to catch, and it survived for the life of both ports.

The interim answer that DID catch it is cheaper and stronger for behaviour: `@gjsify/adwaita-core/conformance` ships libadwaita-derived vector tables that both renderer suites drive their real widgets with, so a divergence fails a unit test naming the input. That covers derivations and state machines; it does not cover layout, spacing or theming, which is what screenshots are actually good at.

So this issue is a fork:

- **(a)** build the harness — GTK via `@gjsify/devtools`' `Screenshot`, browser via Playwright, NativeScript via the devtools agent's rasteriser (android tested, iOS untested) — and wire a per-story comparison; or
- **(b)** decide it is not worth the maintenance and delete the claim from the two `status/status.json` notes and from `packages/nativescript-bridge/AGENTS.md`.

Either is fine. What is not fine is the current state, where the documentation asserts a guarantee nothing provides.

Issue #1051, opened 2026-08-07
iOS symbolic-icon backend needs on-device verification


JumpLink:
`packages/nativescript-bridge/adwaita/src/widgets/icons.ios.ts` landed to close the gap where `nativescriptPlatforms: ['android','ios']` was declared with no iOS variant, so every icon-bearing widget silently rendered nothing on iOS.

**What is covered.** The geometry half is pure and runs in CI on GJS and Node: `widgets/svg-path.ts` reduces the full SVG `d` grammar to absolute M/L/C/Z (H/V to lines, Q/T to cubics exactly, S/T reflections, elliptical arcs via SVG §F.6.5 with the §F.6.6.2 radius correction), and `src/svg-path.spec.ts` sweeps all 182 shipped symbolic icons — 6083 commands, zero non-finite coordinates — plus the hex-colour parser UIKit needs.

**What is not.** The ObjC call sequence has never run on hardware. It mirrors `icons.android.ts` step for step, but these specifically are unverified:

- `UIBezierPath.bezierPath()` and the `addCurveToPointControlPoint1ControlPoint2(end, cp1, cp2)` argument ORDER (endpoint first is the ObjC signature, but worth confirming against the bridge).
- `UIGraphicsBeginImageContextWithOptions(CGSizeMake(size, size), false, scale)` sizing in POINTS with the scale passed separately — Android sizes its bitmap in pixels instead.
- `new ImageSource(uiImage)` accepting a `UIImage` (the NS `ImageSource` wraps the native image type per platform).
- `path.applyTransform(CGAffineTransformMakeScale(s, s))` scaling from the 16×16 authoring grid.
- Whether `fill()` respects per-path `fill-opacity` as passed through `UIColor.colorWithRedGreenBlueAlpha(..., alpha * opacity)`.

The `nativescript-platforms` conformance rule now fails the build if a declared platform loses its variant again, so the *absent* case cannot come back silently. This issue is about the *wrong* case.

To verify: run `showcases/dom/adwaita-widgets-nativescript` on an iOS device or simulator and check that rows, header bars and buttons show their icons.

JumpLink:
**Update — the ObjC surface is now verified; only the runtime is not.**

Two questions were open here: can the macOS test VM stand in for a device, and are the hand-written ambient declarations right?

[...]

PR #1058, opened 2026-08-08
feat(adwaita): hold both renderers to the C source, and make the checks able to see it

JumpLink:
Both Adwaita renderers had drifted from libadwaita, in opposite directions, and nothing in the build could see it. This makes the C source the specification for the split views, the spinner, the view switchers and the split button — and, more importantly, moves each rule to somewhere a test can reach it.

Every claim below was read from `refs/libadwaita/`, not recalled.

## The divergences

| # | What was wrong | Ground truth |
|---|---|---|
| **O1** | Collapsing an overlay split view did **not** hide the sidebar — on a phone the overlay and its scrim simply stayed over the content | `adw-overlay-split-view.c:1457-1458` |
| **O3** | A collapsed overlay ignores the fraction and clamps the view width. GTK draws **280** DIP on a 360 DIP phone; we drew 180 | `:462-463` |
| **O5** | The off-screen pane stayed keyboard- and screen-reader-reachable | `:330-331`, `:1460-1461` |
| **O7** | Once collapsed, the sidebar kept `adw-overlay-active` and its alignment forever | `:718-740` |
| **O12** | *(web)* `pin-sidebar` was missing from `observedAttributes`, so its read branch was dead after construction | `:990-992` |
| **N1** | A collapsed navigation split view with a lone child rendered **blank** | `adw-navigation-split-view.c:389-401` |
| **N2** | With `sidebar-position: end` the navigation split view was fully **inverted** | `:306-316` |
| **N3** | …and its master/detail animation ran backwards | `:1414-1428` |
| **N7** | A collapsed pane lost its full width one frame later — **a regression from this branch's own width work** | `:538` |
| **S1** | The NativeScript spinner was accent-blue. `Adw.Spinner` is neutral — the paintable strokes with the widget's **text** colour | `adw-spinner-paintable.c:432` |
| **V1** | The inline view switcher dropped toggles libadwaita draws: `populate_group` gates on `visible` **alone** | `adw-inline-view-switcher.c:370-378` |
| **A1** | `splitButtonArrowIcon('none')` returned the hamburger. Inside `splitbutton { … }` the stylesheet overrides it to `pan-down-symbolic` | `_buttons.scss:621-623` vs `:454-456` |

Two of these are worth calling out because of *how* they were hiding.

**A1 was shipped, and a conformance vector was pinning it.** The vector cited `_buttons.scss:451-453` — the plain `menubutton` block, which does not even contain an icon declaration — while the winning rule sits 167 lines further down with four element selectors against two. A vector derived from the same reading of the source as the implementation cannot catch a *misreading* of that source. It only pins the misreading down and ships it under the word "conformance".

**S1 and V1 were contradictions the repo had already written down.** `adwaita-web`'s SCSS carries the comment *"Adw.Spinner is a neutral indicator, NOT accent-coloured"* while the NativeScript theme set the accent blue. `@gjsify/adwaita-core` documents the inline switcher's looser gate in `buildInlineToggles` and ships `INLINE_TOGGLE_VECTORS` asserting it — the widget never asked. In both cases the disagreement was in the tree, in prose, and no check could reach it.

## The mechanisms

The fixes matter less than these. Each one makes a *class* of defect visible rather than one instance.

- **Rules moved out of unimportable classes.** A NativeScript widget that `extends GridLayout` evaluates the bare `@nativescript/core` specifier at module-eval, so no spec can import it. That is why N7, S1 and V1 all survived. The split-view state now lives in a new pure `split-view-state.ts` driving the real adapter (the `MockSplitView` it replaces is the anti-pattern `navigation-stack.ts` already records an incident for); `appliesDerivedSidebarWidth` and `switcherButtonVisible` moved to their pure siblings.
- **Cite the selector that wins the cascade.** Recorded in the conformance header next to the avatar incident, with the instruction that actually prevents it: grep the whole stylesheet for the node you are citing, not the block you found it in first.
- **`AdwArrowIcon` makes both renderer tables total.** A sixth upstream glyph is now a compile error in both, not a blank arrow.
- **The openh264 guard is universal.** `gdk-pixbuf2 → libheif → libopenh264.so.8` is a **hard** `Requires`, so `install_weak_deps=False` does nothing (measured in podman on the exact `fedora:43` image: 565 → 385 packages, `openh264` survives). The repo already had the real fix in `release.yml` and `node-gi.yml` — and not in `prebuilds.yml` or `emulated-build.sh`. Two of four sites carrying a guard *is* the failure, so `check-ci-image-packages.mjs` now requires the flag on **every** `dnf install` under `.github/`, with no allowlist and no known-pullers table: both are hand-maintained, and this bug is a hand-maintained guard going stale.
- **The coverage matrix stopped scoring by filename.** It marked a widget core-backed only via a `<widget>.ts` naming convention, which is not how the NS side is built — shared pure siblings (`chrome.ts`, `icon-path.ts`, `split-view-base.ts`) serve several widgets each, so seven widgets read as carrying two copies while already sharing one. Following the import graph one hop subsumes the old lookup entirely. **22 → 26 of 38.**
- **Two tracked build artifacts untracked.** The storybook theme copy had drifted **six weeks** — the device rendered against a theme missing 228 lines of later work, and nothing anywhere compared the two. Playwright's HTML report was committed by accident in #292 and has re-dirtied the tree on every run since.

## Honesty about what this does not do

- **The screenshot claim is corrected, not implemented.** Nine places promised the renderers "screenshot-compare 1:1". No such harness exists — and the avatar divergence is precisely what one is supposed to catch. #1054 fixed the three sites #1052 names; these are the other nine. Behaviour parity is held by the conformance vectors; the pixel half remains unmeasured, and now says so.
- **`build-scss.mjs` is portable, not Node-free.** dart-sass gates its four file-path APIs behind the `node` export condition `--app gjs` omits, so bundling it would have built green and thrown at runtime. `compileString` fixes that (all four outputs byte-identical, source map included), but the real blocker is one level down: `sass.node.js` populates the global `fs` that every read goes through and `sass.default.js` does not. `status/open-todos.md` now counts five calls and records why this one is not equivalent to the other four.
- **`.docker/ci-fedora.Dockerfile` has the same openh264 exposure at four more lines and is deliberately not enforced.** It states that it installs exactly what `main.yml` pulled, and parity with the known-good environment is a contract to change on purpose. The check prints it as a derived line every run, so it is visible rather than silent.
- **Text direction is pinned to `'ltr'`** and the RTL vector rows are skipped with that reason. NativeScript surfaces no direction to read, and guessing one would be a second invented rule.
- **iOS remains unverified** (#1051). No device.

## Counter-tests

Every guard was shown to have teeth, not just to pass:

- reverting the `observedAttributes` line fails 2 web tests;
- reintroducing the two NativeScript split-view regressions fails 10;
- forcing the inline switcher back to the classic gate fails exactly the one test that says an inline toggle survives;
- the corrected arrow vector fails against the old implementation (`open-menu-symbolic` where `pan-down-symbolic` is expected);
- removing either openh264 flag exits 1 naming the file and line — proven for both the YAML and the shell-script corpus.

`adwaita-core` **3401**, `adwaita-nativescript` **2629** (from 2524), web suite green, typecheck and lint clean.


JumpLink:
**Verified on Android.** The responsive sidebar width was measured on a running device, not only unit-tested — thanks to @JumpLink for pointing out that Android can check this, since the width logic lives in the shared `split-view-base.ts` and is therefore the same code iOS runs.

Emulator `Medium_Phone_API_36`, 1080×2400 at density 420 (411 DIP). Widths read from `uiautomator dump`, not estimated from a screenshot:

| Widget | Container | Formula | Expected | Measured |
|---|---|---|---|---|
| the story's `AdwOverlaySplitView`, portrait, inside its card | 349 DIP | `trunc(349 × 0.25) = 87` → clamp lifts to the minimum | 180 DIP | **473 px = 180.2 DIP** |
| the storybook's own `AdwNavigationSplitView`, landscape, full width | 914 DIP | `trunc(914 × 0.25) = 228`, inside [180, 280] | 228 DIP | **599 px = 228.2 DIP** |

One pixel of rounding on each. In both cases the content pane starts exactly where the sidebar ends and the two tile the container.

Two things this shows that the unit tests cannot:

1. **Both branches of the clamp are exercised on a real device** — the minimum lifting a too-small fraction, and the fraction applying once the container is wide enough — and both widget rules are exercised, the overlay one and the navigation one.
2. **The width is re-measured, not computed once.** The storybook shell was constructed in PORTRAIT. Had its width been derived at construction time it would read 180; it reads 228, a value only obtainable from the post-rotation 914 DIP. So `layoutChanged` fires and the container is measured again.

For contrast, the fixed 240 DIP this replaces would have been 630 px in **both** orientations — 58% of the portrait screen, squeezing the story's content pane to about 286 px.

What this does NOT cover: the iOS symbolic-icon rasteriser (#1051). Its geometry half is shared and CI-tested, but the UIKit call sequence is a different code path from Android's and still needs hardware.

JumpLink:
### Device verification: what was and was not done

**Done — the NativeScript app builds and packages with all of this.** `nativescript build android` against freshly built `@gjsify/adwaita-nativescript` / `@gjsify/adwaita-core` succeeds (19.4 s, 105 MB APK) with the new `split-view-state.ts`, the rewritten `AdwSplitViewBase`, the refreshed theme and the corrected arrow table. That is a check the unit suites do not perform — they never see the NS build graph.

**Not done — nothing was run on a device this round.** The emulator would not start here: four attempts, each dying within about two minutes, in and out of the sandbox, with no error in its own log. I also briefly mis-read its liveness — my `pgrep` pattern was matching my own shell command lines rather than a QEMU process, so an early "it's alive" reading of mine was wrong. Nothing about the code changed as a result; it only cost time.

**What stands in its place:**

- The earlier device measurement on this branch is unaffected and still recorded above: portrait overlay split view **473 px = 180.2 DIP** on a 349 DIP card, landscape shell **599 px = 228.2 DIP** on 914 DIP — the latter only obtainable from a post-rotation re-measure, which is what made the `layoutChanged` binding provable rather than plausible.
- Every fix in this PR has a counter-test showing it has teeth: reverting the `observedAttributes` line fails 2 web tests; reintroducing the two NativeScript split-view regressions fails 10; forcing the inline switcher back to the classic gate fails exactly the one test that says an inline toggle survives; the corrected arrow vector fails against the old implementation; removing either openh264 flag exits 1 naming the file and line.

The three claims that genuinely want eyes rather than assertions — the spinner reading as neutral, the collapsed overlay measuring 280 DIP rather than 180 on a phone, and the six-week-stale theme finally being current on device — are the ones still outstanding. They are cheap to check by hand with `npm run run:android` in `showcases/dom/adwaita-storybook-nativescript`.

Issue #1069, opened 2026-08-08
showcase downloads the whole tree before the check that rejects the run


JumpLink:
`gjsify showcase <gjs-only> --runtime node` resolves and downloads the full dlx tree — including the pinned native `@gjsify/node-gi` bridge — and only then reads the declaration that rejects the run.

## Reproduction

macOS, `@gjsify/cli@0.31.0`:

```
$ npx @gjsify/cli@latest showcase adwaita-storybook --runtime node
Need to install the following packages:
@gjsify/cli@0.31.0
Ok to proceed? (y)
gjsify install: resolving done
gjsify install: downloading done
"adwaita-storybook" does not support --runtime node.
  Declared runtimes: gjs.
  Run it on one of the declared runtimes, e.g. --runtime gjs.
```

The error is correct and well-worded — that is not the complaint. The two `gjsify install:` lines above it are: a complete resolve + download finished before anything asked whether the run was possible.

## The ordering

`commands/showcase.ts`, `runShowcaseOnRuntime()`:

```ts
// 283-295 — installs first
const nodeGiSpec = cliVersion ? `@gjsify/node-gi@${cliVersion}` : '@gjsify/node-gi';
pkgDir = await resolveShowcaseDir(showcase, cliVersion, [nodeGiSpec]);
...
// 299-312 — asks whether it was allowed, second
const declared = readDeclaredRuntimes(pkg);
const support = checkRuntimeSupported(runtime, declared, showcase.name);
if (!support.ok) { console.error(support.message); return process.exit(1); }
```

`resolveShowcaseDir` → `ensurePkgDir` → the full dlx install path. The sharp edge is `extraSpecs`: `@gjsify/node-gi` is added **because** `--runtime node` was requested, so the CLI downloads the native `gi://` bridge specifically to support a runtime the showcase is about to say it does not support. Measured for this case: the showcase tarball is 310 KB with `"dependencies": {}`, `@gjsify/node-gi@0.31.0` is 3.77 MB — ~4 MB plus a full resolve round-trip, and it scales with showcases that do have runtime deps.

## The declaration is already available without installing anything

The registry's full packument carries the field:

```console
$ curl -s https://registry.npmjs.org/@gjsify/example-gtk-adwaita-storybook \
  | jq '.versions["0.31.0"].gjsify.example'
{ "runtimes": [ "gjs" ] }
```

[...]

Issue #1065, opened 2026-08-08
Windows CI never loads the node-gi addon — the gate asserts a file, not a load


JumpLink:
No CI leg on any platform ever `require()`s the `@gjsify/node-gi` addon on Windows. Both Windows assertions we have are satisfiable by a host that cannot load a single `gi://` namespace — which is how #1063 shipped in 0.31.0 and reached a user.

## The two gates, and what each actually proves

`.github/workflows/cli-cross-platform.yml`, Windows leg:

**1. `node-gi: a prebuild exists for this platform` (line 237).** The comment is explicit about its own scope:

> This is the cheapest possible form of that check and needs no GTK and no display: install the package and look for the addon for THIS host triple.

It stats `prebuilds/win32-x64/node_gi.node`. In #1063 that file was present and correct; `LoadLibrary` failed on its *dependency closure*. The gate cannot distinguish those two states, and the failing one is the one users hit.

**2. `Showcase: the full --runtime node journey` (line 334).** Deliberately runs `express-webserver`:

> `express-webserver` is the showcase for this, deliberately not the [GUI ones] ... it proves dlx cache -> install -> node bundle -> real HTTP response with no display and no GTK on the runner. The GUI showcases need [a display]

That choice is right for what it targets — it caught the dlx/junction class and is why #1064's EPERM could be attributed to a stale CLI rather than a live regression. But it means the end-to-end Windows path we prove is the one path that never touches `gi://`.

Net: every Windows showcase a user is likely to *want* (three-geometry-teapot, excalibur-jelly-jumper, any GTK/Adwaita demo) is unproven, and was broken on release.

`prebuilds.yml` builds the win32 prebuild, and `release.yml` runs a per-target load test — but against a runner carrying gvsbuild, so it proves the addon loads *when a GTK is present*, never that an end user has one. #1063 is precisely the gap between those.

## What this is an instance of

The same shape as #996 (*"the bundle gates count files where they must assert a decode"*) and the same lesson as the darwin work: a gate that measures a *host property* instead of the *behaviour under test* is green for reasons unrelated to the thing it is named after. `dir-link.ts:63-74` says it outright about the sibling case:

> nothing in CI runs on Windows, so an injected `true` is the ONLY way the unlink-then-rename ordering above is ever EXECUTED, and an unexecutable Windows branch is exactly how the `'dir'` bug shipped.

Windows CI now exists — but the branch it does not execute moved from `dir-link` to `loadNative`.

## Proposal

[...]

Issue #1064, opened 2026-08-08
self-update reports success while PATH still resolves the old gjsify


JumpLink:
`gjsify self-update` prints "Updated @gjsify/cli to v0.31.0" while the `gjsify` that PATH resolves stays at v0.26.0 — and it already holds both pieces of evidence that this will happen.

## Reproduction

Windows 11 VM, `gjsify` on PATH from an earlier `npm install -g`:

```
> gjsify self-update
Current @gjsify/cli: v0.26.0

Warning: no @gjsify/cli install found under C:\Users\JumpLink\.local\share\gjsify\global.
self-update only manages installs created by install.mjs or `gjsify install -g`.
If you installed via `npm install -g`, remove that and use:
  curl -fsSL https://github.com/gjsify/gjsify/releases/latest/download/install.mjs -o /tmp/g.mjs && gjs -m /tmp/g.mjs && rm /tmp/g.mjs
Fetching dist-tags for @gjsify/cli@latest ...
Latest matching --tag latest: v0.31.0
Installing @gjsify/cli@0.31.0 + runtime dependencies ...
gjsify install: resolving done
gjsify install: downloading done
  • C:\Users\JumpLink\.local\bin\gjsify  →  ...\global\node_modules\@gjsify\cli\dist\cli.gjs.mjs

Updated @gjsify/cli to v0.31.0.
Runtime dependencies were resolved alongside the bundle.
```

Next command in the same shell:

```
> gjsify showcase excalibur-jelly-jumper
Could not resolve showcase "excalibur-jelly-jumper": EPERM: operation not permitted, symlink
  'C:\Users\JumpLink\.cache\gjsify\dlx\<hash>\<prep>' -> 'C:\Users\JumpLink\.cache\gjsify\dlx\<hash>\pkg.tmp-<ts>'
```

That EPERM is the bug `utils/dir-link.ts` was written to kill. Its header quotes the failure nearly verbatim and explains the cause (a directory *symlink* needs elevation or Developer Mode on Windows; a junction does not). `dir-link.ts` is absent from `v0.26.0` and present from `v0.28.0` onward — so the command that ran was **0.26.0**, five releases behind what `self-update` had just reported as installed. `cli-cross-platform.yml`'s Windows journey leg exercises `dlx` on 0.31.0 and is green, which independently rules out a live regression here.

## The defect

[...]

Issue #1063, opened 2026-08-08
node-gi on Windows cannot load: nothing installs @gjsify/gtk-runtime-win32-x64


JumpLink:
Every GUI showcase dies at addon load on a clean Windows host, because nothing ever installs the GTK runtime bundle that `@gjsify/node-gi`'s DLL closure needs.

## Reproduction

Stock Windows 11 VM, Node 24.18.1, no gvsbuild, no system GTK:

```
> npx @gjsify/cli@latest showcase three-geometry-teapot --runtime node
Running showcase: three-geometry-teapot [gjsify 0.31.0] on node (...\@gjsify\example-dom-three-geometry-teapot\dist\gjs.node.mjs)

Error: Das angegebene Modul wurde nicht gefunden.
\\?\C:\Users\JumpLink\.cache\gjsify\dlx\<hash>\<prep>\node_modules\@gjsify\node-gi\prebuilds\win32-x64\node_gi.node
    at loadNative (.../@gjsify/node-gi/index.js:103:66)
  code: 'ERR_DLOPEN_FAILED'
```

Identical failure for `showcase excalibur-jelly-jumper --runtime node`. Both on the freshly published 0.31.0.

## What is actually broken

`ERR_DLOPEN_FAILED`, not `ENOENT` — the prebuild **is** there and `existsSync` passed. What cannot be resolved is its *dependency closure*: `glib-2.0-0.dll`, `gobject-2.0-0.dll`, `girepository-2.0-0.dll`, … The German OS string is `LoadLibrary`'s "The specified module could not be found", which on Windows means a transitive DLL, not the named file.

The machinery to fix that exists and is correct. `index.js:37` calls `maybePrependGtkRuntimeDllPath()` before `loadNative()` at `index.js:117`, exactly as designed, and that function prepends the bundle's `gtk/bin` to `PATH` so the later `LoadLibrary` resolves. It just never does anything here:

```js
// gtk-runtime.js:183-185
export function maybePrependGtkRuntimeDllPath() {
    if (process.platform !== 'win32') return;
    const bundle = resolveGtkRuntimeBundle();
    if (!bundle) return; // strict no-op when no bundle is present
```

`resolveGtkRuntimeBundle()` probes four candidates. On an end user's machine only the fourth can hit — `require.resolve('@gjsify/gtk-runtime-win32-x64')` — and it throws into `// Not installed — fine.` (gtk-runtime.js:50).

It is not installed because **nothing declares it**:

```console
[...]

PR #1070, opened 2026-08-08
fix(cli): check before acting, and report only what happened

JumpLink:
Two commands that reported the wrong thing, both found while testing 0.31.0 on a fresh Windows VM and a Mac. One theme: **do the cheap check before the expensive work, and never report success you did not achieve.**

## `showcase` downloaded the tree before the check that rejects the run (#1069)

```
$ npx @gjsify/cli@latest showcase adwaita-storybook --runtime node
gjsify install: resolving done
gjsify install: downloading done
"adwaita-storybook" does not support --runtime node.
```

The error was already correct — the two lines above it were the bug. `runShowcaseOnRuntime` installed first and read the declaration second, and the sharp edge was `extraSpecs`: `@gjsify/node-gi` is added **because** `--runtime node` was requested, so the CLI fetched the native `gi://` bridge to support a runtime the showcase was about to refuse. Measured: a 310 KB showcase with `"dependencies": {}`, plus 3.77 MB of node-gi.

The full packument carries the declaration verbatim, so one metadata GET answers it:

```console
$ curl -s https://registry.npmjs.org/@gjsify/example-gtk-adwaita-storybook \
  | jq '.versions["0.31.0"].gjsify.example'
{ "runtimes": [ "gjs" ] }
```

Pre-flight runs on the dlx path only — a local showcase resolves with no network, so there is nothing to pre-empt there and the on-disk check is already free.

**Deliberately not fixed by adding `runtimes` to `showcases.json`.** That is a second copy of a declaration the package owns, maintained by hand, and `utils/discover-showcases.ts` already records where that goes: `needsWebgl` was deleted rather than wired up because "a field nobody reads is a field nobody maintains", proven by `excalibur-jelly-jumper` shipping `"needsWebgl": false` while Excalibur 0.32 is WebGL2-only.

**The fail-safe direction is the whole contract**, so it is split into a pure function and pinned by tests: the pre-flight may only ever turn a run that *would* have failed into a faster failure. Every uncertain branch — offline, unpublished pinned version, missing field, malformed document — answers "cannot say" and lets the install proceed to the authoritative on-disk check. In particular a pinned version that is not published does **not** fall back to `latest`, because a different version's declaration is a different answer.

## `self-update` reported success while PATH kept the old binary (#1064)

```
> gjsify self-update
Current @gjsify/cli: v0.26.0
Warning: no @gjsify/cli install found under C:\Users\JumpLink\.local\share\gjsify\global.
...
Updated @gjsify/cli to v0.31.0.

> gjsify showcase excalibur-jelly-jumper
Could not resolve showcase "...": EPERM: operation not permitted, symlink '...'
```

That EPERM is the bug `utils/dir-link.ts` exists to kill — absent from v0.26.0, present from v0.28.0. So the command that ran was v0.26.0, five releases behind what `self-update` had just called installed, and the user re-hit a fixed bug and reported it as new. (I filed and withdrew a dlx issue over it; `cli-cross-platform.yml`'s Windows journey leg exercises dlx on 0.31.0 and is green, which independently rules out a live regression.)

It already held both halves of the mismatch — `readCurrentVersion()` → v0.26.0, and `existsSync` on the target prefix → false, which it even warned about — and still printed unconditional success. Now the post-link check is fatal and names both paths and both versions.

`binDirOnPath` (which `install -g` already uses, and `self-update` did not) answers "is our dir on PATH". Necessary but not sufficient: the dir can be on PATH and still lose to an earlier entry. `resolveBinOnPath` answers the question that matters, honouring `PATHEXT` so the `.cmd` shim `bin-shim.ts` writes is found where an extension-less name is not executable.

### One thing the tests caught that I had wrong

Both new functions take `platform` as an argument — and therefore need the *target* platform's path module, not the host's. `path.resolve('C:\\tools')` on posix returns `<cwd>/C:\tools`, so a function that accepts a `platform` and then calls the host's `path` only pretends to be portable; it silently mangles Windows input everywhere except Windows. My first version did exactly that and the win32 specs failed on it, which is the argument for writing them: nothing in CI resolves a bin on a Windows host, so an injected `'win32'` is the only way that branch is ever *executed* — the same reasoning `dir-link.spec.ts` gives, and an unexecutable Windows branch is how the `'dir'` bug shipped in the first place.

## Mechanism

The CI step `Showcase: --runtime node does not demand gjs` asserted only that the message appears, which the wasteful behaviour satisfied. It now also asserts the **absence** of the `gjsify install:` lines, turning "the error is right" into "the error is right and nothing was downloaded to produce it" — the property that actually regressed.

That needed a second version bound, and the node-gi step below already learned why one bound cannot serve two fixes ("one bound served both and therefore lied about one of them"), so the comparison became a shared shell function with `LOWEST_FIXED` (#889) and `LOWEST_PREFLIGHT_FIXED` (this change) as separate values. Both arm themselves on the next release with no edit here.

## Verification

- `npm run check` (tsc --noEmit) clean
- `npm run test:node` — **1998 pass**, up from 1970: 19 new for the PATH resolver and verdict, 9 for the pre-flight
- `gjsify lint` clean on all touched files
- workflow YAML parses

## Not in this PR

**#1063** (every `gi://` showcase dies on a clean Windows host) is **blocked on a decision, not effort** — and my proposed fix in that issue was wrong. `docs/node-gi-platform-notes.md` states in bold that *"a bundle must NEVER be a dependency of `@gjsify/node-gi`"*, backed by a measurement: #910 did exactly that and produced wrong method entries and a 29-minute timeout, #920 reverted it. I corrected the issue and left the real open question — precedence between a bundle and the GTK the addon was linked against — for an explicit decision rather than a drive-by reversal of an incident-backed rule. #1065 (a Windows CI leg that actually loads the addon) is independent and still worth landing.

### DIFF
--- a/.github/workflows/cli-cross-platform.yml
+++ b/.github/workflows/cli-cross-platform.yml
@@ -173,15 +173,21 @@ jobs:
       # release that fixes it, with no edit here and no hand-maintained
       # "currently skipped" list to drift.
       #
-      # LOWEST_FIXED is a BOUND, not a prediction of the next version number:
-      # the fix merged after 0.25.1, and every later release compares >= 0.25.2
-      # in semver order (0.26.0 included).
+      # TWO bounds, because this step now gates TWO fixes that shipped in
+      # DIFFERENT releases — the node-gi step below learned the same lesson the
+      # hard way ("one bound served both and therefore lied about one of them").
+      # Each is a BOUND, not a prediction of the next version number:
+      #   * LOWEST_FIXED — the runtime gate must not demand gjs for
+      #     `--runtime node`. Merged after 0.25.1, so >= 0.25.2 (#889).
+      #   * LOWEST_PREFLIGHT_FIXED — the declaration check must run BEFORE the
+      #     install, not after it. Merged after 0.31.0, so >= 0.31.1 (#1069).
       - name: 'Showcase: --runtime node does not demand gjs'
         id: showcase
         continue-on-error: true
         working-directory: gjsify-diag
         env:
           LOWEST_FIXED: '0.25.2'
+          LOWEST_PREFLIGHT_FIXED: '0.31.1'
         run: |
           set -eux
           out="$(npx gjsify showcase webrtc-loopback --runtime node 2>&1 || true)"
@@ -190,12 +196,19 @@ jobs:
           # Compared in node (present by definition here); `sort -V` is not
           # portable across git-bash and BSD userland. Values travel through
           # the environment, never through string interpolation.
-          HAS_FIX="$(INSTALLED="$INSTALLED" node -e '
-            const parse = (v) => v.split("-")[0].split(".").map(Number);
-            const [a, b] = [process.env.INSTALLED, process.env.LOWEST_FIXED].map(parse);
-            const cmp = a[0] - b[0] || a[1] - b[1] || a[2] - b[2];
-            process.stdout.write(cmp >= 0 ? "1" : "0");
-          ')"
+          #
+          # One comparison, two call sites — a shell function for the same reason
+          # the node-gi step below uses one: two inline copies serving two
+          # different bounds is how the wrong fix got demanded of a release.
+          has_fix() {
+            INSTALLED="$1" BOUND="$2" node -e '
+              const parse = (v) => v.split("-")[0].split(".").map(Number);
+              const [a, b] = [process.env.INSTALLED, process.env.BOUND].map(parse);
+              const cmp = a[0] - b[0] || a[1] - b[1] || a[2] - b[2];
+              process.stdout.write(cmp >= 0 ? "1" : "0");
+            '
+          }
+          HAS_FIX="$(has_fix "$INSTALLED" "$LOWEST_FIXED")"
           if echo "$out" | grep -q "Missing system dependencies"; then
             if [ "$HAS_FIX" = "1" ]; then
               echo "::error::@gjsify/cli $INSTALLED demanded gjs for --runtime node — the runtime gate ran too early"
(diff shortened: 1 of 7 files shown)


JumpLink:
CI proves both halves of the new gate on the Windows leg, against the PUBLISHED 0.31.0 ([run 31256677472](https://github.com/gjsify/gjsify/actions/runs/31256677472)).

The defect, quantified better than the issue estimated — this is 0.31.0 answering `showcase webrtc-loopback --runtime node`:

```
gjsify install: resolving 2 package(s)
gjsify install: resolving done
gjsify install: downloading 43 package(s)
gjsify install: downloading done
"webrtc-loopback" does not support --runtime node.
  Declared runtimes: gjs.
```

**43 packages downloaded** to produce an error that needed one metadata field.

And the new assertion executed rather than sitting inert, correctly reporting it as a known gap instead of failing a release that predates the fix:

```
::warning::@gjsify/cli 0.31.0 predates the showcase pre-flight (< 0.31.1) — known gap, not failing
```

So the gate arms itself: from the first release carrying the fix, the same install lines fail the job. That is the property #1069 asked for — "the error is right **and** nothing was downloaded to produce it" — and it is now executed on a real Windows host, not merely written.

Local verification: `tests/e2e/showcase-runtime/run.mjs` 10/10 pass. That suite's ordering test is deliberately hermetic ("the showcase dir resolves through the workspace symlink — local-first, no dlx, no network"), which is exactly why the pre-flight is gated on the dlx path: a workspace showcase never reaches it, so the test stays network-free.

JumpLink:
CI found a real regression in the first version of the PATH check, and it is worth recording what it was.

`tests/e2e/global-install-engine` failed on `self-update repairs a missing engine even when the cli version matches`. That test sets `GJSIFY_GLOBAL_PREFIX` and `GJSIFY_GLOBAL_BIN_DIR` — the escape hatches `defaultGlobalLayout` documents for exactly this purpose — and installs into a temp prefix it never puts on PATH. My check dutifully reported that PATH does not point there, which is **true and useless**: the caller chose that location.

So the verdict is now scoped to the default layout. That is not a nudge to make a test pass — "which `gjsify` will PATH resolve next" is a question about the *user's* install, and a caller that redirected placement is managing it themselves. `globalLayoutIsDefault()` is injectable and pinned by four cases.

This is the false positive I flagged in the PR description as accepted risk (npx/dlx and friends). I was wrong to accept it unexamined; the e2e caught the concrete instance.

Worth noting how it nearly hid: re-running the e2e locally after the fix still failed, because the suite launches `packages/infra/cli/lib/index.js` and `npm run check` is `tsc --noEmit`. The gate was in `src/` and not in `lib/`. After `npm run build` it passes — the reminder that the thing under test is never the source you just edited.

Now green locally:
- `global-install-engine` 3/3
- `showcase-runtime` 10/10
- unit **2002 pass**, up from 1998
- `tsc --noEmit` clean

PR #1075, opened 2026-08-08
feat(node-gi): decide which GTK wins, per OS and on purpose

JumpLink:
Settles the precedence question `docs/node-gi-platform-notes.md` carried as open, and with it the Windows breakage from #1063 — on a design Pascal proposed rather than the one I first wrote in that issue, which was the fix #910 already tried and #920 reverted.

## The knot

Two GTKs can be reachable at once: the host's, and a batteries-included bundle. The answer was implicit — *a bundle, if one is present* — and the rule keeping it safe was **"a bundle must NEVER be a dependency of `@gjsify/node-gi`"**, written after #910 made it one and a CI job that had *compiled* the addon against Homebrew GTK re-exec'd onto the bundle's typelibs: wrong method entries, then a 29-minute timeout.

That prohibition fixed the from-source case and created a worse one at the other end. Nothing installs the bundle, Windows ships no system GTK, so on a clean host every `gi://` showcase dies with the addon file present the whole time:

```
Error: Das angegebene Modul wurde nicht gefunden.
\\?\C:\...\prebuilds\win32-x64\node_gi.node   code: ERR_DLOPEN_FAILED
```

## The untangling — [ADR 0023](https://github.com/gjsify/gjsify/blob/feat/gtk-source-policy/docs/adr/0023-gtk-source-precedence.md)

**Who installs a bundle is not node-gi's business.** It still declares none. Whoever *ships an app* declares it; for a showcase that is the CLI, which assembles the dlx tree and already supplies `@gjsify/node-gi` the same way. A from-source build therefore never receives a bundle it did not ask for — #910's accident removed at the source rather than caught downstream.

**Which one wins is a policy**, per-OS and overridable via `GJSIFY_GTK_PREFER`:

| platform | default | why |
|---|---|---|
| `linux` | system first | the package manager resolved GTK *and* its typelibs — the combination the distro tested |
| `win32` | bundle first | there is no system GTK; a host with gvsbuild has it deliberately and stays the fallback |
| `darwin` | bundle first | Homebrew is present yet measurably not findable (gjs's rpath points into glib's keg alone) |

Preference is an **order, not an exclusion**: a missing preferred source falls back rather than failing.

**The defaults are spelled out per OS rather than derived** from "does a bundle ship for this platform". Both give identical answers today. The derived rule is a trap: the day someone builds a Linux bundle (Flatpak, a CI image) it would silently flip Linux from system-first to bundle-first with nobody deciding it.

**One hard rule sits above the policy.** A from-source addon does not get a bundle *by default* — that pairing is an ABI error, not a preference; it is #910 itself. Provenance is answered *before* the addon loads, which it must be: the Windows DLL search path has to be set before `dlopen`. `nativeCandidates()` is pure and ordered, so probing which binary *will* load cannot disagree with which one does. An explicit `GJSIFY_GTK_PREFER=bundle` may lift the veto — building against a bundle on purpose is legitimate, and a variable someone set is consent; what #910 was is an accident nobody chose.

## Diagnosis

`ERR_DLOPEN_FAILED` used to propagate raw. `loadNative()` had a careful message for *addon missing* and none for *addon present, closure missing* — the case users actually hit. Now it names GTK, the package that supplies it, and (win32) the MSVC runtime, and reports which source the policy picked.

That lives in `load-diagnostics.js` rather than `index.js` on purpose: `index.js` loads the addon at evaluation time, so anything there is importable only by a process that already succeeded at the thing being diagnosed. My first version put it in `index.js` and the test could not import it.

## Mechanism (#1065)

The Windows gate asked whether the prebuild *file* exists and said so — "needs no GTK and no display". That is exactly the assertion that stays green while `LoadLibrary` fails. And the full-journey step runs `express-webserver`, deliberately chosen for needing neither display nor GTK, so the one end-to-end Windows path we proved never touched `gi://`.

Two steps close it:

1. **The addon LOADS with no system GTK** — install the published prebuild *plus* the published bundle, resolve `GLib` and *call into it* (an unbacked typelib resolves and then dies in its first call, so presence is not the question). The runner must stay gvsbuild-free; installing GTK there would make the step measure the machine instead of the artifact, which is the failure mode of the gate above it.
2. **A bundle-less load is diagnosed, not just failed** — remove the bundle, assert the message names the package and the MSVC runtime. If it loads anyway, the runner has a GTK we did not expect and the step reports that it *proved nothing* rather than passing.

Both version-bound like their neighbours, so a release predating the fix reports a known gap and the gate arms itself on the next one.

## Verification

Everything decidable off-host is pinned from Linux, across all three platforms — the only way these branches are ever executed, since no job here has printed the Windows or macOS wording:

- `test/gtk-source-policy.test.mjs` — 18 cases: per-OS defaults, fallback ordering, the #910 veto (including "answers none rather than the bundle", because failing loudly beats a 29-minute timeout), override semantics, and two cases asserting today's Linux behaviour is unchanged
- `test/addon-load-failure.test.mjs` — 9 cases: cause preserved verbatim, different advice when a bundle *was* selected and still failed, no MSVC hint off win32, no pointer at a linux bundle that does not exist
- node-gi unit **55 pass** · CLI unit **2002 pass** · `showcase-runtime` e2e **10/10** · `package-outputs` e2e **19/19** (it covers `files:`, which the new module had to be added to) · `tsc --noEmit` clean · lint clean

What cannot be proven off-host — that Windows actually resolves the closure — is what the two CI steps are for.

## Not covered

The Linux row of the policy table is currently unobservable: no Linux bundle ships, so `hasBundle` is always false there. The tests pin the intended behaviour, but only a Linux bundle would exercise it for real.


JumpLink:
CI executed both new steps on windows-latest ([run 31262272250](https://github.com/gjsify/gjsify/actions/runs/31262272250)), and between them they prove the fix and reproduce the bug.

**The mechanism works.** With the published prebuild and the published bundle installed, on a runner carrying no gvsbuild:

```
addon loaded and GLib is callable
the addon loaded against the bundle, with no gvsbuild on this runner
```

So the loader machinery in `gtk-runtime.js` was always correct — nothing was wrong except that nobody installed the bundle. That is the whole of #1063, now measured rather than argued.

**And the bug reproduces verbatim.** Remove the bundle and the same runner gives back the report from Pascal's VM, in English this time:

```
Error: The specified module could not be found.
\\?\D:\a\gjsify\gjsify\gjsify-diag\node_modules\@gjsify\node-gi\prebuilds\win32-x64\node_gi.node
  code: 'ERR_DLOPEN_FAILED'
::warning::@gjsify/node-gi 0.31.0 predates the load diagnosis (< 0.31.1) — known gap, not failing
```

The step correctly identified the failure, found that 0.31.0's message names neither the bundle nor the MSVC runtime, and reported the version-bound gap instead of failing a release that predates the fix. From the next release the same message fails the job.

## One thing I had wrong, found while checking this

Both steps were `continue-on-error: true` like every step in this job — and I had added neither to the summary table nor to the `FAILED=1` gate at the end. They would have reported a regression forever and left the job green: precisely the shape of the existence gate they were written to replace. Fixed in bf4933c's follow-up; win32-only steps are `skipped` on macOS, which is accepted there and only there.

Worth stating plainly because it is the second time in this pass that the *mechanism* needed a mechanism — a gate that does not gate is indistinguishable from a passing test until something actually breaks.

PR #1094, opened 2026-08-09
feat(node-gi): ship GStreamer in the batteries-included runtime bundles

### LINKED ISSUE #1092: gtk-runtime bundles ship no GStreamer, so WebAudio cannot work on the batteries-included runtime
`gjsify showcase excalibur-jelly-jumper --runtime node` on Windows 11 died in `new AudioContext()` with `Failed to require Gst 1.0: Typelib file for namespace 'Gst', version '1.0' not found`.

#1089 made that non-fatal — a context now constructs and the app runs silent. This issue is the other half: making audio actually **work** on the bundles.

## Measured

Every `.typelib` in the shipped `@gjsify/gtk-runtime-win32-x64`, listed on the Windows 11 test VM:

```
Adw-1, cairo-1.0, DBus-1.0, DBusGLib-1.0, fontconfig-2.0, freetype2-2.0, Gdk-4.0,
GdkPixbuf-2.0, GdkPixdata-2.0, GdkWin32-4.0, Gio-2.0, GioWin32-2.0, GIRepository-2.0,
GIRepository-3.0, GjsifyHttpSoupBridge-1.0, GjsifyWebrtc-0.1, GL-1.0, GLib-2.0,
GLibWin32-2.0, GModule-2.0, GObject-2.0, Graphene-1.0, Gsk-4.0, Gtk-4.0, GtkSource-5,
Gwebgl-0.1, HarfBuzz-0.0, libxml2-2.0, Pango-1.0, PangoCairo-1.0, PangoFc-1.0,
PangoFT2-1.0, PangoOT-1.0, Rsvg-2.0, Vulkan-1.0, win32-1.0, xfixes-4.0, xft-2.0,
xlib-2.0, xrandr-1.3
```

**No `Gst`, no `GstApp`.** GStreamer appears in no seed list of either builder, and `REQUIRED_NAMESPACES` / `WINDOWING_REQUIRED_NAMESPACES` do not mention it. The darwin bundles are the same — their `brew install` line has no `gstreamer`.

`@gjsify/webaudio` needs `gi://Gst` **and** `gi://GstApp` (the latter for the appsrc/appsink decode pipeline).

## Where the binaries come from

**win32 — nowhere, today.** CI downloads the prebuilt `GTK4_Gvsbuild_<ver>_x64.zip`, and that release publishes exactly two assets: `GTK3_Gvsbuild_…` and `GTK4_Gvsbuild_…`. Two options:

1. **Build it with gvsbuild.** `gvsbuild/projects/gstreamer.py` defines `gstreamer`, `orc`, `gst-plugins-base/good/bad` at 1.28.5 with `-Dintrospection=enabled`, so typelibs are guaranteed and the ABI matches the rest of the bundle. Cost: the Windows bundle job stops being "download a zip" and becomes a source build.
2. **Official MSVC binaries** from `gstreamer.freedesktop.org/data/pkg/windows/`. 1.28.6 ships only `gstreamer-1.0-msvc-x86_64-1.28.6.exe` for x64 — no separate devel package any more, and whether that installer carries `girepository-1.0/*.typelib` is **unverified**. Extracting an installer in CI is also fragile.

**darwin — straightforward.** Homebrew's `gstreamer` formula bundles base/good/bad/ugly/libav with introspection. One formula added to the (deliberately verbatim, duplicated) `brew install` lines in `node-gi.yml` and `release.yml`.

## What the change needs

- seeds: gstreamer core + plugins-base libraries, in the `--windowing` superset
- `WINDOWING_REQUIRED_NAMESPACES` += `Gst`, `GstApp`, so the typelib planner cannot drop them quietly
- the **plugins** are nested native modules, like the gdk-pixbuf loaders in § 2b — on darwin they need the same `@loader_path` relocation walk
- `gst-plugin-scanner` (libexec) and the registry
- `maybeWireGtkWindowingEnv()` += `GST_PLUGIN_SYSTEM_PATH` / `GST_PLUGIN_SCANNER`, same shape as the existing `GSETTINGS_SCHEMA_DIR` / `GDK_PIXBUF_MODULE_FILE` wiring
- license collection (generic on darwin via the keg walk; win32 needs the same treatment as the other seeds)
- a CI assertion that a real `AudioContext` decodes something on the published bundle — the equivalent of the windowing proof, or the typelibs can ship with nothing behind them

## Payload

Worth stating before it is built: GStreamer plus plugins is not small, and every consumer of "batteries-included GTK" would carry it whether or not the app plays audio. A separate `@gjsify/gst-runtime-<platform>` installed on demand by the launcher (which already resolves `@gjsify/gtk-runtime-*` per platform) would keep the GTK bundle unchanged. Noting the alternative; the decision to put it IN the GTK bundle was taken deliberately.

JumpLink:
Closes #1092. **Draft** — neither platform can be verified from the dev machine (no macOS; no MSVC/Python on the Windows test VM), so CI is the judge here.

## The gap

Measured on the published `@gjsify/gtk-runtime-win32-x64`, listed from the Windows 11 VM: **40 typelibs, not one of them GStreamer's.** So `new AudioContext()` threw `Failed to require Gst 1.0` on a machine whose GTK worked perfectly. #1089 made that non-fatal — this makes audio actually work.

## What ships

- `Gst` + `GstApp` join `WINDOWING_REQUIRED_NAMESPACES`. That is what stops them being a *promise*: the typelib planner drops any typelib it cannot back, so without the libraries the bundle would simply have no Gst — green and silent, exactly the shape the 0.27.1 tarballs had when they carried `Adw-1.typelib` with no libadwaita.
- **Two** library seeds per platform. The closure walk reaches gstbase/audio/pbutils/tag/video from libgstreamer but **not** libgstapp — a gst-plugins-base library only an appsrc/appsink user links, which is precisely what the decode pipeline is.
- The **plugins**, which are the part that does the work, get the treatment the gdk-pixbuf loaders already get: darwin copies → relocates to `@loader_path/..` → re-signs → verifies with everything else; Windows copies and walks their backing DLLs into `bin/`.
- `gst-plugin-scanner`, plus `GST_PLUGIN_SCANNER` / `GST_PLUGIN_SYSTEM_PATH` in node-gi's loader. Without the scanner GStreamer degrades to in-process scanning — which works until the first plugin that crashes on load takes the app down instead of itself. `SYSTEM_PATH` rather than `GST_PLUGIN_PATH` (the user's own additive override) also suppresses the compiled-in default, so a foreign GStreamer on the host cannot mix its plugins with the bundle's.

**Everything the prefix has, not a curated element list.** Guessing which codecs an app will meet fails *silently* when the guess is wrong. The size is printed in the build log instead, so the payload is a number somebody can act on.

## Windows builds it from source

Nothing prebuilt exists to lift. The gvsbuild release publishes exactly `GTK3_Gvsbuild` and `GTK4_Gvsbuild`; gstreamer.freedesktop.org dropped its separate devel package, so whether its installer `.exe` carries typelibs is unverified — and an installer is the wrong thing to extract in CI.

gvsbuild defines `gstreamer` / `gst-plugins-base` / `gst-plugins-good` with `-Dintrospection=enabled`, which guarantees the typelibs and matches the ABI already in the prefix. It **will rebuild dependencies** (the extracted zip carries no build markers), so:

- the step is cached, and the key is scoped to the **one** job that builds GStreamer — sharing it with the two jobs that don't would let them save a Gst-free prefix under it and poison the windowing job;
- `release.yml` gains the prefix cache it never had, or every release would pay the build again.

macOS is one formula: brew's `gstreamer` bundles base/good/bad/ugly/libav with introspection, added to both verbatim formula lists.

## What to watch in CI

1. Does gvsbuild complete inside the timeout, and what does it drag in?
2. `Gst-1.0.typelib` + `GstApp-1.0.typelib` in the prefix — asserted right after the build, so it fails there rather than three jobs later with "no element decodebin".
3. Bundle size on both platforms.
4. The darwin relocation verification over the new plugin images.

Still missing, and I would rather add it once the above is green: a proof that actually **decodes** something on the published bundle — the audio equivalent of the windowing proof. Without it the bundle can ship every plugin and still be wired wrong.

### DIFF
--- a/.github/workflows/node-gi.yml
+++ b/.github/workflows/node-gi.yml
@@ -1868,12 +1868,16 @@ jobs:
           # ("Doesn't match target Visual Studio version of 17") and failed with
           # "Unable to find applicable Visual Studio". Hardcoding 18 would break the
           # same way at the next bump; vswhere already knows the answer.
+          # `--vs-ver` takes gvsbuild's own names (vs2022, vs2026), NOT the major
+          # version — 18 was rejected with "Choose from: vs2013 … vs2026".
+          # `catalog_productLineVersion` is exactly that year, so vswhere answers it
+          # directly and no table has to be kept in sync with the runner image.
           $vswhere = "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe"
-          $vsMajor = (& $vswhere -latest -property installationVersion).Split('.')[0]
-          Write-Host "gvsbuild --vs-ver $vsMajor"
+          $vsVer = "vs$(& $vswhere -latest -property catalog_productLineVersion)"
+          Write-Host "gvsbuild --vs-ver $vsVer""
           # Same prefix the zip was extracted into, so one tree carries GTK + Gst and
           # the bundle builder's single closure walk sees both.
-          gvsbuild build --configuration release --vs-ver $vsMajor --build-dir C:\gtk-build gstreamer gst-plugins-base gst-plugins-good
+          gvsbuild build --configuration release --vs-ver $vsVer --build-dir C:\gtk-build gstreamer gst-plugins-base gst-plugins-good
 
       - name: Assert GStreamer landed in the prefix (typelibs + plugins)
         shell: pwsh
--- a/.github/workflows/node-gi.yml
+++ b/.github/workflows/node-gi.yml
@@ -1831,7 +1831,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: C:\gtk-build\gtk\x64\release
-          key: gvsbuild-gtk4-gst-${{ env.GVSBUILD_VERSION }}-x64-v1
+          key: gvsbuild-gtk4-gst-gi-${{ env.GVSBUILD_VERSION }}-x64-v2
 
       # Build-time closure SOURCE only. Verbatim from the `windows` job.
       - name: Download + extract gvsbuild GTK4 (build-time closure source)
@@ -1915,7 +1915,15 @@ jobs:
           # an absent libvpx simply disables the vpx plugin — which is the state we
           # want anyway. -good is still built, for the parsers and decoders the audio
           # path genuinely needs: wavparse, flac, mpg123, autodetect.
-          gvsbuild build --configuration release --vs-ver $vsVer --skip libvpx --build-dir C:\gtk-build gstreamer gst-plugins-base gst-plugins-good
+          # `--enable-gi` IS THE WHOLE POINT and defaults to FALSE. gvsbuild's
+          # gstreamer projects gate introspection on it — without it they pass
+          # `-Dintrospection=disabled`, so the build succeeds and produces no
+          # .gir/.typelib at all. Measured: a complete 16-minute GStreamer build
+          # whose prefix contained zero Gst typelibs. The assertion below is what
+          # caught it, at the prefix, instead of three jobs later as "no element
+          # decodebin" — typelibs are the only reason we build this from source
+          # rather than lifting a binary.
+          gvsbuild build --configuration release --vs-ver $vsVer --enable-gi --skip libvpx --build-dir C:\gtk-build gstreamer gst-plugins-base gst-plugins-good
 
       - name: Assert GStreamer landed in the prefix (typelibs + plugins)
         shell: pwsh
--- a/.github/workflows/node-gi.yml
+++ b/.github/workflows/node-gi.yml
@@ -1236,7 +1236,7 @@ jobs:
       HOMEBREW_NO_INSTALL_CLEANUP: '1'
     steps:
       - name: Install GTK4 + Adwaita + GtkSource + icons + GObject-Introspection (build-time closure source)
-        run: brew install gtk4 libadwaita gtksourceview5 adwaita-icon-theme gobject-introspection cairo pango gdk-pixbuf graphene pkg-config
+        run: brew install gtk4 libadwaita gtksourceview5 adwaita-icon-theme gobject-introspection cairo pango gdk-pixbuf graphene pkg-config gstreamer
 
       - name: Use Node.js 24
         uses: actions/setup-node@v4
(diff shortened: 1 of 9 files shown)


JumpLink:
## Tested on the win11-gjsify VM — the crash it fixes, measured

Not a review, just a datapoint from a real Windows desktop (QXL, no gvsbuild, no MSVC).

**Before**, on `main`'s `gtk-runtime-win32-x64-windowing`, `excalibur-jelly-jumper` did not start at all:

```
Error: Failed to require Gst 1.0: Typelib file for namespace 'Gst', version '1.0' not found
    at ensureGstInit (...)
    at new AudioContext (...)
    at new Sound (...)
```

Unhandled — the process died before a window existed.

**After**, with this branch's bundle from [run 31319774720](https://github.com/gjsify/gjsify/actions/runs/31319774720) (`gtk-runtime-win32-x64-windowing`, 44 DLLs, `Gst-1.0` + `GstApp-1.0` + `GstBase-1.0` typelibs), pointed at via `GJSIFY_GTK_RUNTIME` with everything else unchanged: **the crash is gone and the app comes up** — window titled "Jelly Jumper — Excalibur.js", header bar with a live audio/mute button, so `AudioContext` constructed against real GStreamer.

Setup, if you want to reproduce: published `@gjsify/example-dom-excalibur-jelly-jumper@0.33.0` `dist/gjs.node.mjs` run from inside `packages/node-gi/node-gi` with `NODE_GI_NATIVE=prebuild` and `main`'s CI addon staged in `prebuilds/win32-x64/`.

### One thing this does NOT unblock

The window's canvas still reads **"No GL implementation is available"** — unrelated to GStreamer, and it hits `three-geometry-teapot` and `three-postprocessing-pixel` identically. Two causes stacked: this VM's QXL adapter registers no OpenGL ICD, *and* the bundle ships `epoxy-0.dll` (GL **dispatch**) with no GL **implementation** behind it. The windowing builder already seeds `libEGL`/`libGLESv2`, but the gvsbuild GTK4 ZIP carries no ANGLE, so those patterns match nothing and the bundle promises "windowing" while resolving no GL.

I left that alone — it is the same "what does the win32 bundle ship" question this PR is in, so it is yours to take or leave rather than mine to patch underneath you. Written up with the measurements in the webgl-on-win32 entry of `status/open-todos.md` via #1096, which also corrects two docs that listed ANGLE as already shipped.

Happy to re-run anything here on request.

PR #1096, opened 2026-08-09
fix(webgl): link the redistributable CRT on win32

JumpLink:
## What

The win32 `@gjsify/webgl` exploratory leg built `gwebgl.dll` against the **debug CRT**, producing an artifact that cannot load on any Windows machine without Visual Studio installed. Fixed with `--buildtype=release`, plus an assertion that keeps it fixed.

Also corrects two docs that listed ANGLE among the win32 GTK bundle's payload.

## How it was found

Measured on the `win11-gjsify` VM — a real Windows desktop, no MSVC, no Python — against the *published artifact* of the green dispatch (run 30808747504):

```
gwebgl.dll imports:
  KERNEL32.dll          in-system
  VCRUNTIME140D.dll     MISSING   <-- debug CRT
  epoxy-0.dll           in-bundle
  gdk_pixbuf-2.0-0.dll  in-bundle
  glib-2.0-0.dll        in-bundle
  gobject-2.0-0.dll     in-bundle
  ucrtbased.dll         MISSING   <-- debug CRT
```

Every GTK-side dependency resolved from the batteries-included bundle. The library was **two files short of working**, and said so as `Failed to load shared library 'gwebgl.dll'` — naming the dependent, never the missing dependency.

`meson setup` was called without `--buildtype`. meson defaults to `debug`, and MSVC's reading of `debug` is not "keep the symbols" but `b_vscrt=mdd`. `VCRUNTIME140D` / `ucrtbased` ship only with Visual Studio, and Microsoft's terms forbid redistributing them — so this is a licence gate as much as a portability one.

## Why CI was green

`windows-latest` **has** Visual Studio, so the debug CRT resolves there and the load test passes on the one artifact no user can load. That is not a gap in the probe — the runner is unrepresentative in exactly the dimension a load test cannot see.

So the new check **reads the import table** rather than loading anything: redistribution is a static property of the file, identical on every host. It runs *before* the load test, matches the debug CRT by name (`vcruntime\d+d|ucrtbased|msvcp\d+d|concrt\d+d`) rather than a loose `*d.dll` that `epoxy-0.dll` would trip, and locates `dumpbin` via `vswhere` the way `node-gi.yml` already does.

The generalisable lesson, recorded in `status/open-todos.md`: a CI host provisioned as a *developer* machine silently satisfies developer-only dependencies, so any artifact leaving it needs at least one check that **inspects** rather than **executes**.

## The ANGLE correction

While measuring the GL showcases on the same VM: the `--windowing` bundle's 41 DLLs include `epoxy-0.dll` — GL *dispatch* — and no GL *implementation*. The builder does seed `libEGL`/`libGLESv2`, but those patterns match nothing, because the gvsbuild GTK4 release ZIP carries no ANGLE. A seed that silently matches nothing is how the bundle came to promise "windowing" while shipping no GL.

Two places described that seed list as shipped payload; both are corrected. **No bundle-contents change here** — that decision belongs with the other "what does the win32 bundle ship" work.

## Verification

- `--buildtype=release` verified by dispatching `prebuilds.yml` on this branch and re-reading the resulting artifact's import table on the VM (result in a comment below).
- The three required checks are unaffected: this touches a `workflow_dispatch`-only leg plus docs.

## Not fixed here

`three-geometry-teapot` still cannot render on a GPU-less Windows host, for a reason one layer below webgl: `Gdk.Display.create_gl_context()` fails with `No GL implementation is available`. Two independent causes — the VM's QXL adapter registers no OpenGL ICD, and the bundle ships no software GL. Both are recorded in the ledger with the measurement.

### DIFF
--- a/.github/workflows/prebuilds.yml
+++ b/.github/workflows/prebuilds.yml
@@ -2090,7 +2090,26 @@ jobs:
           rem to the vcvars call that establishes the rest of the toolchain env.
           rem What it is NOT: /w, /std:cNN, or any diagnostic suppression.
           set INCLUDE=%GTK_PREFIX%\include;%INCLUDE%
-          meson setup build . -Dprebuilt_vala_c=vala-c-export || exit /b 1
+          rem --buildtype=release is LOAD-BEARING ON WINDOWS, and on no other
+          rem platform. meson's default buildtype is `debug`, whose MSVC meaning
+          rem is not "keep the symbols" but `b_vscrt=mdd` — link the DEBUG CRT.
+          rem MEASURED on the win11-gjsify VM against the artifact of the green
+          rem dispatch (run 30808747504): that gwebgl.dll imports
+          rem VCRUNTIME140D.dll and ucrtbased.dll, which ship ONLY with Visual
+          rem Studio and which Microsoft's redistribution terms forbid shipping.
+          rem Every other import (epoxy-0, gdk_pixbuf-2.0-0, glib-2.0-0,
+          rem gobject-2.0-0) resolved from the batteries-included GTK bundle, so
+          rem the DLL was two files away from working and failed with a bare
+          rem "Das angegebene Modul wurde nicht gefunden" naming gwebgl.dll —
+          rem the dependency, never the dependent.
+          rem
+          rem WHY THE LOAD TEST BELOW CANNOT CATCH IT: windows-latest HAS Visual
+          rem Studio, so the debug CRT resolves there and the probe goes green on
+          rem the one artifact no user can load. That is not a gap in the probe —
+          rem it is the runner being unrepresentative in exactly the dimension the
+          rem probe cannot see, which is why the assertion after it reads the
+          rem import table instead of loading anything.
+          meson setup build . -Dprebuilt_vala_c=vala-c-export --buildtype=release || exit /b 1
           meson compile -C build || exit /b 1
 
       - name: Collect @gjsify/webgl prebuilds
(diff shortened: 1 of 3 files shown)


JumpLink:
## Verified on the win11-gjsify VM

Dispatched `prebuilds.yml` on this branch ([run 31320051929](https://github.com/gjsify/gjsify/actions/runs/31320051929)) — the win32 pair is green, and the new step ran:

```
13  success  Verify the typelib records a .dll leaf
14  success  Assert the prebuild links the REDISTRIBUTABLE CRT
17  success  Load-test the win32-x64 prebuild through node-gi
```

Then downloaded **both** artifacts and compared them on the VM (no MSVC, no Python — a plain Windows desktop).

**Import table:**

| before (`buildtype` unset → `debug`) | after (`--buildtype=release`) |
|---|---|
| `VCRUNTIME140D.dll` ← VS-only | `VCRUNTIME140.dll` |
| `ucrtbased.dll` ← VS-only | `api-ms-win-crt-runtime-l1-1-0.dll` |
| | `api-ms-win-crt-string-l1-1-0.dll` |
| `KERNEL32`, `epoxy-0`, `gdk_pixbuf-2.0-0`, `glib-2.0-0`, `gobject-2.0-0` | unchanged |

**Actually loading it** — `requireNamespace` + a non-null `getGType`, the same assertion the CI load test makes:

```
--- OLD-debug ---
GLib-GIRepository-WARNING: Failed to load shared library 'gwebgl.dll'
  referenced by the typelib: 'gwebgl.dll': Das angegebene Modul wurde nicht gefunden.
FAIL: typelib found but gwebgl.dll never loaded

--- NEW-release ---
PASS: Gwebgl.WebGLRenderingContext GType = [External: 281d57123c0]
```

Same host, same GTK bundle, same addon, same probe — the only variable is the CRT. That is the whole defect and the whole fix.

Worth noting what the error message did *not* say: it named `gwebgl.dll`, the dependent, and never `ucrtbased.dll`, the thing actually missing. That is why this survived a green load test plus a written-up dispatch — the message points at the file you already know is there.

PR #1115, opened 2026-08-11
feat(showcase): make the LDraw viewer render

JumpLink:
The `three-loader-ldraw` example opened its window and drew nothing but its clear
colour. Three independent core defects were behind that, each of them silent —
no exception, no GL error, no log line. All three are fixed at the core, with
tests that assert the class rather than the one symptom, and the example is
promoted to a showcase.

## The three defects

**`@gjsify/fetch` — `Response.body` was not identity-stable.** Every read built a
fresh `ReadableStream` over the same already-flowing `Readable`, so
`res.body === res.body` was false: the first reader drained all 249 772 bytes and
every later one got an immediate `done`. three.js' `FileLoader` reads `.body`
three times in its feature guard before taking a reader, so it took the third,
empty one. The loader then parsed an empty document into a group with zero
meshes and the camera collapsed onto the origin.

**`@gjsify/webgl` — seven sized internal formats were never registered**
(`R16F`, `RG16F`, `RGB10_A2UI`, the four `_SNORM`). Consumers read these off the
context, so a missing name is not an absent feature: three passed `_gl.RG16F`,
i.e. `undefined`, into `texStorage2D` → `GL_INVALID_ENUM`, a texture without
storage, and `GL_INVALID_OPERATION` on everything after it.

**`@gjsify/webgl` — `vertexAttribDivisor` did not record the divisor.** The
driver drew instanced; our own `_checkVertexAttribState` still believed the
attribute was per-vertex, demanded 1488 bytes from a 384-byte instance buffer,
raised `INVALID_OPERATION` and returned *without* calling the native draw. The
draw was dropped by us, over a capable driver. `getVertexAttrib` with
`VERTEX_ATTRIB_ARRAY_DIVISOR` reads the same field and only had the WebGL1 ANGLE
spelling wired, so on WebGL2 it raised `INVALID_ENUM`.

## Tests

`@gjsify/fetch` 71 (Node) + 116 (GJS) green, `@gjsify/webgl` 940 green. The new
specs assert the whole spec table of sized internal formats, the divisor
round-trip, and an end-to-end instanced draw with an instance-sized mat4
attribute buffer — not just the two names that happened to have a loud consumer.

## Showcase

Moved to `showcases/dom/three-loader-ldraw`, published, and registered in all
seven places a showcase needs: the CLI's `showcases.json`, a `@gjsify/website`
dependency, the catalog row, its own docs page, `ShowcaseEmbed`'s union + switch
arm + `assetBase`, the website asset copy list, and a `ShowcaseSlideshow` slide —
the slideshow is what puts it on the landing page.

`docs/examples-showcases.md` still said to add a showcase as a CLI dependency and
called discovery a dynamic scan of the CLI's deps; that stopped being true when
discovery moved to the static manifest. Corrected, with the website touch points
spelled out.

## Two things reviewers should know

**The linux-x64 prebuild is committed here.** The constants live in the Vala
source and `commit-prebuilds` is main-only, so without it the two format specs
would be red on this PR. The bot refreshes all sixty targets after the merge.

**`gjsify onboard` must run for `@gjsify/example-dom-three-loader-ldraw`** before
the release that ships it — dropping `"private"` publishes a new npm name, and it
sorts ahead of `example-dom-three-postprocessing-pixel`. Ledgered in
`status/open-todos.md`.

## Still open

`PMREMGenerator.fromScene()` produces an all-zero environment map, so the model
renders unlit. The entire PMREM pass now completes without a single GL error,
which makes it a different class of defect from the two fixed above and an
open-ended one. Everything else is verified correct: with lights added
temporarily the scene matches the upstream reference exactly — red body, black
roof, transparent windows, blue seat, yellow headlights. Those lights are *not*
in this PR; papering over a core bug inside an example is what this repo forbids.


JumpLink:
Rebased onto `main` (now at v0.35.0). One conflict, in the showcase's
`package.json`: the release bumped its `version` while this branch rewrote the
same tail. Resolved by keeping the showcase shape and taking 0.35.0.

Two things the rebase surfaced that are worth calling out, because both were
silently wrong in the branch as first pushed:

- **`ce4eb6aa1 fix(example-dom-three-postprocessing-pixel): build a valid URI`
  applies verbatim here.** This showcase built its `assetBase` the same way —
  `file://${GLib.filename_from_uri(...)}` — which makes `C:` the URI hostname on
  win32 and refuses every model fetch. Now `new URL('./', import.meta.url).href`,
  the same spelling as the pixel showcase, and `GLib` is no longer imported.
- **`c702876ba docs(showcases): align every showcase README (#1113)`** landed a
  common README structure after this branch was written. The new README followed
  the old shape, which would have re-introduced exactly the drift that PR removed;
  it now carries the same Targets / Prerequisites / Run / Layout / Related
  sections as its siblings.

Re-verified after the rebase: install, build (gjs + node + browser), type check,
format and lint all clean, and the GJS window still loads and renders the model.
`main` did not touch the webgl prebuilds, so the linux-x64 binary in this branch
is still the newest.

PR #1139, opened 2026-08-13
fix(cli): read the CLI's own version from one resolver

JumpLink:
## What broke

`gjsify showcase adwaita-storybook` — the first tab of the project's own home
page — fails against **published 0.38.0** on a machine that has run gjsify before:

```
$ gjsify --version
0.38.0
$ gjsify showcase adwaita-storybook
Gjs-WARNING: JS ERROR: ImportError: Unsupported URI scheme for importing: node
gjs exited with code 1
```

It ran a cached **0.37.0** bundle, whose `dist/gjs.js` was byte-identical to its
`dist/gjs.node.mjs` (`md5 492063d6…` for both) and therefore carried
`node:module`. 0.38.0 had already fixed that — its two bundles differ and the gjs
one has no `node:` specifier.

## Why it ran the wrong version

`showcase` PINS the showcase package to the CLI's own version, and read that
version with a fixed-depth manifest read:

```ts
const pkgUrl = new URL('../../package.json', import.meta.url);
```

`@gjsify/cli` is dual-entry. From `lib/commands/` that depth is right; from
`dist/cli.gjs.mjs` — the entry a **globally installed `gjsify` runs** — it lands
one directory above the package, where no manifest exists. The read is inside a
`try`, so it does not throw. It answers "no version", `dlxSpec` degrades to the
bare package name, and `dlx` serves whatever it resolved for that name once.

The banner said so the whole time, on the same commit and the same machine:

| tab | banner |
|---|---|
| `npx @gjsify/cli@latest` | `Running showcase: … [gjsify 0.38.0] on node` |
| `bunx @gjsify/cli@latest` | `Running showcase: … [gjsify 0.38.0] on bun` |
| **`gjsify`** | `Running showcase: … (via gjsify dlx @gjsify/example-…)` — **no version** |

Clearing the dlx cache makes 0.38.0 resolve and the showcase render, which is
what confirms it is resolution and not a bad release.

## Why no CI leg caught it

Every runtime the test matrix exercises invokes the `lib/` entry, where the fixed
depth happens to be correct. The one entry that is broken is the one a user
installs.

## The fix

`cliVersion()` in `utils/publish-headers.ts` is already the depth-independent
resolver — a compile-time define for the bundle, an upward manifest walk for
`lib/` — and its own doc comment already said *"deliberately NOT a fixed
`../../package.json` read"*. A fourth copy grew next to it anyway. It is now
exported and used.

## Mechanism, because prose did not hold the line

- **`scripts/check-cli-own-version-read.mjs`** fails any first-party CLI source
  that re-introduces the fixed-depth read, wired into `audit-runtimes.yml`
  beside the other one-rule-one-spelling guards. Verified both ways: green on
  this branch, red on the parent commit, pointing at `showcase.ts:25`.
- **`publish-headers.spec`** asserts a real version — this spec runs as a bundle
  in `dist/`, the exact layout where the old spelling silently answered nothing.
- **`docs/code-anti-patterns.md`** carries the incident, so the rule keeps its
  reason.

## Verification

Measured against published packages on Linux x64, all 6 homepage showcases ×
their declared runtimes plus `create my-app` on all four tabs: green, including
`adwaita-storybook` on gjs once the stale cache is gone (window verified via
`GJSIFY_DEVTOOLS` screenshot, not just exit code).

Not a defect, recorded so it is not re-diagnosed: `deno run -A
npm:@gjsify/cli@latest` resolves the previous release for the first 24 h after a
publish — deno's `--min-dep-age` default, visible as `[gjsify 0.37.0]` on the
deno tab. `--reload --min-dep-age 0` gets 0.38.0, which is what
`showcase`'s own pinned-invocation hint already tells users.

### DIFF
--- a/.github/workflows/audit-runtimes.yml
+++ b/.github/workflows/audit-runtimes.yml
@@ -121,6 +121,17 @@ jobs:
       - name: Check the PR-title type list still matches commitlint
         run: node scripts/check-pr-title-types.mjs
 
+      # Same shape as the two above: a rule that only prose enforced, broken by a
+      # copy that grew next to the resolver whose comment forbids it. `@gjsify/cli`
+      # is dual-entry, so a fixed-depth read of its own manifest is right from
+      # `lib/` and wrong from `dist/cli.gjs.mjs` — the entry a globally installed
+      # `gjsify` runs. It fails silently (the read is inside a `try`), and the
+      # version it fails to produce is what `showcase` pins its dlx spec to, so
+      # 0.38.0 served a cached 0.37.0 showcase that crashed on a bug 0.38.0 had
+      # fixed. Invisible to every CI leg, all of which use the `lib/` entry.
+      - name: Check the CLI reads its own version through one resolver
+        run: node scripts/check-cli-own-version-read.mjs
+
       # Comment volume has no natural limit, so nothing failed while `cli-fail.ts` grew
       # a 47-line yargs research diary above a 3-line function (ratio 9.3), while
       # `rolldown-plugin-gjsify` reached one comment line per code line, or while
--- a/docs/code-anti-patterns.md
+++ b/docs/code-anti-patterns.md
@@ -123,3 +123,39 @@ for the raw call. A guard watching another mechanism is the smell the governance
 rules name, and the cost of this one would be a grep over the tree that cannot
 tell a display string (where the host separator is arguably right) from a value
 about to be split.
+
+## A dual-entry package reading its own manifest at a fixed depth
+
+**Rule: `@gjsify/cli` learns its own version from `cliVersion()`
+(`utils/publish-headers.ts`) and nowhere else. A
+`new URL('../../package.json', import.meta.url)` read is wrong by construction
+in this package — machine-checked by
+[`scripts/check-cli-own-version-read.mjs`](../scripts/check-cli-own-version-read.mjs).**
+
+The CLI ships TWO entries from one package: `npm install` runs the tsc output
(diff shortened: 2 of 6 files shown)


JumpLink:
Reproduced on a **second OS**, which rules out anything host-specific.

macOS 15.7.9, `gjsify` freshly installed today at **0.38.0**:

```
$ gjsify showcase adwaita-storybook
Running showcase: adwaita-storybook (via gjsify dlx @gjsify/example-gtk-adwaita-storybook)
                                     ^ no [gjsify 0.38.0] — the version read failed here too
Gjs-WARNING: JS ERROR: ImportError: Unsupported URI scheme for importing: node
gjs exited with code 1
```

And the cache confirms the mechanism rather than implying it — the only copy on
that machine is the one a previous session left behind:

```
$ for d in ~/.cache/gjsify/dlx/*/*/node_modules/@gjsify/example-gtk-adwaita-storybook; do
    grep -m1 '"version"' "$d/package.json"; done | sort | uniq -c
  2   "version": "0.37.0",
```

So a 0.38.0 CLI ran a 0.37.0 showcase on macOS for the same reason it did on
Linux: the spec was never pinned, so `dlx` reused what it had cached for the
bare name. On both hosts the `gjs` tab printed no version while `npx`/`bunx`
printed `[gjsify 0.38.0]`.

Same sweep also measured, for the record: the four GUI showcases are green on
macOS under `gjs`/`node`/`bun` (window verified by console screenshot), so this
is the only showcase failure there that this PR addresses.

Issue #1061, opened 2026-08-08
@gjsify/cli specs leave the process in a deleted CWD, breaking a later spawn (flaky on darwin-x64)


JumpLink:
## What happened

`macOS suites` run [31246195788](https://github.com/gjsify/gjsify/actions/runs/31246195788), `@gjsify/cli`, darwin-x64:

```
❌ include-args still satisfies the workflow's contains() gates  (604ms)
affected exited 1: ENOENT: process.cwd failed with error no such file or directory,
the current working directory was likely removed without changing the working directory, uv_cwd
    at includeArgsFor (packages/infra/cli/dist/test.node.mjs:915:29)
❌ [Node.js 24.18.0] 2 of 1974 tests failed
```

**darwin-arm64 passed the same commit**, and a re-run of the same job on the same
commit passed too. So it is order- or timing-dependent, not deterministic.

## Why it is not the classifier spec's own fixture

`AFFECTED_ARGV(cwd)` passes the fixture root as an **argument** (`--cwd <root>`),
not as a spawn option — so the child inherits the PARENT's working directory for
`process.cwd()`. `affected-classifier.spec.ts` has exactly one fixture `root` and
one `teardown`, and the teardown runs after the failing test (it is `✔` in the
same log). The directory that vanished is the test process's own CWD.

## Where it can come from

Two specs in the same bundle move the process CWD:

- `packages/infra/cli/src/config.spec.ts:32` — `chdir(dir)` … `finally { chdir(prevCwd); rmSync(dir) }`
- `packages/infra/cli/src/run-node-resolve.spec.ts:92` — `chdir(root)` … `finally { chdir(previous) }` then `rmSync(root)`

Both restore in the right order in isolation. The hazard is that `prevCwd` /
`previous` is whatever the CWD happened to be at entry: if any spec is entered
while the CWD is already a temp fixture, the restore puts it back into a
directory a later `rmSync` removes, and every subsequent spawn inherits a dead
CWD. That is a process-global resource being saved and restored by more than one
spec, which is the whole bug class regardless of which spec loses the race.

## Suggested shape of a fix

Make the CWD not process-global for these tests — pass `cwd` as a spawn option
where a child is involved, and for the two `chdir` sites either assert at entry
that the CWD still exists, or capture the CWD **once** at module load and always
restore to that, rather than to whatever was current.

## Provenance

Found while fixing #1039's OS-axis regressions (PR #1059). Not caused by that
[...]

Issue #1044, opened 2026-08-07
One test body per runtime: shared partials, and the closure rule that must come with them


JumpLink:
Filed for handoff rather than losing it in a session, in the shape of #999.

Goal: a GObject-TypeScript app should run on every runtime and OS the project claims — whether through GJS directly or through `@gjsify/node-gi`. Testing that is one mechanism, not two: **the runtime in which an API is native proves the TEST is correct; every other runtime proves OUR implementation is correct.** `tests/AGENTS.md` rule 3 already states it for the node/gjs pair. A test written for Node, run under GJS, exercises `@gjsify/napi` and the polyfills. A GObject test written for GJS, run under Node, exercises `node-gi`. Same rule, opposite direction — and no third concept is needed, because the build target already decides which side is native.

## The model already exists

`packages/node-gi/example/harness.mjs`, verbatim from its header:

> ONE unchanged `gi://` source builds + runs with byte-identical output on FOUR runtimes […] gjs is the REFERENCE: node/bun/deno output must equal the gjs output exactly.

Scenario table × available runtimes × `buildAppForRuntime`. That is the shape to generalise. The work is lifting it out of one example, not inventing it.

Note the tradeoff before copying it wholesale: comparing runtime output against gjs is only worth it where the expected value is impractical to write down. It has a failure mode fixed expectations do not — **if gjs is wrong, the comparison is green.** Prefer fixed expectations; keep the differential comparison as a supplement.

## Measured first, because the obvious framing is wrong

There is almost no duplicated test *body* to remove. Measured on this tree:

- `tests/e2e/`: **117** suites. An exhaustive all-pairs similarity scan tops out at 55% Jaccard — **no pair is one test written twice.** The runtime-looking pairs are two scenarios each (`cli-only` vs `cli-only-pnp` share 15 lines = 2% of the former; `plugins-by-name-gjs` asserts a GJS-loader-only mechanism — that the plugin was pre-bundled and cached — which the Node suite cannot have).
- `*.node.spec.ts`: **0 files exist anywhere.** All **18** `*.gjs.spec.ts` are solo — no same-base sibling. The suffix is a runtime gate on a GJS-only subject, not a fork with a second copy to fold in.
- `tests/integration/`: 29 of 35 packages already follow the one-source-two-builds pattern. `tests/integration/axios/package.json` is the reference: one `src/*.spec.ts` set, `build:test:node` + `build:test:gjs`, both run.

[...]

Issue #1012, opened 2026-08-05
Nothing stops a new CLI call site from bypassing the spawn teardown contract


JumpLink:
Follow-up to #1010, which was one instance of this class.

## The gap

`packages/infra/cli/src/utils/spawn.ts` owns the GJS teardown contract and states it in a measured table:

| host | caller ends with | path |
|---|---|---|
| Node | anything | async `spawn` |
| GJS | `process.exit(…)` | async `spawn` |
| GJS | a normal `return` | blocking `spawnSync` |

It even makes the declaration **mandatory** — `SpawnToCompletionOptions.completion` is a required field, "on purpose: a new call site must state which side of the table it is on rather than silently inherit a default that may be wrong for it".

That is only binding on code that *uses* the helper. One import bypasses it entirely:

```ts
import { spawn } from 'node:child_process';
```

Which is exactly what `run-lifecycle-script.ts` did, for years, while its own header asserted the opposite of the truth ("no GLib-mainloop intermingling from `ensureMainLoop`"). The cost was `gjsify pack` parking for 5m30s after finishing its work, on any package with a `prepack` — including the production `gjsify run publish:app` chain.

## What is measured, and what is not

Both `spawn()` and `_execImpl()` (behind `exec`/`execFile`) call `ensureMainLoop()` in `@gjsify/child_process`, so on paper both arm the hook.

**Only the async `spawn` path is confirmed to hang.** The `execFile`-based commands were probed and exit cleanly — `gjsify gsettings` compiled a schema and exited in 0.45 s. Whatever releases the loop there is not established; it should be, because a rule written on the wrong axis is worse than none.

## The three sites that still bypass it

| file | spawn at | verdict |
|---|---|---|
| `commands/test.ts` | 243 | **undecided** |
| `commands/storybook.ts` | 245, 268 | **undecided** |
| `commands/flatpak/check.ts` | 145 | **undecided** |

Each resolves a promise on child exit rather than exiting there, and in all three the `process.exit()` calls sit *above* the spawn — so control flow, not line order, decides. Lexical analysis cannot classify them, which is why #1010 did not try. Each needs one run under GJS to land on a side of the table.

(`trust.ts` / `onboard.ts` use `spawnSync` — blocking, unaffected. `foreach.ts` imports the type only.)

## What to build, once those three have verdicts

[...]

PR #1141, opened 2026-08-13
feat: make a green run prove what it actually did

### LINKED ISSUE #1043: @gjsify/unit re-derives the runtime, and gets Bun wrong
Filed for handoff rather than losing it in a session, in the shape of #999.

`@gjsify/unit` carries its own runtime detection, and it is the second — worse — copy of a question `packages/infra/rolldown-plugin-gjsify/src/utils/runtime.ts` already answers correctly. The consequence is not cosmetic: **a Bun run currently reports itself as Node.js and silently takes every Node branch.**

## What is wrong today

`packages/gjs/unit/src/index.ts`:

- `export type Runtime = 'Gjs' | 'Deno' | 'Node.js' | 'Unknown' | 'Browser' | 'Display'` — there is no `'Bun'`.
- `getRuntime()` probes `globalThis.Deno?.version?.deno`, then `process.versions.gjs`, then `process.versions.node`. Bun sets `process.versions.node` for npm compatibility, so it falls into the Node arm and returns `Node.js <version>`.
- `runtimeMatch()` matches by substring: `onRuntime.find((r) => currRuntime.includes(r))`. So under Bun every `on(['Node.js'], …)` block runs, and no `on(['Bun'], …)` block is even expressible.

## What already exists and is right

`packages/infra/rolldown-plugin-gjsify/src/utils/runtime.ts`:

```ts
export function hostRuntime(): 'gjs' | 'node' | 'bun' | 'deno' {
    if (isBun()) return 'bun';
    if (isDeno()) return 'deno';
    if (isGjs()) return 'gjs';
    return 'node';
}
```

`isBun()` tests `globalThis.Bun`, and its docblock states the reason the order matters: *"FIRST in `hostRuntime` because Bun also fakes `process.versions.node` for npm compatibility — a bare `process.versions.node` read is a false Node positive there."* `isDeno()` carries the same note. The four predicates are documented as mutually exclusive.

This is the single source of truth the test framework should be reading.

## The task

Make `@gjsify/unit` derive its runtime from `hostRuntime()` instead of re-deriving it, and widen `Runtime` accordingly. Two things to decide while doing it, neither obvious:

1. **Dependency direction.** `@gjsify/unit` is `packages/gjs/unit`, `hostRuntime()` lives in `packages/infra/rolldown-plugin-gjsify`. A test framework depending on a bundler plugin is the wrong shape; the predicates may need to move to a smaller shared home first. Both packages are already in `GLOBAL_TRIGGERS` (`affected.ts`), so CI selectivity is not a constraint either way.
2. **`on()` matching stays substring-based.** With `'Bun'` added, probe/name ordering becomes semantic rather than cosmetic. Either keep substring matching and pin the order deliberately, or switch to exact matching on the normalised `hostRuntime()` value and keep the version string separate.

## Why this is the first step

It blocks #1044 (shared test partials): an axis that cannot be *named* cannot record that it ran. 

## Not in scope

Adding a bun/deno leg to any package's `test` scripts. This issue is detection only.

JumpLink:
Three mechanisms that stop a green run from claiming more than it did, plus the defects each one surfaced. All three sit in the same seam: the runner and the CLI both had ways to finish successfully while having silently skipped, parked, or misattributed work.

## 1. A run is held to the axes it declared (#1044)

`on()` returns silently when the host is not on the named axis — correct, and the reason `on('Deno', …)` costs nothing under Node. The gap was that a gate which *should* have fired and did not looked identical: a miss did `++countTestsIgnored`, the count was printed, and the exit code read `countTestsFailed` alone. An axis whose tests had stopped being registered left nothing behind and never once stopped a merge — strictly worse than a deleted spec file, which at least shows up as gone.

`RunOptions.requireAxes` turns the record into a verdict. The declaration is **host-conditional**: `test.mts` is built once and runs on every leg, so an axis this host does not match claims nothing, and one entry can name every axis it serves. `packages/gjs/unit` declares all four.

Two details the ledger turns on, both of which I got wrong in the first draft:

- **`tests` is the delta the matched blocks EXECUTED**, not the `it()` calls they contain. A block that matched and then registered nothing scores zero — that *is* the silent shape. Counting gates would have made it look like coverage.
- **Only the axis that MATCHED is credited.** `on(['Node.js', 'Gjs'], …)` under Node exercised Node and nothing else; booking those tests against `'Gjs'` too would let the Node leg satisfy a Gjs declaration.

Enforcement is one line in `run()`, and a declaration whose enforcement has been removed is invisible from inside a run — the shape of the two inline release gates that shipped dead and cost all three v0.28.0 publish legs. So `tests/e2e/unit-axis-declaration` asserts it from **outside**, on exit codes, in both directions and on both the node and gjs legs.

## 2. The flatpak spawns go through the teardown contract (#1012)

All four `gjsify flatpak` sub-commands async-spawned through `node:child_process` directly, bypassing `utils/spawn.ts`. Each handler ends with a `console.log` and a normal return, so all four sit on the `'return'` row — under GJS an async spawn leaves `ensureMainLoop`'s loop armed and parks the process at 0 % CPU after the child has exited and been reaped. Same class as the `gjsify pack` 5m30s hang.

`flatpak check` is the sharpest: its failure path ends `return process.exit(1)`, so **only the success path could park**. The hang would have appeared on a clean run and nowhere else.

`check` also needs the child's output suppressed unless a check fails, and `stdio: 'pipe'` could not serve it — piping hands back a live `ChildProcess` the blocking path cannot produce, which is why it is `'exit'`-only. So `spawnToCompletion` gains **`stdio: 'capture'`**: collects both streams into `SpawnCompletion`, emits neither, works on both paths, and therefore composes with `completion: 'return'`.

**The guard itself is deliberately not here.** Re-measuring found eight further bypass sites, not the three the issue lists, and the check it asks for needs ledger entries that *state each teardown path*. In every one of the eight the `process.exit()` calls sit above the spawn, so control flow decides and lexical analysis cannot. Writing eight reasons I have not verified would seed the ledger with guesses that read as reviewed. Measured table and the proposed `unreviewedBaseline` shape are in [#1012](https://github.com/gjsify/gjsify/issues/1012#issuecomment-5280957464).

## 3. A destroyed process CWD is charged to the test that destroyed it (#1061)

A spec that `chdir`s into a temp directory and removes it leaves the whole **process** in a deleted CWD, and every later `process.cwd()` — including one inside a spawned child — dies with `ENOENT … uv_cwd`. Specs share one process, so the cost lands on whatever runs next.

The runner now probes as each test ends and reports the **transition**, naming the offender:

```
❌ A chdirs into a temp dir and removes it — this test left the process in a
   deleted working directory — every later `process.cwd()`, including one
   inside a spawned child, now fails with ENOENT.
```

Latched, so it fires once rather than once per remaining test.

**This does not fix the darwin-x64 flake, and the PR does not claim to.** While building it I established that #1061's stated hypothesis is impossible: both specs it names run *after* the failing suite in `test.mts`'s namespace order, `runTests` is strictly sequential, and both restore the CWD before their `rmSync` anyway. No spec ahead of the failing suite calls `chdir` at all. The cause is still unknown — the mechanism makes the next occurrence name its own culprit instead of failing an innocent spec in another suite. Detail in [#1061](https://github.com/gjsify/gjsify/issues/1061#issuecomment-5280950990).

## Two ceilings raised, both deliberately

Flagging these because both gates exist to only ever tighten, so a raise should be argued rather than noticed.

- **`packages/infra/cli` comment ratio, 0.424 → 0.426.** The tree sat *exactly* on its ceiling. Replacing four hand-rolled spawn promises with `spawnToCompletion` removed ~40 lines of boilerplate **code** while adding ~26 lines of rationale, so the ratio rose largely from *deleting code*. Measured: cutting every comment this PR adds to that tree still leaves it 5 lines short of 0.424, so no amount of trimming reaches the old value. I did trim first — the four call sites originally repeated the same rationale four times, which is the duplication the budget is for; one canonical explanation now lives in `utils/spawn.ts` and the call sites point at it.
- **`tests/AGENTS.md`, 7817 → 8067 bytes.** Also exactly on its ceiling. `requireAxes` is a new declaration kind, and rule 6 is the place that enumerates its neighbours (`it.skip` / `it.failing` / a runner warning) precisely so the right one gets picked. A fourth concept absent from that list is one nobody finds.

## Also

- Closes #1043 — it was already done on `main` (`b4eb2302a3`, PR #1133) and the issue had not been closed; same for the harness-duplication half of #1044, where `tests/e2e/mock-registry.mjs` now serves 30 suites and `git grep -l ustar tests/e2e/` returns exactly one file.

## Measured

| | |
|---|---|
| `@gjsify/unit` node leg | 288 pass |
| `@gjsify/unit` gjs leg | 286 pass |
| `@gjsify/cli` node leg | 2055 pass |
| `tests/e2e/unit-axis-declaration` | 5 pass (node + gjs) |
| `tests/e2e/unit-fail-attribution` | 5 pass |
| `check-e2e-suite-coverage` | 118/120 run, 2 ledgered |
| `check-e2e-harness-duplication`, `check-comment-budget` | green |

Refs #1044, #1012, #1061. Closes #1043.


JumpLink:
### The declaration verified on all four runtimes, not just the two in the body

`packages/gjs/unit` is in `CROSS_RUNTIME_PACKAGES`, so its `requireAxes: ['Gjs', 'Node.js', 'Bun', 'Deno']` has to hold on every leg. Measured locally:

| leg | axis ledger | summary |
|---|---|---|
| Node.js 24.15.0 | `Node.js: 2 tests from 3 gates` · `Deno: 0 … 1 stood down` | 288 completed |
| Gjs 1.88.1 | `Gjs: 2 tests from 3 gates` · `Deno: 0 … 1 stood down` | 286 completed |
| Bun 1.3.14 | `Bun: 2 tests from 3 gates` · `Deno: 0 … 1 stood down` | 288 completed |
| Deno 2.9.4 | `Deno: 2 tests from 3 gates` · `Bun: 0 … 1 stood down` | 288 completed |

Each leg credits **its own** axis and stands the foreign one down — which is the host-conditional property the design rests on, and the `3 gates → 2 tests` gap is the deliberate one: `axis-ledger.spec.ts` includes a gate that matches and registers nothing.

**A note on how this was measured, because it nearly produced a false alarm.** The first Bun and Deno runs reported `[Node.js 24.3.0]` and `[Node.js 26.3.0]` — the exact symptom #1043 was closed for. The cause was local, not in the tree: `packages/gjs/runtime/lib/esm/` held a 285-byte `index.js` from 2026-07-30 and **no `detect.js` at all**, because `detect.ts` arrived with #1133 and my working copy had never rebuilt that package. Rebuilding it produced the table above. Source was correct throughout; the artefact was three weeks stale, and it is the second time in this session that a stale `lib/` impersonated a defect (the other cost two phantom `process-stub-banner` failures in `@gjsify/cli`).

Issue #1175, opened 2026-08-14
node-gi: an instance ignores a shadowed prototype method


JumpLink:
Under `@gjsify/node-gi`, an instance does not resolve a GI method through the JS prototype chain. Assigning to the prototype **sticks** — and the instance ignores it.

Measured side by side, same script, GJS 1.88.1 / girepository 2.88.3 / Node 24.15.0:

```js
import Gio from 'gi://Gio?version=2.0';
const proto = Gio.ZlibDecompressor.prototype;
const spy = function () { return 'SPY'; };
proto.convert = spy;
const d = new Gio.ZlibDecompressor({ format: Gio.ZlibCompressorFormat.GZIP });
```

| | gjs | node-gi |
|---|---|---|
| `proto.convert === proto.convert` | true | true |
| after assignment, `proto.convert === spy` | true | true |
| `d` has own `convert` | false | false |
| **`d.convert === spy`** | **true** | **false** |

So the write is accepted and observable on the prototype, the instance has no own property that would shadow it, and yet the instance still reaches the native method. Calling `d.convert(...)` increments a spy **zero** times.

## Why this matters more than monkey-patching

The failure mode is silent in the dangerous direction. A test that installs a spy sees `proto.convert === spy` and concludes it is instrumented; every subsequent call bypasses it and the spy observes nothing. `packages/node/zlib/src/gio-codec.gjs.spec.ts` does exactly this to assert the member scan never hands `convert()` an unbounded slice:

```
✖ never passes an unbounded input slice to convert() — Expected 0 to be greater than 0
```

That test passes on gjs and fails through the bridge — not because the invariant broke, but because the instrument is inert there. A spy-based test that happened to assert `calls === 0`, or that only checked a maximum, would have gone GREEN while measuring nothing.

## Scope

Anything relying on JS-level interception of GI methods: spies, patches for a workaround, decorating a method to add logging, and the `Gio._promisify` shape if it is ever applied to a prototype rather than looked up dynamically. Plain calls are unaffected — this is not a marshalling defect. `convert()` itself marshals correctly through the bridge: `[FINISHED, 23, 3]` for a 23-byte gzip member of `'abc'`, and `GLib.Bytes` honours a `byteOffset`-bearing subarray (both measured while chasing #1173).

## How it was found

[...]

PR #1179, opened 2026-08-14
fix(zlib): decode every gzip member, and let CI see it

### LINKED ISSUE #1173: node-gi harness: main red since 2026-08-13, hidden by the workflow's path filter
`main` has been red on the `node-gi` track since 2026-08-13, and nothing said so — because `node-gi.yml` only runs when its own paths are touched.

## What fails

`node-gi consumer harness (proof set / Node+Bun+Deno)`, step `Run the node-gi consumer harness (proof set, --require-pass)`:

```
=== grouped ===
  PASS                       7  @gjsify/sqlite, @gjsify/http2, @gjsify/tls, …
  other                      1  @gjsify/zlib
  assertion-mismatch         1  @gjsify/node-globals

  @gjsify/zlib — partial
```

`partial` means the unit summary reported `failed > 0`, so the suite ran and tests failed.

## It is not zlib

`@gjsify/zlib`'s own suite is green on both runtimes, measured just now:

```
✔ [Node.js 24.15.0] 53376 completed
✔ [Gjs 1.88.1]      53384 completed
```

So the failure is specific to running it **through the reverse bridge**, which is what the harness does (`--alias node:<name>=@gjsify/<name>`, run on node+bun+deno).

## Why nobody noticed — the same shape as #1149 and #1028

`.github/workflows/node-gi.yml` is path-filtered:

```yaml
paths:
  - 'packages/node-gi/**'
  - 'scripts/node-gi-consumer-harness.mjs'
  - '.github/workflows/node-gi.yml'
```

The harness runs OTHER packages' suites through the bridge, so anything outside those three paths can break it invisibly. Last green run on `main`: `010f11a0f4` (*chore: release v0.38.1*, 2026-08-13 11:37 UTC). It has not run on `main` since, because no merge since then touched a filtered path. A docs-only edit to `packages/node-gi/AGENTS.md` is what triggered it again and surfaced this.

This is the third instance of the class in the tree: #1028 (`runE2E` firing only on `tests/e2e/**` touches), #1149 (the affected-bundle hook's four-path trigger), and now this one. The pattern each time: **a guard whose trigger is narrower than what it guards.**

## Candidates

Exactly two commits after the last green run touched code every suite runs through — both in `packages/gjs/unit/src/index.ts`:

| commit | | |
|---|---|---|
| `3d06560a82` | #1141 *make a green run prove what it actually did* | +205 lines (axis ledger) |
| `ce3ae86274` | #1163 *name the failing test…* | +105 lines (failure recap) |

Two mechanisms I checked and **ruled out**:

- `requireAxes` / `failUnexercisedAxes` (from #1141) — `packages/node/zlib/src/test.mts` does not declare `requireAxes`, so that path cannot fire for it.
- The recap's marker (from #1163) — the harness's failure collector matches `❌ ⏱ ✗ ✘` (`node-gi-consumer-harness.mjs:330`) and the recap emits `✖` (U+2716), a different character. It is invisible to the parser, and the recap only prints when failures already exist.

So the cause is still open. Reproducing needs the node-gi native addon; there is no prebuilt one in a plain checkout, so this wants its own pass rather than a guess.

## Two things worth deciding separately

1. **The break.** Bisect between `010f11a0f4` and `main` with the harness, or run it locally once the addon is built.
2. **The trigger.** As long as `node-gi.yml` only fires on its own paths, the next break from outside is equally invisible. The harness's real input set is "every package in the proof set, plus everything they run through" — closer to the closure reasoning #1149 just applied to the affected bundle than to three globs.

JumpLink:
`gunzip()` cannot read concatenated gzip members on any host that has a native `DecompressionStream`, while both of its siblings can. **This is `main`'s red `node-gi` leg.** Closes #1173.

## What it is

A spec-conforming `DecompressionStream('gzip')` decodes exactly ONE member and errors on whatever follows. Node's `zlib.gunzip()` decodes them all — multi-member gzip is legal and produced by `gzip -c a b`, bgzip and some HTTP servers. `decompress()` preferred the Web API whenever the global existed, which made the async one-shot the only one of three siblings that stopped at the first member:

```
✔ gunzipSync should decompress concatenated gzip members          ← member walk
❌ gunzip async should decompress concatenated gzip members
     TypeError: failed to write whole buffer
       at Object.transform (ext:deno_web/14_compression.js)       ← the platform's stream
✔ createGunzip (stream) should decompress concatenated gzip …     ← member walk
```

That stack is what settled it. Four earlier hypotheses of mine died on measurement first — flake (the re-run reproduced it exactly), `convert()` marshalling (`[FINISHED, 23, 3]`, correct on node *and* deno), `GLib.Bytes` ignoring `byteOffset` (honoured), and a node-gi engine change (`packages/node-gi` was never touched in the window).

The Web path stays the fast path, gzip included — it is the platform's own C zlib, and on a host with no Gio it is the only path there is. The member walk runs only after it has *actually failed*, so single-member gzip pays nothing, and corrupt input still rejects because `gunzipWithGio` re-decodes and surfaces GLib's error.

## Why it shipped: a cell no gate could reach

The Web branch was never executed by any test.

- On **GJS** there is no native `CompressionStream`, so the branch is dead code.
- On **Node** the specs import `node:zlib`, which resolves to *Node's own zlib*, not this file — tests/AGENTS.md rule 3: the Node leg proves the test, the GJS leg proves our implementation.

The first host ever to run it was the node-gi reverse bridge, where the runtime's `DecompressionStream` wins over our polyfill — months after the code landed, and reported as a *bridge* defect rather than ours. The consumer harness earned its keep here.

So the capability check is asked **per call** instead of snapshotted at module load. That is what makes both branches reachable from a spec on any host, and three new tests stub a platform stream that fails the way Deno's does:

- the fallback decodes both members
- it still decodes a *single* member (the fallback is not multi-member-only)
- it still **rejects** corrupt gzip rather than swallowing it

## Measurements

Same tree, only the fix differing:

| | node | bun | deno |
|---|---|---|---|
| before | part 53374/53375 | pass | part 53374/53375 |
| after | pass\* | pass 53376/53376 | **pass 53376/53376** |

\* locally node reports `53384/53385`; the one remaining failure is #1175's `convert()` spy, a separate finding in an `on('Gjs')` spec that CI's node leg does not execute.

A/B on the new tests: without the fix two of three fail with the production error `failed to write whole buffer`; the third passes either way by design, since it guards against a fallback that swallows errors.

zlib's own suites: **Node 53376 ✔**, **Gjs 53387 ✔** (+3).

Closes #1173

---

# The second half of #1173: the trigger that could not see this fix

`node-gi.yml` guards nine other packages' suites through the reverse bridge, and its trigger named three node-gi paths. **This PR touches `packages/node/zlib` only, so under that filter no job in the workflow it repairs could observe the repair.** A gate that cannot see its own fix is not a gate.

#1174 answered #1173 with a nightly and a `workflow_dispatch`, on the reasoning that a PR-side fix was unaffordable: *"this builds an addon and runs three runtimes over nine suites, which no unrelated PR should pay for."* That treats the run as indivisible, and it is not:

| | jobs | cost | can a consumer change break it? |
|---|---|---|---|
| `ubuntu-latest` | 7 | ~21 job-min, ~8 min wall | **yes — this is the consumer guard** |
| macOS ×10 / Windows ×6 / arm64 | 16 | scarce runners | no: dyld + PE relocation, marshalling under another pointer width |

So the trigger becomes a **deny-list** (`paths-ignore`: markdown, `docs/`, `website/`, `refs/`) and a five-second `scope` job keeps the OS matrix on the narrow set. An allow-list fails silent — an input nobody listed produces no run, and no run reads as no problem. A deny-list fails towards running, which is the only direction a guard may fail in.

The gate is **per-RUNNER, not per-job**, deliberately: a per-job "could a consumer change move this?" list is a second copy of the dependency graph, and the copy is what drifts. ADR 0018 § 5 already settles the residual — the OS axis is verified on main and nightly, not per PR. The nightly stays; it is what bounds a break when nobody pushes at all.

`cancel-in-progress` also comes into line with `main.yml` (`pull_request` only). A still-PENDING run is evicted when a newer run joins the group, so a landed commit can end up with no run at all — and widening the trigger puts more main pushes in this group.

## And the reason nobody noticed for a day

`main.yml`'s coverage summary already listed every workflow with **no run for this commit**. `node-gi.yml` was in that table every day from 2026-08-13 — indistinguishable from `release.yml`, which is post-merge by definition. The list said which workflows were quiet; it never said which were **red**.

Each row now carries the conclusion of that workflow's last completed run on `main`, and a push to main annotates the ones that are not green. Measured by running the step's script against this repo at `63a18ebcde`:

```
::warning title=node-gi has no run here and last failure on main::https://github.com/gjsify/gjsify/actions/runs/31825211678

| node-gi | `.github/workflows/node-gi.yml` | [failure](…/runs/31825211678) · 2026-08-14 |
| napi    | `.github/workflows/napi.yml`    | [success](…/runs/31825211729) · 2026-08-14 |
```

Not annotated on a pull request — the row still prints, but warning there tags every open PR with someone else's breakage, which is how a red check gets read as noise.

## The scope job, measured on both branches

`Scope the OS legs` runs in 4 s and needs no checkout — it reads the changed-file list from the API and fails open when it cannot. Both branches were run against the real file lists of four merged/open PRs before this was pushed:

| PR | what it touches | verdict |
|---|---|---|
| #1179 (this one) | `packages/node/zlib` **+ `.github/workflows/node-gi.yml`** | `engine=true` — node-gi's own sources changed |
| #1174 | `packages/node-gi/**`, the harness | `engine=true` |
| #1178 | `packages/infra/cli` only | `engine=false` — Linux legs run, OS matrix does not |
| #1180 | `refs/gtkx` only | `engine=false` (and `paths-ignore` means it never triggers at all) |

The `true` branch is also proven live: this PR's run shows `Scope the OS legs  pass  4s` with the full matrix behind it.

## What this does not close

- `paths-ignore` means a markdown-only PR no longer triggers the workflow at all. That is the trade for the deny-list, and the nightly covers it.
- The four other path-filtered workflows (`napi`, `prebuilds`, `deploy-docs`, `build-ci-image`) keep allow-lists. Each guards roughly what its paths name, so none is a measured instance of the class — but nothing yet *checks* that, and the coverage annotation above is what would surface it if one of them goes red quietly.

Closes #1173.


JumpLink:
**Note on the cancelled `node-gi` run** (31844496172), so it is not read as a failure.

The previous commit's `node-gi` run — [31842206245](https://github.com/gjsify/gjsify/actions/runs/31842206245) — was **fully green across all 23 jobs**, including the consumer harness this PR exists to fix. The only commit since is `5f6d069a41`, which changes one line of `docs/ci-selective.md` and nothing else.

Cancelling it is the rule in this PR applied to itself: `docs/**` is in the new `paths-ignore`, so once this merges a docs-only commit does not trigger the workflow at all. Re-running 23 jobs — including ten macOS and six Windows legs — to re-prove an unchanged tree is exactly the cost the `scope` job exists to stop, and the runner pool is currently the binding constraint on every open PR here.

For the record, what the green run established:

| leg | result |
|---|---|
| `node-gi consumer harness (proof set / Node+Bun+Deno)` | **pass** — the fix, proven through the reverse bridge |
| `sqlite suite as node-gi consumer` | pass |
| `Scope the OS legs` | pass, 4 s, `engine=true` |
| all macOS ×10 / Windows ×6 / arm64 | pass |

A `workflow_dispatch` baseline on `main` at the same time ([31842428180](https://github.com/gjsify/gjsify/actions/runs/31842428180)) has that same harness **failing**, 22/23 — which is the measurement that makes this PR's claim checkable rather than asserted.

PR #1183, opened 2026-08-14
fix(node-gi): resolve members via the prototype

### LINKED ISSUE #1175: node-gi: an instance ignores a shadowed prototype method

JumpLink:
An instance did not resolve a GI method through the JS prototype chain: assigning to
`Ns.Class.prototype` stuck and was observable there, and the instance still reached the
native method. A spy therefore reported itself installed and then measured nothing —
a test asserting `calls === 0` would have gone green having measured nothing.

`Closes #1175`

## What was measured first

The issue's script, unchanged, on gjs 1.88.1 / girepository 2.88.3 and Node 24.15.0
(`NODE_GI_NATIVE=build`), before any edit:

```
$ gjs -m repro.mjs                    $ node repro.mjs
proto.convert === proto.convert: true   proto.convert === proto.convert: true
proto.convert === spy: true             proto.convert === spy: true
own convert on instance: false          own convert on instance: false
d.convert === spy: true                 d.convert === spy: FALSE
```

Row 1 was true on node for the wrong reason: `typeof Gio.ZlibDecompressor.prototype.convert`
was `undefined` on both sides of the `===`, while gjs answers `function`.

## The hypothesis in the issue is wrong

There is no V8 named-property interceptor to make non-masking — there is no V8 at all:

```
$ grep -rn "SetHandler|NamedPropertyHandler|PropertyHandlerFlags|InstanceTemplate|ObjectTemplate|v8::" src/
(no matches)
```

The engine is pure N-API/node-addon-api. The masking happens one layer up, in the L1
runtime (`gi.js`): a wrapper is `new Proxy({[HANDLE]: handle}, …)` — a bare target whose
[[Prototype]] is `Object.prototype`, with no link to the class — and `makeClass` built
`ctor.prototype` as an object that never received a single method. So the `get` trap fell
straight through to `native.callMethod`, and the class prototype was an inert object that
happened to accept writes. Both halves had to be fixed: an instance that consults a
prototype chain is useless if the prototype is empty (`original.apply(...)` in a
wrap-the-method idiom would then throw on `undefined`).

## What the change does

- **`Ns.Class.prototype` resolves lazily.** It is a Proxy whose `get`/`getOwnPropertyDescriptor`/`has`
  traps MATERIALIZE the introspected method for a name and define it on the target — the JS
  twin of the SpiderMonkey `resolve` hook gjs installs on every GObject prototype. Lazy, not
  eager: `Gtk.Widget` carries hundreds of inherited methods and a program touches a handful.
  Methods land as writable/enumerable/configurable data properties, so the
  replace-then-restore idiom works.
- **Every wrapper carries a prototype**, resolved from the handle's runtime GType: a
  registerClass subclass's own prototype when we registered that GType, else the introspected
  class's. Memoized per GType name.
- **Lookup order on a wrapper** is now: own JS field → class prototype chain → `runAsync` /
  the GObject.js shims → GObject property → inherited `Object.prototype` member →
  `Gio._promisify` registry → introspected method → `undefined`. The own-field read moved
  ahead of the prototype (ordinary JS shadowing — without the move, patching a prototype would
  have started shadowing instance expandos).
- **Prototype functions are returned unwrapped.** The old userProto branch returned
  `(...args) => desc.value.apply(proxy, args)`, which destroyed identity; `this` comes from the
  call site anyway.
- **Two native entry points**: `hasClassMethod(ns, type, name)` (presence by TYPE, so a
  prototype can answer before any instance exists) and `classInfoForTypeName(gtypeName)`
  (runtime GType name → introspected class, over the same nearest-ancestor walk
  `CallMethod`/`HasMethod` use; `FindNearestObjectInfo` was de-`static`'d rather than copied).
  `classInfoForTypeName` searches only LOADED namespaces, so wrapping an object never pulls
  in a typelib as a side effect.
- **The namespace cache is keyed by namespace alone.** It was `Ns@version`, which handed
  `requireGi('Gio','2.0')` and `requireGi('Gio')` two independent namespace objects — so
  `Gio.File !== Gio.File` across the two spellings and a prototype patched through one was
  invisible through the other. A repository cannot hold two versions of a namespace anyway.

## What broke while doing it, and why the fix is what it is

The first cut made the prototype win over `objectPrototypeShim`, and four tests went red:

```
✖ bind_property_full: a transform-to actually transforms a bound value — null !== 'true'
✖ bind_property_full: [false, …] leaves the target unchanged
✖ bind_property_full: bidirectional transform-from converts back — null !== 'n:5'
✖ bind_property_full runs the JS transform (GClosure IN-arg) — false !== true
```

Measured cause, not guessed:

```
$ node -e "… native.hasClassMethod('GObject','Object', m)"
bind_property_full true    bind_full false           connect false
bindPropertyFull   true    block_signal_handler false  emit false
```

`bind_property_full` IS introspected — as its GIR shadow `bind_with_closures`, whose GValue
write-back a marshaled GClosure cannot satisfy, which is why L1 shims it. So `materialize`
installs the SHIM for those names and only then asks the typelib.

## Measurements (all on this host: gjs 1.88.1, girepository 2.88.3, node 24.15.0, bun 1.3.14, deno 2.9.4)

| | verdict |
|---|---|
| the #1175 script, gjs vs node | all four rows identical, `d.convert === spy: true` on both |
| `npm test` (×4 consecutive) | `tests 536 · pass 523 · fail 0` (13 skipped need `--expose-gc`) |
| `npm run test:gc` | `tests 536 · pass 536 · fail 0 · skipped 0` |
| `npm run test:dbus` | `tests 8 · pass 8 · fail 0` |
| `npm run test:conformance` | 27 programs × gjs/node/bun/deno — **all green**, ledger still empty |
| consumer proof set (`sqlite`, `http2`, `zlib`) on node/bun/deno | sqlite 105/105, http2 102/102, zlib 53376/53376 on node + bun |

New conformance program `prototype-chain` byte-compares the four rows of #1175 plus the
surrounding behaviour (the prototype carries `convert`/`reset` and NOT `notAMethod`, the spy is
actually reached, a counting wrapper counts real calls, restore puts native dispatch back).
Golden = the gjs output; node/bun/deno match it byte-for-byte.

### The zlib spec (issue's motivating case)

`packages/node/zlib/src/gio-codec.gjs.spec.ts` is unchanged — **no `it.failing` was needed**.
Two things had to be separated:

1. The spec is `on('Gjs')`-gated, and through the consumer bridge `@gjsify/runtime` reports
   Node.js, so the harness does not execute it at all (measured: 53376/53376 pass and none of
   its three test names appear in the output). Its absence there is an axis-gate fact, not a
   pass.
2. Its body, run through the bridge with ONLY the axis gate widened (a scratch entry, not
   committed), passes on node, bun and deno: `pass 8/8` three times over.

That measurement is discriminating — with `origin/main`'s `gi.js` restored under the same
addon it reproduces the issue's failure exactly:

```
@gjsify/zlib   ok   part 5/6   assertion-mismatch
  never passes an unbounded input slice to convert() — Expected 0 to be greater than 0
```

## What this does NOT close

- **Class prototypes are not chained to their base classes.** `Gio.ZlibDecompressor.prototype`'s
  [[Prototype]] is still the vfunc chain-up proxy, not `GObject.Object.prototype`. Consequences,
  both measured: `instanceof` still resolves through `g_type_is_a` (unchanged), and one interface
  method materializes once PER implementing class where gjs hands every implementor the same
  function object (`Gio.ZlibCompressor.prototype.convert === Gio.ZlibDecompressor.prototype.convert`
  is true on gjs, false here). Deliberately NOT asserted in the conformance program: a ledger
  entry for that one row would have excused the WHOLE program on node/bun/deno and the fix would
  then be gated by nothing. It is recorded in the README instead.
- **GObject properties are not surfaced as prototype accessors** — they still resolve on the
  instance only, so `Cls.prototype.someProperty` is `undefined` where gjs has an accessor.
- **A private concrete GType resolves to its nearest INTROSPECTABLE ancestor**, not to the
  interface gjs would use: `Object.getPrototypeOf(Gio.File.new_for_path('/tmp'))` is
  `unknown_GLocalFile` on gjs, and here that instance gets `GObject.Object.prototype`. That is
  why `Gio._promisify` keeps its per-class registry as well as writing to the prototype.
- **Pre-existing, unrelated, reported for completeness**: `@gjsify/zlib` on **deno** fails one
  test locally (`gunzip async should decompress concatenated gzip members — failed to write
  whole buffer`) — byte-identical before and after this change, verified by restoring
  `origin/main`'s `gi.js` under the same addon. `scripts/check-comment-budget.mjs --check`
  reports `scripts` 42 lines over its ceiling on a clean `origin/main` checkout too; this PR
  does not touch that tree and `packages/node-gi` stays inside its own ceiling (3 lines spare).
- `packages/node-gi/AGENTS.md` grew by ~550 bytes, so `status/agent-context-budget.json` is
  raised in the same commit; the detail lives in the package README, not in the context file.

Overlaps with open PR #1179 (`fix/zlib-async-gzip-members`) only by subject matter — this PR
touches no file under `packages/node/zlib`.


JumpLink:
**The red `node-gi consumer harness` here is `main`'s, not this PR's** — measured rather than assumed:

```
▶ @gjsify/zlib … node:partial bun:pass deno:partial
  @gjsify/zlib — partial
=== grouped ===
  PASS   7  @gjsify/sqlite, @gjsify/http2, @gjsify/tls, @gjsify/ws, @gjsify/dom-elements, @gjsify/crypto, @gjsify/string_decoder
```

Byte-for-byte the signature of #1173, which `main` still carries: a `workflow_dispatch` baseline on `main` at 21:26 ([31842428180](https://github.com/gjsify/gjsify/actions/runs/31842428180)) fails the same job with the same package, 22 of 23 green. This branch is based on that `main`, so it inherits it.

#1179 fixes it (`gunzip()` took the platform's `DecompressionStream`, which decodes exactly ONE gzip member) and its own run has the harness **passing**. Merge order is therefore #1179 first, then a rebase here — no change is needed in this PR.

Worth noting for whoever reads this later: the harness's node column is what `--require-pass` gates, so `@gjsify/node-globals … bun:partial` in the same output is not what reddened the job.

Issue #1157, opened 2026-08-13
A whole-tree metric can turn main red with every PR green


JumpLink:
## The shape

`check-comment-budget --check` — part of the required check `Detect runtime-triplet drift` — measures a **ratio over a whole tree**, not over a diff. Two PRs that each touch the same tree are each measured against `main + their own changes` and never against each other. Both pass; the second merge lands a state neither run measured.

Nothing catches this before it is on `main`, because the ruleset has:

```
strict_required_status_checks_policy: false
```

so a branch is not required to be up to date with `main` before merging, and the second PR's green check is a statement about a tree that no longer exists at merge time.

## Measured, not hypothesised

Found while #1152 and #1156 were open simultaneously, both writing to `scripts/`. First attempt was arithmetic over the two PRs' reported deltas; that gave the wrong answer twice over. A probe worktree that actually merged the real branches gave:

| state | ratio | verdict |
|---|---|---|
| `main` | 5014/8678 = 0.578 | ok |
| `main` + #1152 + #1156 (pre-trim) | 5060/8671 = **0.584** | **over 0.583** |
| `main` + #1152 + #1153 + #1156 | 5019/8662 = 0.579 | ok, 30 lines spare |

Two things the arithmetic could not have known:

1. **Removing CODE fails the check.** It is a ratio. #1156 deleted ~10 code lines and its own headroom dropped from an estimated 35 to an actual 8 — without adding a single comment. "I only deleted things" is not a safe assumption, and the failure message currently says nothing about it.
2. **There was a third writer.** #1153 touches eight files under `scripts/` and removes comment lines net, which is what makes the three-way merge pass while the pair overflows. The set that overflows is therefore not the set that lands — only a probe merge of the real branches can say, and any estimate over pairwise deltas is blind to whoever else is open.

## Scope

`scripts/` is where it surfaced, but nothing about the mechanism is specific to it. Every area in `status/comment-budget.json` is a shared tree, and several sit at **exactly** their ceiling because `--update` re-baselines to the measured value: `packages/infra/cli` 0.426/0.426, `packages/node-gi` 0.434/0.434, `website` 0.170/0.170. On a tree with zero headroom, any two concurrent PRs that add a comment collide.

Same reasoning applies to any other whole-tree or whole-repo metric behind a required check, not only this one.

## Options, not a recommendation

[...]

PR #1184, opened 2026-08-14
fix(node-gi): pair the pump's prepare with a check

### LINKED ISSUE #1145: devtools methods hang when the bus name IS the application ID
Measured while running the 0.38.0 cross-OS showcase matrix. **This is a defect in the
MEASURING instrument, which is why it is worth more than its size** — it produced a wrong
verdict about a working application, and every GUI cell in that matrix is proven through
this interface.

## The symptom

`adwaita-storybook` under `--runtime node` on Linux:

- `org.freedesktop.DBus.Peer.Ping` answers immediately
- `Introspect` returns the full devtools interface
- `ListToplevels`, `GetStatus` and `Screenshot` all run into the timeout

The process stays alive at ~0.5 % CPU with a flat RSS — it is not spinning and it is not
dead. `bun` and `deno` on the *same bundle* answer all three.

## The correlation

| showcase | D-Bus name | GApplication ID | methods answer |
|---|---|---|---|
| `adwaita-storybook` | `org.gjsify.AdwaitaStorybook` | `org.gjsify.AdwaitaStorybook` | **no** |
| the `dom` showcases | `gjsify.examples.*` | (different) | yes |

The one that hangs is the one whose devtools bus name **is** the application ID. That is a
correlation on a sample of a few, not yet a cause — but it is the obvious first thing to
A/B: export devtools under a name distinct from the app ID and see whether the hang moves
with the name.

Plausible mechanism to check first: when GApplication already owns that name, the devtools
export lands on the same connection/name pair and the method dispatch ends up queued behind
GApplication's own handling rather than on the devtools object — consistent with Ping (a
different, always-answered interface) working while our methods do not.

## Why it matters more than one red cell

The wrong verdict it produced was `NO_WINDOW` for a showcase that in fact renders a full
Adwaita window. The independent cross-check that finally showed the truth bypassed devtools
entirely (Windows `EnumWindows` + a real screen capture). So for as long as this is open,
a devtools timeout cannot be read as "the app has no window", and every GUI conclusion
drawn through devtools needs a second, non-devtools witness.

That is the same axis as #1141 (a green run must prove what it actually did), one layer
out: here a run *fails* while proving nothing.

## Note on the harness, not the bug

Related but separate, and already known: `Screenshot`'s `scope` in-arg is dead — a caller
asking for a widget silently gets the active window (`status/open-todos.md`).

JumpLink:
`adwaita-storybook` under `--runtime node` on Linux answered `Peer.Ping` and `Introspect`
and timed out on every `org.gjsify.Devtools` method. **The bus name has nothing to do with
it, and nothing in the devtools dispatch path was wrong.** The app was deadlocked inside
`gtk_window_present()` — in `@gjsify/node-gi`'s libuv↔GLib pump — and the instrument was
reporting that correctly. (A second, unrelated devtools defect surfaced while measuring and is
fixed here too; see the last section but one.)

## The A/B the issue asked for — hypothesis DISPROVED

The hypothesis was "the devtools export lands on the same connection/name pair GApplication
already owns, so dispatch queues behind GApplication's own handling". Ruled out on the first
two measurements:

- **The `dom` showcases are the same shape.** `gjsify.examples.canvas2d-fireworks` *is* their
  GApplication id and *is* the devtools dest — the issue's "(different)" column is not a real
  difference. So "bus name == app id" does not separate the working row from the failing one.
- **`Introspect` proves the object is registered; `Ping` proves nothing about dispatch.**
  Both are answered by GDBus in the connection's worker thread, in C. A registered object's
  method call is scheduled onto the GMainContext captured at `register_object` time. So the
  symptom says "that context is not being iterated", never "the name is wrong".

Then the real A/B (`gdbus call … GetStatus`, 3 runs each, this Wayland session, gjs 1.88.1 /
gtk 4.22 / Fedora 44):

| storybook bundle | `@gjsify/node-gi` addon | result |
|---|---|---|
| published 0.38.0 | published 0.38.0 prebuild | **HANG 3/3** |
| published 0.38.0 | built from `main` (pre-fix) | **HANG 3/3** |
| published 0.38.0 | built from this branch | OK 3/3 |
| built from `main` | built from `main` (pre-fix) | OK 3/3 |

Row 4 is why this looked bus-name-shaped: it is a race window, and `main`'s bundle happens to
miss it. Rows 2 and 3 isolate the change to `src/loop.cc`.

## The cause

`gdb` on the hung process, main thread:

```
#0  __syscall_cancel_arch
#3  pthread_cond_wait          libc
#4  wl_display_read_events     libwayland-client
#7  wl_display_roundtrip_queue libwayland-client
#8  wsi_wl_display_init        libvulkan_radeon
#11 gdk_vulkan_context_surface_attach   libgtk-4
#13 gsk_gpu_renderer_realize            libgtk-4
#16 gtk_window_realize                  libgtk-4
#27 _gtk_window_present                 libgtk-4
#31 gi_callable_info_invoke     libgirepository-2.0
#34 nodegi::CallMethod          node_gi.node
```

No other thread is inside libwayland — the main thread is waiting for a designated reader
that is itself. `PumpArmWakeups()` (the pump's wake-up hint pass) ran
`g_main_context_prepare()` + `g_main_context_query()` and stopped, with this comment:

> The context is left "prepared" without a matching check/dispatch — safe: the next
> `g_main_context_iteration()` (ours or a blocking loop's) simply re-prepares.

It is not safe. **A GSource may hold a lock between `prepare()` and `check()`**, and GDK's
Wayland event source is exactly that shape: `prepare()` calls `wl_display_prepare_read_queue()`
and only `check()` calls `wl_display_read_events()` / `wl_display_cancel_read()`. The
query-only pass leaves libwayland's `reader_count` incremented forever, so the next
`wl_display_roundtrip*` on that thread blocks with no way out. And it compounds: glib's
`g_main_context_prepare_unlocked` skips a source only while `G_SOURCE_READY` is set, so every
later pass takes the slot again.

Three conditions, all of them necessary — which is what made the correlation table look like
a name problem:

1. **Wayland.** GdkX11's event source holds nothing across the pair. Same fixture with
   `GDK_BACKEND=x11`: `present()` returns.
2. **node.** The `uv_prepare`/`uv_check` pump is node-only; bun and deno drive GLib through
   complete `g_main_context_iteration()` calls, which always check.
3. **GTK initialised at `g_main_depth() == 0`.** `PumpArmWakeups()` no-ops at depth > 0, so
   the leak needs GDK's source attached *before* the blocking run. `runApplication()`
   (`@gjsify/adwaita-app`, used by the storybook) calls `app.register(null)` before
   `runAsync()`, and `g_application_register` emits `::startup` → `gtk_init`. The `dom`
   showcases go straight to `await app.runAsync([])`, so GDK comes up inside the loop and
   they never meet it. **That**, not the bus name, is the difference between the two rows.

24 lines reproduce it with no storybook, no devtools and no bus name at all —
`fixtures/wayland-prepare-check-app.mjs`.

## The fix

`PumpArmWakeups()` closes its own cycle: `prepare` → `query` → `check(fds, revents = 0)`.
Zero revents is the correct reading — the pass did not poll, so nothing is ready and every
source that took something in `prepare()` gives it back. No `dispatch` pairs with it:
`g_main_context_prepare()` clears `pending_dispatches` on entry (glib `gmain.c`), and
`PumpDrainContext()` — a full iteration — runs on the same libuv turn.

## The regression test

`packages/node-gi/node-gi/test/wayland-prepare-check.test.mjs` + its child fixture. It spawns
the child because the pre-fix failure is a hard block of the main thread inside libffi:
`node:test`'s own timeout cannot reclaim a thread parked in `pthread_cond_wait`, so the whole
runner would wedge.

Verified in both directions on this branch, against a `main` addon rebuilt from
`git checkout -- src/loop.cc`:

```
$ NODE_GI_NATIVE=<pre-fix addon> node --test test/wayland-prepare-check.test.mjs
✖ the uv pump leaves no GSource prepared across a Wayland roundtrip (45028.505845ms)
  AssertionError: gtk_window_present() never returned: the uv pump left the default
  GMainContext prepared, so GDK's Wayland source still holds libwayland's reader slot
  and the GSK renderer's roundtrip deadlocks (gjsify #1145).
  REGISTERED

$ NODE_GI_NATIVE=build node --test test/wayland-prepare-check.test.mjs
✔ the uv pump leaves no GSource prepared across a Wayland roundtrip (235.748142ms)
```

Full suite with the fix: `npm test` 537 tests / 0 fail (was 536), `npm run test:dbus` 8/8.

**The guard needs a compositor, so it gets one.** Every other GTK step in `gtk-smoke` pins
`GDK_BACKEND=x11`, where this is invisible, so a new step runs the single test under headless
weston — a display AXIS, not a second way to do the same thing. weston is baked into
`.docker/ci-fedora.Dockerfile` beside Xvfb (its headless backend needs no GPU, no seat, no
DRM); `check-ci-image-packages.mjs` insisted on that and was right. The job keeps a
`command -v weston || dnf install` line that covers exactly one window and retires itself:
`build-ci-image.yml` pushes the `:44` tag only from `main`, so the run that ADDS a package to
the image still runs on the old one. `NODE_GI_REQUIRE_WAYLAND=1` turns the test's self-skip
into a failure there, so a compositor that never came up cannot be reported as a pass.

**The weston invocation is the one thing in this PR not measured locally** — this box has a
real Wayland session and no weston. It tries both the `--backend=headless` and legacy
`headless-backend.so` spellings and prints weston's log before failing. `node-gi.yml` is
path-filtered, i.e. advisory, so a bad first run cannot block anything.

## A second defect, found while measuring this one — and fixed here

`Gio.Application` derives its object path with dots → slashes **and hyphens → underscores**
(gio `gapplicationimpl-dbus.c`). Both devtools derivations — `resolveBusAddress` for the
bridge's prediction, `installDevtools`' bus-less fallback for the export itself — spelled the
dots half only, so they agreed with each other and disagreed with the app.

`-` is not a legal D-Bus object-path character, and getting it wrong is SILENT. Measured on
gjs 1.88.1:

```
$ gjs -c "…impl.export(conn, '/gjsify/examples/canvas2d-fireworks/devtools')…"
(gjs:118714): GLib-GIO-CRITICAL **: g_dbus_interface_skeleton_export:
              assertion 'g_variant_is_object_path (object_path)' failed
HYPHEN PATH: exported OK          ← it does not throw; it exports NOTHING
is_object_path(hyphen)=false
is_object_path(underscore)=true

$ gjs -c "…new Gio.Application({application_id:'gjsify.examples.canvas2d-fireworks'})…"
path=/gjsify/examples/canvas2d_fireworks
```

Every `gjsify.examples.*-*` showcase carries such an id. On a session bus the app itself is
fine (it uses the authoritative `get_dbus_object_path()`), but `gjsify debug` predicted a path
that can never exist; on the bus-less peer transport (macOS/Windows) the export produced no
object while reporting success. One derivation now — `appIdToObjectPath` in
`@gjsify/devtools-protocol`, used by both sides — with a spec asserting the mapping and that
the result matches the object-path grammar.

## What this does NOT close

- **`Screenshot(scope)` in `status/open-todos.md` is untouched** — nothing in that entry is
  resolved by this, so nothing was deleted. Its two unproven shapes (a non-active toplevel, a
  child widget) still wait on `tests/e2e/devtools-export/` becoming listable, which is its own
  entry and a different environment problem. The issue's closing note ("Screenshot's `scope`
  in-arg is dead") is itself stale — #1153 fixed that.
- **`@gjsify/devtools` gets no change.** The A/B says it was never wrong here: it timed out
  because the app's main loop was gone. What is worth carrying forward is that a devtools
  timeout means *"this app is not dispatching"*, which is a strictly stronger statement than
  "no window" — the 0.38.0 matrix's `NO_WINDOW` verdict for Linux/node was a misreading of a
  real hang, not an invented one.
- **The published 0.38.0/0.39.0 prebuilds still carry the bug**; consumers get the fix with the
  next release train.

Closes #1145


JumpLink:
**The red `node-gi consumer harness` is inherited from `main`, not caused by this PR** — verified from the log rather than inferred, because this branch does touch `src/loop.cc` and could plausibly have moved it:

```
=== grouped ===
  PASS   7  @gjsify/sqlite, @gjsify/http2, @gjsify/tls, @gjsify/ws, @gjsify/dom-elements, @gjsify/crypto, @gjsify/string_decoder
  @gjsify/zlib — partial
```

Identical signature and identical PASS set to `main`'s own `workflow_dispatch` baseline at 21:26 ([31842428180](https://github.com/gjsify/gjsify/actions/runs/31842428180)) and to #1183's run — i.e. #1173, which `main` still carries and #1179 fixes. Rebase after #1179 lands and this goes green; nothing to change here.

JumpLink:
Rebased onto `main` after #1179, #1183 and #1181 landed. Two notes worth keeping.

**The harness is no longer inherited-red**: #1179 fixed `@gjsify/zlib`, so this branch's copy of that failure is gone with the rebase.

**The rebase hit the interlock #1181 introduced hours earlier, and the numbers are worth recording** — this is the first real instance of the shape #1157 described:

```
<<<<<<< HEAD
    "packages/node-gi/AGENTS.md": 10251,     ← #1183, already on main
=======
    "packages/node-gi/AGENTS.md": 10487,     ← this branch
>>>>>>>
```

Both PRs grew the same context file, each measured its own ceiling against its own base, and **both were green**. The merged file measures **11034 bytes — larger than either competing value**. So the combined state was over both ceilings and no CI run had ever looked at it.

With the pre-#1181 ledger this would have merged cleanly (the ceilings carried slack, and the raise would have been "already covered") and `main` would have gone red on a required check for a reason appearing in neither diff. Instead **git refused the merge**, and the resolution is the one the failure message names: `node scripts/check-agent-context-size.mjs --update`, which writes the value measured on the merged content.

Cost, as advertised in #1181: one extra step in a rebase. Value: the failure surfaced in a working copy instead of on `main`, where per #906 it would have reddened every other open PR for a reason none of them contained.

PR #1190, opened 2026-08-15
feat(adwaita-nativescript): apply the window insets

### LINKED ISSUE #1128: Android: go properly edge-to-edge, and move to targetSdk 36 / minSdk 26
## What you see

Open any NativeScript dialog — the appearance dialog from #1127 is the easiest — and the modal scrim dims the app content but **stops short of the status bar and the bottom gesture area**. It reads as "an overlay inside a rectangle" rather than "the app is dimmed".

It is not the scrim that is too small. The ROOT VIEW is: the platform is already drawing edge-to-edge and the app lays its content out as if it were not.

This is pre-existing and app-wide — `AdwAlertDialog`, `AdwAboutDialog`, `AdwPreferencesDialog` and `AdwToastOverlay` all share the edge. #1127 only made it visible.

## The obvious fix does not exist any more

The tempting cheap version is "colour the system bars to match the scrim". On `targetSdkVersion 35` that is not available. From [Android 15 behavior changes](https://developer.android.com/about/versions/15/behavior-changes-15):

> Apps are edge-to-edge by default on devices running Android 15 if the app is targeting Android 15 (API level 35).

> The following APIs are deprecated and disabled: `R.attr#navigationBarColor` (for gesture navigation), `R.attr#statusBarColor`, `Window#setDecorFitsSystemWindows`, `Window#setNavigationBarColor` (for gesture navigation), `Window#setStatusBarColor`

> Transparent by default. The top offset is disabled so content draws behind the status bar unless insets are applied. `setStatusBarColor` … have no effect on Android 15.

So the only route is the real one: draw edge-to-edge deliberately and apply window insets as padding where content needs it.

**There is currently no inset handling anywhere in the tree.** A grep over `packages/nativescript-bridge/` and the showcases' `App_Resources` for `WindowInsets` / `fitsSystemWindows` / `setDecorFitsSystemWindows` / `systemui` returns nothing.

## Scope

1. **Insets in the shared chrome**, not per app. `AdwToolbarView` / `AdwHeaderBar` want the top inset as padding; whatever sits at the bottom wants the navigation inset. Getting it into the shared widgets is what fixes every showcase and every consumer at once — doing it per app is how it drifts.
2. **Dialogs and overlays must span the full window** so their scrim covers the bars. Note the trap already hit in #1127: the dialog cannot be installed by WRAPPING the root (NS refuses to re-parent, on every `onNavigatingTo`); it is a spanning child.
3. **`targetSdkVersion` 35 → 36, `minSdkVersion` 24 → 26.**

## Why the SDK bump belongs with it

Google Play requires API 36 for new apps and updates from **2026-08-31**, with an extension possible to 2026-11-01 ([Play Console Help](https://support.google.com/googleplay/android-developer/answer/11926878?hl=en)). Nothing here is published to Play, so that is not binding — but there is no reason to sit on the older level either, and edge-to-edge is enforced at 35 already, so the inset work is required at both levels.

`minSdkVersion` 24 (Android 7.0, 2016) → 26 (8.0) costs **0.5 percentage points** of reach (96.6% → 96.1%, Statcounter April 2026 via [apilevels.com](https://apilevels.com/)) and drops two years of legacy. 28 would cost 3.1 points for nothing we need.

**minSdk is NOT the blocker for edge-to-edge** — `WindowInsetsCompat` and `WindowCompat` work down to API 21. The two changes travel together for convenience, not because one enables the other.

## Risks to weigh

- **We would be ahead of the framework.** NativeScript's own project template (`tools/assets/App_Resources/Android/app.gradle` on current `main`) still ships `compileSdkVersion 35` / `minSdkVersion 24` / `targetSdkVersion 35`. Going to 36 means being the first to find whatever NS has not fixed there.
- **The Android runtime is behind.** `@nativescript/android` is pinned at `9.1.0-alpha.6` while `9.1.0-alpha.9` is published; SDK 36 support may well depend on it. `@nativescript/core` is already on the newest published version of its line (`9.1.0-alpha.11`), so there is nothing to gain there.
- A missed inset shows as content under the clock or under the gesture bar, which is easy to miss on one screen and obvious on another. Verification has to walk the collapsed AND expanded layouts, not just one.

## Reference to work from

`references/nativescript/systemui` in the easy6502 tree (updated 2026-08-12) is the NativeScript package for system bars and insets — the closest thing to a worked example. `references/nativescript/nativescript` is current as of the same date for searching the framework's own inset plumbing.

## Related

Blocked-adjacent to #1126 (the 22 unreviewed NativeScript theme classes) only in that both touch `theme/adwaita.css`; otherwise independent.

JumpLink:
Closes #1128.

Android draws edge-to-edge from API 35 whether an app asks or not, and the APIs that used to colour the bars instead are deprecated **and disabled** there. Nothing in this tree ever applied an inset, so the chrome ran under the clock and under the gesture pill — and every dialog's scrim stopped short of both, which is how the issue was reported.

## Measured on a booted API-36 emulator

Same tree, only this change differing. 1080×2400 at 420 dpi → scale 2.625, so the 24 dp system bars are **63 px**. Colour runs down x=30:

| edge | before | after |
|---|---|---|
| **top** | nothing pays it — the sidebar list starts at `y=63`, directly under the status bar | **63 px** of chrome colour at `y=63–125`, then the header |
| **bottom** | nothing pays it — content runs to `y=2336`, under the gesture pill | **63 px** paid at `y=2274–2336` |

Both edges land on 63 px, and 63 is the platform's reading — there is no constant of that value anywhere in the code.

## The CSS was a feature-shaped no-op

`theme/adwaita.css` carried `padding-top: var(--safeAreaInsetTop, 0)` with a comment saying *"the host app feeds those vars"*. Nothing in the repo ever fed them — and nothing could have: **this NS CSS subset has no custom properties at all**, which is the same reason the accent has to be inlined as a literal (the ambient declaration in `ns-core.d.ts` says so). It is gone; the widget applies the real reading.

## Where the padding goes

Each edge is paid by the bar sitting on it, and by the toolbar view's **own** padding where that edge has no bar — never by the content view, which belongs to the consumer and may already carry padding the chrome would silently overwrite. The decision itself is the pure module that landed with the first half of this work; this PR is the platform half plus the wiring.

## One listener, many widgets

`setOnApplyWindowInsetsListener` **replaces** whatever was there, while an app has as many toolbar views as it has panes (the storybook has two). So the native listener is installed once and `WindowInsetsBroadcast` fans the reading out.

Its two load-bearing behaviours are exactly the ones a device cannot check cheaply and a Linux runner can, so both are tested:

- **a late subscriber gets the last reading at once** — insets dispatch once, early; a pane built afterwards would otherwise sit un-inset until the next rotation, which is the shape of bug that looks like "sometimes it works"
- **an unchanged reading is dropped** — Android re-dispatches insets on every layout pass, and writing padding schedules another layout pass

6 new tests, green on Node **and** GJS.

## iOS

Written, and **unverified on device** — it says so in its own header. It exists because the `nativescript-platforms` conformance rule demands both variants of any `.android.ts`, and that rule exists because `icons.android.ts` once shipped without an iOS twin and iOS silently rendered no icons at all.

## SDK bump

`targetSdkVersion` 35 → 36 (Play requires it for updates from 2026-08-31) and `minSdkVersion` 24 → 26 (0.5 points of reach, two years of legacy). Neither *enables* the inset work — `WindowInsetsCompat` works down to API 21 and edge-to-edge is enforced at 35 already — they simply travel together.

## Verified

`gjsify workspace @gjsify/adwaita-nativescript run test` green (Node + GJS) · package build clean · `gjsify format --check` green tree-wide · agent-context ceiling raised by one reviewed line (+106 B for a new subsystem) · APK built, installed and screenshotted on the emulator in both states.


JumpLink:
**Re-verified on the emulator after rebasing onto main** (which now carries the four new storybook stories, #1189), so one build proves both halves.

Top edge, colour runs down x=30 on the fresh APK:

```
  y   0- 62  h=63  (250, 250, 251)   status bar
  y  63-125  h=63  (255, 255, 255)   the inset, paid by the top bar in chrome colour
  y 126-...        the sidebar list
```

63 px = 24 dp at this device's 2.625 scale, identical to the first run — so the reading is the platform's, not a constant.

**A stale artefact almost made me report a defect that was not one.** The first rebuild showed the new `Controls` category sorting LAST in the NativeScript sidebar, after Feedback — which reads exactly like `orderCategories` failing to place it. It was not: `@gjsify/storybook-core`'s `lib/` still held the pre-#1189 array, because the app bundles the built package and I had only rebuilt `adwaita-nativescript`. Rebuilt `storybook-core` + `storybook-nativescript`, re-prepared, and the sidebar reads Presentation → Boxed Lists → **Controls** → … → Feedback, which is the declaration. Screenshots of both states are in my notes; the lesson is the one already in the tree — it never runs the source you edited.

JumpLink:
Self-review found one thing worth the CI restart: the iOS path re-read the safe area only on `resume`, which misses the case that actually changes it — **rotation while the app is in the foreground**. Landscape puts a cutout on the leading edge and shrinks the bottom bar, and the widget would have kept portrait's numbers until the app was backgrounded and returned.

It now listens to `orientationChanged` as well as `resume`. Still UNVERIFIED ON DEVICE and still says so — but knowingly-incomplete is a different thing from unverified, and only one of the two is acceptable to merge.

JumpLink:
CI caught a lint error in my own code, and it turned into a better change than the fix.

`oxlint`'s `unicorn/no-useless-spread` flagged `for (const listener of [...this._listeners])` — *"for…of can iterate over iterable, it's unnecessary to convert to an array"*. My comment defended the copy as protection against **mutation during iteration**, and **that defence was wrong**: a Set iterator is live, so deleting an entry mid-dispatch is already safe, and my existing test (which unsubscribes the first listener while it runs) passes with or without the copy.

The copy protects something narrower and real: **adding**. `subscribe()` replays the last reading immediately, so a live iterator reaches that same listener again in the very same dispatch — one reading, two calls, and the second call writes padding, which schedules the layout pass `publish`'s dedupe exists to avoid. A toolbar view built from another view's inset handler is how that happens on a device.

So: the comment now says the reason that actually holds, the rule is disabled on that one line with it, and there is a test.

The test was rot on its first run, for the right reason — my inner `subscribe` fired during the initial replay rather than during a dispatch, which is a different situation that passes either way. With the ordering fixed, the A/B is decisive:

| | result |
|---|---|
| with the copy | `[24]` |
| without | `[24, 24]` |

Local sweep after: `gjsify lint` 0 errors, NS build (tsc) + suite green on Node and GJS, `format --check` tree-wide.

PR #1192, opened 2026-08-15
fix(adwaita): retire the last three two-copy widgets

JumpLink:
The three the generated matrix still shows implemented on both renderers **without** a shared behaviour. Each carried a real divergence, not just a duplicated line. With #1191 this closes the set: 43 of 43.

## `adw-button` — two tables that disagreed

| | classes it knew |
|---|---|
| browser | flat · suggested-action · destructive-action · **circular** · pill |
| NativeScript | flat · suggested-action · destructive-action · — · pill |

`circular` existed on one renderer and not the other, and neither file could see the other. The table moves to `@gjsify/adwaita-core` — both spellings resolve, since the browser hands in `suggested` and NativeScript the class itself — and **NativeScript gains `circular`** because the shared list says it exists.

The **composition** rule deliberately stays with the renderers: GTK style classes are a set (`.pill.suggested-action` is a real button) and the browser composes independent attributes, while the NS widget exposes one `variant` because its markup surface is one property. That is a surface difference, not a drifted rule.

ADR 0004 says a widget with genuinely trivial behaviour does not need a core class — so this is a table and two functions, not a state machine.

## `adw-status-page` — one rule, spelled twice

`text ? … : …` on one side, `title.length === 0` on the other. It is the same `string_is_not_empty` closure the action rows already bind their labels to; both now call `stringIsNotEmpty`. That is also why a title of `'   '` stays **visible** — it reads the first byte, it never trims.

## `adw-about-dialog` — the NativeScript port showed every label

Unconditionally. A dialog without a version rendered a blank line where the version belongs. `aboutDialogVisibility` is what the browser port has answered with all along (`adw-about-dialog.c` binds each part's `visible` to a closure over the property that feeds it). Same input, two different dialogs, and nothing said so. The setters now share one `_render()` that collapses what has nothing to say.

## Verified

- **12 new vectors** for the class table, driven from the core suite **and** through a real `<adw-button>` in the browser suite
- the rows the browser element cannot express — the long spellings NativeScript uses — are partitioned out and **the partition is asserted**, so a row cannot fall out of coverage unnoticed. That assertion caught its own first draft being wrong (it claimed `destructive-action` was unreachable; it is not).
- `@gjsify/adwaita-core` suite green (Node + GJS) · `@gjsify/adwaita-nativescript` suite green (Node + GJS) · browser suite green · `check-adwaita-conformance-drivers` 155 tables, every one driven or explained · `check-adwaita-reset-components` 65 elements · `tsc --noEmit` · `gjsify format --check` tree-wide

### DIFF
--- a/packages/nativescript-bridge/adwaita/src/widgets/adw-about-dialog.ts
+++ b/packages/nativescript-bridge/adwaita/src/widgets/adw-about-dialog.ts
@@ -18,6 +18,7 @@
 // Reference: refs/libadwaita/src/stylesheet/widgets/_dialog.scss
 // Copyright (c) GNOME contributors (libadwaita). LGPLv2.1+.
 
+import { aboutDialogVisibility } from '@gjsify/adwaita-core';
 import { Button, GridLayout, ItemSpec, Label, ScrollView, StackLayout, type EventData } from '@nativescript/core';
 
 /** Event name emitted when the dialog is closed. */
@@ -77,6 +78,12 @@ export class AdwAboutDialog extends GridLayout {
         closeButton.set('androidElevation', 0);
         closeButton.addEventListener('tap', () => this.close());
         card.addChild(closeButton);
+
+        // A dialog nobody has written to yet is SEVEN empty labels, all visible —
+        // the same blank lines the visibility rules exist to remove. Every setter
+        // re-renders the whole card, so one call here is what covers the state
+        // before the first one.
+        this._render();
     }
 
     private _addCardLabel(card: StackLayout, className: string): Label {
@@ -95,7 +102,7 @@ export class AdwAboutDialog extends GridLayout {
 
     set applicationName(value: string) {
         this._applicationName = value ?? '';
-        this._nameLabel.text = this._applicationName;
+        this._render();
     }
 
     /** A glyph standing in for the app icon (no icon-theme lookup in the NS subset). */
@@ -105,7 +112,7 @@ export class AdwAboutDialog extends GridLayout {
 
     set applicationIcon(value: string) {
         this._applicationIcon = value ?? '';
-        this._iconLabel.text = this._applicationIcon;
+        this._render();
     }
 
     /** The version string. */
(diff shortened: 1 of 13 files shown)


JumpLink:
Rebased onto main (which now carries #1191) and fixed the type error CI caught: `copyright` is not a field of `AdwAboutDialogProps` — upstream has no copyright LABEL, it derives the string into the Legal page this flat-card port does not build (`legalSectionVisible` takes it as a separate argument). It keeps the non-empty rule directly; `comments` and `website` now read the core's `commentsLabel`/`websiteRow` instead of a local length check, which is more delegation than the first draft had.

Worth recording how it got past me: I ran the NS package's `run test`, which builds a bundle and runs it — the **tsc** pass lives in `run build`. Same shape as the stale-`lib/` trap right behind it: switching branches to build the website regenerated `adwaita-core`'s `lib/types` from a branch without `button.ts`, so the next NS typecheck failed against artefacts rather than sources. Rebuilt the core, then the package: both green, plus the browser suite, the core suite and `format --check`.

JumpLink:
Self-review found the same defect I was fixing, one level earlier: the NativeScript about dialog's constructor never rendered, so a **freshly built dialog is seven empty labels, all visible** — the exact blank lines `aboutDialogVisibility` exists to remove. The setters covered every field a consumer writes; nothing covered the state before the first write.

One `this._render()` at the end of the constructor. Every setter re-renders the whole card anyway, so that is the complete fix rather than a per-field one.

Worth naming as a shape: applying a visibility rule in the SETTERS leaves the constructed state uncovered, and the constructed state is the one a screenshot of an empty dialog would have shown.

JumpLink:
Correction to this PR's body: it says *"With #1191 this closes the set: 43 of 43"*. **The derived matrix says 42 of 43.** I asserted the number instead of regenerating it.

The one left is `adw-menu-button` on the NativeScript side: it reaches the core only through `AdwImageButton`'s helpers, which is TWO hops, and `generate-status.mjs` deliberately does not count those — a transitive walk would report a widget as core-backed because something four modules away imports core. Its own `AdwMenuItem` interface was still a third declaration of a shape the core owns, missing `icon`.

Fixed in #1196, which also makes the generated "the remainder still carry two copies" sentence conditional — at 43 of 43 it was naming a backlog that does not exist.

PR #1198, opened 2026-08-15
ci: un-gate the three whole-tree checks

### LINKED ISSUE #1187: A whole-tree check inside a per-workspace gate: `format --check` never ran on the PR that broke it
## What happened

`main` went red on `Check Fedora 44` → `Format (oxfmt --check)` after #1183 merged:

```
packages/node-gi/node-gi/conformance/programs/prototype-chain.conf.mjs (0ms)
Format issues found in above 1 files.
Finished in 362ms on 2896 files using 4 threads.
```

Fixed by #1186. This issue is about why the PR was green.

## The mechanism

#1183's diff is `packages/node-gi/**`, `packages/node-gi/AGENTS.md` and `status/agent-context-budget.json`. Every one of those is on the affected-classifier's IGNORE list — `packages/node-gi/**` (own CI), `**/*.md`, `status/` — so `changes` returned skip-all and **the entire GJS matrix skipped**, `Check Fedora 44` among it.

And `Check Fedora 44` is where `gjsify format --check` runs. That step is **whole-tree**: 2894–2896 files, `packages/node-gi/**` included. So a PR whose every path is "not worth building for" skipped the one check that covers those paths anyway.

The classifier is not wrong: `packages/node-gi/**` genuinely cannot change what `main.yml` BUILDS or TESTS, which is what IGNORE means (`docs/ci-selective.md` says so explicitly, and names the two-part claim membership requires). What it cannot express is that one STEP inside a gated job has a different subject from the job around it.

Same shape as #1173 and #1157, one layer in: **a guard whose subject is wider than its trigger.** Here the trigger is per-workspace and the subject is the tree.

## Scope — which steps are in this position

Worth enumerating before fixing, rather than patching the one that bit. Any step in a selectively-gated job whose input is the tree rather than the closure. `gjsify format --check` is one. `gjsify lint` is likely a second (oxlint takes a path set, not a workspace set). The type-check half of that job is correctly selective and should stay.

## Options, none free

- **Move the whole-tree steps to `audit-runtimes.yml`**, where the other whole-tree checks already live (`check-comment-budget`, `check-agent-context-size`, `check-workflow-run-syntax`) and which has no `paths:` filter. Cost: that workflow states "no install and no build" as its own constraint — deliberately, so it cannot reintroduce the staleness circularity `verify-committed-bundles.mjs` exists to break — and `gjsify format` needs the CLI. Adding an install there is a real change to that workflow's character.
- **A small always-run job in `main.yml`** that installs and runs only the whole-tree steps. Cost: one more job with an install on every PR, including docs-only ones.
- **Make the classifier able to say "this path still needs the format step"** — a third category beside IGNORE and GLOBAL_TRIGGERS. Cost: a new concept in a table whose own documentation warns that silent mis-categorisation is the most common way to leak slowdowns back into the PR loop.
- **Accept and detect**: `main`'s push run is unconditional, so a formatting break is caught within one merge. That is what happened here, and the cost is a red `main` plus, per #906, every open PR failing on `main`'s reason while looking like its own diff is broken.

The last option is in force today by default, and — as in #1157 — it was chosen by nobody.

## Not urgent, but do not lose it

The immediate break is fixed. What remains is that the next PR touching only IGNORE-listed paths can land unformatted code exactly the same way, and there is currently nothing between it and `main`.

JumpLink:
Closes #1187.

## The enumeration the issue asks for

Every step inside a `skip-all`-gated job that makes a claim about the **whole tree** rather than about a set of workspaces:

| step | lived in | claim |
|---|---|---|
| `Lint (oxlint, errors only)` | `check` | lint is clean |
| `Format (oxfmt --check)` | `check` | the tree is formatted |
| `Check dependency consistency` | `build` | no dependency range has drifted (incl. the TypeScript `^6.0.3` invariant) |

The third one is **not named in the issue**. It matters: `website` is both a root workspace member and a classifier-IGNORE path, so a website-only dependency bump skipped it. Fixing only lint and format would have left the class open — which is exactly what #1187 asks to avoid.

Checked and deliberately NOT moved: `verify-package-outputs.mjs`, `verify-tarball-outputs.mjs`, `verify-committed-bundles.mjs`. All three read built workspace output, which a change to an IGNORE path cannot alter.

## Why the merge button was green

`gate` counts `skipped` as a pass (`main.yml`, the `Verdict` step), and `check` is skipped whenever the classifier says `skip-all`. So a PR touching only `packages/node-gi/**`, `packages/napi/**`, `website/**`, `docs/**`, `status/**` or `**/*.md` merged with neither claim tested.

Three PRs paid for it:

- **#949 / #960** — over-width files landed green. The red surfaced on **#957**, an unrelated PR that happened to touch a non-ignored path, and on `main`'s own push run behind it.
- **#1183** — touched only `packages/node-gi/**`, skipped the matrix, landed unformatted code, reddened `main`. Patched afterwards by #1186.

Exposure, measured against `.oxfmtrc.json`'s ignore patterns: the IGNORE trees that actually carry lint/format-eligible source are `packages/node-gi` (181 files), `packages/napi` (71) and `website` (18). Neither `node-gi.yml` nor `napi.yml` has any lint or format step (grep: zero hits).

## Why a new job, and not a different `if:`

`check` is `needs: [changes, setup, build]` **and** requires `needs.build.result == 'success'`. `build` is itself gated on `skip-all`, so on the PRs that matter `check` is unreachable regardless of its own condition. The steps had to leave.

`tree-checks` therefore takes `needs: [setup]` — `setup` has no `if:` and is never skipped — and carries no classifier gate anywhere in its expression.

It reuses the existing `./.github/actions/gjsify-setup` composite with `bootstrap-bundles: 'false'`, so this adds **no new install path**. The CLI it drives is the published bootstrap (`$GJSIFY_BOOTSTRAP`), because a tree that is installed but not built has no other since ADR 0002.

That is honest for what the job claims: the CLI only assembles argv, while the **engines** are the workspace's own pinned `oxlint` / `oxfmt`, and `upgrade --check` reads the tree's `package.json` files. Whether the CLI's own `lint`/`format` commands still work is a different claim, held by `tests/e2e/{oxc,oxc-plugin,lint-gjs-launcher}` in the `e2e` job.

## Measured before pushing

Against the **published 0.39.0** bundle — the exact CLI the job will use — with a discriminator on each, because a lint that silently loses its rules is precisely the failure mode here:

```
gjsify lint                     exit 0   (0 errors, warnings visible)
  + a file with a bare TODO     exit 1   gjsify(todo-needs-anchor)
                                         → the JS-plugin host really loads under the
                                           published GJS bundle, rather than the Rust
                                           rule subset passing quietly on its own
gjsify format --check           exit 0   2956 files, 216ms
  + one unformatted file        exit 1   names the file
gjsify upgrade --check          exit 0   109 deps consistently declared
```

`gate`'s own coverage auditor was run locally against the edited file: `gate covers all 11 jobs`.

## Cost

One install on a docs-only PR — the tradeoff #1187 names and does not resolve. `node_modules` is cached on the same key the other six jobs share, so it is a warm no-op after the first run on a given lockfile.

## Also

- `status/open-todos.md` — the section describing this gap is deleted, per the file's header. Its incident moved one hop into the new job's comment rather than being dropped.
- `docs/lint-format.md`, `docs/ci-selective.md` — both described the old arrangement.


JumpLink:
## Live evidence, from the PR merged a few minutes ago

#1197 was a docs-only PR (`status/`, `docs/`, `.github/workflows/release.yml` — all classifier-IGNORE paths). Its GJS checks, on a tree WITHOUT this change:

```
SUCCESS  setup
SUCCESS  changes
SKIPPED  Build Fedora 44
SKIPPED  Check Fedora 44          <- lint + oxfmt --check lived here
SKIPPED  Test / Cross-runtime / macOS / E2E / Examples / Browser
SUCCESS  CI coverage summary
SUCCESS  CI gate (GJS)            <- required check, green
```

`mergeStateStatus: CLEAN`, 34 checks, 0 failing — with `gjsify lint`, `oxfmt --check` and `gjsify upgrade --check` none of them having run.

On this branch the same workflow reports `SUCCESS  Whole-tree checks` beside `setup` and `changes`.

That is the before/after on two real runs rather than a description of one.

PR #1201, opened 2026-08-15
fix(fs): close the four divergences from Node

### LINKED ISSUE #1046: fs: four pre-existing divergences from Node, measured on both legs
Surfaced while verifying #1039 (the fd-first redesign) and deliberately left out of that PR's scope. Each one was **measured** on both legs — the same probe built with `--app node` and `--app gjs` — rather than read off the source, so the "Node" column is real `node:fs` v24.15.0 and the "gjsify" column is `@gjsify/fs` at 1039's head.

| | Node v24.15.0 | `@gjsify/fs` |
|---|---|---|
| `typeof fs.mkdtemp` (the callback form) | `function` | **`undefined`** |
| `copyFileSync` of a 4755 file, mode of the copy | `755` | **`4755`** |
| `filehandle.readableWebStream()`, first `read()` | resolves (`done`) | **never settles** |
| `fs.fstat(fd)` / `fsync` / `readv` / `writev` / `cp` / `glob` / `opendir` / `statfs` with no callback | `ERR_INVALID_ARG_TYPE` | returns, then `callback is not a function` as an unhandled rejection |

Notes on each:

1. **`fs.mkdtemp` is not exported.** Only `mkdtempSync` and `fsPromises.mkdtemp` exist. The bundler already says so out loud — building anything that imports it prints `IMPORT_IS_UNDEFINED: Import 'mkdtemp' will always be undefined because there is no matching export in 'lib/esm/index.js'` — so a consumer's `fs.mkdtemp(prefix, cb)` is a `TypeError` at the call. `mkdtempAt()` (added in #1039) is the shared body; the callback wrapper is the missing piece.

2. **`copyFile` carries setuid/setgid onto the copy.** `copy(2)` semantics in Node create the destination with the source's *permission* bits masked of the special bits; ours preserves the whole mode. A copy of a setuid binary into a writable directory is a privilege-escalation shape, so this one is worth doing first.

3. **`readableWebStream()` returns a source-less `ReadableStream`.** `new Ctor()` with no underlying source: `for await (const chunk of fh.readableWebStream())` hangs forever. Node's reads the handle to EOF and closes. Better to throw `ERR_METHOD_NOT_IMPLEMENTED` than to hang, if a real implementation is not wanted yet.

4. **The nine required fd/dir callbacks accept a missing callback.** #1039 fixed the two that are *genuinely* optional in Node (`fs.close(fd)` and `writeStream.close()`, both silent there — see K-16). The rest must raise `ERR_INVALID_ARG_TYPE` and instead return normally and then reject inside a promise chain, which GJS cannot even report: it has no host hook for `unhandledRejection`, so the runner never sees it. Node's own `validateFunction(cb)` at the top of each is the shape to copy.

JumpLink:
Closes #1046. All four measured against **node v24.15.0** on both legs, and pinned as **K-18…K-21** in `fs-semantics.spec.ts`.

## 1. `copyFile` carried set-user-ID onto the copy

`Gio.File.copy` reproduces the source mode whole. Node's `copyFile` does not — copying a setuid binary into a writable directory left it setuid, which is why #1046 ranks it first.

**The mask is measured, and the obvious one is wrong:**

| src | Node dst | |
|---|---|---|
| `4755` | `0755` | set-user-ID dropped |
| `2755` | `0755` | set-group-ID dropped |
| `1755` | **`1755`** | **sticky is KEPT** — `& 0o777` would break this |
| `7755` | `1755` | only the two ID bits go |
| `0666` under umask `0027` | `0666` | the umask does not apply |
| pre-existing dst `0600`, src `0644` | `0644` | overwrite is no different |

Applied only when a bit is actually set — without them `Gio.File.copy` already lands on Node's answer. `copyFile` (callback) and `fsPromises.copyFile` both delegate to `copyFileSync`, so one site covers all three.

### NOT applied to `fs.cp` / `cpSync`

The earlier note saying `cp.ts` needs the same fix is **wrong**, and following it would have introduced a divergence. Measured on the same Node, source `4755`:

```
cpSync(file, file)   dest absent   -> 4755   KEPT
cpSync(file, file)   dest present  -> 0755
cpSync(dir, dir, {recursive:true}) -> 0755   for every file inside
```

`@gjsify/fs` already matches the first row. Reproducing the other two means teaching the walk whether it is the top-level entry or a descendant — a change to `cpOneSyncFile`'s callers, not a mask — and it is not a security shape (an existing destination means the file was already there). Full table in `status/open-todos.md`.

## 2. `fs.mkdtemp` did not exist

Only `mkdtempSync` and `fsPromises.mkdtemp` were exported, so `fs.mkdtemp(prefix, cb)` was a `TypeError` at the call. The bundler had been printing `IMPORT_IS_UNDEFINED` on every build that imported it — a warning, not a gate.

The callback form wraps `mkdtempSync`, which wraps the shared `mkdtempAt()`. No second copy of the atomic `mkdir(2)` retry loop, so the 0700 ceiling cannot drift on one path only.

## 3. `readableWebStream()` returned a stream with no underlying source

`for await (const chunk of fh.readableWebStream())` never settled. A hang is the worst shape a missing feature can take — no error, no stack, no timeout, and a consumer that looks like it is working — so this is implemented rather than made to throw.

Measured rules, all four now held:

- it streams from the handle's **current position**: after `read(Buffer.alloc(6), 0, 6, null)` on `"hello world"` Node yields `"world"`, and so does this
- a second call throws `ERR_INVALID_STATE` — *"Invalid state: The FileHandle is locked"*. The lock is taken by the **call**, not by the first read, and is not released when the stream ends
- a closed handle throws `ERR_INVALID_STATE` — *"…is closed"* — and **not** the `EBADF` that `_assertOpen` would raise
- the handle is not closed for you

One chunk per pull, so a slow consumer does not pull the whole file into memory.

`ERR_INVALID_STATE` is a plain `Error`, measured — the guess that it is a `TypeError` was wrong (`e instanceof TypeError` is `false`, `e.name` is `'Error'`).

## 4. Callback-form entry points accepted a missing callback

…and then called it inside a promise chain, so the mistake surfaced as an unhandled rejection — which **GJS cannot report at all**, having no host hook for the event. That invisibility is why it survived every previous round.

**The issue names nine. The class is twenty-four**, and the extra fifteen are the point:

The first sweep found sites by grepping for the `callback!` non-null assertion. K-20 then went red on `fs.chmod` — which has no `!`, because its type already declares the parameter required, and a type is not enforced at a JS boundary. So the audit was redone from the other side: every exported callback-form entry point in the package, checked against Node one at a time.

- `stat`, `lstat`, `readdir`, `realpath` share `parseOptsCb` → closed in **one** place
- `open`'s `switch` left its callback unassigned for `fs.open(path)`; repaired by typing it `| undefined` rather than asserting it away — a definite-assignment `!` there would have put the class straight back
- `opendir` is the one whose message says `"callback"`; the other twenty-three say `"cb"`

**Two stay unvalidated, deliberately.** `fs.close(fd)` is genuinely optional in Node and silent (K-16 pins it). `fs.rm` is not validated by Node either — it returns and then **crashes** inside `node:internal/fs/rimraf` at `callback(err)`. Matching that is a decision, not a bug fix; recorded in `status/open-todos.md`.

## Verification

```
gjsify lint                 0 errors
gjsify format --check       clean, 2956 files
gjsify upgrade --check      OK, 109 deps
gjsify workspace @gjsify/fs check   exit 0
test:node                   935 completed
test:gjs                    944 completed
verify-committed-bundles    all bundles reproduce byte-identically from source at HEAD
```

Test lists diffed before and after: **504 named tests both times, none added, none dropped** — the differing totals are `@gjsify/unit`'s per-gate accounting, not lost coverage.

`@gjsify/web-streams` is a new **devDependency** (tier 1, same as `@gjsify/fs`): K-21 needs a global `ReadableStream` and GJS exposes none. Production code still resolves the constructor off `globalThis`, so an app that never calls the method does not bundle the WHATWG implementation — the *test* is what supplies it.

`packages/infra/cli/dist/affected.gjs.mjs` is rebuilt by the pre-commit hook because `packages/node/fs/package.json` is in its closure.

## Before merging

K-18 is gated `it.failing(…, { when: !CREATE_KEEPS_SPECIAL_BITS })` — the special bits are a Linux surface, and darwin's BSD `open(2)` masks them. That gate is exactly the kind #1039 got wrong in both directions, so **the macOS and Windows legs need to run on this branch** before it lands. Dispatching both.


JumpLink:
## Self-review found the central claim wrong — force-corrected in `ee9292ca05`

An adversarial review of this PR against real `node:fs` refuted its own premise. **Node's `copyFile` masks nothing.** libuv `fchmod`s the SOURCE mode onto the destination *before* its copy loop, and what removes the bits is the kernel on the following `write(2)` — `should_remove_suid()`. That predicate has two conditions a `& ~0o6000` cannot express:

```
len 3  src 2644 -> 2644   S_ISGID SURVIVES without S_IXGRP     mask gave 0644
len 3  src 6644 -> 2644   both set: only S_ISUID goes          mask gave 0644
len 3  src 2744 -> 2744   user-exec does not count             mask gave 0744
len 3  src 2614 -> 0614   nor other-exec; S_IXGRP does         (agrees)
len 0  src 4755 -> 4755   NO WRITE, NO STRIP                   mask gave 0755
len 0  src 7755 -> 7755   same                                 mask gave 1755
```

In **every** diverging row bare `Gio.File.copy` already returned Node's answer — so the first commit turned correct behaviour into wrong behaviour in six cases while fixing one. It is the exact trap this PR's own text warned about for `cp` ("masking there would introduce a divergence rather than close one"), walked into one function over.

**K-18 could not see it**: every case wrote `'x'` and used x755 modes, so group-exec was always set and the source never empty. It now carries ten rows and content *length* is part of the case.

### Six more findings from the same review

| | |
|---|---|
| `readableWebStream()` | the comment asserted Node's is "a plain (non-BYOB) one" and built a default stream on that. Measured: `getReader({mode:'byob'})` returns a `ReadableStreamBYOBReader` on v24.15.0. A default stream **rejects** it — a divergence introduced inside the PR that exists to remove them. Now `type: 'bytes'`, answering `byobRequest`, which also drops the 64 KiB allocation per pull |
| handle closed mid-stream | ended the stream with an error; Node resolves `{done:true}`. An ordinary `close()` during iteration threw at the consumer |
| `mkdtemp(…, {encoding:'buffer'})` | could not work — `'buffer'` is not a TextDecoder label, so `decode()` threw a RangeError and left the directory created. Fixed in `mkdtempSync`, so both forms get it |
| `mkdtemp` double-callback | a throwing user callback was re-entered with its own exception as `err`. Every sibling has the same shape; recorded in `status/open-todos.md` rather than ridden in here |
| `watchFile` | missed by the callback audit. Node validates its listener, and names it `"listener"` — a third spelling |
| `describeReceived` | diverged for bigint (`10n`), for strings past 25 chars, and for strings with an apostrophe (Node swaps the quote, does not escape) |

Plus: the post-copy failure reported the **destination** in `err.path` where the copy's own throw reports the source; the `typeof cb !== 'function'` probe in `write`'s error path became guaranteed-false once `requireCallback` ran before it; and the **live counts** in `requireCallback`'s docblock are gone — AGENTS.md forbids them because they drift, and all three were already wrong.

### Now

```
test:node   958 completed
test:gjs    966 completed
gjsify lint 0 errors      format --check clean      verify-committed-bundles OK
```

macOS + Windows legs re-dispatched on the new head.

PR #1218, opened 2026-08-16
ci: run the OS suites before the merge, not after

JumpLink:
## The gap

`windows-suites.yml` and `macos-suites.yml` run on `push` to `main`. Their `pull_request` trigger was path-filtered onto their own file — **a filter satisfiable only by editing the filter**. In the `on:` block they read as PR-covered; in a PR's check list they were simply absent, which is indistinguishable from a check that passed.

Concretely: #1209 changed the CLI's install classifier, showed *all checks passed* with no Windows package suite among those checks, merged, and left the Windows leg red across eight further merges until #1217 corrected `classifyInstall`'s answer for the target platform. The defect was on the win32 leg, in a suite ADR 0018 created to catch exactly that, and the PR that introduced it could not have seen it.

## The A/B

Before this branch, on `origin/main`:

```
PR-TRIGGER-PARITY FAILURES on 2 finding(s):
  - `.github/workflows/macos-suites.yml` runs on `push` to `main` with no path filter,
    but its `pull_request` trigger has `paths: [.github/workflows/macos-suites.yml]` …
  - `.github/workflows/windows-suites.yml` … (same)
exit 1
```

After:

```
pr-trigger-parity: OK. 10 workflow(s) pushing to main offer the same coverage on pull_request.
exit 0
```

## Why the cost argument no longer applies

ADR 0018 § 5 justified the asymmetry with billing. Re-measured 2026-08-16, no premise held **for this repository**:

| Claim | Measurement |
|---|---|
| Windows bills 2x, macOS 10x | Multipliers price *billed* minutes. `gjsify/gjsify` is public (`visibility: public`) and both legs use standard images (`windows-latest`, `macos-latest`, `macos-15-intel`). |
| The runner pool is scarce | `node-gi.yml` already spawns Windows and macOS jobs on every pull request under a `paths-ignore` deny-list. This is one more job on a class the PR already occupies. |
| It would slow PRs down | A successful run of either leg finishes well inside `main.yml`'s. Neither is the critical path. |

The **required-check** half of the reasoning is untouched: `main` still carries exactly three required checks, and these two stay ADVISORY. Advisory is a statement about the merge button — it was never a reason to withhold the measurement from the PR.

## Why no path filter

Both runs END in `audit-runtimes.mjs --check`, whose subject is the whole tree. Gating a job that contains a whole-tree check on paths is the shape that made #1183 green while `main` went red. Deny-list or allow-list, a filter here would re-arm that.

## The mechanism

New repo-scoped conformance rule **`pr-trigger-parity`**: a workflow running on `push` to `main` must carry the identical path filter on `pull_request`. Identity rather than a subset test — deciding whether one glob list covers another needs a solver, and every conforming workflow here already spells both sides the same, so **the rule needs no exception list**.

It states what it does *not* check: whether a filter's globs still cover the workflow's inputs (`napi.yml` is the open instance and passes), `types:` narrowing, and required-check status.

`tests/e2e/ci-pr-trigger-parity` pins every verdict against a synthetic workflow — including the exact pre-fix shape — so the rule is known to be able to fail, plus a real-tree assertion that earns its emptiness.

## Also in here

- **`bundled-license` had no print block.** Selected by `CHECK_RULES` since #1208 and fetched in neither branch, so it printed no summary on success and reached a reader only via the accountant, which labels its findings a *reporter bug* rather than a licence drift. Fixed alongside, since this PR adds the rule next to it.
- **Docs corrected where they stated the old policy as fact**: ADR 0018 (superseded in place, re-measurement appended — not edited out of the Decision), `docs/ci-selective.md` (which already named this as its one real unclosed parity gap), `packages/node/AGENTS.md`, `@gjsify/unit`'s failure-naming rationale.
- **Two ledger entries deleted, verified against the tree**: the GTK `license: MIT` entry (all three manifests now say `SEE LICENSE IN gtk/THIRD-PARTY-NOTICES.md`, held by `bundled-license.mjs`) and the doc-revert entry (`check-doc-revert.mjs` exists and its header records that the signature the entry proposed was measured backwards — it was still publishing a wrong instruction). `priorities.md`'s point 3 restated the first of those and is gone with it.
- **The third ledger guard is measured and rejected.** `priorities.md` nominated a quote-anchor check as "the next measurement". Implemented sentence-scoped it flags 42 of 98 pairs, the sample false without exception: it cannot tell a QUOTE from a MENTION, and a ledger is mostly mentions.

## What to watch after merge

macOS **concurrency**, not minutes, is the scarce resource. If PR runs start queueing, narrow the macOS matrix on PRs (arm64 only, both arches on `main` + nightly) — do not withdraw the leg, which is the arrangement that produced the four darwin defects ADR 0018 records.


JumpLink:
Zwei Nachträge, beide vom PR selbst gefunden:

**1. Formatierung** — `oxfmt --check` fiel auf `pr-trigger-parity.mjs`. Behoben.

**2. Ein rohes NUL-Byte, das genau die Klasse dieses PRs war.** Beim Formatieren stellte sich heraus, dass mein `join(' ')` als `join('\0')` committet war — NUL statt Leerzeichen. Durchgekommen sind damit: `tsc`, `oxlint`, die 13 Fälle der eigenen e2e (ein `join` ist deterministisch, egal mit welchem Separator — beide Seiten des Vergleichs waren sich weiter einig) und `audit-runtimes --check`. `oxfmt --check` fiel zwar, aber an der Einrückung, und das Reformatieren hat das Byte **erhalten**. Einziges Symptom irgendwo: `git diff` meldete `Bin 11257 -> 11318 bytes` für eine handgeschriebene `.mjs`.

Ein Sweep über den getrackten Baum fand eine zweite, ältere und absichtliche: `packages/node/http/src/index.browser.spec.ts` hatte das Byte eingebettet, um zu prüfen dass ein Header-Wert mit NUL abgelehnt wird — zwei Zeilen unter einer Schwester-Assertion, die `\r` korrekt als Escape schreibt. Steht jetzt als `\0` (nachweislich derselbe 8-Zeichen-String) und macht die Spec nicht länger binär für git.

Neuer Schritt in `audit-runtimes.yml`: `check-source-control-bytes.mjs`. **Auf NUL verengt, und die Verengung ist gemessen statt angenommen** — ein erster Entwurf über alle C0-Bytes lieferte 37 Treffer, davon 34 legitime rohe ESC in ANSI-Farb-Fixtures (`tests/integration/{chalk,debug}`, `packages/node/util`). Die sind der Gegenstand der Assertions um sie herum. Ein Check, der auf korrektem Code feuert, ist einer, den man sich abgewöhnt.

PR #1232, opened 2026-08-21
fix(cli): resolve node-script deps from the CLI

JumpLink:
Closes the cold-tree half of the defect #1231 made diagnosable.

## The defect

The bootstrap ADR 0002 documents — `gjs -m install.mjs` → `gjsify install --immutable` → `gjsify run build:infra` — fails at its **third step** on a Node-less host, from a fresh clone. Measured on postmarketOS/aarch64 (gjs 1.88.1, musl, no node) against `3ad411530`, in a `git worktree` with no `lib/esm` anywhere:

```
gjsify run --node-script: failed to bundle …/create-gjsify/scripts/process-template.mjs
[@gjsify/create-app] gjsify run build exited with code 1
```

`node scripts/x.mjs` in a package script is re-entered as `gjsify run --node-script` (`ensureGjsifyShimOnPath()`), which bundles the script `--app gjs`. That build substitutes `node:fs` → `@gjsify/fs` and injects a register subpath per global `--globals auto` detects — and every one has to **resolve**. Anchored at the script, they resolve to the workspace's own packages, which a cold clone has not built.

**Not an ordering bug**, which is what it looks like at first: `create-app` is merely the first `node scripts/*.mjs` the chain reaches. The same failure hits `check-refs-pin.mjs`, so `gjsify workspace @gjsify/rolldown-native build:prebuilds` cannot start either. And the set is not small — `--globals auto` expands **one** detected global to **twenty-one** through the committed closure map for a one-line `export {}` entry, spanning `web-globals`, `abort-controller`, `dom-exception`, `perf_hooks`, `dom-events` and more.

## The fix

A repo build script is **toolchain, not user code**: it does not need the workspace's `@gjsify/fs`, it needs a working one. `BuildAction.bundleFileForGjsCached` — the shared helper behind node-scripts, plugin loading and config loading — now passes its own directory as `toolchainAnchorDir`, and `unresolved-workspace-import` consults it **after** normal resolution returns null.

- **Project first, always.** A workspace that has the package keeps using it; a warm tree is byte-unchanged.
- **`gjsify build` proper never sets the anchor**, so a user's missing dependency stays fatal — the guard's whole reason for existing.
- Same two-anchor shape `commands/tsc.ts` already uses for `@gjsify/tsc/bundle`, whose invariant `tests/e2e/node-free-bootstrap` states in so many words: on a Node-free host only the CLI-relative anchor can answer.

## Alternatives measured and rejected

- **Build the register closure first.** Works — I got `build:infra` green that way — but it needs twenty packages built before the first script runs, i.e. most of the polyfill tree, early, on every host including the ones that never bundle a script. It also needs a list that is bundler-determined and therefore not derivable from manifests, so it would drift the day someone adds an import to `@gjsify/fs`.
- **Drop the expanded half of the globals set.** `auto-globals.spec.ts` ("closure-map expansion vs generator bypass") pins that the expansion is a *seed* the pure iterative loop reaches anyway — an expanded global is one the injected register genuinely references. Dropping it trades a build error for a runtime `ReferenceError`.
- **A narrower globals policy for repo scripts.** Measured not to be a one-flag change: `--globals node` pulls the Node register group just as wide.

## Tests

- Five unit cases pin the fallback: resolves from the anchor, inert without one, miss when the anchor lacks it too, a throwing resolver (`ERR_PACKAGE_PATH_NOT_EXPORTED`) is a miss, and the project-first ordering.
- e2e `node-script-cold-workspace` drives the real `--node-script` path under `gjs -m dist/cli.gjs.mjs` against a planted unresolvable `@gjsify/fs`, with fake `node`/`npm`/`npx` on PATH that exit 127 and announce themselves. It asserts the script actually RAN, not merely bundled.

The fixture plants a broken package rather than reproducing a cold clone (install + a multi-minute build) because what matters is one bit — the project's copy does not resolve — and that bit is cheap to state exactly.

## Verification

Type-checks and formats clean. **Not verified end-to-end on the measuring host**: building a CLI from this source needs a fully built tree, and `node --test` needs the Node this host deliberately does not have. CI runs both suites.

Still open, recorded in `status/open-todos.md` via #1231: no CI leg executes the third bootstrap step on a tree with no `lib/esm` and `node` off PATH. This PR makes that step pass; it does not yet make it observed.


JumpLink:
## Rebased onto `main` (`24f9cef5`) + review applied

### Rebase

Two conflicts, both in the files #1236 also touched:

- **`package.json`** (`test:e2e`) — both sides appended. Took `main`'s line and inserted
  `tests/e2e/node-script-cold-workspace/run.mjs` after `tests/e2e/node-script/run.mjs`;
  `ci-workflow-rev-pin` and `gi-runtime-prologue` from `main` are both still there.
- **`status/agent-context-budget.json`** — not hand-merged. Took `main`'s value through the
  rebase, then regenerated with `--update` after the doc edits were final;
  `--check` is green.

`build.ts`, `app/gjs.ts`, `src/index.ts` and the plugin `AGENTS.md` auto-merged, and both new
plugin options survive: `giSystemProbes` (#1236) is still threaded through `processStubPlugin`,
`toolchainAnchor` through `unresolvedWorkspaceImportPlugin`. `verify-committed-bundles.mjs`
reports every committed bundle reproduces byte-identically.

### One blocker I could not reproduce — measured

The claim that the fallback is **unreachable** because `aliasPlugin` handles `node:fs` first and
its uncaught `this.resolve` rejects: **not reproduced.** I ran a probe under real GJS
(gjs 1.88.1, the `linux-x64` prebuild, `bundleWithPlugins` with a `pre` alias-shaped plugin and a
`post` guard-shaped one):

```
ALIAS saw node:fs
GUARD ran for @gjsify/definitely-not-there        <- POST_GUARD_RAN = true
GUARD nested resolve REJECTED: NotFound("@gjsify/definitely-not-there")
GUARD rescuing from toolchain
ALIAS nested resolve returned {"id":".../rescue.mjs"}
BUILD OK   CODE HAS RESCUED: true
```

The guard **does** run inside the alias's nested `this.resolve` (`skipSelf` skips only the
calling plugin), so the catch fires and the rescue reaches the alias. The negative control —
same probe, catch removed — fails as
``plugin `gjsify-alias` threw an error / Caused by: plugin `gjsify-unresolved-workspace-import`
threw an error``, i.e. exactly what the commit message describes.

**But the underlying core defect is real and is now fixed** (`716f5aca`). The bridge violates its
own declared contract — `NativePluginContext.resolve` is documented "Returns null when the
specifier doesn't resolve" — and **nine** `this.resolve`/`ctx.resolve` call sites in
`rolldown-plugin-gjsify` assume it, none with a catch, one a bare presence probe
(`Boolean(await ctx.resolve(NAPI_BARE_SPECIFIER, importer))` at `napi-node-addon.ts:820`, which
threw instead of answering `false` under GJS). Fixed JS-side in `isResolveMiss` rather than in
`session.rs`, so no Rust prebuild rebuild is needed: **only** `NotFound` becomes `null`;
`InvalidPackageConfig`, `PackagePathNotExported`, `IOError`, `MatchedAliasNotFound`, a
`BuildError` from another hook, and the two bridge-internal faults keep rejecting.

### The findings

| # | What changed |
|---|---|
| 1 | Not reproduced (above). Core fix done anyway — repairs all nine call sites. |
| 2 | Anchor is now an explicit `resolveFromToolchain` opt-in on the `opts` object; all three callers set it with their own reason. The three prose sites no longer claim "always toolchain" — they say the discriminator is "the CLI is bundling this on the project's behalf". Gating node-scripts on the CLI's own tree was tried and **does not work**: in the ADR 0002 bootstrap the CLI is a global install and the script is in the user's clone, so nothing relates them. Contained by project-first + the new warning instead, and that is written down. |
| 3 | `this.warn` on every rescue, naming package, importer and anchor; deduplicated by `resolvedCache`. Verified end-to-end under GJS, and the e2e now asserts it — "rescued" and "warm" are no longer indistinguishable. |
| 4 | Anchor path folded into the cache key, anchor **mtime** into `isBundleFresh`. |
| 5 · 12 · 15 · 17 | Catch kept (it is load-bearing) but no longer a swallow: the error is retained, the fallback is **not** consulted after a throw, and `UnresolvedWorkspaceImportError` carries it as `cause` plus one message line. Its comment now states the true reason. |
| 6 | The inert-probe test records calls and asserts the list is empty, instead of asserting inside a callback whose throw the code swallowed. |
| 7 · 14 · 19 | Resolved by construction: `createRequire` is gone. The anchor is a file path used as the **importer of a second `this.resolve`**, so the rescue uses the target's own `conditionNames` — no `require`-branch `cjs-compat.cjs` can reach a GJS bundle, and there is no second resolver with a dead `null` arm to test. |
| 8 | One `fileURLToPath`, inside a catch: a non-`file:` CLI origin loses the fallback rather than breaking config loading. |
| 9 · 10 · 16 | The mock context is importer-aware and the anchor answers with a **distinct id**, so order changes the answer. New cases: project wins while the anchor also hits; rescue + cache + warning; rejecting resolver -> no rescue, `cause` preserved; the anchor probe declines itself. |
| 11 | Skip guard gone. `gjs --version` and the CLI bundle are hard asserts inside the test; only the non-Linux leg skips. |
| 13 | Header now says the fake `node` is a belt against a future regression, not proof — `runNodeScript` branches on `hostRuntime()`, not PATH. |
| 18 | The count lives in exactly one place (`AGENTS.md`); the code comments point at it. |

### Mutation testing (the point of the above)

| mutation | before | after |
|---|---|---|
| delete the fallback block | unit **green** | unit **2 fail** |
| consult the anchor before the project | unit green, e2e green | unit **4 fail** |
| rescue after a caught resolver error | no coverage | unit **1 fail** |
| revert the bridge fix | — | e2e **1 fail** |
| delete the fallback block | e2e red | e2e **1 fail** |

### Measured

```
gjsify workspace @gjsify/cli run test            2407 completed, exit 0
node --test tests/e2e/node-script-cold-workspace  1/1
node --test tests/e2e/unresolved-workspace-import 6/6
node --test tests/e2e/node-script                13/13
node --test tests/e2e/run-command                16/16
node --test tests/e2e/build-cache                12/12
node --test tests/e2e/cli-config                  2/2
node --test tests/e2e/ci-classifier-output-coverage 8/8
node scripts/verify-committed-bundles.mjs        byte-identical
node scripts/check-agent-context-size.mjs --check green
gjsify tsc --noEmit (cli, rolldown-plugin-gjsify, rolldown-native)  clean
gjsify format --check / gjsify lint              clean on the changed files
```

Not run here: the full `test:e2e` matrix and the non-Linux legs — CI covers those.

JumpLink:
## The three QEMU legs were red on a gate defect, not on this PR's code

`fead025f` fixes it here rather than in a separate PR, because this PR's own run is the proof.

### What was measured

All three emulated legs failed at **"Gate on the glibc floor of what this leg just built"** while their build step was **green**:

| | main, job [96741121999](https://github.com/gjsify/gjsify/actions/runs/32484927927) (ppc64) | this PR, job [96779305698](https://github.com/gjsify/gjsify/actions/runs/32484927927/job/96779305698) (ppc64) |
|---|---|---|
| `PREBUILD_SKIP` | `[]` | all ten packages |
| collect steps | all ran | all `SKIP … — unchanged` |
| eight upload steps | success | **skipped** |
| gate | success | **failure** |

This PR edits TypeScript under `packages/infra/rolldown-native/src/ts/`. That matches the workflow's `packages/infra/rolldown-native/src/**` trigger, so the run starts — but it changes no native source, so the `changes` job correctly decided to build nothing the emulated legs produce, and the uploads correctly skipped with it. The gate then read the resulting zero as *"the collect steps did not run"*, which is the one zero it was written to catch.

Two correct mechanisms contradicting each other. `rolldown-native` **was** in the build set; no emulated leg builds it (it has no `-linux-ppc64`/`-s390x`/`-riscv64` platform package), so those legs had a candidate set of eight, all skipped.

### Which layer, and why not the trigger

Fixed in **`scripts/check-staged-prebuild-libc.mjs`**, not in `on: paths:`.

Narrowing the trigger would not have fixed it. Any PR touching only `oxfmt-native`'s or `rolldown-native`'s **Rust** sources reproduces this exactly — both are genuine prebuild inputs that legitimately start the workflow, and neither is built on an emulated leg. The trigger decides how *often* the shape occurs; only the gate can decide what it *means*. Fixing the trigger alone would have hidden this instance and left the defect.

### The fix — the gate is told what the leg was supposed to build

Both gate steps now pass `PREBUILD_TARGET` (`linux-${{ matrix.arch }}`) and `PREBUILD_REPORT` (`needs.changes.outputs.report`, the classifier's own per-package decision — the same output `prebuilds-summary` renders, so nothing re-derives *which packages this workflow gates*).

Three rules, none of them "pass when in doubt":

1. **Every freshly staged directory is measured**, skip list or not — a partial skip is never a blanket exemption.
2. A candidate for this leg's target that staged nothing and is **not** `build: false` is **neither built nor skipped**, and is an error even when other packages on the same leg measured fine.
3. **Zero passes only when every candidate is accounted for as skipped**, and it names each one. Without the two values the gate cannot explain a zero, so the zero stays an error — the fail-closed direction, which is also what a broken wire degrades to.

### Verified locally, both shapes, against this tree

Same input, old gate vs new — the accounting comes from the real classifier run on this PR's own file list, which reproduces CI's `PREBUILD_SKIP` exactly:

```
# 1. all candidates skipped (ppc64 leg) — must pass, and must SAY it skipped
check-staged-prebuild-libc: 0 targets measured for linux-ppc64 — every one of the 8 package(s)
this leg builds was SKIPPED AS UNCHANGED by the `changes` job, so no fresh bytes exist to measure:
  skipped  @gjsify/http-soup-bridge-linux-ppc64  (no change under packages/node/http-soup-bridge/)
  skipped  @gjsify/http2-native-linux-ppc64      (no change under packages/node/http2-native/)
  skipped  @gjsify/lightningcss-native-linux-ppc64 (no change under packages/infra/lightningcss-native/)
  skipped  @gjsify/sab-native-linux-ppc64        (no change under packages/node/sab-native/)
  skipped  @gjsify/terminal-native-linux-ppc64   (no change under packages/node/terminal-native/)
  skipped  @gjsify/tls-native-linux-ppc64        (no change under packages/node/tls-native/)
  skipped  @gjsify/webgl-linux-ppc64             (no change under packages/framework/webgl/)
  skipped  @gjsify/webrtc-native-linux-ppc64     (no change under packages/web/webrtc-native/)
  This is the accounted-for zero. The other one — collect steps that were supposed to run
  and did not — is still a hard error […]
exit=0     (old gate on the identical input: exit=1)

# 2. same leg, everything was to be BUILT, nothing staged — must still fail
check-staged-prebuild-libc: 8 package(s) this leg was told to BUILD staged nothing for linux-ppc64.
  Neither built nor skipped is the state that must never pass […]
exit=1
```

Eight candidates for `linux-ppc64` — exactly the eight upload steps the QEMU leg carries. `oxfmt-native`, `rolldown-native` and `webkit-native` are correctly not candidates there.

### Mechanism

`tests/e2e/staged-prebuild-libc-gate` (11 cases, ~1 s) drives the **real** script through both halves of every shape: all-skipped passes *and* says so, told-to-build-staged-nothing fails, a partial skip still measures the built one, another leg's target is never a finding, and no target / `[]` / malformed report all fail closed. One case holds `prebuilds.yml` to passing both keys in every leg that runs the gate, derived from the workflow. And a **discriminator**: it stages a real committed artifact with a deliberately false glibc floor, so the passing cases cannot be satisfied by a gate that measures nothing.

### Note on this PR's run

Editing `prebuilds.yml` is a shared input, so `changes` will now classify **every** package as build and the emulated legs will do a full rebuild (riscv64 ≈ 96 min). That exercises the *built* path end to end; the *all-skipped* path is covered by the e2e suite and by the local run above, and the next PR that touches no prebuild input will exercise it in CI.

### One finding left open (not fixed here)

`prebuilds.yml`'s `on: paths:` does not list the scripts it runs: `scripts/check-staged-prebuild-libc.mjs`, `scripts/audit-runtimes.mjs`, `scripts/audit-test-scripts.mjs`, `scripts/check-ci-image-packages.mjs`, `scripts/clear-committed-platform-exemptions.mjs`, `scripts/clear-satisfied-gir-gaps.mjs`. That is the same shape as the incident recorded at the `tests/e2e/…/**` entries — editing a gate's content does not run the workflow that runs it. Adding all six would make each a shared input (a full emulated rebuild per edit), so it is a cost decision worth taking on its own rather than smuggling into this PR.

PR #1234, opened 2026-08-21
ci: gate what the workflows actually build

JumpLink:
## What changed

**`napi.yml`'s trigger becomes a deny-list.** It guarded `packages/napi` with a
four-path allow-list (`packages/napi/**`, `packages/node-gi/**`,
`packages/infra/rolldown-plugin-gjsify/**`, the workflow itself) while its jobs read
most of the tree: `gjs -m install.mjs`, `gjsify foreach build:gjsify` over the whole
workspace, a `@gjsify/create-app` build, the published-CLI version out of
`packages/infra/cli/package.json`, two caches keyed on `gjsify-lock.json`, and
`.github/scripts/install-published-cli.sh`. `packages/napi` is not a workspace member,
so `main.yml` neither builds nor tests it — this file is the only CI the whole N-API
pillar has, and an allow-list fails *silent*: an input nobody listed produces no run,
and no run reads as no problem. Same class as #1028 and #1173. It now ignores
`**/*.md`, `docs/**` and `website/**` and nothing else, and carries the nightly
`node-gi.yml` already has (a Fedora or nodejs.org update can red this pillar with no
commit in the repository at all).

`refs/**` is the one entry deliberately *not* copied across from `node-gi.yml`, and as
accepted over-triggering rather than a dependency — see below.

**`NODE_TEST_REV` is now held to the `refs/node` gitlink by a new conformance rule.**
The workflow said the two "MUST match" and nothing checked it; they did not match. The
new `workflow-rev-pin` rule pairs a named workflow `env:` with a `refs/` gitlink,
fails when they disagree, and is *selected* by `audit-runtimes --check`, so it runs on
every PR — which is where the submodule sweep that caused the drift lands. It reads the
gitlink out of `.git/index` **as a file** (`scripts/manifest-conformance/git-index.mjs`),
never through `git ls-files`; the reason is a Windows leg with a checkout and no git
binary, measured below.

**`openssl` joins the CI image**, retiring the sole recorded blocker on
`@gjsify/integration-tls-session`. The suite is *not* re-added to `main.yml`'s
`--include` in the same commit; that is mechanical, not slack (below).

**Four claims made false by the trigger change move with it**: the
`pr-trigger-parity` docblock, `docs/ci-selective.md`, `status/sections/priorities.md`
and a parenthetical in `release.yml` all described `napi.yml` as the un-covered
allow-list. The open instance of "nothing checks that a filter's globs cover what the
workflow guards" is now `deploy-docs.yml`. A fifth (`node-gi.yml`'s `scope` comment
citing a superseded ADR sentence) is corrected in its own commit.

## Measured

Every claim above, with the command that produced it.

**The jobs read far more than the four allow-listed paths**

```
$ grep -nE -e 'create-app|infra/cli/package.json|install-published-cli' \
           -e 'install.mjs|foreach build:gjsify|gjsify-lock.json' \
           .github/workflows/napi.yml
307:          key: node-modules-v1-${{ hashFiles('gjsify-lock.json') }}
332:          key: napi-libs-v2-${{ hashFiles('gjsify-lock.json') }}-${{ hashFiles('packages/*/*/src…
336:          B=$(gjs -m install.mjs --fetch-only)
344:          CLI_VERSION="$(node -p "require('./packages/infra/cli/package.json').version")"
356:          "${GITHUB_WORKSPACE}/.github/scripts/install-published-cli.sh" "${CLI_VERSION}"
383:          gjsify foreach build:gjsify --exclude '@gjsify/tsc'
509:          gjsify workspace @gjsify/create-app build
624:          key: node-modules-v1-${{ hashFiles('gjsify-lock.json') }}
649:          key: napi-libs-v2-${{ hashFiles('gjsify-lock.json') }}-${{ hashFiles('packages/*/*/src…
653:          B=$(gjs -m install.mjs --fetch-only)
661:          CLI_VERSION="$(node -p "require('./packages/infra/cli/package.json').version")"
673:          "${GITHUB_WORKSPACE}/.github/scripts/install-published-cli.sh" "${CLI_VERSION}"
699:          gjsify foreach build:gjsify --exclude '@gjsify/tsc'
```

(comment lines and the workflow's own header trimmed from the paste.) Hits in three of
the five jobs; none of those paths could trigger the workflow.

**`packages/infra/{tsc,resolve-npm}` are NOT inputs** — a claim the previous revision
of this branch made and a reviewer refuted; re-measured and dropped:

```
$ sed -n '383p' .github/workflows/napi.yml
          gjsify foreach build:gjsify --exclude '@gjsify/tsc'
$ node -p "Object.keys(require('./packages/infra/resolve-npm/package.json').scripts).join(',')"
clear,check,test,build          # no build:gjsify, so the foreach skips it entirely
```

The `!packages/infra/tsc/lib` / `!packages/infra/resolve-npm/lib` carve-outs in the
napi-libs cache key are evidence of the same thing: this job never produces those
trees. The comment now says that instead.

**`refs/**` cannot change a run** — the second refuted claim, re-measured:

```
$ sed -n '153,159p' .github/workflows/napi.yml
          rm -rf refs/node && mkdir -p refs/node && cd refs/node
          …
          git fetch -q --depth 1 --filter=blob:none origin "$NODE_TEST_REV"
```

The job wipes the submodule and clones nodejs/node at the env pin, so a gitlink bump
produces a byte-identical run. `refs/**` stays off the deny-list because
`paths-ignore` has no negation (all-or-nothing) and erring towards *running* is the
only direction a guard may fail in.

**The pin and the gitlink had diverged**

```
$ git ls-tree HEAD refs/node
160000 commit 0618e9f0fc7606f7a7981b6a0bf28ec1606ae2b3    refs/node
$ grep -n NODE_TEST_REV .github/workflows/napi.yml    # before this PR
103:  NODE_TEST_REV: ed63b195cd61367db7b611cb601352197569ee0e
$ grep -rn NODE_TEST_REV --include='*.mjs' --include='*.md' .    # what enforces it
(nothing)
```

Both paths into the same oracle read a different one:
`packages/napi/napi/scripts/conformance.mjs` copies its addon sources out of
`refs/node/test/{js-native-api,node-api}` (so `test:conformance:update` regenerates
goldens from the *gitlink*), while CI clones nodejs/node at the *env*.

Why the drift was invisible, and why the pin could move safely:

```
$ for s in ed63b195… 0618e9f0…; do
    t=$(gh api repos/nodejs/node/git/trees/$s --jq '.tree[]|select(.path=="test")|.sha')
    gh api repos/nodejs/node/git/trees/$t \
      --jq '.tree[]|select(.path=="js-native-api" or .path=="node-api")|.path+" "+.sha'
  done
js-native-api 6b6d6ab43e5e6b7ff6bd8acdc0b33e9646e8bd31
node-api      62cdfca1a92fab04d26dc8869ef4aaae90923576     # identical for BOTH revisions
```

So `NODE_TEST_REV` moves to the gitlink with no change to what CI fetches. The
divergence was never a decision:

```
$ git log -p -- refs/node | head -20
commit 4752e94a  chore: release v0.23.0
-Subproject commit ed63b195…
+Subproject commit 0618e9f0…
$ gh api repos/nodejs/node/compare/ed63b195…0618e9f0… --jq '{ahead:.ahead_by}'
{"ahead":109}
```

**The new rule is not decorative**

```
$ node scripts/audit-runtimes.mjs --check --strict ; echo $?
workflow-rev-pin: OK. 1 workflow revision pin(s) name the commit the matching `refs/` gitlink does.
0

# same command with the env reverted to ed63b195:
$ node scripts/audit-runtimes.mjs --check --strict ; echo $?
  - [workflow-rev-pin] `.github/workflows/napi.yml`'s `NODE_TEST_REV` and the `refs/node`
    gitlink name DIFFERENT upstream commits:
      workflow env: ed63b195…
      gitlink:      0618e9f0…
1
```

It is a *new* rule rather than an extension of `refs-pin`, which owns the same
directory, because `refs-pin` compares the gitlink to a working COPY and is therefore
registered-but-not-selected by `--check` (`scripts/audit-runtimes.mjs`, `CHECK_RULES`
docblock). A check that never runs on a PR would not have caught this.

**Which leg proves the comparison**: `audit-runtimes.yml` → `Detect runtime-triplet
drift`, `runs-on: ubuntu-latest`, triggered on bare `pull_request:` and `push: [main]`
with **no** `paths` filter, running `node scripts/audit-runtimes.mjs --check --strict`.
It printed `workflow-rev-pin: OK. 1 workflow revision pin(s) …` on this PR (job
96718292411). `windows-suites.yml` and `macos-suites.yml` run the same `--check` as
their last step, so the rule now executes on three legs, not one.

**The rule reads `.git/index`, and does NOT skip when `git` is missing.** The first
revision of this branch shelled out and reddened `Node-pillar suites (win32-x64 /
cmd.exe)`:

```
[workflow-rev-pin] threw while running: Error: spawnSync git ENOENT
  at gitlinkSha (…/manifest-conformance/rules/workflow-rev-pin.mjs:65:17)
```

`windows-suites.yml` deletes every `\Git\` entry from PATH on purpose (npm dispatches
package scripts through `%COMSPEC%`, so leaving Git for Windows' `chmod`/`cp`/`rm`/`sed`
reachable would turn cmd.exe-only failures green — the measurement that leg exists for),
and that takes `git.exe` with it. The job still holds a **complete** checkout; only the
binary is gone.

So the two options were: (a) stop selecting the rule where git is absent, or (b) stop
needing git. **(b)**, because (a) is not actually available and would be wrong if it
were — there is one `CHECK_RULES` list shared by every leg, so "not selected here" would
have to become a capability probe, and a rule that returns *not applicable* wherever the
binary is missing goes green on that leg forever while comparing nothing. That is the
same shape as the `refs-pin` problem this rule was created to avoid, one level down.
`scripts/manifest-conformance/git-index.mjs` parses index versions 2/3/4, follows a
worktree `.git` **file**, reads `extensions.objectFormat` for the hash width, and
**refuses** a split index by name instead of reporting the submodule as absent. It is
also what the repo asks for: shelling out where a file read answers the question is a
documented anti-pattern here. `refs-pin` keeps its subprocess — it asks the working copy
things no index can answer (`rev-parse HEAD`, `describe --tags`) — and is not selected
by `--check`, so no PR leg can hit the missing binary through it.

Proof that the binary-free reading is faithful, and that the rule still bites without it:

```
$ diff <(git ls-files -s | awk '$1=="160000"{print $2" "$4}' | sort) <(node -e '…readIndexGitlinks…')
                                            # 93 gitlinks, byte-identical
$ GIT_INDEX_FILE=$T/.git/index git update-index --index-version 4 && <same diff>
                                            # identical again on a real v4 index
$ GIT_INDEX_FILE=$T/.git/index git update-index --split-index && <read>
THREW: … is a SPLIT index (core.splitIndex), whose entries live in a shared index this
       reader does not follow. Run `git update-index --no-split-index`, or teach …

$ PATH=<only a node symlink>  node scripts/audit-runtimes.mjs --check ; echo $?
NO GIT ON PATH - good
workflow-rev-pin: OK. 1 workflow revision pin(s) name the commit the matching `refs/` gitlink does.
0

# same git-less PATH, with NODE_TEST_REV reverted to ed63b195:
$ PATH=<only a node symlink>  node scripts/audit-runtimes.mjs --check ; echo $?
audit-runtimes --check: DRIFT DETECTED.

WORKFLOW-REV-PIN FAILURES on 1 finding(s):
  - `.github/workflows/napi.yml`'s `NODE_TEST_REV` and the `refs/node` gitlink name
    DIFFERENT upstream commits:
    workflow env: ed63b195cd61367db7b611cb601352197569ee0e
    gitlink:      0618e9f0fc7606f7a7981b6a0bf28ec1606ae2b3
    …reads the gitlink locally and the env in CI, so the same oracle runs against two
    upstreams and only notices on the day their files differ…
1
$ grep -c 'UNREPORTED FINDING' <that output>
0
```

**The second half of the same defect: the rule had no print block.** Its summary was
fetched into the `run.ok` branch, and the failure branch was never written — so its very
first finding reached the log through the accountant, labelled *"this is a REPORTER bug,
not a new drift"*. Fixed as that message instructs: a `WORKFLOW-REV-PIN FAILURES` block
next to the others, and the id added to `REPORTED_RULE_IDS` **together with** it. The
`grep -c` above is the check that it took.

**The macOS lever this file actually has**

```
$ python3 -c "import yaml; d=yaml.safe_load(open('.github/workflows/napi.yml')); \
  print([(k, v.get('runs-on'), list(v.get('strategy',{}))) for k,v in d['jobs'].items()])"
[('build-test','ubuntu-latest',[]), ('consumer','ubuntu-latest',[]),
 ('nodegi-shim','ubuntu-latest',[]), ('macos','macos-latest',[]), ('windows','windows-latest',[])]
```

One macOS job, no matrix — so ADR 0018 § 5's "narrow the macOS MATRIX" fallback (which
is about `macos-suites.yml`'s two-arch matrix) has nothing to narrow here. The comment
now names what is left to trade: the gate set that job runs.

**The other three CI-incompatible preconditions still hold** (the past-tense edit was
wrong and is reverted to present tense):

```
$ sed -n '40p;132p;188p' status/integration-coverage.md | cut -c1-150
**HELD OUT of CI** … external precondition — `ghcr.io/gjsify/ci-fedora` ships neither podman nor doc…
**HELD OUT of CI** … not selectable at all: the root manifest excludes `tests/integration/nativescri…
**HELD OUT of CI** … external precondition — `Cannot find module '../../../build/Release/node_datach…
```

Only `tls-session`'s is retired.

**The `openssl` absence, and why the suite waits one merge**

```
$ podman run --rm fedora:44 sh -c 'rpm -q openssl openssl-libs; command -v openssl'
package openssl is not installed
openssl-libs-3.5.7-2.fc44.x86_64
(nothing on PATH)
$ node scripts/check-ci-image-packages.mjs ; echo $?
0        # image → jobs: no job under .github/ dnf-installs openssl, so baking it breaks nothing
```

`build-ci-image.yml` gates both its GHCR login and `push:` on
`github.event_name != 'pull_request'`, while `main.yml`'s `integration` job resolves
the mutable `ghcr.io/gjsify/ci-fedora:<major>` tag at job start. A commit doing both
would run the suite against the image from *before* the package: red on its own PR,
then red on main. `dash` hit the same wall in cc775993. The suite returns in the first
commit after `build-ci-image` republishes both majors — one cause per commit.

**Machinery**

```
$ node scripts/audit-runtimes.mjs --check --strict          → exit 0
$ node --test tests/e2e/ci-pr-trigger-parity/run.mjs        → 13/13
$ node --test tests/e2e/ci-workflow-rev-pin/run.mjs         → 14/14
$ node scripts/check-ci-image-packages.mjs                  → exit 0
$ python3 -c "import yaml,glob; [yaml.safe_load(open(f)) for f in glob.glob('.github/workflows/*.yml')]"
                                                            → all 14 parse
```

## Deliberately left out

- **`@gjsify/integration-tls-session` is not re-added to `main.yml`'s `--include`.**
  Measured above: it would be red on this PR and red again on main. It is one commit
  after the image republishes, not a follow-up someone has to remember — the note sits
  in `status/integration-coverage.md`, `status/open-todos.md` and the `integration`
  job's own comment.
- **`deploy-docs.yml` keeps its allow-list.** It is the open coverage instance and it
  is named as such in three places now, but a deny-list there re-triggers a Pages
  DEPLOY rather than just a check, so it is a separate decision with a separate cost.
- **No `packages/napi` code changed.** This PR is entirely about what CI runs and what
  the tree claims about it.
- **The `refs/node` gitlink is untouched.** Aligning the two by moving the env pin
  forward is provably neutral (identical suite trees); moving the gitlink backwards
  would have been a submodule rewrite with no measurement behind it.


JumpLink:
Two things this PR's own CI run measures that could not be measured before pushing it:

1. **The deny-list works.** This PR touches no file under `packages/napi/**`, `packages/node-gi/**` or `packages/infra/rolldown-plugin-gjsify/**` — under the old allow-list `napi.yml` would not have run at all. It did: `Build & gates & conformance (GJS / Fedora 44)`, `better-sqlite3 consumer gate`, `node-gi N-API fidelity under shim` and `macOS build + gates (GJS / macos-latest arm64)` are all in this PR's check list.

2. **The new rule runs on a PR, not just locally.**
```
$ gh run view <audit run> --log | grep workflow-rev-pin
Detect runtime-triplet drift  …  workflow-rev-pin: OK. 1 workflow revision pin(s) name the commit the matching `refs/` gitlink does.
Manifest checks (Windows)     …  workflow-rev-pin: OK. 1 workflow revision pin(s) name the commit the matching `refs/` gitlink does.
```
Both legs of `audit-runtimes.yml`, which has no `paths` filter — so a submodule sweep cannot land without passing it.

`Lint commit messages`, `Detect runtime-triplet drift`, `Manifest checks (Windows)`, `changes`, `Detect affected prebuild packages`, the two musl prebuild legs and `Build & push Fedora 44 image` (which builds the Dockerfile carrying the new `openssl`) are green; the long N-API and main legs are still running.

JumpLink:
## win32 failure, diagnosed and fixed (force-pushed, rebased onto `main`)

`Node-pillar suites (win32-x64 / cmd.exe)` was red on **two** defects, both introduced by
this branch. Both are fixed in the amended `ci: hold NODE_TEST_REV to the refs/node pin`.

### 1. The rule shelled out to `git`, and that leg has no git binary

```
UNREPORTED FINDING(S) — this is a REPORTER bug, not a new drift:
  - [workflow-rev-pin] threw while running: Error: spawnSync git ENOENT
      at gitlinkSha (…/manifest-conformance/rules/workflow-rev-pin.mjs:65:17)
```

`windows-suites.yml` strips every `\Git\` entry from PATH before anything runs — on
purpose, because npm dispatches package scripts through `%COMSPEC%` and a leg that kept
Git for Windows' `chmod`/`cp`/`rm`/`sed` reachable would turn cmd.exe-only failures green.
That strip takes `git.exe` with it. The job still has a **complete checkout**; what it
lacks is the binary — and `audit-runtimes.mjs --check` is its last step.

**The obvious fix — catch the ENOENT, return "not applicable" — is not the fix.** A rule
that no-ops wherever git is absent goes green on that leg forever while comparing
nothing, which is the same defect class this rule was written to end. Nor was
"deselect it there" available: `CHECK_RULES` is one list shared by every leg, so it would
have had to become a capability probe with the same outcome.

So the rule **stopped needing git**. New `scripts/manifest-conformance/git-index.mjs`
reads `.git/index` as a file: index versions 2/3/4, a worktree `.git` *file* pointer,
`extensions.objectFormat` for the hash width, and a split index **refused by name**
rather than silently reported as a missing submodule. `refs-pin` keeps its subprocess —
it asks the working copy things no index can answer (`rev-parse HEAD`, `describe`) — and
is registered-but-not-selected, so no PR leg reaches it.

Faithfulness, against real git:

```
$ diff <(git ls-files -s | awk '$1=="160000"{print $2" "$4}' | sort) <(readIndexGitlinks)
                     # 93 gitlinks, byte-identical — live v2 index
$ GIT_INDEX_FILE=$T/.git/index git update-index --index-version 4   # → same diff, identical
$ GIT_INDEX_FILE=$T/.git/index git update-index --split-index       # → refused by name
```

### 2. The rule had no print block

The harness said so itself, and it was right: the summary had been fetched into the
`run.ok` branch and the failure branch was never written, so the rule's first real
finding arrived labelled *"this is a REPORTER bug"*. Fixed exactly as that message
instructs — a `WORKFLOW-REV-PIN FAILURES` block next to the others, with the id added to
`REPORTED_RULE_IDS` **together with** it.

### Which leg proves the comparison

`audit-runtimes.yml` → **`Detect runtime-triplet drift`**, `runs-on: ubuntu-latest`, bare
`pull_request:` + `push: [main]`, **no `paths` filter**, step
`node scripts/audit-runtimes.mjs --check --strict`. It already printed
`workflow-rev-pin: OK. 1 workflow revision pin(s) …` on this PR (job 96718292411).
`windows-suites.yml` and `macos-suites.yml` run the same `--check` as their last step, so
after this fix the rule executes on three legs instead of dying on one.

### Proof it still bites — run with **no git on PATH**, i.e. the win32 environment

```
$ PATH=<a dir holding only a `node` symlink>  node scripts/audit-runtimes.mjs --check
NO GIT ON PATH - good
workflow-rev-pin: OK. 1 workflow revision pin(s) name the commit the matching `refs/` gitlink does.
EXIT=0

# same PATH, NODE_TEST_REV temporarily reverted to ed63b195 (then restored):
audit-runtimes --check: DRIFT DETECTED.

WORKFLOW-REV-PIN FAILURES on 1 finding(s):
  - `.github/workflows/napi.yml`'s `NODE_TEST_REV` and the `refs/node` gitlink name DIFFERENT upstream commits:
    workflow env: ed63b195cd61367db7b611cb601352197569ee0e
    gitlink:      0618e9f0fc7606f7a7981b6a0bf28ec1606ae2b3
    `packages/napi/napi/scripts/conformance.mjs` (addon sources → golden diff) reads the gitlink locally and the env in CI, so the same oracle runs against two
    upstreams and only notices on the day their files differ. Point both at the same commit — and when
    the bump is deliberate, regenerate whatever the old revision produced in that same change.

A workflow `env:` that hard-codes an upstream revision and the matching `refs/` gitlink are two names for ONE commit, …

EXIT=1
$ grep -c 'UNREPORTED FINDING' <that output>
0
```

`EXIT=1`, named under its **own** header, and the accountant is silent — which is the
half that was broken before.

### Also in this push

- Rebased onto `main` (#1233 and #1235 landed); the `status/` edits merged without
  conflict.
- `tests/e2e/ci-workflow-rev-pin/run.mjs` → **14/14** (was 7/7): the new cases drive the
  index reader against hand-built v2/v3/v4 buffers, a worktree `.git` file, a `link`
  extension, a non-`DIRC` file, a version this reader does not implement, and a directory
  with no checkout at all — where the rule must produce a **finding**, never a quiet pass.
- Fixed a broken markdown table in `scripts/manifest-conformance/README.md` (the
  `refs-pin` vs `workflow-rev-pin` paragraph had been inserted between two table rows).
- `node scripts/audit-runtimes.mjs --check --strict` → 0 · `oxfmt --check` clean ·
  `oxlint` clean (only the two pre-existing warnings in `audit-runtimes.mjs`) · all 14
  workflow YAMLs parse · `check-comment-budget.mjs` unchanged-or-better on both touched
  trees.

JumpLink:
**Confirmed on CI.** `Node-pillar suites (win32-x64 / cmd.exe)` → **pass** (13m21s,
[job 96729605575](https://github.com/gjsify/gjsify/actions/runs/32468318552/job/96729605575)),
and its `Conformance audit (os-axis included)` step printed:

```
09:45:07  workflow-rev-pin: OK. 1 workflow revision pin(s) name the commit the matching `refs/` gitlink does.
```

That line is the point: it comes from a host with a full checkout and **no `git` on PATH**,
so the rule really compared the pin there rather than skipping. `Detect runtime-triplet
drift` (ubuntu, every PR, no `paths` filter) prints the same line, and `macos-suites.yml`
runs the same `--check`.

One unrelated `spawnSync git ENOENT` remains in that log at 09:34:16 — it is
`scripts/install-git-hooks.mjs` during `gjsify install`, pre-existing on `main`, and it
already degrades to `[gjsify install] … exited 1 — git hooks may not be active.` rather
than failing the job. Out of scope here; noted so it is not mistaken for a leftover of
this one.

PR #1237, opened 2026-08-21
feat(cli): add gjsify dev, a watch-rebuild loop

JumpLink:
## What this adds

`gjsify dev` — the inner development loop: watch, rebuild, relaunch. It exists
because `gjsify build --watch` cannot serve the host it matters most for:
that flag drives rolldown's watcher API, which only the npm engine exposes
(`bundler-pick.ts` refuses under `@gjsify/rolldown-native`), so on a Node-free
GJS host the loop was manual — edit, Ctrl-C, rebuild, relaunch.

`gjsify storybook --watch` already had the way around it: `fs.watch` plus an
in-process re-entry of the ordinary build command, which asks the bundler for no
watcher API at all. That loop is lifted into `utils/watch-loop.ts` and `dev` is
its second caller — one implementation, two callers, no drifting copy.
`run.ts`'s tokenizer moves to `utils/simple-command.ts` for the same reason.

WHAT to build is not re-declared: `utils/dev-plan.ts` reads the project's own
`build:gjs` / `build:node` script and puts the CLI flags on top, so the dev loop
and `gjsify run build` cannot drift into different bundles. All seven templates'
`dev` script becomes `gjsify dev --runtime gjs`.

## The part that took the work: the GJS host

A first pass of this PR was Node-green and completely broken on GJS — the CLI
printed its two lines, launched the app and exited 0, orphaning it. Three
distinct defects, none of which a Node-only suite can fail for:

| | |
|---|---|
| **nothing holds a GJS process open** | An `fs.watch` handle is a Gio file monitor, not a Node handle count, and the supervised spawn's `ensureMainLoop()` declines at `GLib.main_depth() !== 0` — which is where a supervisor always reaches it, because GJS spins the default context itself while the entry module's top-level await is pending. Fixed at the core: `holdMainLoop()` in `@gjsify/utils`. |
| **a watch loop's resolution is not completion** | `gjsify run`'s in-process fast path `process.exit()`ed on it, so `gjsify run dev` killed the loop it had just started — and that is the documented entry point (`gjsify create` prints it, `getting-started.mdx` documents it). The loop now declares itself through `utils/daemon-command.ts`. |
| **`process.on('SIGINT')` was a listener nothing emitted** | Registering worked, `listenerCount` counted, and the kernel's default disposition ran instead of the handler. Fixed at the core: `@gjsify/process` arms a `GLibUnix.signal_add()` source on the first listener (SIGHUP/SIGINT/SIGTERM only — the rest are renumbered on some architectures). |

Plus one host-independent defect: the watch dir defaults to the entry point's
directory, so a project whose entry sits at its root watches the root and the
build writes `dist/` **into** it. One edit then rebuilt every ~5 s forever.
`isSelfWrite()` excludes the resolved output and its directory — the latter only
when it is not the watched dir itself, where excluding it would exclude the
sources too.

`tests/e2e/dev-command` now drives **both** hosts, which is the whole point: a
Node-only suite would have reproduced, in the replacement, exactly the gap it was
written to close.

## Measured

Fedora, node 24.19.0, **gjs 1.88.1** (the previous body said 1.86 — that was
wrong, and it is corrected in the commit).

The branch was rebased onto `main` after #1236/#1234 landed, and everything below
was re-run on the rebased tree — including a full `--with-dependencies` rebuild,
so nothing here is a number from the old base.

```
gjsify workspace @gjsify/cli run check              exit 0
gjsify workspace @gjsify/utils run check            exit 0
gjsify workspace @gjsify/process run check          exit 0
gjsify workspace @gjsify/cli run test               2377 completed, 0 failed
gjsify workspace @gjsify/utils run test             255 completed, 0 failed (gjs)
gjsify workspace @gjsify/process run test           171 completed, 0 failed (node + gjs)
node --test tests/e2e/dev-command/run.mjs           7/7  (5 node rows, 2 gjs rows)
node --test tests/e2e/storybook-on-node/run.mjs     2/2
node --test tests/e2e/storybook-no-stories/run.mjs  1/1
node --test tests/e2e/storybook-declaration/run.mjs 8/8
node --test tests/e2e/build-watch/run.mjs           2/2
node --test tests/e2e/run-command/run.mjs           16/16
node --test tests/e2e/spawn-gjs-teardown/run.mjs    3/3
node --test tests/e2e/gi-runtime-prologue/run.mjs   exit 0  (new on main)
node scripts/check-agent-context-size.mjs --check   exit 0
node scripts/check-spawn-teardown-contract.mjs      exit 0
node scripts/check-e2e-harness-duplication.mjs      exit 0
node scripts/verify-committed-bundles.mjs           all bundles reproduce
gjsify lint                                         exit 0 (pre-existing warnings only)
```

### The GJS rows are discriminators, not decoration

Each fix was removed on its own, `dist/cli.gjs.mjs` rebuilt, and the suite re-run:

```
holdMainLoop() neutered    ✖ stays alive… — "timeout: no rebuild under GJS" (241 s)
run.ts daemon guard gone   ✖ survives `gjsify run dev`… — "did not survive its first build"
signal source never armed  ✖ stays alive… — "dev did not exit on SIGINT" (after GJS-A/GJS-B)
isSelfWrite() bypassed     ✖ rebuilds once per edit… — 119 launches for one edit, still climbing
```

Restoring each turned its row back green (7/7 above).

### Root-cause traces

The `holdMainLoop()` diagnosis, from three one-file `--app gjs` bundles under
`gjs -m`:

```
spawn() at module top level, then `await`   alive at 12 s (timeout)
`await Promise.resolve()`, then spawn()     alive at 12 s (timeout)
`await` a 50 ms timer, then spawn()         EXIT 0, child orphaned
```

and the third printed `depth after timer await: 1`.

## Left out deliberately

- **No unit row for `holdMainLoop()`.** Arming the GJS main-loop hook inside the
  test process is exactly the hang the depth guard exists to prevent. It is
  covered end-to-end by the GJS legs of `tests/e2e/dev-command` instead.
- **Only SIGHUP/SIGINT/SIGTERM.** `g_unix_signal_add()` also accepts
  SIGUSR1/SIGUSR2/SIGWINCH, but those are renumbered on some architectures, and a
  hardcoded table would arm the *wrong* signal there rather than fail. Stated in
  `@gjsify/process`'s README as a boundary, not left as an oversight.
- **`ensureMainLoop()`'s own behaviour is unchanged**, down to making the arming
  decision once at creation. Widening its depth guard would have reached
  `@gjsify/unit`'s runner, which is what that guard is for; `holdMainLoop()` is
  opt-in so only a caller that knows it will never unwind stacks a loop.
- **The GJS e2e rows are gated** on `gjs`, the committed CLI bundle and a
  `rolldown-native` prebuild for the arch — the same way other suites gate
  optional runtimes. A host that cannot answer the question skips rather than
  reporting a false red.
- `packages/infra/cli/AGENTS.md` grew 589 B; `status/agent-context-budget.json`
  carries the one reviewed line for it.


JumpLink:
## The 12 CI failures were three defects

### 1. The build failed on TS2305 — and the cause was `build:infra`'s ORDER

```
src/utils/watch-loop.ts:15:10 - error TS2305:
  Module '"@gjsify/utils/main-loop"' has no exported member 'holdMainLoop'.
```

`holdMainLoop` is exported, and `@gjsify/utils`' exports map routes `./main-loop`
at `./lib/types/main-loop.d.ts`. The source was fine; the **declaration file the
CLI compiled against was not the one this branch produces**.

`build:infra` built `@gjsify/cli` at clause 11 and `@gjsify/utils` at clause 17 —
the CLI compiled ten clauses before the library whose new export it imports.

One correction to the original diagnosis, from the job log: this is **not only a
cold-tree defect**. The failing leg reported `build cache cold or INCOMPLETE`, so
`packages/gjs/utils/lib` was present but **stale** — last build's `main-loop.d.ts`,
which has no `holdMainLoop`. That is why the error is TS2305 ("no exported member")
and not TS2307 ("cannot find module"). Both codes are the same defect seen from
different cache states, and the warm-stale one also fires on the incremental legs,
not just the cold ones nobody runs per-PR. Reproduced locally in both shapes.

**Fix:** a `gjsify workspace @gjsify/utils build:types` pre-pass before the CLI
clause — the same shape the four link-time deps (`semver`, `npm-registry`, `tar`,
`workspace`) already use, and it stays inside the bundler-free prefix because
utils' `build:types` is `gjsify tsc`. utils' compiled sources import no other
workspace package, so its declarations can be emitted at any point in the chain.

Verified the way CI would: `packages/gjs/utils/lib` and `packages/infra/cli/lib`
deleted, then the clauses run in `build:infra` order.

- before the fix, cold: `TS2307: Cannot find module '@gjsify/utils/main-loop'`
- after: `utils build:types` exit 0, then `cli build` exit 0

### 2. `tests/e2e/dev-command/` ran nowhere

`scripts/e2e-shard.mjs` shards the suites named in `package.json#scripts.test:e2e`
rather than globbing `tests/e2e/*`, so a suite missing from that script is executed
by no leg. 444 lines covering the watch/rebuild/relaunch loop — including both GJS
cases — ran nowhere.

It needs nothing the shared batch does not provide, so it is now listed in
`test:e2e` in its alphabetical slot rather than exempted in
`scripts/e2e-unlisted-suites.mjs`.

### 3. `tests/e2e/dev-command/run.mjs` was not formatted

Three over-width statements. Formatted with the **freshly built 0.41.0 workspace
CLI** — the globally installed `gjsify` here is 0.40.0, and that drift has
previously made `gjsify format` report success while writing nothing. Confirmed by
a changed file hash, then `format --check` clean.

## The mechanism, not just the instance

`status/open-todos.md` carried an entry asking for exactly this gate, and it had
already measured why the obvious answers do not work: deriving the order
topologically puts `@gjsify/canvas2d-core` first and `@gjsify/unit` last while
canvas2d-core's specs import unit (a devDependency edge no production sort sees,
and the dev graph is cyclic), and two cheaper rules were rejected for crying wolf
— 4 and 6 false positives. Both were missing the same fact: **which files a
package's `build:types` actually compiles**.

`scripts/check-build-infra-order.mjs` now carries that rule alongside its existing
bundler-free rule, so it runs in a required check with no new workflow. It reads
each clause's tsconfig `files`/`include`/`exclude`, scans only those files, and
clears a dependency when an earlier clause emitted its types, when `-d` builds it
as a production dep (`@gjsify/process` → `@gjsify/string_decoder`), or when its
declarations are committed (`@gjsify/resolve-npm` ships `lib/` via a `!lib/`
gitignore rule).

Getting it precise needed three things the naive version got wrong, each measured:

- **comments and template literals must be blanked.** `gjsify debug`/`gjsify browse`
  generate consumer entry files as template literals containing real
  `import … from '@gjsify/devtools-mcp'` lines. Nothing type-checks those.
- **import statements must be line-anchored.** `@gjsify/tar` carries the error text
  `"import '@gjsify/compression-streams/register' on GJS to register it"`, which a
  free-floating regex reads as an import.
- **`@ts-ignore` must be honoured.** `rolldown-plugin-gjsify` imports `@gjsify/console`
  under an explicit "resolved by Rolldown at user-build time, not by tsc here".

On this tree it now reports **zero** problems. As a discriminator, with the ordering
fix reverted it reports exactly one:

```
✗ clause 11 runs `@gjsify/cli build`, whose tsc compiles
  packages/infra/cli/src/utils/watch-loop.ts — it imports '@gjsify/utils/main-loop',
  but @gjsify/utils is built at clause 17.
```

and swapping `unit`/`runtime` back to their pre-#1133 order reproduces that incident
too. The open-todos entry is retired; both incidents live in the script header.

(A trap worth recording: a regex comment-stripper eats `/**/` inside the glob
`src/**/*.ts`, silently rewriting it to `src*.ts` — the file list collapses to the
`files` entries and the check passes for the wrong reason. The stripper is
string-aware, and the rule fails loudly if it resolves no inputs at all.)

## Re-verified rather than trusted

The tree did not compile in CI, so the two review blockers about `gjsify dev` under
GJS were re-run against a compiled tree. Real output:

| check | result |
|---|---|
| `gjsify workspace @gjsify/cli run check` | exit 0 |
| `gjsify workspace @gjsify/cli run test` | **2421 completed**, exit 0 |
| `gjsify workspace @gjsify/utils run test` | **255 completed** (Gjs 1.88.1), exit 0 |
| `gjsify workspace @gjsify/process run test` | **171 completed** (Gjs 1.88.1), exit 0 |
| `node --test tests/e2e/dev-command/run.mjs` | **7/7 pass**, 0 skipped |
| `node --test tests/e2e/run-command/run.mjs` | **16/16 pass** |
| storybook e2e (`-declaration`, `-no-stories`, `-on-node`) | **11/11 pass** |
| `gjsify lint` | exit 0 (pre-existing warnings only) |
| `gjsify format --check` | clean |

The dev-command suite's two GJS cases — `stays alive, rebuilds and relaunches, and
stops the app on SIGINT` and `survives 'gjsify run dev'` — both **ran** (not
skipped) and passed. The three storybook suites cover the donor: `storybook.ts`
lost 92 lines to the shared loop and still works.

JumpLink:
## A fourth defect, which only became visible once the build worked

With the build fixed, 19 checks went green and one turned up a **real bug in this
PR's own new code** — not a leftover from the three failures above:

```
✖ watch-loop: self-write filter › ignores the bundle and what the bundler writes beside it
✖ watch-loop: self-write filter › ignores the whole output directory
✖ watch-loop: self-write filter › excludes only the file when the output IS in the watched dir
❌ [Node.js 24.19.0] 3 of 2404 tests failed
```

`Node-pillar suites (win32-x64 / cmd.exe)` only got there because the CLI now
compiles — previously the job died at the build, so these three never ran. The
earlier failure was masking a second one.

**The defect.** `isSelfWrite()` resolved the *changed* path but compared it
against a raw `output`:

```ts
const changed = resolve(dir, filename);
if (changed === output || changed.startsWith(`${output}.`)) return true;
const outDir = dirname(output);
```

On POSIX that matches only because every caller happens to pass an
already-absolute path — `dev.ts` uses `resolve(cwd, plan.bundle)`, `storybook.ts`
`resolve(cwd, args.out)`. On win32 `resolve()` additionally rewrites separators
and prefixes the drive, so `D:\tmp\project\dist\out.mjs` was compared against
`/tmp/project/dist/out.mjs`: never equal, so **every self-write read as a source
edit** — precisely the rebuild-forever feedback the filter was added to stop, on
the one platform where nothing caught it.

**The fix** resolves both operands the same way. Under `path.win32` semantics the
old function got 6 of the 12 spec assertions wrong; the new one gets 0. Linux is
unchanged: `@gjsify/cli` still reports 2421 completed, and
`tests/e2e/dev-command` still 7/7 with both GJS cases running.

Worth noting for the record: on Linux **all five** of these cases passed, which is
the recurring lesson that a green Linux leg is not evidence about the OS axis.

Re-run after the fix — `dev-command` 7/7, `run-command` + `build-watch` 18/18,
`storybook-declaration` + `storybook-no-stories` 9/9, `foreach-command` 22/22,
`gjsify format --check` clean, and both `build-infra-order` rules pass.

JumpLink:
Rebased onto `main` (`0245982b`, picking up #1232/#1238/#1239 — one conflict, in
`test:e2e`, resolved by re-inserting `dev-command` into main's updated list;
coverage now reads 131/133).

## [1] The order gate failed its own suite — three defects, none of them the check's model

The invariant is right and unchanged. What was wrong was the reading around it:

- **`gjsify tsc -p .` names a DIRECTORY.** I read it as a config file and threw
  `EISDIR`, which is what surfaced as the `does not mistake a bundler-shaped word`
  failure. Now resolved to `<dir>/tsconfig.json`, the way tsc does.
- **A package with no readable tsconfig gives the rule nothing to assert.** I was
  reporting that as a finding; the suite's synthetic roots have no sources at all.
  "Cannot assert" and "is broken" are different answers.
- **`git ls-files` failed outright** — the fixtures are temp dirs, not work trees.

**On the git call you flagged: you were right, and I took the reader.** My first
instinct was a `try/catch` returning an empty set, which is defensible only because
it can make the rule *stricter*, never blinder — but it is still the
"skip where the binary is absent" shape, and `scripts/manifest-conformance/git-index.mjs`
says in its own header why that is this repo's most expensive defect. So the gate now
reads the index. `readIndexGitlinks` already walked every entry (version-4 paths are
stored as deltas, so it must) and discarded all but the gitlinks; I factored that walk
out and added `readIndexPaths` beside it, keeping the file's "ONE parser" promise
rather than adding a second reader. Measured with `git` absent from PATH entirely:

```
git on PATH? NO
build-infra-order: type-ordering rule read 337 compiled file(s) across 23 clause(s)
EXIT=0
```

Fully working, not degraded — so the stripped-PATH Windows leg is covered too.

**Nothing was loosened.** The discriminator still fires: reverting the ordering fix
still produces exactly one finding, naming file, specifier and clause. And the rule
now has the coverage it was missing — four tests in
`check-build-infra-order: the type-dependency order`, covering both directions
(rejects a clause compiling against a package built later; accepts it behind a
`build:types` clause) and both false-positive guards (a specifier that is only text
in a generated template literal; one carrying a `@ts-ignore`). The vacuous-pass guard
moved to the shape that is genuinely blind — configs that resolve but expand to zero
files, i.e. a broken glob — and the suite now asserts the real monorepo run reads a
non-zero count, so "checked nothing" cannot read as "checked and clean".

`tests/e2e/release-bundle-gate/run.mjs`: **38/38 pass**.

## [2] `gjsify dev` does not orphan the app — the test's liveness probe was lying

I have to correct the diagnosis here, because the evidence points the other way.

**Both tests exercise the same path.** The `under GJS` case that PASSED in the same
container does the identical thing — same `process.kill(session.child.pid, 'SIGINT')`,
same `isAlive` helper, same "was orphaned when dev exited" assertion. So there is no
second interrupt path to find, and the difference cannot be one.

**What actually differs is the reaper.** `isAlive` used `kill(pid, 0)`, which succeeds
on a **zombie** — the pid exists until someone reaps it. A desktop session reparents an
orphan onto an init that reaps within milliseconds; the CI container's pid 1 is the job
shell, which reaps nothing. So a child `dev` had already killed on its way out stayed
visible to the probe forever.

Reproduced deliberately, with `unshare -Urpf --mount-proc` giving the same non-reaping
pid 1:

```
old probe:  ✖ stops the app it launched when it is interrupted (20861ms)
            Error: timeout: the app (pid 47) was orphaned when dev exited (state: Z)
new probe:  ✔ stops the app it launched when it is interrupted (773ms)
```

`state: Z` — the process was already dead. That 20861 ms also matches the container's
20765 ms. So `gjsify dev` does stop the app it launched; there is no orphaned live
process and no second kill site was added.

The probe now reads `/proc/<pid>` state and counts only `Z` as gone, falling back to
signal 0 where there is no `/proc`. **This cannot mask a real orphan**: a RUNNING
process reports `R`/`S`/`D` and still fails the test — and the message now names the
state it saw, so the next occurrence says which of the two it is rather than assuming.

Two smaller corrections for the record: the app in this test is a plain long-running
script that writes a pid file, not a GTK process, so no GUI leak is possible from it;
and I checked for stray processes from my own runs while iterating — none
(`unshare` also confines them to a namespace that dies with the test).

`tests/e2e/dev-command/run.mjs`: **7/7 pass**, and **7/7 again under the non-reaping
pid 1**, with both GJS cases running rather than skipped.

## One thing I could not clear, and it is not from this branch

`@gjsify/cli`'s own suite reports 4 failures locally after the rebase, all in
`unresolved-workspace-import: toolchain fallback` — tests that arrived with #1232
minutes ago. `unresolved-workspace-import.{ts,spec.ts}`, `bundler-pick.ts` and
`rolldown-native/src/index.ts` are all **byte-identical to `origin/main`** on this
branch, so this branch cannot be their cause; it looks like local build state that a
partial rebuild does not reach (the first symptom was a genuinely stale
`rolldown-native` missing `isResolveMiss`, which rebuilding fixed). Flagging it rather
than claiming a pass I did not see — CI on a clean tree is the authority.

Local: `gjsify format --check` clean, `check-e2e-suite-coverage` 131/133,
`check-build-infra-order` both rules pass, `audit-runtimes --check --strict` exit 0.

Issue #996, opened 2026-08-05
The bundle gates count files where they must assert a decode


JumpLink:
Two GTK-runtime defects share one cause, so they share an issue: the release gate cannot see either of them, because **a file count is not a capability**.

## The gate hole

`packages/node-gi/scripts/verify-bundle-manifest.mjs` asserts that every `windowingData.verified` set has `files > 0`. The darwin manifest passes with **iconFiles 860 / verified icons 863 while zero of those files can be decoded**. Same false-positive shape as `Pixbuf.get_formats()` listing `svg` on win32 while every decode fails, and the same class ADR 0018 names ("do not let a file count stand in for a load").

### Design (settled; stays inside the gate's documented shape)

The gate is a pure **manifest reader** on purpose — "the build log is not shipped". So:

- The **builder**, which runs on the target platform, performs a real decode of an icon it has just bundled (load the addon, `GdkPixbuf.Pixbuf.new_from_file` on a bundled `.svg` **and** `.png`, assert width/height > 0) and records the result in the manifest, e.g. `windowingData.decodeProbe = { svg: '16x16', png: '16x16', loaderDir, formats: [...] }`.
- The **gate** asserts that record exists and passed. Exactly the pattern `windowingData.verified` already uses: "requiring the RECORD means a bundle built by an older builder, or with the gate bypassed, cannot publish."

Builders: `packages/node-gi/scripts/build-gtk-runtime-darwin.mjs` (`windowing.verified` is set around line 720) and `packages/node-gi/gtk-runtime-win32-x64/scripts/build-gtk-runtime.mjs`.

### Why it was not done blind

1. Gate and builders must land **together** — a gate requiring a record no builder writes fails every release leg immediately.
2. There is **no linux gtk-runtime bundle** by design (only darwin-arm64, darwin-x64, win32-x64; linux uses the system GTK), so the builder half cannot be verified on a Linux workstation at all.
3. `release.yml` is `on: release` / `workflow_dispatch` only, so **a PR never executes this gate**. Shipping a new assertion into it blind is exactly how the two original inline `node -e` blocks shipped dead on arrival — the failure that cost all three v0.28.0 publish legs while every bundle they gated was correct (#980).

So: implement with a real darwin/win32 run behind it (a `workflow_dispatch` on a macOS/Windows runner, or the local macOS VM — note it is darwin-**x64** while CI publishes darwin-arm64), and land gate + both builders in one PR.

[...]

PR #1239, opened 2026-08-21
fix(node-gi): assert a decode, not a file count

JumpLink:
## What changed and why

`verify-bundle-manifest.mjs` gated the GTK-runtime publish on every declared
`windowingData.verified` set having `files > 0`. The published **darwin-x64 0.28.0**
bundle passed that perfectly — `iconFiles: 860`, `verified icons: 863` — and **zero of
those files decoded**: the addon kept its absolute Homebrew install names, so a Mac with
Homebrew glib loaded two GObject registries, type identity failed across the boundary,
and `Pixbuf.new_from_file()` on the bundle's own `open-menu-symbolic.svg` returned
`-1x-1`. Every count in the manifest was correct. A file count is not a capability
(ADR 0018). `release.yml` is `on: release` / `workflow_dispatch` only, so no pull request
has ever executed this gate.

The gate is deliberately a pure **manifest reader** (the build log is not shipped), so
the decode has to happen where the target platform is — in the builder — and travel to
the gate as a **record**.

- **`packages/node-gi/scripts/decode-probe.mjs` (new)** — points node-gi at the freshly
  built bundle and decodes a `.svg` and a `.png` through the bundle's own loader
  (the bundled PNG when one ships; otherwise the decoded SVG `savev`'d to a temp PNG and
  read back, recorded as `source: 'saved-from-svg'`, which still drives the bundle's own
  png loader). A **child process** writing through a **file**: activating a bundle
  re-execs the process on darwin, and the two-registry state aborts rather than throws.
- **Both builders** run it under `--windowing` and fail the build on a bad record, so
  `--addon` is now required there. darwin probes the relocated
  `<stage>/{node_gi.node, gtk/}` sibling layout — the shape a consumer actually loads.
- **`decodeProbeProblems()` is the ONE definition of "passed"**, imported by the builders
  (fail the build) and by the gate (fail the publish), so the two cannot drift. It
  re-derives the verdict from the recorded **dimensions**, so `ok: true` over a `-1x-1`
  decode still fails; a **missing** record fails too rather than degrading to
  "unverified" — the shape the arch guard once got wrong.
- **Workflow wiring**: the four `--windowing` bundle legs now take the addon from the
  matching `node-gi-prebuild-*` job. A GitHub artifact belongs to the run that produced
  it, so `release.yml` cannot reach `node-gi.yml`'s upload.

### The probe must measure the tarball, not the runner

A pixel count is only evidence if the pixels came from the bundle. Two additions make
that checkable, and both close paths that are reachable **without any bug**:

1. **The child no longer inherits the builder's environment** (`probeChildEnv`). Dropped:
   `DYLD_FALLBACK_LIBRARY_PATH`, `DYLD_LIBRARY_PATH`, `GI_TYPELIB_PATH`,
   `GDK_PIXBUF_MODULE_FILE`, `GDK_PIXBUF_MODULEDIR`, `GSETTINGS_SCHEMA_DIR`,
   `XDG_DATA_DIRS`, `FONTCONFIG_PATH`, `FONTCONFIG_FILE`, `GST_PLUGIN_SYSTEM_PATH`,
   `GST_PLUGIN_PATH`, `GST_PLUGIN_SCANNER`; plus every `PATH` entry at or under the build
   prefix. `node-gi.yml` exports `DYLD_FALLBACK_LIBRARY_PATH=$BREW_PREFIX/lib` by hand —
   the incident `docs/node-gi-platform-notes.md` records as *"CI's own YAML was the reason
   no test could catch it"* — and appends `<gvsbuild prefix>\bin` to `GITHUB_PATH` one
   step before the win32 build, where a DLL **missing** from the bundle would resolve
   from the prefix and the probe would record a pass. `maybeWireGtkWindowingEnv()` sets
   its variables only *if unset*, so an inherited value beats the bundle's own data.
   Scrubbed in the probe rather than per workflow: one place, both platforms, and it also
   makes the PR leg and the release leg measure the same thing (only one of them has the
   prefix on `PATH`).
2. **The record states which GTK answered** — `platform`, `gtkSource`,
   `bundleIsProbeTarget`, read from `gtk-runtime.js` *after* the addon loaded — and on
   darwin/win32 the gate requires `gtkSource === 'bundle'` plus a resolved dir identical
   (realpath-compared at record time) to the probed one. Setting `GJSIFY_GTK_RUNTIME` +
   `GJSIFY_GTK_PREFER` proves nothing on its own: `index.js` wraps
   `activateBundledGtkRuntime()` in a never-fatal `try {} catch {}`, the function returns
   `null` whenever the policy did not pick the bundle, and `resolveGtkRuntimeBundle()`
   falls through to `prebuilds/`, the sibling monorepo package and the installed optional
   dep. In every one of those the **host** GTK decodes the file sitting at the bundle's
   path and records real, passing dimensions. linux stays permissive — there is no linux
   bundle by design, and the unit test's real-decode leg probes the system stack on
   purpose. A record with no `platform` fails closed, exactly as an absent record does.

The paths are compared in the child and only the verdict travels: this record ships to
consumers inside `manifest.json`, which is why the addon is recorded as a bare basename.

## Measured

All on this Fedora linux-x64 workstation.

```
$ node --test tests/e2e/release-bundle-gate/run.mjs
ℹ tests 34   ℹ pass 34   ℹ fail 0

$ node --test packages/node-gi/node-gi/test/gtk-runtime-decode-probe.test.mjs
ℹ tests 13   ℹ pass 12   ℹ fail 0   ℹ skipped 1

$ node --test packages/node-gi/node-gi/test/gtk-runtime-bundle-gates.test.mjs
ℹ tests 32   ℹ pass 32   ℹ fail 0

$ for s in ci-pr-trigger-parity ci-workflow-rev-pin ci-runner-arch \
           ci-classifier-output-coverage prebuild-change-gate affected-classifier-bundle; do
    node --test tests/e2e/$s/run.mjs; done
ci-pr-trigger-parity           pass 13  fail 0
ci-workflow-rev-pin            pass 14  fail 0
ci-runner-arch                 pass 29  fail 0
ci-classifier-output-coverage  pass  8  fail 0
prebuild-change-gate           pass 37  fail 0
affected-classifier-bundle     pass 14  fail 0

$ node scripts/audit-runtimes.mjs --check      # exit 0
$ gjsify format --check                        # All matched files use the correct format.
$ gjsify lint                                  # exit 0, 0 errors (59 pre-existing warnings, none in touched files)
```

**Revert-proof — the tests bind to the implementation, not to themselves.** Deleting the
`platform`/`gtkSource`/`bundleIsProbeTarget` block from `decodeProbeProblems()` and
re-running:

```
gtk-runtime-decode-probe.test.mjs   pass 9  fail 3
  ✖ a record that cannot say WHICH GTK decoded the file fails closed
  ✖ a decode the HOST GTK answered is not a passing bundle probe
  ✖ a decode through a DIFFERENT bundle than the probed dir fails
release-bundle-gate/run.mjs         pass 31 fail 3
  ✖ rejects a probe the HOST GTK answered, not the bundle
  ✖ rejects a probe that resolved a different bundle than the one it decoded
  ✖ rejects a probe that cannot say which GTK decoded the file
```

(The same experiment on the gate half: restoring `main`'s `verify-bundle-manifest.mjs`
takes the suite from 34/0 to 25/9 — every decode-probe case goes red.)

**The scrub, measured directly** (a poisoned parent env, the exact value `spawnSync`
receives):

```
$ GI_TYPELIB_PATH=/host/typelibs DYLD_FALLBACK_LIBRARY_PATH=/opt/homebrew/lib \
  XDG_DATA_DIRS=/host/share PATH="/opt/homebrew/bin:$PATH" node -e '…probeChildEnv…'
parent GI_TYPELIB_PATH = /host/typelibs
child  GI_TYPELIB_PATH = undefined
child  DYLD_FALLBACK   = undefined
child  XDG_DATA_DIRS   = undefined
parent PATH has brew   = true
child  PATH has brew   = false
child  keeps HOME/PATH = true
```

`probeChildEnv` takes `platform` as a parameter rather than reading `process.platform`,
so the win32 branch (`;`, backslashes, case-insensitive) is executed from Linux in the
unit test — the same rule `decideGtkSource()` follows.

## Deliberately left out

- **No darwin or win32 decode has ever been RUN.** There is no linux GTK bundle by design
  and this host cannot build the node-gi addon (`cairo-devel` absent, no sudo), so the
  real-decode test is skipped here. Everything above is the probe's pure half plus the
  gate. The first real decode is the next `node-gi.yml` macOS/Windows bundle leg; a
  failure there is the probe reporting #996's defect, not a bug in the probe.
- **`publish-gtk-runtime-darwin` now depends on the whole two-arch
  `node-gi-prebuild-darwin` matrix**, so an Intel Homebrew hiccup skips *both* darwin
  bundle publishes, not only x64. That is wider than this job's own `fail-fast: false`
  note promises, so both comments now say so rather than contradicting each other. Not
  repaired by splitting the matrix: `publish-node-gi` already needs the whole matrix, so
  that scenario already degrades the release, and a split would put a third verbatim copy
  of the brew formula line on the release path — the copy that job's header argues
  hardest against.
- **The PNG fallback is documented rather than gated.** `runDecodeProbe` round-trips the
  SVG when the bundle ships no `.png`. Asserting `source === 'bundled'` would only catch a
  hand-edited record — a bundle that ships PNGs whose loader is broken already fails on
  the recorded `0x0`/`-1x-1` dimensions. So the claim was corrected instead: the docs and
  both READMEs now say "a `.svg` and a `.png` through the bundle's own loader,
  round-tripping the svg when the bundle ships none".
- **`scripts/check-comment-budget.mjs` (advisory, not a required check)** reports
  `packages/node-gi` at 0.442 against a 0.434 ceiling. `main` is already over at 0.438;
  this PR adds 78 of the 151 over-ceiling lines. They are incident text — the two-registry
  measurement, the hand-exported `DYLD_FALLBACK` incident, why each fallback path is
  reachable — which the repo's own rule says not to compress away. Six lines of overlap
  between the header, the gate and the probe were cut instead.

Refs #996


JumpLink:
## The five red legs were TWO defects, not one — and only one of them was the probe

Rebased onto `origin/main` and pushed two commits. Short version: **four of the five legs
never decoded anything** (a path bug in the probe, now fixed), and **the fifth decoded and
told the truth** — the win32 bundle really cannot read its own SVG icons, and has not been
able to since PR #774. The gate was not relaxed anywhere.

---

### 1. darwin (4 legs) — the probe never ran. `fix(node-gi): resolve a pinned addon path`

```
probe aborted: the native addon exists but could not be loaded.
  addon: packages/node-gi/node-gi/prebuilds/darwin-arm64/node_gi.node
  cause: Cannot find module 'packages/node-gi/node-gi/prebuilds/…'
A bundle WAS selected, so it is likely incomplete for this target.
Reinstall @gjsify/gtk-runtime-darwin-arm64, or set GJSIFY_GTK_PREFER=system
```

`node-gi.yml` runs the darwin builder with
`--stage packages/node-gi/node-gi/prebuilds/darwin-<arch>` — **repo-relative**. The builder
passed it to the probe, which set it as `NODE_GI_NATIVE`. `loadNative()` then asked two
questions that mean different things about the same string:

| | resolves against |
|---|---|
| `existsSync(candidate)` | the process **cwd** → passed |
| `require(candidate)` | a **module specifier**, i.e. `@gjsify/node-gi`'s own directory and `node_modules` → `Cannot find module` |

The addon was on disk the whole time and was never opened. Two fixes:

- **`native-paths.js`** resolves a pinned `NODE_GI_NATIVE` path, so the check and the load
  name one file. `build` / `prebuild` and an absolute pin are untouched.
- **`decode-probe.mjs`** resolves `--bundle` and `--addon` in parent and child. Both leave
  the process again as env — one into a `require()`, one into a lookup that **re-execs** on
  darwin — so a probe whose meaning depends on its caller's cwd is measuring the caller.

**The misleading diagnostic is fixed at the core too**, since that message is what a stranger
would have acted on. `load-diagnostics.js` now discriminates: `MODULE_NOT_FOUND` /
`ERR_MODULE_NOT_FOUND` (plus the codeless wording Bun, Deno and the GJS napi shim raise) is a
**resolution** failure and says so, naming the fix; the GTK dependency-closure explanation
stays for the dlopen failure it was written for (#1063).

### 2. win32 (1 leg) — the probe ran, and it is right. `fix(node-gi): point the win32 loader cache home`

```
svg: share/icons/Adwaita/scalable/devices/audio-headphones.svg -> 0x0
  "Unable to load image-loading module: <bundle>\pixbufloader_svg.dll: … not found."
png: share/icons/Adwaita/16x16/devices/audio-headphones.png -> 16x16   ← fine
gtkSource: bundle, bundleIsProbeTarget: true, loaderModules: 1
formats: ["svg","png","gif","jpeg","tiff"]                             ← still advertises svg
```

`<bundle>\pixbufloader_svg.dll` has never been written — the module is staged at
`lib/gdk-pixbuf-2.0/2.10.0/loaders/`. § 4a rewrote every module line of `loaders.cache` to a
**bare leaf**, stating that "gdk-pixbuf resolves the DLL paths in it relative to
`GDK_PIXBUF_MODULEDIR` when set (node-gi sets it)". Read against **gdk-pixbuf 2.44.6**, the
version gvsbuild 2026.6.0 pins, that is false twice:

- `build_module_path()` (`gdk-pixbuf-io.c`) is the only thing that touches a relative cache
  entry, and it joins it with `gdk_pixbuf_get_toplevel()` — on win32 the package installation
  directory of the gdk_pixbuf DLL, i.e. **the parent of `bin/`**. The leaf resolved to the
  bundle root.
- `GDK_PIXBUF_MODULEDIR` is read by `queryloaders.c`, the **generator**, where it only picks
  the directory to scan. The runtime reads exactly one variable, `GDK_PIXBUF_MODULE_FILE`.

So **every SVG icon in every win32 windowing bundle since PR #774 has failed to decode**, and
each reason nothing noticed is its own lesson: PNG/JPEG/GIF/TIFF are `-Dbuiltin_loaders=all`
inside `gdk_pixbuf-2.0-0.dll` and never consult the cache; the cache still *parses*, so
`Pixbuf.get_formats()` still advertises `svg`; and the gate was a `Test-Path` on the cache
file. A window renders — its symbolic icons just come back empty. This is the darwin half of
#996 with a different mechanism, and it is exactly what this PR exists to surface.

Fix + **mechanism**, in `scripts/pixbuf-loader-cache.mjs`:

- module lines are rewritten to `lib/gdk-pixbuf-2.0/2.10.0/loaders/<leaf>`, relative to the
  toplevel gdk-pixbuf actually joins against. **Forward slashes**, and that is not style:
  `scan_string()` runs the token through `g_strcompress()`, where the honest Windows spelling's
  `\2` (from `\2.10.0`) opens an **octal escape** and is silently mangled rather than rejected.
- `loaderCacheProblems()` asserts that **every** module line resolves to a file that exists
  under the bundle. A bare leaf, the query tool's un-rewritten absolute path, a stale subdir and
  a loader the copy loop missed all fail the build, by name. The probe decodes one format; this
  reads every line.
- `gtk-runtime.js` and § 5c carried the same false claim as justification. Corrected in place
  rather than deleted.

---

### Measured (Fedora linux-x64 — no darwin or win32 decode was run here)

```
node --test .../test/addon-load-failure.test.mjs        15 pass  0 fail
node --test .../test/pixbuf-loader-cache.test.mjs        9 pass  0 fail
node --test .../test/gtk-runtime-decode-probe.test.mjs  13 pass  0 fail  1 skip
node --test .../test/gtk-runtime-bundle-gates.test.mjs  32 pass  0 fail
node --test .../test/gtk-source-policy.test.mjs         18 pass  0 fail
node --test tests/e2e/release-bundle-gate/run.mjs       34 pass  0 fail
node --test tests/e2e/prebuild-loader-path/run.mjs      43 pass  0 fail
node scripts/audit-runtimes.mjs --check                 exit 0
gjsify format --check                                   clean
```

**The path bug is provable here without a darwin host and without an addon**: a 4-byte file is
refused by `dlopen` and never by the module resolver, so `ERR_DLOPEN_FAILED` is positive
evidence that a relative path survived the trip. `RELATIVE paths reach the child as paths, not
as module specifiers` drives `spawnDecodeProbe` with the exact CI shape.

**Revert-proof**, all three A/B'd:

| reverted | result |
|---|---|
| `resolve(prefer)` in `nativeCandidates` + the probe's `resolve()` | 2 cases red (26/2) |
| the `isResolutionFailure` early return | 3 cases red (12/3) |
| the cache rewrite back to a bare leaf | 4 of 9 cases red, incl. *"a bare leaf FAILS the build — the exact cache that shipped"* |

### What is still NOT proven

No Windows or macOS host was available, so **whether the win32 bundle now decodes is not proven
by anything in these commits**. The decode probe on the next `windows-gtk-windowing-runtime`
leg is the oracle. A still-red probe there means the diagnosis above was wrong — **not** that
the gate should move.

Likewise the darwin legs: the path fix lets the probe *run*. What it then finds is #996's
original question, and if the darwin bundle genuinely cannot decode its own icons, that is the
defect this PR was built to make visible, and it should stay red until the bundle is fixed.

PR #1258, opened 2026-08-22
docs(release): what 0.42.0 is about

JumpLink:
The prose preamble for the next release body.

`gjsify ship` landed in 0.39.0, is exercised by this repo's own tests — and the first two projects that pointed it at *their own* `package.json` both hit a wall inside the first minute. The text says which two walls those were, plus the gap that made one of them unfixable from the consumer side:

- **`ship` could not find its own desktop template** — `ENOENT` on *every* project through the GJS entry point, while the Node path (CI, `npm install -g`) worked fine. That split is exactly why it went unnoticed for three releases.
- **`bundledTypelibs`** — a GI library that arrives as an npm prebuild has no `gir1.2-…` package anywhere. Naming a package that does not exist would have turned a clear refusal into an install that succeeds and an app that dies at its first import.
- `Gda-6.0` in the typelib table, and `ship.description` as a plain string.

Plus the new package **`@gjsify/gtk-host`** (ADR 0027) and `FindWidget` / `SendKey` in `@gjsify/devtools`.

## On the cycle

The previous prose went out with 0.41.0 and would not be reused — the file counts only if git says it changed since the last tag. So this **replaces** it rather than appending.

`node scripts/check-changelog-references.mjs --release-notes 0.42.0` reports only that the top `CHANGELOG.md` section is still 0.41.0; `--write` regenerates that in the `before:git:beforeRelease` hook during the cut. No bare `@` or `#` appears in the text, so the reference detector cannot abort the cut.

https://claude.ai/code/session_0124tpVQDCtmrh9p4McqjBty


JumpLink:
Not a change request on what is here — the `ship` narrative is the right lead. But `next.md` is published as the whole 0.42.0 body, and the "Elsewhere" section is currently one line while **twelve** other PRs merged into this cycle. Three of them are things a user would want to know, and one is a new command.

Rather than describe the gap, here is prose you can fold in or ignore.

---

### New command: `gjsify dev`

Watch, rebuild, relaunch — the inner loop, on a host where `gjsify build --watch` cannot run at all. That flag drives rolldown's watcher API, which only the npm engine exposes, so on a Node-free GJS machine the loop was manual: edit, Ctrl-C, rebuild, relaunch.

`gjsify dev` asks for no watcher API. It watches with `fs.watch` and rebuilds by re-entering the ordinary build command, so it works the same on gjs, node, bun and deno. What it builds is not declared twice: it reads your own `build:gjs` / `build:node` script and layers flags on top, so the dev loop and `gjsify run build` cannot drift apart. All seven templates now scaffold a `dev` script. (#1237)

### Every `--app gjs` bundle shipped an empty prologue

The byte-1 GI prologue was built across three PRs and then wired to nothing — `app/gjs.ts` never passed it to `processStubPlugin`, and the generator returns `''` for empty input by design. So every unit test of the generator stayed green while every bundle ever built carried nothing.

It carries the system GI libdir candidates now, each prepended at startup only on a host that actually shows them. This is what lets a `--app gjs` bundle find a Homebrew or MacPorts GTK on a machine the build never saw. (#1236)

### A build script can bundle itself on a Node-less host again

The ADR 0002 bootstrap died at its third step on a fresh clone with no `node`: a repo script re-entered as `gjsify run --node-script` has to resolve one `@gjsify/<pkg>/register` subpath per injected global, and in a cold clone those workspace packages have no `lib/esm` yet.

A build script is toolchain, not user code — it does not need *the workspace's* `@gjsify/fs`, it needs a working one. Toolchain bundles now fall back to the CLI's own directory once workspace resolution comes up empty. Project first, always: a workspace that has the package keeps using it, and `gjsify build` proper never sets the anchor, so a user's missing dependency stays fatal. (#1232)

### Every SVG icon in every Windows windowing bundle failed to decode

Since #774, and nothing saw it. Three things hid it: PNG, JPEG, GIF and TIFF are compiled into `gdk_pixbuf-2.0-0.dll` and never consult the loader cache; the cache still parsed, so `Pixbuf.get_formats()` kept advertising `svg`; and the release gate was a file-count check.

A file count is not a capability. The gate decodes a real `.svg` and a real `.png` through the bundle's own loader now, with the host GTK scrubbed out of the probe child's environment so the prefix on `PATH` cannot answer for the bundle. Running it also closed the darwin half of #996 in the other direction — both arches had been decoding correctly for some time, and the count-based gate could not see that either. (#1239)

---

Also merged this cycle, if you want them in a list rather than prose: #1245 (`url.searchParams` mutations now write back into `href`), #1250 (`@gjsify/domparser` parses real HTML5 and runs selectors), #1233, #1234, #1235, #1240, #1244.

One aside, unrelated to the content: the PR title is German. Per the repo's own convention that is the one place it should be English — the body is fine either way.

PR #1259, opened 2026-08-22
ci: run build:infra on a cold tree with no node

JumpLink:
Closes the observation gap behind #1231 / #1232: the defect they fix broke `build:infra` for a Node-less contributor, and no CI leg could have noticed.

## What is already covered, and the one dimension it misses

The `build` job's **"Bootstrap the build toolchain (cold trees only)"** step does run `build:infra` under gjs on a cold build-output cache. It differs from a Node-less contributor's tree in exactly one way, and that way is the whole defect: **`node` is on PATH there**, so a package script's `node scripts/x.mjs` runs natively and is never bundled for GJS. The re-entry through `gjsify run --node-script` — where the `@gjsify/*` substitutions and every `--globals auto` register subpath must RESOLVE against a tree that has not been built — was unreachable from CI by construction.

That is how it stayed broken: `status/open-todos.md` recorded the opposite as measured green (the measurement deleted the two native facades, which leaves every workspace `lib/esm` in place — not the cold case it reads as), and the same failure blocked `check-refs-pin.mjs`, so `build:prebuilds` could not start either.

## Design, and what it costs

- **A clone, not `git clean`.** The checkout is warm by then — the setup composite restores `node_modules` and the build-output cache — and the rest of the graph depends on that. A clone costs a hardlinked object store.
- **`node_modules` is hardlinked across, not installed.** A cold clone is defined by having no `lib/esm`, *not* by having no dependencies; a contributor runs `gjsify install` first anyway. The workspace symlinks are relative (`../../packages/node/fs`), so the copy re-points them at the clone's own **unbuilt** packages — exactly the shape under test. Repeating the ~2 GB extract this workflow already documents would buy nothing. `.cache` is dropped: it carries the build cache, which would hand the clone the outputs it must not have.
- **Both premises are asserted, not assumed** — no `lib/esm` anywhere, and `node_modules` actually came across. A warm clone would make the job green while testing nothing, which is the failure mode it exists to remove.
- **Fake `node`/`npm`/`npx` exit 127 and announce themselves**, so a silent fallback to Node fails the job rather than passing it. Same reason `workspace-node-free-gjs` does it, rather than editing PATH.
- **Wired into the gate's `needs:`** (and `ci-summary`'s). A job outside the gate cannot block a merge, and the gate's own self-check enforces that; leaving it out would restore exactly the dishonest green this removes.

The CLI under test is the bundle **this tree's** `build` job produced, not the published one `install.mjs` fetches — the question a PR must answer is whether its own source bootstraps a cold clone. The published-bundle half is `install.mjs`'s contract and belongs to `tests/e2e/install-script`.

## Verification

**The first run of this PR is the test of the job**, and I would rather say so than imply more. What I could verify from the measuring host is the premise it rests on: the workspace symlinks are relative, so a clone that receives `node_modules` resolves `@gjsify/*` to its own unbuilt packages. The container, `testuser`, the CI `node_modules` and the built bundle do not exist there, so the step body itself is unproven until it runs here.

Rebased onto `bed3e93d1`; after the rebase I re-checked the two things a moved `main.yml` could break — the gate still covers all 15 jobs, and the step this job's comment refers to still exists.


JumpLink:
This closes the one thing `status/open-todos.md` § "Nothing runs `build:infra` on a cold tree with no `node`" says is still open, so **that entry should go in this PR** — otherwise it lands as exactly the stale-reads-like-live-work claim the ledger's own § "This ledger goes stale silently" describes.

Context, since the entry was rewritten a couple of hours ago in #1231 and you may be looking at the older text: the *defect* was closed by #1232 (a toolchain bundle falls back to the CLI's own directory once workspace resolution returns null). What survived was only the measurement gap — no suite executes the third bootstrap step, because `bootstrap-cold-tree` asserts the `--print-plan` branch without spawning, `node-free-bootstrap` covers the install, `node-script-cold-workspace` drives one planted package rather than a whole cold tree, and every CI host has `node`. This leg is the missing half.

One thing worth asserting while you are in there, if it is cheap: all 23 register subpaths `--globals auto` injects sit in `@gjsify/cli`'s transitive dependency closure, and that is *why* the anchor can answer on a cold tree. A leg that goes green without ever exercising the fallback — say because the tree it builds is not actually cold — would look identical to one that does. `--verbose` printing `closure map expanded 1 → 21 global(s)` is the cheapest discriminator I know of.

PR #1275, opened 2026-08-22
feat(rolldown-plugin-gjsify): Blueprint in libraries

JumpLink:
Blueprint was quietly an application-only feature, and the way it failed pushed people toward code that cannot be translated.

## The defect

The `.blp` transform is installed by the `--app gjs|node|browser` build factories and by nothing else. A Blueprint file imported from a package built with `--library` therefore reached rolldown's JavaScript parser — and a `.blp` opens with `using Gtk 4.0;`, which is *valid JavaScript*: a `using` resource declaration with no initializer.

```
PARSE_ERROR: Using declarations must have an initializer.
```

Nothing in that message names Blueprint.

## Why it mattered more than the error

The error at least stops you. What it pushed people to do instead did not: a widget shared between applications had to assemble itself in TypeScript, and a caption assigned from TypeScript carries no `translatable` attribute, so `xgettext` never sees it.

`LoadingStack` — used by consumer apps for every loading/error view — said **"Something went wrong"** in English in all of them, permanently. That is invisible: an untranslat**able** string looks exactly like one nobody has translated yet, so it never gets filed.

## The fix

`--library` builds now run the same transform, placed next to `cssAsStringPlugin()`, which exists for the identical reason (an asset extension reaching the JS parser).

`LoadingStack` is the first widget to use it: three pages in `loading-stack.blp`, error title `_("Something went wrong")`.

## Verification

Not "it builds". The emitted library module carries the compiled Builder XML with `translatable="yes"`, and the widget was **constructed under a real display**:

```
PROBE visible=loading | pages=loading,content,error | errorTitle=Probe title
    | errorDesc=Probe description | visibleAfterShowError=error
    | contentChild=content probe
```

That step is the point: the package's 45 tests pass on GJS and Node and **none of them construct** `LoadingStack` — the test entry says so explicitly ("need a display to exercise"). That gap is how the untranslatable caption survived.

## Keeping the door shut

Two lint rules in `oxlint-plugin-gjsify`, both measured against this workspace before being switched on:

| | Learn6502 | steuererklaerung/desktop | bauplaner/cli |
|---|---|---|---|
| `prefer-blueprint-template` | **0** | 16 | 15 |
| `no-literal-widget-label` | **0** | 454 | 219 |

Learn6502 carries a whole application in 24 Blueprint files and reports zero under both, which is what makes these rules a measurement rather than an opinion.

- **`prefer-blueprint-template`** requires *both* signals — constructs a `Gtk`/`Adw` type **and** parents it — because constructing alone is a model or an adjustment and parenting alone is a reparent. A template that fills data-driven children at runtime is the intended pattern and stays silent.
- **`no-literal-widget-label`** covers the half that survives a template: a bare literal in a prose property. Wrapping it as `_("…")` silences it, which is why there is no autofix.

Both were tightened by their own findings: `Adw.Application` is excluded (an application object is not a widget and legitimately builds its own `Gtk.CssProvider` and `Adw.AboutDialog`), and `**/*.spec.ts` joined the scoped-off paths.

gjsify itself now reports **0** under both rules — the repo obeys the rule it enforces.


JumpLink:
## Red for two real reasons, both mine, and the second one is the interesting one

**1. `blueprint-compiler` is not on the macOS or Windows runners.** By putting a `.blp` into `@gjsify/adwaita-app` — a package built on all three platforms — I made the compiler a hard build requirement everywhere, not just where an app ships. `BlueprintCompilerNotFoundError` on darwin-arm64, darwin-x64 and win32-x64.

**2. The consumer gates bootstrap from the PUBLISHED CLI** (ADR 0002), which is 0.42.0 and has no library-mode transform. So during a cold bootstrap the `.blp` reached the JavaScript parser and `better-sqlite3` / `node-gi N-API` failed on the *very parse error this PR fixes*.

That second one is a general rule worth writing down: **a capability cannot be used by the tree that introduces it until it has shipped once.**

**3.** My rules also fired across `examples/` and `showcases/` — I had measured only `packages` and `website` locally. Demos are now scoped off alongside fixtures and stories; their captions are the subject matter.

## What changed

The capability stays — it is correct, and application consumers (which build with their own installed CLI) need it. What went is the premature first consumer.

- `LoadingStack` reverted to constructor-assembly, with the measured reason recorded at the widget and two reasoned `oxlint-disable-next-line` suppressions. `reportUnusedDisableDirectives` is `error` here, so those retire themselves the day the conversion becomes possible.
- **`tests/e2e/library-blueprint/`** carries the capability instead: a library fixture whose emitted module must hold the compiled Builder XML with `translatable="yes"`, plus the discriminator that a malformed `.blp` fails through *blueprint-compiler* and not through the JavaScript parser — a JS-parser message would mean the transform never ran and the first assertion passed for an unrelated reason.

Probed by deleting the fix and rebuilding: both e2e assertions go red, then green again. Whole-tree lint and `format --check` are clean.

One thing I got wrong twice and fixed: `oxlint-disable-next-line` applies to the line immediately after the comment, so a directive at the top of a multi-line comment block suppresses the comment's own continuation, not the code. Both suppressions are single-line now, verified by the fact that the rules report zero *and* the unused-directive check stays silent.

JumpLink:
## Review round: 44 findings raised, 7 survived, all 7 fixed

Ran an adversarial review over seven lenses — rule precision in both directions, the walker, plugin ordering, whether the e2e can pass vacuously, every factual claim in the prose, and the suppressions. Each finding went to a skeptic instructed to refute it and to default to refuted when it could not reproduce. **37 died there.** These 7 survived, each reproduced by a fixture before and after the fix.

### Two were defects in what the rules claim to do

**`declaresTemplate` could be silenced by accident.** It searched the whole class subtree for the identifier `Template`, so anything merely *using* the word switched the rule off for the enclosing class:

```ts
export class Panel extends Gtk.Box {
    static { GObject.registerClass({ GTypeName: 'Panel' }, this); }
    describe() { return { Template: false, name: 'panel' }; }   // ← silenced the class
    build() { const l = new Gtk.Label({}); this.append(l); }
}
```

A rule whose whole value is that it cannot be switched off by accident had an accidental off switch. It now looks only where a template can be declared. Verified: that fixture reports again, a genuine `Template` in `registerClass` stays silent.

**`add_response` was invisible** — and it is the only way an `Adw.AlertDialog` gets its buttons. So a dialog whose `heading` the rule *did* report still shipped English-only buttons, in the same function. 24 live call sites across the two consumer apps. Now covered, with a message that says to wrap in place, since a response label has no declarative form.

### Three were precision

Neither signal can type-check its operand — the rule sees `new Gtk.X` and a method *name*, never a GType. So `Gtk.StringList` filled with `.append()` read as assembling an interface, which is exactly the model case the header promises is silent. A deny-list of known non-widget constructions closes it.

`add_action` left `PARENTING_METHODS` entirely: it is `Gio.ActionMap.add_action` and has nothing to do with parenting, so **every widget that registers an action** was reported.

`walk` descended into `Program.tokens` and `Program.comments` — its node test is structural, and a token object has a `type` string too. The label rule walks the statements now.

### One was a wrong message

It suggested `set_title: _("…")` in the `.blp`. Blueprint has no `set_title`; it has `title`. The message names the property.

### And the new coverage immediately found a real defect one package over

`errorDialog` hardcoded its `"OK"` button with no caller override. Library code has no text domain of its own, so that caption was untranslatable in **every consumer application**. It takes an `okLabel` now — the same hook `confirmDialog` already had, and the reason is written down next to it.

That is the rule catching something in the framework on its first day, which is the argument for having it.

### Verification

| | |
|---|---|
| Learn6502 (the reference app) | 0 findings, unchanged |
| this repo, whole tree | 0 findings |
| `tests/e2e/library-blueprint` | 2/2 pass |
| `@gjsify/adwaita-app` | 45/45 on GJS **and** Node |
| `format --check`, agent-context ledger | clean |

All three required checks were already green before this round; pushing it re-runs them.

PR #1276, opened 2026-08-23
fix(gtk-host): restore the constructed default, not the declared one

JumpLink:
## What this is

A renderer that drops a prop hands `undefined`, and GObject cannot store it. `setProp` answered with the **ParamSpec's declared default**. That is the wrong source: what "removed" means is the value the widget would have carried had the prop never been authored — and that comes out of **construction**, not out of the class declaration.

## The measurement

gjs 1.88.1 / GTK 4.22.4 / Adw 1.10, over the 26 shipped descriptors: **953 scalar properties are both readable and writable, and 104 disagree** — 78 once construct-only `css-name` is set aside, since those route to `rebuild` and get the constructed value for free. Twelve property names carry it; four are behavioural:

| property | ParamSpec | constructed |
|---|---|---|
| `GtkWindow.visible` | `true` | `false` |
| `AdwActionRow.activatable` | `true` | `false` |
| `GtkToggleButton.receives-default` | `false` | `true` |
| `GtkListBox.focusable` (7 types) | `false` | `true` |

So `<AdwActionRow activatable={cond}>` with `cond` going `undefined` made a row that had never been activatable **into** one, and removing `visible` **showed a window** — silently, at exit 0, in the package that exists to refuse exactly that. Both are reachable from the Solid and Vue adapters today: Vue signals a dropped prop with `null`/`undefined` through `patchProp`, Solid through `setProp`.

## Why a probe, and why the JS accessor

There is no API for the question — a widget's `_init` runs arbitrary code, so constructing one is the only way to learn what construction leaves behind. It also cannot be read off a `GValue`: under GJS `probe.get_property(name)` is the raw two-argument call (*"At least 2 arguments required, but only 1 passed"*), which is why the first probe script cheerfully reported **0 mismatches over 1106 properties** while comparing nothing at all. The JS accessor is the channel.

One probe per GType, read once, dropped. A `Gtk.Window` probe is `destroy()`ed explicitly — GTK holds a reference to every toplevel, so dropping the JS reference alone would leak one per window GType for the life of the process.

**The probe skips `DEPRECATED` specs, and the suite is why.** Reading `Adw.ActionRow.icon-name` emits a deprecation warning per read, and `installDiagnosticsGate()` failed the new test. Correctly: a host that exists because GTK fails at exit 0 must not add noise to a consumer's render. Skipping costs nothing measurable — `icon-name` is the only deprecated property among the 11 that disagree on this path, and there `''` and `null` both mean no icon.

## Proof

`450 completed`, 0 failures. **A/B checked:** reverting `removedValue` to the ParamSpec turns all three new vectors red (`Expected values to match using ===`). Each vector also asserts *the disagreement itself* before asserting the answer, so a future GTK that changes one of these defaults fails by name instead of leaving the test vacuously green. The two round-trip vectors are opposite polarities (`activatable` true→false, `receives-default` false→true), so a "fix" that merely inverted a boolean satisfies neither.

The pre-existing `GtkLabel:label` removal test is kept and renamed: it is a property where the two sources **agree**, so it cannot tell them apart, and it now says so.

## Two things found beside it

- **`unknownTag` advised a route that cannot resolve it** — "or use a raw GType tag that is present in the installed typelib". `lookupWidget` is a `Map.get` and there is no `GObject.type_from_name` anywhere in the package, so a real installed GType lands right back on the same error. That is ADR 0028's decision rather than an omission (*"As a TAG it does not: createElement looks the GType name up exactly"*), so the message now says being a real GType is not enough — and a test pins `AdwClamp`, in the typelib with no descriptor, as still `unknown-tag`.
- **ADR 0027 § 4 said "thirteen codes today"; `errors.ts` holds eighteen.** Removed rather than corrected, per the precedent in `nativescript-bridge/AGENTS.md`: a hand-kept count is a second truth that drifts, and the file itself is the list.

## Checks run locally

`gjsify workspace @gjsify/gtk-host run build` · `run test` (450) · `gjsify format --check` (clean tree-wide) · `gjsify lint` (0 in this package) · `node scripts/audit-runtimes.mjs --check --strict` · `audit-test-scripts.mjs` · `check-adapter-import-direction.mjs` · `check-refs-citations.mjs`.

One note for review: `packages/framework` is over its advisory comment ceiling. It was already **507 lines over on `main`**; this adds 51 (90 comment lines against 104 code lines), and every one of them is an incident record or a measurement the repo's own rules say to keep. Flagging rather than trimming the incidents away.


JumpLink:
Superseded by #1284: your commit is carried unchanged, rebased onto #1283, plus the two things that only exist in the merge with #1281 — the bare `new klass()` probe is fatal for `AdwLayoutSlot` (g_error, SIGABRT, uncatchable), and `refuses a real installed GType that carries no descriptor` pinned `AdwClamp`, which the generated table has carried since #1281.

PR #1287, opened 2026-08-24
fix(gtk-host): close the two findings on the generated surface

JumpLink:
> Recreated from #1283, which GitHub auto-closed when its base branch was deleted by the #1282 squash-merge. Same commit, rebased onto `main`.

Both findings survived adversarial verification on #1281 and both are reproduced here.

## 1. The table shipped a row that KILLS THE PROCESS

`AdwLayoutSlot::constructed` calls `g_error("AdwLayoutSlot %p created without an ID")`. That is fatal **by contract** — SIGABRT, exit 134, a core dump — so there is no exception, no diagnostic, and nothing a `try` can do. `<adw-layout-slot>` with no `id` ends the consumer's process.

`descriptorProblems()` reports `[]` because it reads policy and never instantiates. Nothing in the 1746-test suite constructed a single one of the 164 generated widgets, which is exactly why a row like this was green.

**Measured**, bare-constructing all 164 rows one per process and resuming past the abort (gjs 1.88.1 / GTK 4.22.4 / Adw 1.10):

| rows | result |
|---|---|
| 163 | construct |
| 1 | `AdwLayoutSlot` — aborts |
| 0 | throw |

So the requirement is **curated as one entry**, not as a policy — GIR says `id` is construct-only and nullable, which is what it also says about properties that construct fine, and only running it tells them apart. `materialize` checks it **before** `g_object_new`, because for this GType there is no after:

```
<adw-layout-slot> cannot be constructed without id. In the installed library this is
not an exception but a g_error(): the process ABORTS (SIGABRT, exit 134, core dump)
with no catch and no diagnostic, so the host refuses here while a refusal is still
reportable. Author id on the element.
```

## 2. Three signal parameters carried uninitialised memory

`readSignal` never read `direction`. The non-`in` parameters of `Gtk.SpinButton::input`, `Adw.SpinRow::input` and `Gtk.Editable::insert-text` were therefore typed from their GIR type and read as ordinary inputs.

**Measured on gjs 1.88.1 / GTK 4.22.4**, by connecting and printing:

```
input:      new_value = 6.9526682391035e-310   typeof number
insert-text: position = 1711500784             typeof number
```

Nothing warns. A handler that reads either gets a plausible number.

They are emitted as `OutParam` — a nominal type nothing can be read out of and nothing but itself assigns to. **Declared**, not dropped, so a later `in` parameter cannot silently shift into the vacated position; and annotating `number` is now a compile error at the one position a reader would have looked.

**`caller-allocates="1"` is NOT rewritten, and that is the sharp edge.** `Gtk.Overlay::get-child-position` hands the handler a live `Gdk.Rectangle` to **fill** — measured as a real boxed wrapper, not garbage — and filling it is the entire purpose of the signal. A fix that blanket-typed every non-`in` parameter as unusable would break it, so a type-test asserts that line still **compiles**.

Across Gtk-4.0 and Adw-1 there are exactly four non-`in` signal parameters: three `caller-allocates="0"`, one `="1"`. The generated file changes by 4 lines.

## The mechanism, not the two fixes

`constructs every row it offers` — the suite now builds all 164 **through the host**, so `materialize` is exercised rather than bypassed. A future GType that aborts fails here with the tag on the last line, instead of a core dump nobody attributes; a future `requiresProps` entry that is not needed fails the `size` assertion.

`refuses a fatal construction by name instead of aborting` is its A/B in both directions: without `requiresProps` the first line does not throw, it ends the process — and the second line proves the entry is a *requirement*, not a ban, by building the slot with an `id`.

Two lines of `Gjs-CRITICAL: Attempting to run a JS callback during shutdown` are the price, and they are documented in place with the measurement: `installDiagnosticsGate` installs a process-global JS log writer and never takes it back. Disposing the widgets makes **eight** rows print it; leaving them alive makes **two** (`GtkColorChooserWidget`, `GtkEmojiChooser`). Neither is the host's doing, and skipping those two rows would trade the noise for the blindness this test exists to remove.

## Proof

- `1750 completed`, 0 failures (was 1746 + 4).
- `check-type-surfaces`: **13/13** negatives fire individually with the annotated code (was 12/12), 10 measurements, exit 0.
- `oxlint`: 0 errors. `gjsify tsc --noEmit`: clean.
- The generator is re-run, not hand-edited: `props.ts` moves by exactly the 4 expected lines.

### DIFF
--- a/docs/adr/0028-widget-table-provenance.md
+++ b/docs/adr/0028-widget-table-provenance.md
@@ -195,6 +195,29 @@ coercer and the verifier.**
    `<gtk-box orientation="vertical">` — the most-written GtkBox attribute there is —
    is a type error. The reader walks parents AND interfaces.
 
+10. **A table row is a claim that the widget can be BUILT, and one row was not.**
+    `AdwLayoutSlot` requires a construct-only `id`; without one, `constructed`
+    reaches `g_error()`, which is fatal by contract — SIGABRT, exit 134, a core
+    dump, and nothing to catch. GIR cannot tell that apart from any other
+    construct-only nullable property, and `descriptorProblems()` reads policy
+    without ever instantiating, so the row was green in a 1746-test suite that
+    constructed **none** of the 164. Bare-constructing all of them measures 163
+    clean and no throws at all, so the requirement is CURATED as one entry
+    (`REQUIRED_CONSTRUCT_PROPS`), refused in `materialize` by name, and the
+    construct-every-row test is what keeps the entry count honest.
+
+11. **A signal parameter has a DIRECTION, and three of them carry nothing.**
+    GIR marks `Gtk.SpinButton::input`, `Adw.SpinRow::input` and
+    `Gtk.Editable::insert-text` with a non-`in` parameter at
+    `caller-allocates="0"`. GJS still passes an argument there and it holds
+    uninitialised memory — measured, `new_value` arrives as `6.95e-310` and
+    `position` as `1711500784`, both perfectly ordinary numbers. Typing them from
+    the GIR type invites a reading that is garbage, and nothing warns. They are
+    emitted as `OutParam` instead: declared, so the following parameters do not
+    shift, and unreadable, so annotating `number` is an error at that position.
+    `caller-allocates="1"` is the opposite case and keeps its type —
+    `Gtk.Overlay::get-child-position` is handed a live `Gdk.Rectangle` to FILL.
+
 ## Consequences
 
 - The normal disagreement between table and runtime is **version skew** — a user's
--- a/packages/framework/gtk-host/src/attrs.ts
+++ b/packages/framework/gtk-host/src/attrs.ts
@@ -19,6 +19,28 @@ import type { HostNode } from './types.js';
  */
 export type NotifyHandler = (pspec: GObject.ParamSpec) => void;
 
+declare const outParam: unique symbol;
+
+/**
+ * A signal parameter GIR declares `out`/`inout` with `caller-allocates="0"`.
+ *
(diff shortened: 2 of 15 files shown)


JumpLink:
Two commits added after the first CI run, and the second is a real defect the first run exposed.

**CI failed with a count and no name.** `constructs every row it offers` reported

```
GTK reported 1 diagnostic(s) that would have passed at exit 0:
```

— followed by a **blank list**, over 164 constructions. The same sweep is silent on this workstation both with a session bus and with `DBUS_SESSION_BUS_ADDRESS` / `XDG_CURRENT_DESKTOP` / `XDG_SESSION_DESKTOP` / `DESKTOP_SESSION` stripped, so guessing which row it was would have been guessing.

**`test(gtk-host): name the row that made GTK talk`** — the gate is process-global and drains once per test, so a count is all it can give for a loop. Draining per row moves the same assertion inside and attaches the widget's name to it.

**`fix(gtk-host): never count a diagnostic it cannot name`** — the blank list was not a formatting accident. `MESSAGE` is a structured field like any other and a record may arrive without one; the writer answered `String(raw ?? '')`, so such a record was **recorded and then rendered as the empty string**. A gate whose whole purpose is to end guessing was counting a failure it could not describe.

A record with no usable `MESSAGE` is now rendered from the fields it does carry. Tested on the renderer directly rather than by emitting a record — `GLib.log_structured_array` wants `GLib.LogField` structs GJS cannot build from object literals (`not a subclass of GObject_Struct`) — across all three shapes that used to render as `''`: no `MESSAGE` key, `MESSAGE: null`, and `MESSAGE: ''`.

`1762 completed`, 0 failures. Whatever the container is emitting, the next run names it.

PR #1293, opened 2026-08-24
feat(gtk-host): render React onto the host ops

JumpLink:
The third adapter over one host: `react-reconciler`'s HostConfig mapped onto the
same host ops Vue and Solid use. `scripts/check-adapter-import-direction.mjs` now
reports **3 adapters carry no widget knowledge**.

## Phase 1 first: can react-reconciler run under GJS at all

Measured before writing a line of adapter, with a probe built by this repo's own
CLI (`--app gjs`) and run under `gjs -m`. **Nothing is missing.** react-reconciler
0.29.2 + react 18.3.1 + scheduler 0.23.2 load and render a tree even with
`--globals none`, i.e. with no gjsify polyfill injected at all:

| build | globals present | result |
|---|---|---|
| `--globals none`, production define | only `setTimeout`/`clearTimeout`/`console`/`process` from the GJS prologue | loads, renders, exit 0 |
| `--globals auto`, production define | `performance`, `queueMicrotask`, `MessageChannel`, `setImmediate`, `navigator`, … (24) | loads, renders, exit 0 |
| `--globals auto`, no define (dev bundle) | additionally `document` | loads, renders, exit 0 — and pulls four typelibs |

`scheduler` is fully feature-detected: `typeof performance === 'object' &&
typeof performance.now === 'function'` (else `Date.now`), `typeof setImmediate`,
then `typeof MessageChannel`, else `setTimeout(…, 0)`. So the historic
MessageChannel requirement is not one — and this repo ships
`@gjsify/message-channel/register` anyway, which `--globals auto` injects.

What Phase 1 *did* find is a build trap, and it is the reason for two flags:

```
--define 'process.env.NODE_ENV="production"'
--exclude-globals navigator
```

`react-reconciler/index.js` picks its bundle from `process.env.NODE_ENV`, and the
development one reaches for `document`, `HTMLCanvasElement` and `Path2D`; even the
production `scheduler` carries `typeof navigator !== 'undefined' &&
navigator.scheduling`. Each is a free identifier `--globals auto` answers with a
GTK-backed DOM register, pulling `gi://Gdk`, `GdkPixbuf`, `Pango` and
`PangoCairo` into a bundle that needs none of them — the same trap the Vue adapter
pays four defines for. A vector asserts `document` and `navigator` are both
`undefined`, so losing either flag fails the suite instead of silently growing
four typelib dependencies.

## The contract is READ, not restated

`react-reconciler` is `module.exports = function $$$reconciler($$$hostConfig)` and
its body opens by destructuring the whole config into module-scope variables. So
the HostConfig surface is not a documentation list — a recording `Proxy` reports
it exactly. Measured on 0.29.2, production bundle:

| | count |
|---|---|
| members the installed reconciler reads | **76** |
| answered by this adapter | **37** |
| gated off by `supportsPersistence: false` / `supportsHydration: false` / absent `supportsTestSelectors` / absent `supportsMicrotasks` | **39** |
| declared but read only by the DEVELOPMENT bundle (which reads 94) | **4** |

`react.spec.ts` asserts all three directions by name, so a version that starts
asking for something new fails a test instead of reading `undefined` inside a
commit — and a misspelled member sits in the declared-but-never-read set instead
of being invisible.

## Four measurements that shaped the mapping

Each has a vector, and each vector was A/B'd — the implementation was broken
deliberately and the named vector confirmed red, then restored.

- **React CLEARS the container before its first commit.** `updateHostRoot` sets
  the `Snapshot` flag whenever the previous render produced no child, and
  `commitBeforeMutationEffects` calls `clearContainer` — in the DOM to discard
  leftover markup. Here the container is a widget the application built and
  filled, so it destroys the SHADOW children only and never `el.foreign`. The
  vector *records the call*, because surviving chrome alone would also be
  explained by React never making it.
- **`removeChild` is a teardown**, like Vue's `remove` and unlike Solid's
  `removeNode`: React moves a node with `insertBefore`/`appendChild` alone. The
  reorder vector holds both halves — the same widget objects survive, and
  `removeChild` was never reached.
- **`render()` flushes synchronously.** A `ConcurrentRoot`'s updates are
  default-lane and go to `scheduler`; with no GLib main loop running, an unflushed
  first render leaves the container empty and says nothing. `flushSync` sets the
  current update priority to `DiscreteEventPriority`, which makes everything
  scheduled inside it a sync lane the same call flushes. A separate vector pumps
  `GLib.MainContext.default()` and proves the *scheduled* path works too — so a
  `setState` from a GTK signal handler lands on the next main-loop iteration, as
  it should in a real app.
- **`shouldSetTextContent` answers `false`, always.** `true` would leave the
  string in `props.children` for the adapter to write, which is a widget-knowledge
  test inside an adapter (ADR 0027 § 7). The host already routes a text node to
  the descriptor's `textSink` and refuses text by tag name where there is none.

## JSX: a separate subpath, and why

`./jsx-runtime` deliberately throws from `jsx()`/`jsxs()`, and that refusal is
correct for the dialects it serves — Solid's reactivity lives in
`babel-preset-solid`'s output, Vue's in its SFC compiler, so reaching those
functions means the pipeline was misconfigured and nothing would render. React is
the opposite case, structurally: its `jsx()` builds a plain, renderer-agnostic
element record and the host mapping happens later, in the reconciler. One module
cannot both refuse the automatic runtime and provide it, so React gets
`./react/jsx-runtime` (+ `./react/jsx-dev-runtime`, which is what TypeScript
appends under `jsx: "react-jsxdev"`): React's own factories, our own `JSX`
namespace over the generated tag list. `./jsx-runtime`'s refusal is unchanged and
a vector holds it in place.

Pointing `jsxImportSource` at `"react"` instead is the 208-tag trap ADR 0028 § 8
measured on Solid: the element list would come from `@types/react` and every HTML,
SVG and MathML tag would type-check clean on a GTK renderer and render nothing.

## Verification

- `gjsify run build:gjsify && gjsify run build:test && gjsify run test:gjs` →
  **1843 assertions from 8 gates, exit 0** (1791 before; 19 new vectors). Every
  vector runs under `installDiagnosticsGate()` through `gated()`, so a GTK
  critical fails the test rather than passing at exit 0.
- A/B: 8 deliberate breaks, 8 named vectors red, all restored —
  `clearContainer` sparing `foreign`, `removeChild`→teardown, removed-prop
  `undefined` vs `null`, `ref` vs the wrapper row, a dead HostConfig member, the
  main-loop pump, `hideTextInstance`, `shouldSetTextContent`.
- `gjsify tsc --noEmit` clean, `gjsify lint` clean, `gjsify format --no-write`
  clean.
- `scripts/check-adapter-import-direction.mjs` (this one caught a real defect: the
  first draft imported `../policies.js` for `addressOf`; it now reads the node's
  own `wrapper`), `scripts/check-type-surfaces.mjs`,
  `scripts/verify-package-outputs.mjs`, `scripts/audit-runtimes.mjs --check
  --strict` — all exit 0.

## Not done, deliberately

- **The React JSX surface has no half in `scripts/check-type-surfaces.mjs`.** Its
  element list *is* covered — same generated `WidgetPropsByTag`/`WidgetClassByTag`
  the `jsx` half holds negative-first — but the React-specific
  `Element`/`ElementType`/`IntrinsicAttributes`/`Ref` plumbing is not. That gate is
  built around two hardcoded halves and a third needs its own probes; a half
  without them is the checked-nothing shape the script exists to refuse. Recorded
  in `status/open-todos.md` rather than half-added.
- **`packages/framework` comment budget** goes from `-223` to `-511` (advisory,
  `--warn` in CI, and the tree was already OVER). Per-file ratios are in line with
  the siblings (`react.ts` 1.34 vs `vue.ts` 1.49); the ceiling is deliberately not
  raised, since raising it would also absorb the pre-existing overage.
- **Lockfile churn beyond the four new entries** (`@babel/types` de-nesting,
  `postcss`/`nanoid` consolidation) is the installer's own dedupe. Verified
  deterministic: a clean lock + the unmodified `package.json` reinstalls with zero
  diff, and the churn reappears identically once the react deps are added.

---

## Update after the first CI run

`Audit runtimes` came back red on one step — `check-agent-context-size`: the
`packages/framework/AGENTS.md` addition was +992 bytes over a ceiling that
ratchets to each file's current size. Fixed by shrinking first, then declaring
what survived:

- The AGENTS.md rule is now **+625 bytes** and carries only the two routing facts
  an agent must not re-derive: the `/react` build recipe and its consequence, and
  that the HostConfig is *read* from the installed `react-reconciler` through a
  Proxy rather than hand-listed. Every measurement — the 76/37/39 census, the four
  measurement-decided behaviours — moved one hop into the README, which that
  section already links.
- The remaining growth is **one reviewed line** in
  `status/agent-context-budget.json`, which is the form that gate asks for.
- The dropped export-name lesson landed in the README instead, next to its Solid
  sibling: TypeScript's automatic runtime emits `jsx`/`jsxs`/`Fragment` from
  `<jsxImportSource>/jsx-runtime` and `jsxDEV` from `/jsx-dev-runtime`, so a
  rename here is a consumer's `MISSING_EXPORT` — exactly what #1285 paid for on
  Solid's `setProp`.

**The process lesson, which is the more useful half.** I had run
`node scripts/audit-runtimes.mjs --check --strict` locally and it passed — but
that script is only *step 4* of that job, and the failure was in step 15. Running
the named script is not running the job. All **45** `run:` steps of
`audit-runtimes.yml` are now executed locally: 44 green, and the 45th is the
Windows-only PowerShell leg that needs `pwsh` on PATH (absent on this Fedora box,
and its real job — `Manifest checks (Windows)` — already passed on this PR).

**Rebased onto the new `main`** (`c42d7b6ce`), which had moved four commits ahead,
two of them in gtk-host (#1285 `setProp` export contract, #1286 generator nicks).
Clean rebase, no conflicts; `gjsify install` reproduces the lockfile with zero
further diff; `gjsify tsc --noEmit` clean, which is the cheap detector for the
duplicate-symbol shape a rebase produces without anyone typing it twice. Suite is
now **1848 assertions from 8 gates, exit 0** (1843 before the rebase — #1285/#1286
added 5).

`check-type-surfaces`, `check-adapter-import-direction`, `verify-package-outputs`,
`gjsify lint`, `gjsify format --no-write`: all green after the rebase.

## One more correction, in its own commit

`status/open-todos.md` still claimed the widget-table generator "does NOT exist
yet ... deliberately absent rather than stubbed" and named its four gates as
pending. All four landed with #1281 — `src/generator/`, `npm run generate`, and
`generator.spec.ts` carrying G1, G3 and G4, with G2 structural in `emit.mts` (the
generator never writes a curated field, so a contradiction is impossible rather
than merely checked). Deleted rather than struck through, per that file's own
rule; the post-generator state is already covered by the "138 of 164 generated
widgets have no measured placement rule" entry, and `generate-status.mjs` confirms
no dangling `open-todos:` anchor to the removed heading.

## Still to come

#1291 and #1292 are stacked ahead of this branch and both touch
`gjsify-lock.json` (#1292 also adds `@babel/*` + `babel-preset-solid`, the
`packages/infra/rolldown-plugin-solid` package, a `bundler-plugins` conformance
rule and a `status/status.json` entry). Once they land I will rebase again and
resolve the lockfile by re-running `gjsify install` rather than hand-merging, then
re-run `gjsify tsc --noEmit`, and keep both `status/` additions.


JumpLink:
Probe-merged this onto the top of my stack (`feat/solid-jsx-build-step` → #1292 → #1291) rather than reasoning about it, because both branches touch `gjsify-lock.json`.

**Only the lockfile conflicts**, and re-running `gjsify install` resolves it — don't hand-merge it. Everything else applies cleanly (`packages/framework/AGENTS.md`, `gtk-host/README.md`, `package.json`, `test.mts`, `status/open-todos.md` all auto-merged).

On the merged tree:

| gate | result |
|---|---|
| `gtk-host` suite | **1848 completed**, exit 0 |
| `gjsify tsc --noEmit` | clean |
| `check-type-surfaces` | exit 0 |
| `check-adapter-import-direction` | exit 0 |
| `audit-runtimes --check --strict` | exit 0 |
| `gjsify lint` | exit 0 |
| solid showcase probe | `PROBE: PASS {"rows":["Row 2","Clicks"],"count":1,"diagnostics":0,"tree":81}` |

One correction to how I first read that, worth stating because it is the trap this repo keeps eliminating: **`audit-runtimes --check --strict` exiting 0 does not cover the `Audit runtimes` job.** `scripts/check-agent-context-size.mjs` is a separate step, and on the merged tree it still reports

```
packages/framework/AGENTS.md   26579   25587  OVER
```

so that failure is real and reproduces outside CI — run that script directly, not only the audit.

JumpLink:
**One item for the rebase after #1291/#1292 land** — recording it here so it does not depend on anyone's memory.

#1292 changes `jsx-runtime.ts` from `export type Element = ElementChild` to `= HostNode`, because Solid's control-flow components (`<For>`, `<Index>`, `<Show>`) are typed against `HostNode` and the first `<For>` in any app was otherwise a type error.

This PR's `react-jsx-runtime.ts` says `Element = ReactElement`. **That divergence is correct and must not be "fixed" for consistency.** A React JSX expression really does evaluate to a `ReactElement`, and React's control flow is plain JS rather than typed components — so the two surfaces answer two different questions. After the rebase the two files will visibly disagree, which is exactly when someone reaches for a one-line consistency change.

Please add that as a comment on the `Element` line in `react-jsx-runtime.ts` during the rebase, naming the Solid sibling and the reason. A rule without its reason gets simplified back into the bug.

(Verified separately: #1292 leaves `AUTOMATIC_RUNTIME_UNSUPPORTED` untouched, so this PR's assertion on that refusal survives the rebase.)

PR #1334, opened 2026-08-26
ci: cancel a PR's superseded runs on every push, not when the next one starts

JumpLink:
## The deadlock, measured twice

`main.yml` already sets `cancel-in-progress` for `pull_request`, but that fires when the
**superseding run starts** — and starting is exactly what a saturated queue prevents.

**2026-08-26:** five parallel PRs (#1329–#1333) filled the runner pool. Run
[32937220245](https://github.com/gjsify/gjsify/actions/runs/32937220245) — branch
`fix/prebuilds-musl-leg`, commit `1fc2ebbdc`, superseded hours earlier — sat `queued`
holding the `main-pr` concurrency group. Its successor could not start, so
`cancel-in-progress` never fired, and `CI gate (GJS)` was never green on that branch until
a human cancelled the run by hand.

`cancel-pr-runs.yml`'s own header had already diagnosed this in 2026-08 while covering only
the merged-PR half of the queue.

## Extended, not doubled

`cancel-pr-runs.yml` gains `synchronize` beside `closed`. It was already the mechanism that
supersedes by an **API call** instead of by a concurrency group, which is the property that
matters: one small job takes a runner long before a 20-job matrix does, so the cancel no
longer depends on the queue having room for the work it unblocks. For `audit-runtimes.yml`,
`build-ci-image.yml` and `commitlint.yml`, which carry no `concurrency:` block at all, it is
now the only superseding mechanism there is.

A second workflow would have meant a second copy of the same six-line API dance and a second
place to get the demarcation wrong. Per `simplicity`, the addition also pays for itself in
deletions: the inline `--jq` selection is **gone** from the `run:` block, and both events now
run through one function instead of one filter that only the `closed` path ever used.

## The `synchronize` window, and why it is not wider

Both events cancel *"the non-completed runs of this PR's head branch that existed at time
T"*. They differ only in T and in what is exempt:

| event | T | exempt |
|---|---|---|
| `closed` | `closed_at` | nothing |
| `synchronize` | `updated_at` | the SHA that is now the head |

Matched by head **branch**, not by head SHA — a force-pushed PR has runs under several SHAs
and a SHA match would leave the older ones running, which is half the backlog this drains.

The `T` bound is load-bearing in two separate ways, and this is where getting it wrong costs
somebody a valid measurement:

1. **Branch names in this repo are reused** (agent worktrees and the release tooling both do
   it), so without the bound a push could cancel a fresh run belonging to whatever reopened
   that name.
2. **Ordering.** Two pushes land in quick succession and the first push's cancel job only
   gets a runner after the second. Its `keepSha` is the *first* head, so every other rule
   agrees that the second push's runs — the current ones — are stale. The bound is the only
   thing that saves them.

Unchanged from the `closed` path and deliberately so: `pull_request` and never
`pull_request_target` (`actions: write` must not be paired with head code; a fork PR's token
is read-only and the step reports that rather than failing), the self-run exclusion, and the
`|| true` per cancel because a run that finishes between the list and the cancel returns 409
— the success case arriving a second early.

The checkout is pinned to the head SHA rather than left on the default merge ref, which is
not guaranteed to still resolve once the PR is closed, and is sparse to the one script, so
the job holding `actions: write` materialises as little head code as it can.

## Proof

The selection is lifted into `scripts/select-superseded-runs.mjs` and put under fixtures in
`tests/e2e/ci-cancel-superseded-runs` (14 tests, registered in `test:e2e`). A workflow
cannot be exercised before it runs for real, on real runs, so the decision is the part that
has to be testable.

Every test was A/B'd: green with the logic, red with the logic deliberately broken. 12 of 13
mutations flip exactly one test each.

| mutation | verdict | test that caught it |
|---|---|---|
| drop the head-SHA exemption | 🔴 | never cancels the commit that is now the head |
| drop the branch match | 🔴 | does not reach a run on another branch |
| drop the time bound | 🔴 | leaves a run that did not exist when this push happened · still stops at the moment the PR closed |
| drop the `completed` filter | 🔴 | does not touch a run that has already completed |
| drop the self-run exclusion | 🔴 | does not cancel the job doing the cancelling |
| exempt the head on `closed` too | 🔴 | cancels the head run too … · reads the close bound from `closed_at` … |
| treat an unparseable timestamp as `0` | 🔴 | refuses a bound it cannot parse instead of selecting nothing |
| accept a missing own run id | 🔴 | refuses to run without its own run id |
| accept any `pull_request` action | 🔴 | refuses an action it has no window for |
| stop unwrapping the API response | 🔴 | takes the event from the environment and the runs from stdin |
| drop the `synchronize` trigger | 🔴 | triggers on both events and invokes this selection for them |
| stop invoking the selection from the workflow | 🔴 | triggers on both events and invokes this selection for them |
| bound the push window by `closed_at ?? updated_at` | ⚪ | *not a defect* — a `synchronize` payload has no `closed_at`, so the fallback restores the same value |

**Two tests were green for the wrong reason on the first draft, and the A/B is what found
them.** "never cancels the commit that is now the head" put the new run's `created_at` four
seconds past the bound, so the *clock* excluded it and the exemption under test was never
consulted; and the self-run case sat in the push window, where `keepSha` covers this job's
own run anyway. Both fixtures now sit exactly on the bound, and the self-run test moved to
the `closed` window — the one where the id check is the only thing between this job and
cancelling itself halfway through the list.

`scripts/check-workflow-run-syntax.mjs` reads the new `run:` block: verified by A/B too —
an unbalanced `(` in the loop makes it exit 1 naming this file and step.

## Not affected

- **`pr-trigger-parity`**: `cancel-pr-runs.yml` has no `push` trigger, so its verdict is
  `not-gated` before and after — confirmed by running `triggerParityRows()` over the tree,
  and the suite is still 13/13 green. The rule is about post-merge coverage a PR cannot
  offer; a `types:` list is not a coverage claim and the rule says so.
- **`docs/ci-selective.md`**: not touched. This mechanism's documentation is the workflow's
  own header, which is where both incidents live; a paragraph there would be a second copy
  of it.

## Deliberately left open

- `scripts/check-comment-budget.mjs` reports `tests` at `-734` where it was already `-706`
  on `main`. Advisory (`--warn` in `audit-runtimes.yml`), and every comment added is the
  justification for one cancel decision.
- `per_page=100` and no pagination — inherited from the `closed` path. A PR whose branch has
  more than 100 in-flight runs behind it is not the case this drains.
- The four workflows with no `concurrency:` block still have none. Naming them in the header
  is what this PR does about it; giving them groups is a separate decision.


JumpLink:
Live verification, on this PR itself: the force-push that rebased onto `main` fired the new `synchronize` trigger, and [run 32948970888](https://github.com/gjsify/gjsify/actions/runs/32948970888) took a runner **~15 s after the push** while eight of this PR's own workflows sat `queued` behind it — which is precisely the gain over `cancel-in-progress`. Checkout, sparse pattern and `node scripts/select-superseded-runs.mjs` all worked on the runner; the selection came back empty and the step reported `No superseded runs for this event — nothing to cancel.` (correct: the pre-rebase commit never spawned runs, the PR was `CONFLICTING` until the rebase).

Rebased onto #1330, whose rewrite of `scripts/check-workflow-run-syntax.mjs` landed meanwhile. Re-verified against that new version, same A/B: an unbalanced `(` in the cancel loop makes it exit 1 naming `cancel-pr-runs.yml` — step "Cancel the runs this event superseded" (shell: bash).

PR #1340, opened 2026-08-26
feat(manifest-conformance): hold the ADR 0019 library boundary

JumpLink:
Answers a direct question — *ts-for-gir is meant to be used as a library, so can more of gjsify move there?* — with measurements, and lands the two things that were cheap and unambiguous. The ts-for-gir half is gjsify/ts-for-gir#439.

## The answer, in one line

**ts-for-gir knows GIR as XML** — parsed headlessly, in CI, with no GTK installed and no typelib loaded. **gjsify knows GI as a loaded runtime.** Every introspection-carrying module in this repo sorts cleanly on that one question, and the sort is not a judgement call: a module that needs an installed library cannot move to a generator that runs without one.

## Candidates, audited

Code lines exclude comments and blanks — several of these files are 60 % comment, and counting prose measures the incident record rather than the coupling.

**Moves, and it is all one thing already decided** (ADR 0029 steps 3–5):

| module | code | GIR-generic | what holds it here |
|---|---:|---:|---|
| `generator/gir.mts` | 249 | 248 | one import, `@gjsify/domparser` |
| `generator/surface.mts` | 196 | 196 | nothing — imports only its two siblings |
| `generator/tsmap.mts` | 100 | 100 | nothing |
| `generator/mini.fixture.mts` | 59 | 59 | nothing |
| `generator/emit-types.mts` | 156 | ~151 | `tagOf` + emitted `../attrs.js`; the dialect half stays |
| `generator/emit.mts` | 103 | 13 | the runtime table is gtk-host's model (ADR 0028 § 1) |

The first four are a closed subgraph — 604 code lines whose only edge outside `generator/` is a single XML-parser import. `tsmap.mts`'s `GIRS_PACKAGES` table (`Gtk` → `@girs/gtk-4.0`) is **ts-for-gir's own naming convention, written on the wrong side of the boundary.**

**Does not move, same reason each time** — full of GI vocabulary, but each needs something a headless generator does not have: `system-gi-dirs.ts` (44/43), `cli/utils/system-gi.ts` (77/77), `node-gi/system-gi.js` (70/70), `gi-typelib.ts` (80/80), `typelib-backers.mjs` (170/~155, parses the typelib *binary* header), both `parseGiSpecifier` copies (22/22 — `gi://` is a GJS import specifier, not a GIR concept), `repo.cc` (245, `gi_repository_require` *is* the runtime), `conformance/` + `registry.ts` (walk the live GObject type system), `docs/gnome-mappings.md` (0 — a polyfill choice table).

`generated.spec.ts` belongs in that second list for the sharpest version of the reason: it asks the *installed* typelib whether every emitted name is real.

**The three `systemGiLibraryDirs()` copies are a real duplicate, and ts-for-gir is not its home.** They exist three times because ADR 0005 Decision 2 forbids a Tier-1 package a `dependencies` edge on `@gjsify/node-gi` — a tier rule, not a technical obstacle. Answering "can it move to ts-for-gir" with yes would export a runtime concern into a generator to dodge a tier rule. The already-tracked shared `@gjsify/system-gi` stays right.

## The dependency-direction check found a hole

ADR 0019 Decision 1 keeps every `@ts-for-gir/*` and `@gi.ts/*` package build-step-free, and the registry confirms it — `@ts-for-gir/lib`, `@ts-for-gir/cli` and `@gi.ts/parser` all resolve `exports["."]` to `./src/index.ts`. So a `dependencies` edge from a published `@gjsify/*` package onto one of them **ships raw TypeScript to every consumer.**

Nothing enforced that. `tier.mjs` collects only `dep.startsWith('@gjsify/')`, so an external `@ts-for-gir/*` edge in `dependencies` was invisible to it, and no other manifest-conformance rule inspects external dependency names — the edge would have passed `audit-runtimes --check` in silence.

The `tier` rule now fails on it **by name**, the same shape it already uses for ADR 0005's node-gi isolation and for the same reason: these are external packages carrying no `gjsify.tier`, so the tier walk structurally cannot see them. Not a new rule and not a fourth required check — one more failure inside the rule that already special-cases a package by name.

- Green on all 202 published packages (whose eight existing `@ts-for-gir/*` edges are all `devDependencies`, the sanctioned seam).
- Red on an injected `dependencies` edge, naming the package, the field and the seam.

**And the good news it makes explicit:** the prize costs no new dependency edge at all. The widget surface arrives as generated `@girs/*` — `.d.ts` plus runtime `.js` — and `@gjsify/gtk-host` already declares eight `@girs/*` packages. ADR 0029 steps 3–5 add zero `@ts-for-gir` dependency.

## Is the rollback still blocked? Yes, and it is live-checkable

`npm view @girs/gtk-4.0 exports --json` re-measured today: still **four** keys, no `./surface`. ts-for-gir is at 4.1.0 with tag `v4.1.0` already cut, so ts-for-gir#438 is unreleased code.

`status/open-todos.md` said what was missing and how to detect arrival, but not how to *cause* it. It now carries the ordered sequence, which hangs off **one human action** — none of it a gjsify PR:

1. A ts-for-gir maintainer cuts 4.2.0 from `main`. `.release-it.json` has `npm.publish: false`; the cut only tags and opens a GitHub release.
2. `release: published` fires `release-types.yml` → `build:types:release` → pushes regenerated `@girs/*` to `gjsify/types@main`. Nothing needs enabling: `.ts-for-gir.packages-all.rc.js` already sets `widgetSurface: true`, and the package template emits `./surface` conditionally on `girModule.hasWidgetSurface`.
3. That push fires `gjsify/types`' own `Release CI` (`on: push: branches: [main]`), which publishes every `@girs/*` to npm. **This is the step that makes the subpath installable, and it is triggered by the push rather than by the release** — so a cut whose `release-types.yml` leg fails publishes `@ts-for-gir/*` and no new `@girs/*`, which looks like a successful release from the tag.
4. Probe *both*: `npm view @girs/gtk-4.0 version` must read 4.2.0 **and** `exports --json` must show the fifth key. Step 3 is where the two can come apart.

**The same release retires a second thing.** `checkTypeSkew`'s `isDegenerate()` detects the `@girs` namespace-version-as-release fallback that ADR 0019 Decision 3 removed in ts-for-gir#436, also unreleased. Once a released `@girs` omits `libraryVersion` where the library declares none, that detector reads for a shape that can no longer be published — it goes away rather than moving anywhere.

## The 97-vs-889 number

`gir.mts` said 97 and ts-for-gir said 889, over the *same* 705 GIRs — this repo's generator literally reads ts-for-gir's `girs/`. Measured both ways: **both are right**, and they differ by exactly the 792 members whose GIR name carries uppercase and whose nick preserves it, which disagree only with a derivation that also lowercases. `97 + 792 = 889`. `nickOf` does not lowercase, so 97 is its number; the pair only looked like a contradiction because the derivation was never named beside the count.

Per the convention #1328 set, the count moves to ts-for-gir — which owns the corpus and now asserts it in `check:girs` — and this comment keeps what it establishes: some nicks retain an underscore the substitution would have replaced, so only the attribute knows which one survived. Full detail, including the invariant that made ts-for-gir's `toLowerCase()` a measured defect, in gjsify/ts-for-gir#439.

## Checks

`audit-runtimes --check` ✅ (tier summary now names the ADR-0019 boundary) · `check-adr-index` ✅ · `check-doc-fences` ✅ · `check-agent-context-size --check` ✅ · `check-doc-revert` ✅. No AGENTS.md touched, so no budget move.

### DIFF
--- a/docs/adr/0019-ts-for-gir-as-library.md
+++ b/docs/adr/0019-ts-for-gir-as-library.md
@@ -137,7 +137,9 @@ it: `@ts-for-gir/lib`, `@ts-for-gir/cli` and `@gi.ts/parser` all resolve
 `exports["."]` to `./src/index.ts`. So a **`dependencies` edge from a published
 `@gjsify/*` package onto one of them publishes raw TypeScript to everyone who installs
 the depending package.** The sanctioned seam is a `devDependency` that gjsify bundles,
-which is what all eight existing edges in this repo already are.
+which is what all seven published edges in this repo already are (all of them on
+`@ts-for-gir/cli`; `@ts-for-gir/lib` itself appears only under the private
+integration test, which declares no tier and publishes nothing).
 
 Nothing enforced that. `scripts/manifest-conformance/rules/tier.mjs` collects only
 `dep.startsWith('@gjsify/')`, so an external `@ts-for-gir/*` edge was invisible to it,
--- a/docs/adr/0019-ts-for-gir-as-library.md
+++ b/docs/adr/0019-ts-for-gir-as-library.md
@@ -129,6 +129,44 @@ is still open".
 Decision 2 is **independent** of this: declaring `gjsify.prebuilds` adds a `.gir`
 to the tarball, not a dependency edge.
 
+### 4. The boundary Decision 1 implies is now enforced, not trusted
+
+Added 2026-08-26, while auditing what else could move to ts-for-gir. Decision 1 keeps
+every `@ts-for-gir/*` and `@gi.ts/*` package build-step-free, and the registry confirms
+it: `@ts-for-gir/lib`, `@ts-for-gir/cli` and `@gi.ts/parser` all resolve
+`exports["."]` to `./src/index.ts`. So a **`dependencies` edge from a published
+`@gjsify/*` package onto one of them publishes raw TypeScript to everyone who installs
+the depending package.** The sanctioned seam is a `devDependency` that gjsify bundles,
+which is what all eight existing edges in this repo already are.
+
+Nothing enforced that. `scripts/manifest-conformance/rules/tier.mjs` collects only
+`dep.startsWith('@gjsify/')`, so an external `@ts-for-gir/*` edge was invisible to it,
+and no other conformance rule inspects external dependency names — a package could
+have taken that edge and passed `audit-runtimes --check` silently. The `tier` rule now
+also fails on it, by name, the same shape it already used for ADR 0005's node-gi
+isolation and for the same reason: these are external packages with no `gjsify.tier`,
+so the tier walk structurally cannot see them.
+
+**This does not restrict where GIR-derived DATA may flow.** `@girs/*` packages ship
+`.d.ts` plus runtime `.js` and are unaffected — which is why ADR 0029's migration of the
+widget vocabulary into `@gjsify/gtk-host` costs no new dependency edge at all: the
+surface arrives as generated types on a package gtk-host already depends on.
+
+### 5. What the boundary answers about migrating more code here
+
+The audit behind Decision 4 sorted every introspection-carrying module in gjsify on one
+question, recorded in `status/open-todos.md` § "What else could move to ts-for-gir":
+**ts-for-gir knows GIR as XML** — parsed headlessly, in CI, with no GTK installed and no
+typelib loaded — while **gjsify knows GI as a loaded runtime**. A module that needs an
+installed library cannot move to a generator that runs without one, so the whole
+typelib-loading cluster (`systemGiLibraryDirs()` and its two mirrors, `gi-typelib.ts`,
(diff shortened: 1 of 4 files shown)


JumpLink:
Pushed a correction to two of my own numbers, both the class this repo already names:

- The candidate table carried a second **GIR-generic** count beside ADR 0029's accepted one, differing on four of seven rows — because it was silently answering a different question (*how many lines mention gjsify*) under the same heading. ADR 0029 now owns the code/GIR-generic columns and the second question gets its own. The pair is more useful than either alone: `emit.mts` references gjsify **nowhere** and still cannot move, because only 13 of its 103 lines are GIR-generic. A module can be free of gjsify imports and still be entirely about gjsify.
- **Seven** published packages carry a `@ts-for-gir/*` edge, not eight — all on `@ts-for-gir/cli`, all `devDependencies`. The eighth was the private integration test, which declares no tier and publishes nothing, so it is not evidence about published packages either way.

PR #1379, opened 2026-08-28
fix(adwaita-web): define the view before its pages

JumpLink:
`customElements.define` upgrades every matching element already in the document,
immediately, so the two calls at the foot of
`packages/web/adwaita-web/src/elements/adw-tab-view.ts` are a sequence and not a pair.
With the PAGE defined first, every declared `<adw-tab-page>` ran its
`attributeChangedCallback` while its `<adw-tab-view>` parent was still an ordinary
`HTMLElement` — and that callback reaches for the parent.

## Measured, before and after

Uncaught page errors on the built documentation site, same build pipeline both times:

| page | before | after |
|---|---|---|
| `/getting-started/` — five `CommandTabs` windows, 19 declared pages | **19** | **0** |
| `/adwaita/theming/` — declares no tab markup (the control) | 0 | 0 |

One error per declared page, all reading
`this.closest(...)?.syncDeclaredPage is not a function`. The control is what makes the
count attributable rather than a number.

After the rebuild the widget still behaves: the first `CommandTabs` window renders
`GJSify · Node · Bun · Deno`, and clicking chip 2 moves `selected` to 2 and swaps the
visible panel.

## Two halves, because the order was only the trigger

1. **The define order.** The view is registered first, so a page never meets an
   un-upgraded parent through this file.
2. **The guard.** `view?.syncDeclaredPage(...)` conflated a MISSING ancestor with one
   that is present but not yet upgraded, and the cast to `AdwTabView | null` asserted
   the second away. `instanceof` narrows for real and keeps TypeScript honest about the
   call. Dropping the notification for an un-upgraded view is correct rather than merely
   quiet: such a view has not adopted the page yet, and `_adoptDeclaredPage` reads every
   attribute in `PAGE_ATTRIBUTES` straight off the element when it does.

Either half alone would clear the site; both are here because the order is the cause and
the guard is the class.

## The other two orderings, walked

- **A page appended by script into a live view** carries no `data-page-id`, so
  `syncDeclaredPage` returns without touching the model. Out of contract — the imperative
  surface is `appendPage` — but quiet, before and after. Pinned by a second test.
- **A view whose pages reach the parser first** cannot happen: the definitions live in an
  ES module, which is deferred until parsing is done. The only reachable orders are
  parse-then-define and create-after-define.

## The tests, red before and green after

- `tests/browser/specs/adwaita-upgrade-order.spec.ts` parses the declared markup and only
  then loads the package — the one order no spec inside the bundle can reach, because by
  the time a spec runs both names are registered. **3 uncaught errors before, 0 after**,
  plus chips and panel visibility, so a guard that merely swallowed the notification would
  not pass either.
- `packages/web/adwaita-web/src/adw-tab-view.spec.ts` stands a page beside a view that
  `document.implementation.createHTMLDocument()` left un-upgraded. A custom-element
  reaction never throws INTO its caller — the exception is reported — so the assertion
  runs off a window `error` listener.

Both files call out the trap: building the two elements with `document.createElement`
after the module has run passes either way, because the race only exists for elements
already in the document when `define` runs.

## Checks

`npx oxfmt --check .` · `npx oxlint packages/web/adwaita-web/src` ·
`check-node-test-registration` · `check-browser-test-registration` ·
`audit-runtimes --check` · `check-adwaita-conformance-drivers` ·
`gjsify tsc --noEmit` in `@gjsify/adwaita-web` and in `tests/browser` — all exit 0.
`npx playwright test --project=chromium` — 4 passed (only Chromium is installed on this
machine; Firefox is the CI default).

https://claude.ai/code/session_01S9HRKqQZskrYLVer1fuFvg

### DIFF
--- a/.github/workflows/audit-runtimes.yml
+++ b/.github/workflows/audit-runtimes.yml
@@ -499,6 +499,16 @@ jobs:
       - name: Check declared keyboard affordances have an implementation
         run: node scripts/check-adwaita-keyboard-contract.mjs
 
+      # An element that reaches for an ANCESTOR is registered after it. `define` upgrades
+      # matching elements in the document IMMEDIATELY, so `adw-tab-page` before
+      # `adw-tab-view` ran every declared page's attributeChangedCallback against a parent
+      # that was still a plain HTMLElement: 337 uncaught errors over 22 built pages, and
+      # the identical shape sat unnoticed in adw-navigation-view.ts. With the `instanceof`
+      # guard in place the ORDER is unobservable — page-first renders identically and the
+      # whole browser suite stays green, measured — so a spec cannot hold it and this does.
+      - name: Check ancestor-reaching elements are registered after their ancestor
+        run: node scripts/check-adwaita-upgrade-order.mjs
+
       # Three READMEs promised the three storybook targets "compare 1:1 by screenshot"
       # (#1052) and no such harness existed. It was not built — three rasterisers cannot
       # be compared pixel-for-pixel, and the NativeScript leg needs a device CI has not
@@ -999,6 +1009,16 @@ jobs:
       # arms and the ledger are in the script header.
       - name: Check declared keyboard affordances have an implementation
         run: node scripts/check-adwaita-keyboard-contract.mjs
+
+      # An element that reaches for an ANCESTOR is registered after it. `define` upgrades
+      # matching elements in the document IMMEDIATELY, so `adw-tab-page` before
+      # `adw-tab-view` ran every declared page's attributeChangedCallback against a parent
+      # that was still a plain HTMLElement: 337 uncaught errors over 22 built pages, and
+      # the identical shape sat unnoticed in adw-navigation-view.ts. With the `instanceof`
+      # guard in place the ORDER is unobservable — page-first renders identically and the
+      # whole browser suite stays green, measured — so a spec cannot hold it and this does.
+      - name: Check ancestor-reaching elements are registered after their ancestor
+        run: node scripts/check-adwaita-upgrade-order.mjs
 
       # Three READMEs promised the three storybook targets "compare 1:1 by screenshot"
       # (#1052) and no such harness existed. It was not built — three rasterisers cannot
--- a/packages/web/adwaita-web/src/adw-tab-view.spec.ts
+++ b/packages/web/adwaita-web/src/adw-tab-view.spec.ts
@@ -75,6 +75,59 @@ export const AdwTabViewTest = async () => {
         });
     });
 
+    // ONE HALF of the upgrade-order defect; the other half is the define order at the
+    // foot of `elements/adw-tab-view.ts`, which no test in this realm can observe —
+    // both names are already registered by the time a spec runs, and `define` is what
+    // upgrades. What IS reproducible here is the STATE that order produced: a real
+    // `<adw-tab-page>` next to an `<adw-tab-view>` that is still an ordinary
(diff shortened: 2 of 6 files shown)


JumpLink:
Adversarial review, measured against a rebuilt site and mutated code. Four commits pushed. Three claims in the description above are wrong and should be corrected in the squash body — left as a comment rather than a body edit, because editing the body re-queues commitlint and has blocked a merge here before.

**1. `/adwaita/theming/` is not a control.** It never loads adwaita-web: `customElements.get('adw-tab-view')` is `undefined` there and the bundle is never requested. Its 0 would stay 0 with the widget on fire. The real control is `/adwaita/` — it loads the package and declares no tab markup. Its only two errors are Starlight's own duplicate `define`s, identical before and after.

**2. The reach was 22 pages, not one.** Rebuilding the site with and without the fix and counting uncaught page errors on every page that declares tab markup: **339 → 2**, of which 337 → 0 are `syncDeclaredPage is not a function`. `/adwaita/boxed-lists/` 36, the home page 32, `/guides/install/` 22, `/getting-started/` 19 (matching your count exactly).

**3. "A guard that only swallowed the notification would leave the tabs untitled" is false.** Stubbing `syncDeclaredPage` to return immediately still renders One/Two/Three — on the declared path the titles come from `_adoptDeclaredPage`, never from the notification. That mutation is caught by the LIVE-title test in `adw-tab-view.spec.ts`, not by the new browser spec.

**The load-bearing claim holds.** "Dropping the notification is correct, not merely quiet" is confirmed independently: on the pre-fix build, with all three notifications dropped by the error, the chips still read `One/Two/Three` and `nPages === 3`. Nothing arrives only through `syncDeclaredPage`. `_adoptDeclaredPage` reads all seven of `PAGE_ATTRIBUTES`, and the model's setters short-circuit on equal values, so the re-notification after the reorder emits no extra `page-updated` (measured: 0 before, 0 after).

**`instanceof` across realms is safe here, measured.** The built site contains exactly one `customElements.define('adw-tab-view')`, in one shared Vite chunk. Loading the bundle twice makes the second copy's `define` throw on the first name it reaches, aborting its evaluation — so only one class copy is ever live. There are no iframes in `website/src` or the storybook; and a view+page pair adopted out of a real second-realm iframe stays internally consistent (the page's callback closes over its own realm's view), so the rename still propagates.

**Two things the tests did not hold, now fixed.**

- The define order had no test. Page-first *with* the `instanceof` guard passes every spec, `adwaita-upgrade-order.spec.ts` included. The guard makes the order unobservable, so no spec can hold it — `scripts/check-adwaita-upgrade-order.mjs` does, and it is mutation-tested against all four shapes.
- The same defect was live in `adw-navigation-view.ts` — `AdwNavigationPage` reaches `closest('adw-navigation-view')` with the same cast and the same define order, reproducing at four uncaught `syncPageProperty is not a function`. The site never showed it only because the gallery clones navigation markup out of a `<template>`.

**One claim that remains wrong, reported rather than fixed:** *"A view whose pages reach the parser first cannot happen."* It can. An inline classic script starting `import()` during a streamed parse reaches a mid-parse `define`; the view then runs `replaceChildren` while the parser's insertion point is still inside it, and every page parsed afterwards is stranded — no chip, no panel, not `hidden`, `nPages` short by one. Measured on both builds: pre-fix it strands *and* throws twice; post-fix it strands **silently**. Pre-existing and unchanged by this PR, but it is a silent wrong render at exit 0 and wants its own issue.

PR #1402, opened 2026-08-28
feat(cli): let onboard select @girs/* via --include

JumpLink:
`gjsify onboard` hard-excluded `@girs/*`. In this repo that is right — those are third-party type packages nobody here publishes. In ts-for-gir it is exactly backwards: its `@girs/*` workspaces **are** its publishables, and `gjsify/types` releases them over OIDC.

## The gap

npm Trusted Publishing cannot create a package that does not exist yet, so a new `@girs/*` name needs a first publish before it can be trusted. The release script in `gjsify/types` says so in a comment, and names this command as the way to do it:

> the first publish of a new `@girs/*` namespace needs this or a prior `gjsify onboard` sweep

A documented path the command could not walk.

Not hypothetical: GNOME 51 realigned Mutter's API version with the GNOME major (18 → 51 in one step), so a single cycle mints `@girs/{clutter,cogl,meta,metatest,mtk,shell,st}-51` plus whatever new libraries the release carries. Thirteen names this time. Each unbootstrapped one fails its publish with a 404 that reads like a broken trusted publisher.

## The change

`--include` takes name globs. Passing **any** pattern also lifts the default `@girs/*` exclusion — otherwise `--include '@girs/*'` would select nothing and the sweep would report *"no publishable workspace packages found"*, a bootstrap command announcing its own gap as closed. An explicit `--include` is a whitelist, so nothing else comes along either.

The empty-selection message now names the filter, so a mistyped glob says what it did not match instead of claiming the repo has no packages.

## Test

One E2E test asserts **both** halves against the mock registry: `@girs/thing-1.0` absent from the default sweep, present under `--include '@girs/*'`, and no `@onb/*` alongside it. 7/7 pass.

```
✔ publishes + trusts only the missing packages, reusing ONE OTP; skips the done one
✔ is idempotent — a second run does nothing and exits 0
✔ --dry-run reports the plan and changes nothing
✔ skips @girs/* by default, and --include '@girs/*' selects it
✔ gates login on whoami — a LIVE token does not trigger login
✔ gates login on whoami — a DEAD token triggers the login flow, then proceeds
✔ --yes fails clearly when a dead token would need an interactive login
```

https://claude.ai/code/session_016ci5gmUojZSbm5JXkgwYKj

### DIFF
--- a/packages/infra/cli/src/commands/onboard.ts
+++ b/packages/infra/cli/src/commands/onboard.ts
@@ -10,7 +10,14 @@
 //   1. Auth gate — `whoami` first. If the token is live, proceed without asking
 //      for credentials; only when it is dead/missing do we run the `gjsify
 //      login` flow (unless `--yes` on a non-TTY, which fails clearly).
-//   2. Enumerate publishable workspaces (non-private, excluding `@girs/*`).
+//   2. Enumerate publishable workspaces (non-private). `@girs/*` is excluded
+//      by DEFAULT, not by rule: in this repo those are third-party type
+//      packages nobody here publishes. `--include '@girs/*'` selects them,
+//      which is what ts-for-gir needs — its `@girs/*` workspaces ARE its
+//      publishables, a new GNOME cycle mints a dozen new npm names, and
+//      gjsify/types publishes them over OIDC, which cannot create a name that
+//      does not exist yet. The release script there names this command as the
+//      way to bootstrap them; until now the command could not select them.
 //   3. Determine each package's state (bounded concurrency) by reading npm's
 //      Trusted-Publisher config through the SAME requester path `gjsify trust`
 //      uses: unpublished (404) / published-but-untrusted / published-and-trusted.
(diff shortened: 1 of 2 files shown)


JumpLink:
Superseded by #1381, which landed the same capability in a better shape while this was in flight.

I hit the gap from the ts-for-gir side — `gjsify onboard` could not select `@girs/*`, and GNOME 51 needs thirteen new names bootstrapped before the OIDC release can publish them. This PR added an `--include` flag that lifted the hardcoded exclusion.

#1381 removed the hardcode outright and made onboard a general monorepo tool: `--include`/`--exclude` as plain name filters, `--packages` for a root with no workspace manifest, and `assertRepositoryAgreement` so a package cannot be trusted for the wrong repo's workflow. That last part matters here specifically — ts-for-gir's ~703 `@girs/*` workspaces publish from gjsify/types, and my version had no guard against trusting them for the wrong one.

It is also the better entry point for the GNOME 51 bootstrap: `--packages '*'` runs directly in gjsify/types, where the packages actually publish from. Nothing here is worth keeping on top of it.

Issue #1414, opened 2026-08-29
One widget vocabulary is a rule for every surface, and three of the four do not hold it


JumpLink:
`scripts/check-vocabulary-alignment.mjs` reads **one** of this repository's four widget-bearing surfaces. It holds `adwaita-web`'s `adw-*` elements against the GIR-derived tag set, and reads neither `@gjsify/adwaita-nativescript` nor `@gjsify/adwaita-react-native`:

```sh
grep -ic nativescript scripts/check-vocabulary-alignment.mjs   # 0
grep -ic adwaita-web  scripts/check-vocabulary-alignment.mjs   # 3   <- the control
```

## Where each surface stands

| surface | widgets | GIR naming | namespace export | declared |
|---|---:|---|---|---|
| `@gjsify/gtk-host` | 168 tags | **holds by construction** — prefix derived from the GType (`src/tags.ts:18`) | n/a | n/a |
| `@gjsify/adwaita-web` | 65 | **10 violate it** (`adw-entry` is `GtkEntry`) | absent | **half** — all 21 non-matching elements declared, but the 10 aliases carry no reason |
| `@gjsify/adwaita-nativescript` | 46 | **4 violate it**, 4 more have no counterpart | absent | **not at all** — outside the check |
| `@gjsify/adwaita-react-native` | 2 | **holds** (`AdwBin`, `AdwClamp`) | absent | **not checked** |
| the documentation | 40 gallery blocks | **4 `Gtk.*` blocks sit under an `Adwaita` heading**; there is no `Gtk` section | — | no |

## The clearest instance needs no cross-runtime argument

```
@gjsify/gtk-host  (Solid, Vue and React alike)   <gtk-entry placeholderText="…" />
@gjsify/adwaita-web                              <adw-entry placeholder="…"></adw-entry>
```

Both appear on the same gallery page, inside one block titled `Gtk.Entry` (`adwaita-framework-snippets.ts:60-62` against `controls.mdx:56`). gtk-host derives its prefix from the GType and says `gtk-entry` without anyone choosing; adwaita-web says `adw-entry`.

One accuracy note, because the sharper-sounding version is false: these are **not** two browser renderers. gtk-host renders to real GTK and adwaita-web to the DOM, and adwaita-web has no Solid/Vue/React binding at all. They differ in render target, not in what the widget *is* — and ADR 0027 § 9's goal names both surfaces in one sentence, so the disagreement is inside the goal's scope.

[...]

Issue #1338, opened 2026-08-26
The `glib:nick` disagreement is measured as 97 here and 889 in ts-for-gir, over the same 40 940


JumpLink:
Two shipped comments measure the same thing over the same corpus and disagree by
a factor of nine. The number is not trivia: it is the stated JUSTIFICATION for
reading `glib:nick` rather than deriving it from the member name, so whichever
one is wrong is an argument resting on a figure nobody can reproduce.

**gjsify** — `packages/framework/gtk-host/src/generator/gir.mts:59-66`, on `nickOf`:

> across every GIR in this workspace 40,940 members carry the attribute and 97
> disagree with it.

**ts-for-gir** — `gjsify/ts-for-gir#438` (merged 2026-08-26), in the emitted
header and in the PR body:

> 889 of the 40 940 members carrying the attribute disagree with the underscore
> substitution, none of them in Gtk-4.0 or Adw-1 — which is how a derived nick
> passes review and breaks elsewhere.

**Why this cannot be explained by different corpora.** That would be the obvious
reading — two repos, two GIR sets. But both quote the SAME denominator, 40 940
attribute-carrying members. If the inputs differed, the total would differ too.
So either the two definitions of "disagree" differ (a plausible candidate: member
rows vs. distinct nicks, or a comparison that also folds case), or one count is
simply wrong.

Note the qualifier only ts-for-gir's carries: *none of them in Gtk-4.0 or Adw-1*.
That is the part that makes the fallback survivable for this repo's own widget
work and lethal for anyone reusing the generator on another namespace — and it is
missing from gjsify's version of the same sentence.

**What to do.** Measure once, keep one copy. Per ADR 0029 the nick vocabulary is
`@girs/*`'s to own, so the surviving measurement should be ts-for-gir's, with the
`gir.mts` comment pointing at it rather than restating a second number. If the
generator moves as ADR 0029 steps 3–5 describe, the gjsify copy disappears with
it — but the contradiction is live in `main` today either way, and a number quoted
as evidence should not have two values.

JumpLink:
Resolved, and both numbers were right — they measure different derivations.

Measured independently over `ts-for-gir/girs/` (the same corpus both comments name):

| | measured |
|---|---|
| GIR files | 705 |
| `<member>` elements | 59 789 |
| carrying `glib:nick` | 40 940 |
| nick ≠ `name.replace(/_/g, '-')` | **97** |
[...]

PR #1415, opened 2026-08-29
docs(adr): a widget-vocabulary rule for every surface

JumpLink:
ADR 0027 § 9 ends by saying the NativeScript/browser horizon *"is **not decided here** … it would need its own ADR."* This is that ADR. It decides the vocabulary the horizon would need, states the rule **once** so it binds surfaces that do not exist yet, and refuses six things that look like progress and are not.

Docs only: one new ADR, its index row, one ledger entry. No widget renames, no namespace objects, no constructor changes.

Design for #1414.

## The rule, stated once

> A widget-bearing surface **1.** names a widget after the library that owns its GType; **2.** exports its widgets through a namespace object (`Gtk`, `Adw`); **3.** declares every divergence from 1 and 2, with a reason. An **undeclared** divergence is what goes red. **And the documentation follows the same split as the code.**

Written about *a surface*, not about any of the four that exist — because ADR 0032 deferred *"the NativeScript component vocabulary"* and *"a `.web` export over `@gjsify/adwaita-web`"* separately, and a deferral scoped to a port has no claimant. The fourth surface appeared three days ago in one PR.

## Where each stands (measured, with the command beside every number)

| surface | GIR naming | namespace | declared |
|---|---|---|---|
| `gtk-host` (168 tags) | **holds by construction** (`src/tags.ts:18`) | n/a | n/a |
| `adwaita-web` (65) | 10 violate it | absent | **half** — 21 declared, but the 10 aliases carry no reason |
| `adwaita-nativescript` (46) | 4 violate it, 4 more have no counterpart | absent | **not at all** — outside the check |
| `adwaita-react-native` (2) | **holds** | absent | not checked |
| the docs | `controls.mdx` is 100 % GTK under an `Adwaita` heading | there is no `Gtk` section | no |

**The clearest instance needs no cross-runtime argument**: gtk-host says `<gtk-entry>` for Solid, Vue and React alike; adwaita-web says `<adw-entry>` for the same widget; both sit on one gallery page under one block titled `Gtk.Entry`. (Stated accurately in the ADR: they differ in *render target*, not in what the widget is — adwaita-web has no framework binding, so the "two browser renderers" version would have been false.)

The documentation has the same shape of defect: `controls.mdx` carries **zero** `Adw.*` gallery blocks and two `Gtk.*` ones, under a heading named for a different library — and the section's own rename comment already argued the rule (*"naming it after the design system leaves room for a second one … **beside** it rather than **under** it"*), on a premise that has since stopped holding. Stage 5 applies that sentence a second time.

And **no stated rationale for the flattening exists anywhere** — every ADR, all 14 `AGENTS.md`, `status/`, the READMEs, the website, the three introducing commits. The docs argue against it on the pages that render it. So this overturns a convention, not a decision.

## The staging follows the cost curve, not the severity

They are different surfaces, and this is the substantive change from the first draft:

| surface | widgets | published | downloads/mo | consumers |
|---|---:|---:|---:|---|
| `adwaita-nativescript` | 46 | **49 versions** | **3 761** | 49 TS imports, 9 `.mdx` pages, 108 XML tags |
| `adwaita-react-native` | **2** | **none — npm 404s** | **0** | **0 outside the package** |

`@gjsify/adwaita-react-native` landed in one PR (#1380), both names already correct, `private: false` at 0.44.0 — **it publishes on the next release cut.** Adopting the rule there is free today and costs the NativeScript price afterwards, so it is **stage 1, with the expiry written into the stage**. The worst surface (NativeScript) is stage 3; the clearest to argue (adwaita-web) is stage 2. Ordering by severity would have spent the cheap window on the expensive surface.

That also means the rename rejection is re-scoped explicitly: it is a statement about *cost already paid*, and reading it as "gjsify does not rename widgets" would be reading a measurement as a principle.

**Stage 4 is what makes this a rule rather than a list**: enrolment becomes the property — a package declaring itself a widget surface that no reader covers fails. Without it, "every surface" binds three packages and the next one repeats the defect.

Nine stages, each additive, each with what would go red. The browser `gi://` arm was an open question in the first draft; under the wider rule its red condition is sayable (build `--app browser` from `gi://Adw?version=1`, construct `Adw.ActionRow`, assert an upgraded `<adw-action-row>`; today `Class extends value undefined`), so it is promoted to stage 9 — with the honest caveat that its `Gtk` namespace would be sparse and every absent member has to be a named refusal.

## The `tsc` rejection, rebuilt on evidence rather than on venue

The first draft rejected structural conformance because "the gate job installs nothing". True of `audit-runtimes.yml`'s `check:`, and **not an argument**: `tree-checks` installs, already hosts two gates moved there for precisely that reason, pays its install on every PR regardless, and a 42-widget `tsc` program measures **1.5 s**. What survives is what the check would prove — a 509-member assertion nobody wants; `Partial<>` being vacuous in the useful direction; and `@girs` being a **second reader of our own source**, since `generated/props.ts` imports seven `@girs` packages and `gtk-host` declares eight as runtime dependencies. The independent oracle for "does this match the GTK the host loads" is the runtime `GParamSpec` (ADR 0028 § 5).

Related, and recorded rather than proposed: ts-for-gir is usable as a library (ADR 0019 § 1, confirmed against the registry) — that buys a second reader, not a second source. Shipping the `.gir` inside `@girs/*` is **already rejected by ADR 0019 § 2**, and the sizes agree (5.86 MB package against a 6.20 MB `.gir`; 705 files / 379 MB pooled). The real capability behind it — nicks live only in the typelib, docs only in the XML — points at a companion artifact in `gjsify/ts-for-gir`, not at a stage here.

## Provenance, because it nearly went wrong

A set of counts in the first draft came from a shared checkout **43 commits behind `main`** and was plausible and wrong. The ADR now states where every number was read and pins `@girs/gtk-4.0@4.1.0` for the type-surface figures. One thing that looked like a generator bug is not: `props.ts`'s "194 interfaces" against `grep`'s 198 is exactly the four tag maps the header already excludes; the genuine nit is that it calls them "the **two** tag maps", which is a one-word fix in `emit-types.mts` and its own PR.

## Kept from the earlier drafts

The declared remainder and its discriminated union; the four counterpart-less NativeScript widgets classified from their files (two convergeable, two genuinely original); the measured rename cost; and the **45**-property distance as a number a gate prints rather than one this ADR states.

## Gates

`check-adr-index` (34 ADRs), `check-doc-fences`, `check-refs-citations`, `check-vocabulary-alignment`, `check-storybook-widget-coverage`, `check-adwaita-rn-platform-split`, `check-rn-surface`, `check-agent-context-size`, `check-changelog-references`, `check-source-control-bytes`, `check-website-adwaita-gallery` — all green; `generate-status.mjs` leaves the tree clean.

### DIFF
--- a/docs/adr/0034-widget-vocabulary-convergence.md
+++ b/docs/adr/0034-widget-vocabulary-convergence.md
@@ -1,0 +1,940 @@
+# 34. Every widget surface: named from the GIR, exported as a namespace, remainder declared
+
+- Status: **Proposed**
+- Date: 2026-08-29
+- Deciders: Pascal Garber
+- Related: [ADR 0027 § 9 (the goal)](0027-gtk-host-layer.md), [ADR 0028 § 6 (the alignment mechanism)](0028-widget-table-provenance.md), [ADR 0029 (the vocabulary in `@girs/*`)](0029-girs-widget-vocabulary.md), [ADR 0019 (ts-for-gir as a library; where the `.gir` travels)](0019-ts-for-gir-as-library.md), [ADR 0004 (headless core)](0004-headless-adwaita-core.md), [ADR 0032 (React Native on the host)](0032-react-native-on-the-gtk-host.md), [ADR 0033 (templates preferred)](0033-declarative-templates-preferred.md)
+
+ADR 0027 § 9 records one vocabulary across every surface as a goal and ends: *"The
+longer horizon this points at — generating NativeScript and browser builds from one
+native-authored source … is **not decided here**. … it would need its own ADR."* This is
+that ADR. It does not decide the horizon; it decides the vocabulary the horizon would
+need, and refuses six things that look like progress towards it and are not.
+
+**The rule is stated once and binds every surface** — `@gjsify/gtk-host`,
+`@gjsify/adwaita-web`, `@gjsify/adwaita-nativescript`, `@gjsify/adwaita-react-native`,
+and the next one somebody adds. A rule written as "the NativeScript port should…" binds
+nothing that does not exist yet, and the surfaces this repository keeps growing are
+exactly the ones that were not there when the last rule was written. § 1 is therefore
+surface-neutral; § 2 is a table of where each stands against it today.
+
+The staging follows the **cost curve**, not the severity of the defect, and the two are
+not the same surface. § Context measures both.
+
+### How the numbers here were obtained
+
+Every number below is followed by the command that produces it. This paragraph says where
+each command was run, because during the drafting of this ADR a set of counts was read out
+of a **shared checkout sitting 43 commits behind `main`**, and every one of them was
+plausible and wrong. A count with no stated revision cannot be re-measured; it becomes
+folklore about a tree nobody has any more.
+
+| source | what was read there | how to reproduce |
+|---|---|---|
+| **this repository at `main`** | every widget, tag, element, page and gate count | the quoted command, from the repository root |
+| `refs/libadwaita`, pinned at `42f647ff` | upstream widget existence (`AdwSliderRow`, `.image-button`) | `git ls-tree HEAD refs/libadwaita` states the pin; the submodule must be initialised — **89 of 95 are not by default, and `grep -r` over an uninitialised one searches nothing**, so every claim of absence below carries a control string |
+| **`@girs/gtk-4.0@4.1.0`**, the version `gtk-host/package.json` declares (`libraryVersion` 4.23.0) | the type-surface sizes (509 / 151 / …) | pin the version: `^4.1.0` resolves to **4.3.0** today, and these counts move with it |
+| **the npm registry** | published versions, downloads, package sizes | `npm view <pkg> …`, revision-independent |
+
+The one number that matters most — the property distance, **45** — is deliberately not
+left in prose at all: stage 6 makes a check print it. Every count written into prose in
+this area has drifted at least once, including inside the file whose own rule is that
+counts drift.
+
+Anything unmeasured is marked UNVERIFIED.
+
+## Context
+
+### The same widget, four spellings, on one documentation page
+
+`website/src/content/docs/adwaita/controls.mdx:52-98` renders one gallery block whose
+title is `Gtk.Entry`. Inside it:
+
+| fence | how the widget is named |
+|---|---|
+| markup | `<adw-entry placeholder="Search files…">` |
+| GJS | `new Gtk.Entry({ placeholderText: 'Search files…', widthRequest: 280, halign: Gtk.Align.CENTER })` |
+| Blueprint | `Gtk.Entry { placeholder-text: _("Search files…"); halign: center; }` |
+| NativeScript | `const entry = new AdwEntry(); entry.placeholder = 'Search files…'` |
+
+Two of the four fences spell `Adw` for a widget the block itself titles `Gtk`, and the
+prose two paragraphs above them (`controls.mdx:14-17`) says why that is wrong:
+
+> Neither carries an `Adw` prefix, and that is the point worth knowing about them:
+> Libadwaita ships no entry and no drop down of its own. They are `Gtk.Entry` and
+> `Gtk.DropDown`, and what makes them look Adwaita is the stylesheet.
+
+The gallery's own titles already follow the GIR. Measured: 40 `<AdwWidget>` blocks, 36
+titled `Adw.*` and 4 titled `Gtk.*` — and the four are exactly `Gtk.Button`,
+`Gtk.DropDown`, `Gtk.Entry`, `Gtk.MenuButton`.
+
+```sh
+grep -rhoE '<AdwWidget title="[A-Za-z.]+"' website/src/content/docs/adwaita/ \
+  | sed 's/.*title="//;s/"//' | sed 's/\..*//' | sort | uniq -c
+```
+
+### What each surface ships — measured
+
+| surface | vocabulary | count |
+|---|---|---|
+| `@gjsify/gtk-host` (`src/generated/widgets.ts`) | GIR-derived kebab tags | **168** — 63 `Adw*`, 105 `Gtk*` |
+| `@gjsify/adwaita-web` | `adw-*` custom elements | **65** |
+| `@gjsify/adwaita-nativescript` | `Adw*` view classes in `adw-*.ts` | **46** |
+| `@gjsify/adwaita-react-native` | `Adw*` components | **2** (`AdwBin`, `AdwClamp`) |
+
+```sh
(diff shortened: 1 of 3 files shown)


JumpLink:
## Measured: the typelib is a lossy subset of the GIR, not a supplement

Follow-up on this line in the PR body:

> The real capability behind it — nicks live only in the typelib, docs only in the XML — points at a companion artifact in gjsify/ts-for-gir, not at a stage here.

The first half is inverted, and I think it is worth correcting before anyone plans a stage around it. `g-ir-compiler` compiles `.gir` into `.typelib`, so nothing can be in the output that was not in the input. I ran the counter-check with `g-ir-generate`, which decompiles a typelib back to XML, against the original GIR from the `ts-for-gir` corpus:

| | `Adw-1.gir` | decompiled from `Adw-1.typelib` |
|---|---|---|
| size | 2.36 MB | 626 KB |
| `glib:nick` | 120 | **0** |

Gtk-4.0 answers the same way: `glib:nick` 798 to 0, `default-value` 992 to 0, `<doc>` 18386 to 0, `nullable` 1288 to 0, `deprecated-version` 1098 to 0.

An enum member in the typelib is bare:

```xml
<member name="blue" value="0">
```

The same member in `Adw-1.gir` carries `glib:nick="blue"`, `glib:name="ADW_ACCENT_COLOR_BLUE"`, `c:identifier` and its doc. The typelib does not even keep the C identifier.

Full list of attributes the typelib has and the GIR does not, across both samples: `offset=`, the struct field byte offsets. Everything else the diff turned up is the same fact under an older spelling (`c:prefix` for `c:symbol-prefix`, `glib:ref-function` for `glib:ref-func`, `glib:is-gtype-struct` for `glib:is-gtype-struct-for`).

So among static files, nicks live only in the XML, which is exactly where gjsify/ts-for-gir#438 reads them. Byte offsets are the one thing worth remembering, and they are useless for a `.d.ts`. They would only matter to an FFI consumer reading struct layout without libgirepository.

## Refinement to ADR 0028 § 4

> The GIR — and the typelib, via `ValueInfo.get_name()` — is the only source of the nick list.

The GIR half is right, the typelib half is not an equal second reader. `ValueInfo.get_name()` returns the typelib's member name, which is byte-identical to the GIR's `name` attribute because the compiler copied it there. That is the underscore substitution, not the nick.

Measured over the 718-GIR corpus, deriving `name.replace('_', '-')` and comparing against the recorded `glib:nick`: 42759 members carry the attribute, 84 contradict the derivation. All 84 are in `AgsAudio-6.0`, where the nick itself keeps an underscore (`audio_unit_client_activated` has nick `audio_unit-client-activated`). Small, but it means a typelib-sourced nick list is derivation wearing a runtime costume, and it fails in exactly the places a derived nick always fails.

## What this leaves for the companion artifact

Chasing missing typelibs buys nothing. 233 of 718 GIRs in the corpus have no typelib beside them, and adding one would not add a character to any generated type. The 26 MB of typelibs already committed under `girs/` are read nowhere in `ts-for-gir`; only comments mention them.

The genuine third source is the runtime, not the typelib file. `ts-for-gir`'s own `emit-data.ts` already states the case: GIR's `default-value` and a probed instance disagree in 104 of 953 measured cases. Add ParamSpec min/max, CSS name, accessible role, layout manager, template children, and the one check that can go red for a real reason, whether the installed library has the member at all. None of that is in either file.

Two things I would hold to if that artifact gets built:

1. **Do not generate it on whatever happens to be installed.** The generator runs headless over 475 namespaces; a CI box has maybe a third of them present. A package whose contents depend on the build machine is the failure mode where a green run checked nothing.
2. **Commit the measurement, like `fetch-fedora-girs.sh` commits the GIRs.** A probe run in a pinned container, output checked in, reviewable as a diff. Namespaces without a measurement get nothing. That makes it optional per input rather than optional by luck, and a wrong default shows up in review instead of shipping silently.

And it should stay data plus doc tags (`@min`, `@max`, a measured `@default`), never types. The `.d.ts` is compiled once for every install; the runtime facts are not.

Method, for anyone who wants to repeat it: `g-ir-generate --includedir=girs girs/<NS>.typelib`, then diff the attribute vocabulary of both files.

JumpLink:
## The direction reproduces; the distribution does not

Verified the lossy-subset claim independently before adopting it, and it holds — same method,
`g-ir-generate --includedir=girs`, on this machine:

```
Adw-1.gir                        2 310 423 B, glib:nick = 120
g-ir-generate < Adw-1.typelib      577 131 B, glib:nick =   0
```

`Gtk-4.0.gir`: 799 `glib:nick`, 18 327 `<doc>`, 986 `default-value`, none of them in the typelib.
The correction is now in ADR 0028 § 4 — **#1417** — with the old sentence struck rather than
deleted, since three design passes reasoned from it.

**One claim in the comment above does not reproduce, and it is the one that makes the defect look
negligible.**

> All 84 are in `AgsAudio-6.0`, where the nick itself keeps an underscore

`Avahi-0.6.gir` alone contradicts the derivation **35** times, and in exactly the same shape:

```
name="GA_BROWSER_NEW"           derived "GA-BROWSER-NEW"           nick="GA_BROWSER_NEW"
name="GA_BROWSER_REMOVE"        derived "GA-BROWSER-REMOVE"        nick="GA_BROWSER_REMOVE"
name="GA_CLIENT_FLAG_NO_FLAGS"  derived "GA-CLIENT-FLAG-NO-FLAGS"  nick="GA_CLIENT_FLAG_NO_FLAGS"
```

Same derivation (`name.replace('_', '-')` against the recorded `glib:nick`), over a 705-GIR corpus
(one revision behind the 718 above): **40 940** members carry the attribute, **97** contradict it —
Avahi-0.6 **35**, AgsAudio-6.0 16, AgsAudio-8.0 16, GES-1.0 12, Gom-1.0 7. The 13 GIRs my corpus
lacks cannot explain 35 contradictions in a file both corpora contain, so this is a disagreement
about the measurement, not about the input.

`97` over `40 940` is also the exact pair already recorded in **#1338** ("97 here and 889 in
ts-for-gir, over the same 40 940"), so the number has a second, independent sighting.

Why it matters for the conclusion rather than just for the record: 84 in one obscure namespace reads
as an artefact of that namespace. 97 across five namespaces, including one whose nicks are
uppercase-with-underscores throughout, is a pattern — a derived nick list fails wherever upstream
chose a nick that is not the name with dashes, and nothing warns. Worth reconciling under #1338,
where the same corpus already answers two different ways.

Everything else in the comment I have taken as-is and acted on: the runtime as the genuine third
source, and the two conditions on a measurement artifact (do not generate against whatever happens
to be installed; commit the measurement so a wrong default shows up in review).

Issue #1354, opened 2026-08-27
gjsify ship stages 4 and 5: macOS .app/.dmg and a Windows program directory + installer


JumpLink:
`gjsify ship` packs `.deb`, `.rpm` and a Flatpak bundle out of one staged payload. ADR 0024's
stages 4 and 5 — a macOS `<App>.app`/`.dmg` and a Windows program directory plus installer — are
the two rows that do not exist, and they are the two the desktop story now ends on.

This issue describes only the gap. The design is already written down: ADR 0024
(`docs/adr/0024-ship-installable-artifacts.md`), including the 2026-08-21 amendment (§ A1-A7) that
moved the host boundary, and item 5 of *`gjsify ship` — remaining roadmap* in `status/open-todos.md`.

## What exists

**Stages 2, 3 and 6 have landed.** `gjsify ship` stages the payload once and packs three formats
out of that one tree:

- `.deb` and `.rpm`, written by this tree — `utils/ship/{ar,cpio,rpm-header,deb,rpm,gzip,bytes}.ts`,
  no `dpkg-deb`, no `rpmbuild`, no vendored `nfpm`. Measured today those seven files plus
  `flatpak.ts` are 1 243 lines; ADR § A6 cites 1 036 for the packer surface before Flatpak joined it.
- `--target flatpak`, whose module is `buildsystem: simple` + `cp -a stage/.`, so meson is gone from
  inside the sandbox.
- The two-phase split of § A2: `--stage` assembles, `--from-stage <dir>` packs a stage that arrived
  from another host with no project, no `package.json`, no bundle and no config in reach.
  `.gjsify-ship-stage.json` (schema 2) carries the closure — `{ settings (arch resolved at stage
  time), staged, overlay, namespaces, mtime }` — and each omission was measured to fail silently at
  exit 0: without `staged` the launcher packs 0644, without the pre-rendered overlay the `.deb`
  carries no `/usr/share/doc/<pkg>/copyright` (Debian Policy § 12.5), without `namespaces` its
  `Depends` loses `gir1.2-gtk-4.0` and `gir1.2-adw-1`.
- Host-boundness as data: `HostRequirement` on the format descriptor (`utils/ship/types.ts`), with
  `assertHostCanFinish` and `assertToolsInstalled` as the two refusals — separate, because one needs
  a different machine and the other needs a package.

**The oracles are not ours.** Every e2e assertion runs against a real artifact, read back by a
different implementation family:

| suite | reads the artifact with |
[...]

PR #1418, opened 2026-08-29
feat(cli): a macOS bundle that carries its runtime

JumpLink:
M2a's `<App>.app` was a real bundle whose launcher exec'd the bare name `node` off `PATH` and whose `gi://` imports expected a GTK on the machine. macOS ships neither — so it was an application in shape and not in behaviour, which `formats.ts`'s own `macos-app` row said in as many words: *"whatever the bundle does not carry, it finds on the machine or it does not run — which is what makes M2b (staging a runtime) the milestone that matters"*.

Closes the darwin half of ADR 0024 stage 4 (#1354 M2b).

## What the bundle carries now

`utils/ship/app-runtime.ts` stages four things, each resolved **by name** from the project being shipped and each `null`-not-throw, so a partial bundle is a reported intermediate rather than a failed build:

| Piece | Package | Where it lands |
|---|---|---|
| the interpreter | `@gjsify/node-runtime-darwin-<arch>` | `Contents/MacOS/node` + `Contents/Resources/share/licenses/node/LICENSE` |
| the relocated GTK closure | `@gjsify/gtk-runtime-darwin-<arch>` | `Contents/Frameworks/node-gi/prebuilds/darwin-<arch>/gtk/**` |
| the node-gi addon | `@gjsify/node-gi` | beside that closure, as its **sibling** |
| node-gi's **JavaScript** | `@gjsify/node-gi` | `Contents/Resources/lib/node_modules/@gjsify/node-gi/` |

`resolveNodeRuntime` finally has the caller its header said it did not have (*"⚠️ NOTHING CALLS THIS YET … a promise to a caller that does not exist yet"*). That header is corrected rather than deleted, and so is `packages/node-runtime/AGENTS.md`'s copy of the same claim.

**The fourth row was not in the plan, and without it the milestone is not reached.** `@gjsify/node-gi/*` is external in every `--app node` bundle **by design** — `rolldown-plugin-gjsify/src/app/node.ts`: the reverse-bridge modules *"must resolve at runtime against the consumer's node_modules"* — so a `gi://Gtk` import compiles to `require('@gjsify/node-gi/gi')` in the shipped file, and a `.app` has no consumer `node_modules`. Measured on this workstation, on a bundle staged the M2a way and run from an unrelated directory:

```
$ ./probe-bare.app/Contents/MacOS/ship-window-demo
ship-window-demo: start
Error: Cannot find module '@gjsify/node-gi/gi'
Require stack:
- /…/probe-bare.app/Contents/Resources/lib/app.node.mjs
```

With the package staged, the same bundle runs 1519 lines into `gi.js` before it stops on the stale local addon — i.e. past every resolution question this row exists to answer.

## Tree-preserving, and why it could not go through `bundledTypelibs`

The claim was verified before it was built on. Driving the built planner directly:

```
$ node -e "planStage({typelibFiles:[…3 paths at 3 depths…]})" | grep /gi/
lib/demo/gi/Gtk-4.0.typelib
lib/demo/gi/libpixbufloader-svg.so
lib/demo/gi/loaders.cache
```

`plan.ts` stages each such file as `posix.join(libDir, 'gi', basename(file))` while `discoverTypelibs` walks its directory recursively — so `lib/gdk-pixbuf-2.0/2.10.0/loaders/` is destroyed, `loaders.cache`'s `@loader_path/../../..` entries point three levels above where the decoders now are, and the result has neither a `lib/` nor a `girepository-1.0/` for `resolveGtkRuntimeBundle()`'s existence probe. That measurement is kept as a **CONTROL test beside the staging it justifies**, so the decision cannot be "simplified" back into the bug.

`placeStage` therefore takes the staged runtime as already-stage-relative files, through the same uniqueness check `Layout.metadata` goes through. The staged shape is node-gi's own sibling layout — `prebuilds/<target>/{node_gi.node, gtk/}` — because that is what `resolveGtkRuntimeBundle()` walks, what the addon's `@rpath` of `@loader_path/gtk/lib` reaches, and what `.github/workflows/node-gi.yml`'s macOS legs already assemble by hand before every conformance run.

## The launcher

```sh
#!/bin/sh
# Generated by `gjsify ship` — do not edit.
set -e
here=$(cd -- "$(dirname -- "$0")" && pwd)
contents=$(dirname -- "$here")
XDG_DATA_DIRS="$contents/Resources/share:${XDG_DATA_DIRS:-/usr/local/share:/usr/share}"
export XDG_DATA_DIRS
GJSIFY_GTK_RUNTIME="$contents/Frameworks/node-gi/prebuilds/darwin-arm64/gtk"
export GJSIFY_GTK_RUNTIME
NODE_GI_NATIVE="$contents/Frameworks/node-gi/prebuilds/darwin-arm64/node_gi.node"
export NODE_GI_NATIVE
exec "$here/node" "$contents/Resources/lib/app.node.mjs" "$@"
```

Each variable was checked against node-gi's source rather than assumed:

* **`GJSIFY_GTK_RUNTIME`** is candidate 1 of `resolveGtkRuntimeBundle()`'s four (`gtk-runtime.js:39-64`). Candidates 2-4 walk node-gi's own package directory or `node_modules`; in a shipped `.app` node-gi's package root is `Contents/Resources/lib/node_modules/@gjsify/node-gi`, so every probed path lands under `Resources` and never in `Frameworks`.
* **`NODE_GI_NATIVE`** is the absolute-path form `nativeCandidates()` accepts (`native-paths.js:104-113`), and it returns a pinned path **alone** — so the bundle's addon is not merely preferred, nothing else is tried. `prebuildAddonPath()` would otherwise join `prebuilds/<target>/node_gi.node` onto that same wrong root.
* **`GJSIFY_GI_LIBRARY_PATH`** (emitted only when the app carries GI libraries of its own) is the writer #1410 shipped a reader for. Nothing in this pipeline set it; grep confirms the variable appeared only inside `packages/node-gi` before this PR.

**No `DYLD_*`, with the reason that holds.** The narrow measured fact is that an *inherited* `DYLD_*` is stripped when a *protected* binary is exec'd — this repo depends on the difference today in `utils/bin-shim.ts` and `node-gi/gtk-runtime.js`, both green on the darwin legs. The reason to keep it out of a `.app` is the other one: under hardened runtime a Developer-ID-signed main executable is restricted, so it is stripped at M6 and a launcher depending on it works unsigned and breaks the day it is signed (ADR 0024 § A4, § A16). Everything the launcher does export is read by node-gi in JavaScript and handed to GI through the binding; dyld sees none of it.

## A core fix in the reader that guards the launcher

`interpreterOf` took the basename of a shell token with its quotes still attached, so `"$here/node"` read back as `node"` — neither known interpreter, so `assertLauncherMatchesInterpreter` took its *"a program that is neither"* branch and **passed**. Not a wrong answer: a vacuous one, on exactly the layout that made the check matter. Fixed in `payload.ts` with a surrounding-pair-only strip, and reded (below).

## What a third-party author adds

The list is real, written down and proven, because `@gjsify/gtk-runtime-*` and `@gjsify/node-runtime-*` are resolved by name with no `optionalDependencies` edge anywhere (`docs/publishing.md`): whoever ships an app declares the runtime it ships.

```jsonc
{
  "devDependencies": {
    "@gjsify/node-runtime-darwin-arm64": "^0.44.0",
    "@gjsify/node-runtime-darwin-x64": "^0.44.0",
    "@gjsify/gtk-runtime-darwin-arm64": "^0.44.0",
    "@gjsify/gtk-runtime-darwin-x64": "^0.44.0"
  },
  "dependencies": { "@gjsify/node-gi": "^0.44.0" }
}
```

Written in `website/src/content/docs/ship/index.mdx` (copy-pasteable) and `docs/ship-formats.md` (the model). **Proven by the fixture**: every case in `app-runtime.spec.ts` and `tests/e2e/ship-macos` installs those packages into a throwaway project's own `node_modules` holding nothing of gjsify, so a resolution that only worked from inside this monorepo fails there. `gjsify ship` names what it staged and, for anything missing, the package to install:

```
[gjsify ship] carries its own interpreter from @gjsify/node-runtime-darwin-arm64
[gjsify ship] carries its own GTK closure from @gjsify/node-gi/prebuilds/darwin-arm64
[gjsify ship] carries its own node-gi runtime (16 file(s)) from @gjsify/node-gi
[gjsify ship] carries its own node-gi addon from @gjsify/node-gi
```

and `Layout.runtimeGap` — *"the staged launcher execs an interpreter off `PATH`…"* — is now **conditional**, because printing it over a bundle that carries its own interpreter would be the command saying something untrue about the tree it had just written.

## The proof

**The Linux half.** `tests/e2e/ship-macos` grew from 17 to 28 tests. It reads every staged Mach-O back with `manifest-conformance/lib/binary.mjs`'s `readLibrary()` — ADR 0024 § A4's own instrument, which produced that section's published control (106 images, 0 unresolved non-system dependencies) — and asserts **6 images, all `darwin`/`arm64`, 0 unresolved non-system dependencies** resolved inside the bundle by dyld's own rule (`@loader_path` against the image's directory, `@rpath` against each `LC_RPATH` expanded the same way). The fixture closure is synthetic and its dependencies are real: the deep pixbuf loader reaches its backer as `@loader_path/../../../libglib-2.0.0.dylib` from four levels down, and the addon reaches the closure as `@rpath/libgtk-4.1.dylib` against `@loader_path/gtk/lib`. **The discriminator sits beside it**: flattening the same staged tree to basenames leaves both unresolved, which is what makes "0 unresolved" a measurement of the staging rather than of a path string.

**Every new gate watched red before it counted.** Each mutation below was applied to `main`-plus-this-PR, built, and the suites run:

| Mutation | Red |
|---|---|
| stage the closure flattened (`basename`) | `stageAppRuntime › preserves the closure tree — every path, at its own depth`; `… › gives the two files that get EXEC'd their executable bit` — 2 of 3185 |
| drop the quote strip in `basenameOf` | `readLauncherInterpreters › reads a QUOTED, layout-relative interpreter back to its bare name`; `… › still refuses the mismatch it exists to catch` — 2 of 3198 |
| exec the bare interpreter name again | `ship launcher: the runtime the bundle carries › execs "$here/node" instead of a name macOS cannot resolve` — 1 of 3199 |
| stop exporting the two locators | `… › exports the two locators node-gi resolves the bundle with` — 1 of 3197 |
| stop exporting `GJSIFY_GI_LIBRARY_PATH` | `… › points GI at the libraries the APP carries, not only its typelibs` — 1 of 3198 |
| stage the closure by extension alone | `stageAppRuntime › gives the two files that get EXEC'd their executable bit` — 1 of 3195 |
| never stage node-gi's JavaScript | e2e `stages the JavaScript the bundle require's, where Node looks for it` + `says what it carries…`; unit `preserves the closure tree` — 2 e2e, 1 of 3203 |
| take `files` instead of the reach rule | + `resolveNodeGiPackage › selects what a bundle can REACH, not what npm publishes` — 2 e2e, 2 of 3203 |
| `placeStage` never receives the runtime | 6 of 26 e2e |
| the launcher execs the bare name (e2e) | 1 of 26 e2e |

Green after each revert: **3205 unit assertions**, **28/28** `ship-macos`, **103/103** across `ship-macos` + `ship-layout` + `prebuild-loader-path`, **100/100** across `ship` + `ship-from-stage` + `ship-declaration` + `ship-layout`, **122/122** across the six `ci-*` suites. `packages/infra/cli/tsconfig.json` excludes `src/**/*.spec.ts`, so the suites were run rather than `tsc` trusted.

**A mechanism, not just a fix**, for the one rule that could rot: `resolveNodeGiPackage` selects `package.json` + the root `.js` + `overrides/` rather than the published `files` list (which also carries `src/`, `binding.gyp` and `scripts/install.mjs` — inputs to a `node-gyp` run a `.app` never performs). `tests/e2e/ship-macos` holds that rule against the **real** package's `exports` map, so a node-gi release adding an export outside the set fails there rather than in a `.app` a stranger downloaded.

**And one check this milestone makes non-vacuous:** `assertPayloadMatchesArch` had nothing to read on a darwin stage — a JavaScript bundle has no `cputype`. It now refuses an arm64 closure labelled `x64`, reached through `GJSIFY_GTK_RUNTIME` (the only way an operator *can* produce it: resolution by name derives the target from `--arch`).

## The macOS half — and what it is honestly worth

> **A Linux-green run is not evidence about macOS.** Everything above runs on a host that cannot execute one instruction of the closure it inspects. It is a claim about our wiring: the layout, the modes, the launcher, the arch label, the resolution.

The macOS claim is a new pair of jobs in `node-gi.yml`, modelled on `macos-gtk-batteries-included` / `macos-gtk-windowing`:

* **`macos-app-assemble`** (ubuntu, `ci-fedora:44`, both arches) — downloads the windowing GTK bundle and the relocated addon into node-gi's own `prebuilds/darwin-<arch>/`, fetches the interpreter with `fetch-node-runtime.mjs`, builds this checkout's CLI, and assembles the `.app` + zip. This is ADR 0024 § A2's split doing real work: a leg that built the app on the same Mac that runs it would prove the Mac has a working GTK, which is the opposite of what is wanted.
* **`macos-app-selfcontained`** (`macos-latest` + `macos-15-intel`) — **no checkout, no `setup-node`**. Asserts `brew list --formula` shows neither `gtk4` nor `libadwaita` (hard fail), then asserts that `PATH` reduced to `/usr/bin:/bin:/usr/sbin:/sbin` has **no `node`** — the second absence, and the one this milestone is actually about, since the runner image ships a Node that would let a bare-name launcher pass here and fail on every Mac a stranger downloads it to. Then it unzips the artifact outside any checkout, strips every loader variable, and runs the launcher. It reads back four lines: the interpreter that ran is the one inside the bundle, `GJSIFY_GTK_RUNTIME` points inside the bundle, the Adwaita widget tree is complete, and a real `GdkSurface` was allocated and rasterised through the GSK renderer — the capture path of `test/windowing.test.mjs`, line for line.

The subject is a committed fixture project, `tests/e2e/ship-macos/fixtures/window-app`, so what the leg runs is reviewable rather than written inline in YAML.

**Dispatched, and green on both arches.** Run [33249462533](https://github.com/gjsify/gjsify/actions/runs/33249462533) — `completed/success`, every job.

Both discipline gates fired *before* the app ran, which is what makes the result mean anything:

```
clean host — the GTK toolkit was never brew-installed in this job
no node on the system PATH — whatever runs must come from inside the .app
```

Then, on `macos-latest` (arm64) — the whole of what the `.app` printed:

```
ship-window-demo: start
interpreter: /Users/runner/work/_temp/app/Ship Window Demo.app/Contents/MacOS/node
gtk-runtime: /Users/runner/work/_temp/app/Ship Window Demo.app/Contents/Frameworks/node-gi/prebuilds/darwin-arm64/gtk
ship-window-demo: start
interpreter: /Users/runner/work/_temp/app/Ship Window Demo.app/Contents/MacOS/node
gtk-runtime: /Users/runner/work/_temp/app/Ship Window Demo.app/Contents/Frameworks/node-gi/prebuilds/darwin-arm64/gtk
chrome: ok
render: 480x320 13562
done: 0
a downloaded .app opened and rendered a window on a Mac with no Homebrew GTK and no node on PATH
```

`macos-15-intel` (x64) is the same with `darwin-x64` in the path and `render: 480x320 13572`.

**The first three lines appear twice, and that is node-gi's darwin re-exec, not a duplicated log.** `maybeReexecForGtkRuntime` re-launches the process once with `DYLD_FALLBACK_LIBRARY_PATH` set at launch, because dyld captures that variable only there. Worth reading closely: `process.execPath` is the bundle's own `Contents/MacOS/node` in **both** the parent and the re-exec'd child, so the interpreter that ends up drawing the window is the one the artifact carries, not something the child picked up from the machine.

What the assembling half produced, from the same run:

```
[gjsify ship] staged 1199 file(s) for darwin in ship/stage/
[gjsify ship] carries its own interpreter from @gjsify/node-runtime-darwin-arm64
[gjsify ship] carries its own GTK closure from @gjsify/node-gi/prebuilds/darwin-arm64
[gjsify ship] carries its own node-gi runtime (16 file(s)) from @gjsify/node-gi
[gjsify ship] carries its own node-gi addon from @gjsify/node-gi
[gjsify ship] macos-app: ship/out/Ship Window Demo.app (202592569 bytes)
[gjsify ship] macos-app-zip: ship/out/ship-window-demo-1.0.0-1.arm64.zip (202997419 bytes)
```

1199 files and 193 MiB, assembled on Linux, unzipped on a Mac, opened a window. The Linux half above remains a claim about the wiring; this is the claim on the other axis, and it now holds.

## Also here

* `tests/e2e/macho.mjs` — the synthetic Mach-O writer `tests/e2e/prebuild-loader-path` had inline, lifted and shared. That suite builds *leak* shapes for a check that must refuse them; this one builds a *closure* whose every dependency resolves, and it needed a `cpuType` the inline copy could not give. Two copies of a binary writer are two sets of bytes that drift — the argument `check-e2e-harness-duplication.mjs` makes about the npm harness, one format over. `prebuild-loader-path` stays 43/43.
* `isExecutableAsset` is exported from `plan.ts` rather than restated, so "is this a shared library" has one answer.
* Docs: ADR 0024 (status → stage 4 darwin landed, with what M2b does and does not claim), `docs/ship-formats.md`, `docs/publishing.md`, `website/.../ship/index.mdx`, `docs/release-notes/next.md`, `status/open-todos.md` items 5+6, `packages/infra/cli/AGENTS.md`, `packages/node-runtime/AGENTS.md`.

## Not in scope

No signing (M6 — the `.app` is unsigned by construction and Gatekeeper will quarantine it; `readStage` still validates the pre-sign tree, and the packer still leaves the seam § A17 requires). No `.dmg`. No Windows runtime staging — `app-runtime.ts` is written for it (`GTK_RUNTIME_TARGETS` includes `win32-x64` and the `bin/` vs `lib/` split is read off the target, exercised from Linux), but that layout has no launcher form for a carried interpreter yet.

---

### The macOS evidence and the merged head — the gap is closed

An earlier revision of this body disclosed a SHA gap: the `.app` legs had run against `0c3a61f0` while the head carried an extra ledger commit. **That gap no longer exists.** The branch has since been rebased onto `main` (for #1411) and force-pushed, so the whole chain — including `macos-app-assemble` and `macos-app-selfcontained` — re-runs against the merged head. The quoted output above was produced by the pre-rebase run; the re-run is the one that gates this PR, and if it disagrees with what is quoted, the re-run wins and this section is wrong.

Keeping the note rather than deleting it, because the reasoning outlives the instance: expensive evidence and the merged head can drift apart, and a green check that never ran on the merged tree is the same shape as every other "green that checked nothing".

### A gate invoked in a mode that does not gate

`check-agent-context-size.mjs` failed on this PR after I had run it locally and seen it pass. I ran it **bare**; CI runs it with **`--check`**. Measured on the identical inconsistent state:

```
$ node scripts/check-agent-context-size.mjs            # default 'print' mode
packages/infra/cli/AGENTS.md        21055    20889  OVER
packages/node-runtime/AGENTS.md      8738     8739  STALE -1
exit 0

$ node scripts/check-agent-context-size.mjs --check    # same state
exit 1
```

**The invocation error is mine** — the script's header documents the three modes and names `--check` as "gate (CI)". But the residual edge is worth naming, because it is what made the mistake survive: the non-gating mode prints `OVER` and `STALE` rows, which are failure-shaped, while exiting 0. I piped stdout to `/dev/null` and read only `$?`, so the half that would have told me was discarded and the half I read is uninformative *by design*. Eyeballing the output and trusting the exit code each fail independently; only both, in the right mode, work.

If that is worth hardening, the cheap fix is in this repo's own idiom of naming the hazard in the output — a trailing `print mode — no verdict; run with --check to gate` — rather than making `print` exit non-zero, which would break its use as a report. I have not made that change here; it is a decision about someone else's gate.

The ledger edit itself is the interlock working as designed: the ceiling is EXACT in both directions, so `8738` against a committed `8739` fails as `STALE` too. Unclaimed slack is what let two concurrent PRs each spend the same window and land a state neither run measured (#1157).

**Every repo-local gate has since been re-run with its real CI invocation** — 22 from `audit-runtimes.yml` plus 9 from the other workflows — and again after the rebase.

### `tsconfig.json` excludes the specs, and it cost a vacuous test

While fixing the Bun defect below I declared `resolve` on `StageAppRuntimeInput`, spelled it correctly at the call site in the spec, and it was **silently dropped**: `packages/infra/cli/tsconfig.json` excludes `src/**/*.spec.ts`, so no excess-property check ever ran on it. The test that was supposed to red the new guard passed with the guard removed. `discoveryOptions()` now builds those options once, so a field cannot reach two of the three lookups and miss the third — and the red run is what caught it, not review.

---

## A defect this PR found in the resolver, and fixed at the core

`Cross-runtime (node+bun+deno)` went red on **`Test CLI on Bun`** while Node stayed green. Verified as a regression at job/step level rather than from the workflow verdict: on `main` `a63f022c` that same job ran and `Test CLI on Bun` was `success`.

The failing line named the defect exactly:

```
✖ stageAppRuntime › stages what it found and NAMES what it did not
  Expected: undefined (undefined)
  Actual:   Ship Demo.app/Contents/Frameworks/node-gi/prebuilds/darwin-arm64/node_gi.node (string)
```

Reproduced locally under Bun, then probed. Under Bun, `createRequire(<anchor>).resolve('@gjsify/node-gi')` falls back to Bun's **global install cache** instead of throwing — a sixth anchor `resolveNpmPackage`'s documented list does not have:

```
node 24.19.0  →  MODULE_NOT_FOUND
bun 1.3.14    →  ~/.bun/install/cache/@gjsify/node-gi@0.44.0@@@1/index.js
```

**Not just the empty-directory case**, which would make it a fixture problem. Measured across realistic project shapes:

| project shape | node | bun |
|---|---|---|
| `package.json`, no `node_modules` | MODULE_NOT_FOUND | **cache hit** |
| `package.json` + deps declared, not installed | MODULE_NOT_FOUND | **cache hit** |
| `package.json` + *empty* `node_modules` | MODULE_NOT_FOUND | MODULE_NOT_FOUND |

So a fresh clone, or any project using Bun's auto-install, would have had `@gjsify/node-gi`'s addon **staged from a global cache into a redistributed `.app`** — bytes the author never declared, at whatever version that machine cached, while the `install @gjsify/node-gi` line that should have printed does not. It is #910's shape one level up: an addon paired with a closure it was not built against.

**The fix is `resolveInstalledPackage()`**, beside the resolver whose behaviour it corrects. `resolveNpmPackage` is deliberately left alone — resolving `rolldown` from a cache is fine, because a *tool* runs and is forgotten; *payload* is not. The discriminator is a **control**, not a list of cache paths: hardcoding `~/.bun/install/cache` would be this repo guessing at another runtime's layout and going stale the first time Bun moves it. Asking the same question from a directory inside no project asks the *runtime*; if both anchors answer with the identical path, the project contributed nothing. Under Node the control can only throw, so it is a strict no-op there, and the control anchor need not exist, so it costs no filesystem access. `node-runtime.ts` takes the same guard — an interpreter shipped out of a global cache is one the author never declared.

**Red on both runtimes without the guard, green with it:**

| | Node 24.19.0 | Bun 1.3.14 |
|---|---|---|
| guard removed | **1 of 3209 failed** | **2 of 3206 failed** (guard test + the original defect) |
| guard present | 3211 completed | 3212 completed |

The first cut of that test *was* vacuous on Node — it used a plausible-looking cache string, and `resolveNodeGiAddon` stats the addon before staging it, so a path that does not exist answered `null` and the case passed with the guard removed. The fixture is a real populated directory now, which is what makes it red on both.

## Rebased onto #1411

`main` moved (freedesktop localisation). One conflict, in `assemble()`, where both sides wanted the same lines: #1411 translates the metadata through `localizeMetadata`, M2b renders the launcher from `runtime.launcher`. Both are kept — they are independent.

`packages/infra/cli/lib/` was rebuilt before anything was believed, and the ship e2e suites re-run rather than assumed: **135/135** across `ship`, `ship-macos`, `ship-layout`, `ship-from-stage` and `ship-declaration`. Unit: **node 3216, bun 3217**. Every gate, format and lint re-run clean after the rebase.


JumpLink:
**Three follow-up commits on top of the feature, each with its own measurement.**

`ci(node-gi): match the .app's paths by suffix` — `process.execPath` is `uv_exepath`, which `realpath()`s, while the launcher computes `$here` logically (`cd … && pwd`). An absolute comparison in the macOS leg would have been a bet on no component of the runner's temp directory being a symlink, and `/var` → `/private/var` is that bet lost. The claim is that the interpreter and the closure are inside *this* bundle; a bundle-relative suffix says it without the bet.

`test(ship): name the staged bundle file distinctly` — a local `const bundle` shadowed the suite-level one (the packed artifact) with a path to a file inside the stage.

`test(ship): stop two cases measuring the machine` — the one worth reading. `resolveNpmPackage`'s **last** anchor is the CLI bundle's own `node_modules` chain, which is exactly what makes *"only `@gjsify/cli` installed, no gjsify checkout"* work. So whether `@gjsify/node-gi` resolves from a throwaway fixture is a property of the tree the suite runs in — and this repository has both trees: `node-gi.yml`'s consumer jobs `ln -s` node-gi at the workspace root, `main.yml`'s jobs do not.

Measured rather than reasoned. With that link present, three cases on this branch went red:

```
✖ stageAppRuntime › stages what it found and NAMES what it did not   (on missing.length === 4)
✖ resolveNodeGiPackage › returns null for a package root with no JavaScript at all
✖ CONTROL: a project without the packages stages no runtime …        (on an assert.throws)
```

Each now asserts what the fixture controls instead:

* the count becomes the three package **names**, which are unconditional — both runtime packages are installed nowhere in this tree, and `@gjsify/node-gi` is named through its missing **addon** either way, since `prebuilds/` is gitignored;
* the `.js`-less fixture gets an `.mjs` entry, so it *resolves* and still copies to nothing — the only shape that reaches that floor at all, and the shape node-gi would have if it renamed its entry points (which the exports-map gate would catch first);
* the e2e asserts `startsWith(<the bundle>)` rather than "it resolved", a predicate nothing outside the `.app` can satisfy.

Both suites are green with the link present **and** absent now; before, only without it. 28/28 e2e and 3206 unit assertions in each tree.

Issue #1438, opened 2026-08-31
Two node-gi class-realization gaps make the React Native suite fail 2 of 857 under Node


JumpLink:
`@gjsify/react-native`'s suite is green under GJS and fails **2 of 857** under Node + `@gjsify/node-gi`, measured in a **fresh worktree** with both build halves rebuilt from HEAD. Both failures are node-gi class-realization gaps, and both are diagnosed differentially against the GJS oracle.

| run | result |
|---|---|
| `gjs -m test.gjs.mjs` | 318 test rows, **0 failures** |
| `gjsify run test:gjs-on-node` | **2 of 857** |

> **Corrected twice.** This issue first reported 47 failures, then 33. Both numbers were artefacts of how the measurement was wired, not of `main` — see § *Why the first two numbers were wrong*, which is the more useful half of this issue for anyone measuring this layer on another operating system.

## The two real failures

**1. `lists.spec.ts` — "binds rows into a `Gtk.ListItem`…"**

`GObject.Object.list_properties.call(Gtk.ListItem)` returns `array(9)` under GJS and **`array(0)`** under node-gi. The portable spelling `Gtk.ListItem.list_properties()` returns 9 on both, so the gap is specifically the `.call()` form against a class the process has not realized.

**2. `router.spec.ts` — "emits `popped` on the view and `hidden` on the page"**

`GObject.signal_lookup('popped', Adw.NavigationView.$gtype)` returns **84** under GJS and **0** under node-gi, until some class-struct static realizes the class. node-gi's class proxy takes no `g_type_class_ref` — `src/calls.cc` names this asymmetry itself.

Both are recorded in `status/open-todos.md`.

## Why this matters for the OS axis

ADR 0024 stages 4 and 5 put a shipped `.app` and `.msi` on **Node** — there is no relocatable GJS to put in a bundle, and no GJS host on Windows at all — so on two of the three declared target operating systems the Node path is the only path. With the number at 2 rather than 33, the honest summary is that this layer is **substantially green under Node**, and these two are specific, named, fixable gaps rather than a systemic problem.

A darwin/win32 leg should be read against this baseline: a run showing these two has measured this issue, not the operating system.

## Why the first two numbers were wrong

Both mistakes were the same shape — **a failure count reported without the configuration that produced it** — and both are worth writing down because each one produces a confident, plausible, wrong answer.

[...]

PR #1442, opened 2026-08-31
feat(style): a theme registry over the display

JumpLink:
> Stacked on #1440, which is stacked on #1439. Review the top commit only.

`sheet.ts` owns the *generated* document and deliberately owns nothing else, which left the other half an application needs with no answer at all: its **own** GTK CSS document, and Adwaita's named colours set from its design tokens. `PlatformColor` → "Adwaita named colours" has been sitting in the support table as P3 for exactly that reason.

It is a **registry** and not a setter, because the shape it has to fit is several named themes with one selected at runtime — register a name, ask for the current one, switch from a settings row. **One neutral theme ships now.**

```ts
import { NEUTRAL_THEME, ThemeRegistry } from '@gjsify/gtk-host/style';

const themes = new ThemeRegistry();
themes.register(NEUTRAL_THEME).register({
    name: 'ocean',
    namedColors: { 'accent-bg-color': 'rgb(53 132 228)' },
    css: '.card { border-radius: 16px; }',
    defaultOn: ['linux'],
});
themes.selectDefault(process.platform); // …or themes.select('ocean')
```

`NEUTRAL_THEME`'s document is **empty**, and that is a statement rather than a placeholder: a GTK application should look like the desktop it is on, and the accent belongs to whoever picked it in Settings. Selecting it is how an application returns to Adwaita *exactly* — measurably, because reloading the provider to the empty string restores every named colour to libadwaita's own value.

## Three measurements, and the plausible guess is wrong in all three

**`Adw.StyleManager:accent-color` is read-only** on libadwaita 1.9.3. The ParamSpec carries no `WRITABLE` flag and assignment throws *Property AdwStyleManager.accent-color is not writable*. So "set the accent from the application's tokens" is not a property write at all: it is a CSS custom property, `--accent-bg-color` and its 47 siblings redefined on `:root`, which is what Adwaita's own rules resolve against — measured end to end.

The 48 names are a committed measured table with a **negative direction** beside it, because a name libadwaita does not define is a valid custom property that nothing reads, so a theme setting one silently does nothing. `borders` is the interesting negative row: a real Adwaita colour under the **legacy** `@define-color` mechanism and not a variable — which is why an underscored name is refused with the hyphenated spelling named.

**The priority is the whole design, and it is not `APPLICATION`:**

| Measured | Consequence |
|---|---|
| libadwaita's own stylesheet sits at **exactly** `PRIORITY_THEME` (200) — 199 loses, 201 wins, 200 wins *only* by being added later | a theme document must be strictly above 200 or its effect is a load-order race |
| the generated sheet is at `PRIORITY_APPLICATION` (600), and at equal priority the **later** provider wins | a theme at 600 installed after the sheet clobbers the utility class written at the element |
| priority resolves **before** specificity — low-specificity at 400 beats high-specificity at 200 | the number is the only lever |

`PRIORITY_SETTINGS` (400) is GTK's own named value in the one gap that satisfies both constraints, and is measured to do so **whichever document is installed first**.

**One provider, whose document is replaced.** Remove-and-re-add works too, measured. Replacing is chosen because a provider installed once can never end up at the wrong position relative to the generated sheet, however many times a user switches.

## Shared discipline, and a trap

The containment probe moves to `document.ts` and both callers share it — an application's document is the one this package did *not* write, so it deserves the discipline more than a generated rule does. It is probed at **registration**, so a malformed rule names the line that declared the theme rather than the settings row that switched to it hours later.

**A measurement trap, kept because it cost a wrong conclusion:** reading the switch back from a `Gtk.WidgetPaintable` render node reports the **old** colour after a provider is removed — the node is cached and a removal does not invalidate it, while a subsequent add does. That reads exactly like "removal does not work" and is false. Every vector resolves `var(--name)` and reads `get_color()` instead.

The platform is a **parameter**: `selectDefault(platform)` never reads `process.platform` itself, which keeps a `process` read and its OS declaration out of a package whose job is GTK, and lets a settings screen preview another desktop's look.

## What holds it

19 vectors, and the four decisions were A/B checked: moving the priority to `APPLICATION` reddens two, dropping the name validation reddens two, probing at selection instead of registration reddens one, ordering the palette after the document reddens one. The "loses to the generated sheet" vector selects the theme **after** the sheet is installed on purpose — selecting first let it pass under the `APPLICATION` probe for the wrong reason, which is the whole order-dependence the constant exists to avoid.

`gjsify run test` green in `@gjsify/gtk-host` (2347) and `@gjsify/react-native` (897); `gjsify run check` clean in both.


JumpLink:
### Rebased onto `main`, and one thing found on the way

Rebased past the squashes of #1439 and #1440. Verified at the tree rather than at the conflict report: all three original patches came out **hunk-identical**, `status/open-todos.md` reconciled to main's 154 headings **+ my 2** with none of main's lost, `gjsify run check` clean and both suites green (gtk-host **2347**, react-native **897**).

### A fourth commit: the OS-axis conformance rule matched prose

`audit-runtimes --check` went red on this branch and green on `main`, so I checked rather than assumed. The rule reported `@gjsify/gtk-host` as *"reads the host operating system"* on the strength of two **JSDoc lines** in `theme.ts` — one of which is the sentence saying the module *never* reads `process.platform`. Its patterns are textual, and text does not know a comment from a statement.

I did not reword the documentation to appease the regex, and did not declare `gjsify.os` on `gtk-host` — the package genuinely makes no OS decision (`selectDefault(platform)` takes the platform as a parameter), so that declaration would have been false, and the win32 claim belongs to the strand that measured it (#1446).

**The fix and what it uncovered.** `decidesOnOs` now strips comments, one-directionally: it may never blank a line carrying code, because a hidden decision is a wrong `supported` claim while a surviving comment is only a false report someone argues with. The `startsWith('//')` is load-bearing — `import Gtk from 'gi://Gtk?version=4.0'` *contains* `//`, so a naive "cut at the first `//`" would have deleted the import.

Measured across the tree: **14 matches in 9 packages were comment-only**, and **three** packages dropped to zero sites. One was the false positive. The other two were being caught **by accident, through a comment**, because no pattern matches how they actually decide:

| Package | How it really reads the OS | Matched by |
|---|---|---|
| `@gjsify/os` | `mapSysname(cli('uname -s'))` — no `process` to ask | nothing |
| `@gjsify/child_process` | probes `/proc/self/status`, the macOS version plist, `/system/bin/sh` | nothing |

Both already declare `gjsify.os` correctly, so nothing in the tree was wrong — but fixing the false positive alone would have silently removed the coverage holding those declarations in place. So two patterns join the list, each with a package behind it, as the list's own doctrine requires.

The `uname` pattern is anchored on the **quote**, and the audit itself found why: a bare `\buname\b` also matches `@gjsify/tar`, where `uname` is the POSIX tar header's *user-name* field.

**Net, measured:** the full audit report is byte-identical to main's except one pre-existing non-enforced note, the OS-axis summary is unchanged at **16 declared / 15 deciding / 13 below `supported`**, and no package lost coverage. The commit is separate so it can be reviewed or dropped on its own.

JumpLink:
### The darwin failure was a real defect in my test, and the fix changes the reason, not the number

`Adw.StyleManager:accent-color` is **writable** on the darwin runtime bundle, whose libadwaita is later and grew a setter. It is read-only on the host I wrote this on. The measurement was right; I wrote it as an invariant about the **layer** when it is a fact about the **closure**.

I did not flip the assertion or delete the vector. The claim that holds everywhere is about the effect — and it turns out to be a stronger reason than the one I gave:

> `accent-color` is an **enum**. `AdwAccentColor` has nine members (BLUE, TEAL, GREEN, YELLOW, ORANGE, RED, PINK, PURPLE, SLATE), so on **no** runtime, writable or not, can it carry a design token's `rgb(17 34 51)`.

Setting the accent as a custom property is therefore not a workaround for a locked property; it is the only mechanism that can express the value at all.

**The vector now asserts what is portable:** the property exists, its value type is an enum, and an arbitrary accent set through the registry is what Adwaita resolves — which is what goes red if someone rewires the seam to the property because their own machine happened to allow it. Writability is **read at runtime and printed**, asserted in neither direction, because both answers are correct:

```
[measured] Adw.StyleManager:accent-color is read-only here — both occur across
the shipped libadwaita closures, and this layer depends on neither
```

I do assert the **value type** where I no longer assert the flag, and the difference is a risk class rather than a preference: gaining a setter is an additive change libadwaita can make and did, while changing an enum property into a colour would break its own ABI.

### Why the wrong reason mattered enough to correct everywhere

"The property is locked" and "the property is the wrong shape" lead to different code. The first invites *just set the property* the moment someone's machine allows it — producing an app that ignores its own accent token on every host where it is read-only, and can only ever reach nine colours where it is not. That reasoning had already been passed on as a product fact, so the correction is carried into the module comment, both docs pages, and the multi-theme TODO that repeated it.

### Two things found while fixing it

- **My new vector polluted a later group.** A registry installs its provider on the shared display and nothing takes it off, so a theme left selected outranked the 199/201 providers the priority group installs — reddening a vector in a group this one does not belong to. It now selects the neutral theme afterwards, which reloads the document to empty and restores libadwaita's own values, asserted against a baseline captured before any theme is installed. That is the neutral default's documented guarantee, used to keep one vector from deciding the next one's result.
- **`status/open-todos.md` gains the class**, since this was the third instance in a day of a correctly measured host fact written as a layer invariant — and all three surfaced only because a leg nobody watches went red. The note suggests the mechanism: an assertion whose subject is a runtime-provided flag or version is asserting a property of the closure, and the portable claim is almost always about the effect one layer up.

Local: `check` clean, gtk-host **2349**, react-native **897**, `audit-runtimes --check` exit 0, agent-context ratchet green, commitlint clean across the range. A/B confirmed the new vector reaches its branch.

JumpLink:
### Second defect in the same three lines, and it was masked by the first

`spec === null` was right: `GObject.Object.find_property.call(Adw.StyleManager, …)` resolves under gjs and returns **null** over the reverse bridge — node-gi's class proxy takes no `g_type_class_ref`, so a class the process has not realised reports nothing (#1438, whose original finding is the same shape with `list_properties`). Nothing throws; the caller gets null and reads it as *"the class has no such property"*.

It now goes through **`paramSpecs`**, which calls `list_properties()` directly on the class. That isn't a third spelling — it's the reader `gtk-props.spec.ts` already uses, and that is the evidence the form is portable on this leg: those vectors read `GtkWidget`, `GtkBox`, `GtkLabel`, `GtkCenterBox` and `GtkFlowBox` *as classes* and were among the 2345 that passed while this one failed. (I also confirmed the rebuilt `dist/test.node.mjs` no longer contains the borrowed form. I could not run the leg locally — `@gjsify/node-gi` is deliberately outside the workspace per ADR 0031 — so I leaned on that evidence rather than a rebuild.)

### My sweep gave you a false all-clear, and the rule caught it

I reported "only my site in 3585 tracked files". **That was wrong.** `lists.spec.ts` has two, written as

```js
GObject.Object.list_properties
    .call(Gtk.ListItem)
```

The pattern spans a **line break**, so a line-based grep structurally cannot see it. Both are fixed. Worth stating plainly because I handed you that all-clear as a measurement, and the mechanism is what contradicted it.

### So the class got the guard

`gjsify/no-gobject-method-borrow`, in the oxlint plugin that already exists — no new script, no new CI wiring. Refuses `GObject.Object.<method>.call(…)`/`.apply(…)`, with or without an explicit `.prototype`. **AST-based precisely because the grep failed**: the shape survives reformatting and the search did not.

A/B'd both directions — 4 bad spellings flagged (`find_property.call`, `list_properties.call`, the `.prototype` form, `.apply`), and the direct form, `[].slice.call` and `Object.prototype.hasOwnProperty.call` left silent.

`error` and not `warn` because the defect is **invisible on Linux by construction**: gjs answers correctly, so a green local run and a green main leg prove nothing about it. Only the reverse-bridge and darwin legs can see it, and it reached them twice from two authors who did not know the direct form existed.

### State

Repo-wide `gjsify lint` exits **0** with the rule on and zero occurrences left; gtk-host **2349**, react-native **898**; `audit-runtimes --check` exit 0; agent-context ratchet green; commitlint clean across all 6 commits. `status/open-todos.md` carries the grep-vs-AST lesson alongside the host-fact-as-invariant class.

JumpLink:
### Author's response to the review

Read all 14 commits. **Every finding against my work is correct**, and two go further than I had. One commit of mine is dropped, one is pushed on top.

**Dropped: my `test(style): two assertions that could not fail`.** `0f1163d68` found the same `-1 < 42` vector independently and went past me: I fixed the *assertion* and kept the vector's stated reason, and that reason was itself wrong — "palette FIRST so the theme's own rules can read what it just set" is not a constraint, since `var(--x)` resolves identically before its `:root`. The real reason is the override direction. Mine would have been a weaker duplicate with a false justification still attached, so it is gone rather than merged.

That also retires a claim I made in this PR: my A/B *"ordering the palette after the document reddens one"* validated only that the vector detects index order — not that palette-first matters. The A/B was sound about the assertion and empty about the property.

**Pushed: `828453802 style(lint): format the new oxlint rule source`**, one line, the escaped apostrophe `oxfmt` rewrites. Verified still needed rather than assumed: at `8350a2330` `format --check` exits 1 naming that file as the only offender, and the review never touched it.

### The accent correction — verified, not just accepted

I checked the mechanism myself before agreeing. Both assertions of the old vector ended in `.toBe(false)`, so both print `Expected: false / Actual: true` — byte-identical. And `triggerResult` **throws**, so the `spec === null` assertion two lines above short-circuited the flag read. The reported boolean was the null check; **the writability flag was never evaluated on that leg**. The claim was not a measurement that was too broad — it was inferred from the test's *name*. "Manufactured measurement" is the right name and the more dangerous form.

My conclusion survives only because of the reason I found second — `accent-color` is an enum of nine, so no runtime can carry a token colour — which was load-bearing all along while the writability was a side question we promoted to the main one.

### His work held to the same standard

I A/B'd two of his claims rather than take them on trust:

- **`dispose()`** neutered to a no-op → *"switches at runtime, and back to Adwaita exactly"* reddens. That vector was **green** under exactly that condition before his fix.
- **palette block never emitted** → *"emits the palette first, so a theme's own css can override it"* reddens **by name**. Mine stayed green under the identical sabotage.

I also checked that `65b15fd7a`'s consolidation to one exported table did not drop the load-bearing half: `NOT_ADWAITA_NAMED_COLORS` and the "defines NONE of the names the table claims are absent" vector both survive.

**Nothing I disagree with.** Local on `828453802`: `format --check`, `lint`, `audit-runtimes --check`, `check-doc-fences`, `check-girs-exact-pins`, agent-context ratchet all exit 0; `gjsify run check` clean in both packages; gtk-host **2353**, react-native **898**; commitlint clean.

PR #1475, opened 2026-09-02
fix(node-gi): collected is not yet finalized

JumpLink:
## The bug

A GTK app on Node + `@gjsify/node-gi` died ~1 s after startup on **Windows**, 6 runs out of 6,
with no survivor:

```
(node.exe:608): GLib-GObject-CRITICAL **: g_object_weak_unref_cb: couldn't find weak ref 00007FFCC19F3140(0000023BDE3A2550)
exit=-1073740940     # 0xC0000374 = STATUS_HEAP_CORRUPTION
```

The same app is fine on macOS (5/5) and Linux (6/6), which read for a while as a Windows port
defect. It is not: the defect is platform-independent and Windows' heap is just the one that
notices.

## The invariant that was violated

**An empty `napi_ref` proves the wrapper was COLLECTED, not that its finalizer has RUN.**

V8 resets the weak persistent inside the *first-pass* weak callback, during GC. Node then defers
the finalize callback to a `SetImmediate` (`node_napi_env__::EnqueueFinalizer`). Between those two
points `napi_get_reference_value` already reads empty while `NodeGiInstanceFinalize` has not run.

`SettleCollectedInstance` — the resurrection path — read that empty ref as *"the wrapper is dead
and its teardown is queued"* and freed the `NodeGiInstance` record. Its comment said as much:
*"Collected hit (wrapper GC'd, idle teardown still PENDING)"*. The teardown is only pending if the
finalizer already ran, and nothing established that.

The second act is what makes it fatal rather than merely leaky:

1. `delete old` frees the record;
2. the allocator hands **the identical block** straight back to the `new NodeGiInstance()` twelve
   lines later — measured, same address every run;
3. the stale finalizer then fires against the **live** wrapper's record, sees
   `teardown_queued == false`, and queues a teardown for it;
4. `RunTeardown` removes the *live* toggle ref → the GObject's last ref goes with it → it is
   finalized while JS still holds a reachable wrapper; `napi_delete_reference` runs on the live
   handle; `delete inst` frees the record a second time;
5. one drain later the live External's own finalizer arrives at a freed record and a freed
   GObject. On Linux glibc has already recycled that memory, `G_IS_OBJECT` fails and the process
   takes SIGSEGV. On Windows the freed block still passes `G_IS_OBJECT`, so it walks on into
   `g_object_weak_unref` — whose weak ref Settle removed in step 1 — and prints exactly
   `couldn't find weak ref <callback>(<inst>)` before `g_object_remove_toggle_ref` on a dead
   object corrupts the heap.

That is also why the app **stops crashing under `NODE_GI_TOGGLE_DEBUG=1`**: the log's own
`g_strdup_vprintf`/`g_free` traffic recycles the freed block, the stale finalizer reads garbage
instead of the live record, and the whole chain collapses into a silent no-op. A heisenbug by
construction.

## Minimal reproduction

Deterministic, headless, no GTK, no display — 5/5 SIGSEGV on Linux before the fix:

```js
import { requireGi } from '@gjsify/node-gi/gi';
const Gio = requireGi('Gio', '2.0');

(() => { const c = new Gio.Cancellable(); c.push_current(); })();

globalThis.gc();                          // wrapper collected, finalizer PENDING
const back = Gio.Cancellable.get_current();   // resurrect inside the window
await new Promise((r) => setImmediate(r));    // the stale finalizer runs here
console.log(back.is_cancelled());             // 💥
```

Two details are load-bearing and are both spelled out in the test:

- the vehicle must be a **borrowed** C pointer. `g_cancellable_push_current` stores the pointer on
  a thread-local stack with no ref, so the GObject stays reachable from C at refcount 1 — the only
  shape whose wrapper is weak (collectable) while C can still hand the object back. Anything that
  *refs* its member toggles the wrapper UP and pins it, which is why a churn probe over 8000
  `Gtk.Label` create/remove cycles survives happily;
- there must be **no `await` between `gc()` and the re-fetch**. Any loop turn drains the finalizer
  queue first and closes the window. That is precisely why the existing resurrection case
  (`gc-identity` case 4, which does `await settle()` before re-fetching) never caught this.

Run on **Windows** against the app's own installed (unfixed) `@gjsify/node-gi`, that same
12-line program reproduces the application's crash signature byte for byte — 4 of 5 runs, no
GTK, no React Native, no app:

```
resurrected, cancelled = false
(process:3096): GLib-GObject-CRITICAL **: g_object_weak_unref_cb: couldn't find weak ref 00007FFCC5C43140(00000148DC12C880)
run 1 exit=-1073740940
```

(The callback address differs from the app's `00007FFCC19F3140` only by DLL load base; it is the
same `OnGObjectFinalized`.)

Confirmed with a temporary probe build:

```
(PROBE) make:      new inst 0x18fa3060 for obj 0x18f9eea0
(PROBE) settle:    delete old inst 0x18fa3060 (teardown_queued=0 obj 0x18f9eea0)   <- freed while pending
(PROBE) make:      new inst 0x18fa3060 for obj 0x18f9eea0                          <- same address
(PROBE) finalize:  inst 0x18fa3060 (teardown_queued=0 gobject 0x18f9eea0)          <- stale, hits the live record
GLib-GObject-CRITICAL **: g_object_get_qdata: assertion 'G_IS_OBJECT (object)' failed
```

## The fix

Ownership of the free follows the **finalizer**, never the collection:

- `teardown_queued == true` ⇒ the finalizer already ran and nobody else will free the record, so
  Settle frees it as before;
- otherwise the finalizer is still pending and **it** frees the record; Settle marks the record
  `settled` and leaves it allocated.

Settle additionally disarms every handle the record still carries (`gobject`, `toggle_added`,
`handle_ref`), so a finalizer that ever reached a settled record would find nothing to act on.
`NodeGiInstanceFinalize` frees a `settled` record and returns.

`teardown_queued`'s comment is corrected while we are here: it is not a dedupe flag, it is the
answer to "has the finalizer run".

If a process dies before draining its finalizer queue a settled record is not freed — the same
bounded leak-at-exit the shutdown-dropped teardowns already accept, and at most one record per
racing resurrection.

## Verification

| | |
|---|---|
| `node --test --expose-gc` (full node-gi suite, Linux) | **571 tests, 567 pass, 0 fail**, 4 skipped |
| new case against the **unfixed** addon | fails — `TypeError: invalid GObject handle` (the live wrapper had been torn down under it) |
| new case against the fixed addon | passes |
| minimal repro, unfixed | 5/5 SIGSEGV |
| minimal repro, fixed | 5/5 exit 0 |
| the real app on Windows 11 (win32-x64, Node 24.18.1), unfixed | 6/6 `STATUS_HEAP_CORRUPTION` |
| the real app on Windows 11, fixed | **pending** — to be run against this PR's `node-gi-prebuild-win32-x64` once CI publishes it, and reported here |

The regression test is `gc-identity` case 4b, next to case 4 whose blind spot it covers.


JumpLink:
## Windows acceptance, measured

Run on the win32-x64 VM (Windows 11 25H2, Node 24.18.1) against the real application — the CORRECTIV desktop host, React Native on GTK4 through `@gjsify/react-native`. Same machine, same app build, same interactive session; the only variable is the addon.

| addon | runs | crashed | rendered + captured |
|---|---:|---:|---:|
| installed `@gjsify/node-gi@0.45.0` (unfixed) | 13 | **11** | 2 |
| this PR's `node-gi-prebuild-win32-x64` | 12 | **0** | **12** |

Every crash is the same signature the PR body opens with: `STATUS_HEAP_CORRUPTION` (`exit=-1073740940`), preceded by `g_object_weak_unref_cb: couldn't find weak ref`.

The two addons are byte-different and both 510464 bytes, so the staging was verified by hash rather than by size: `cc373c73765829db…` (this PR's artifact, staged) against `63a0376d2f5a00e8…` (the one it replaced).

**The crash rate is load-dependent, which is worth recording.** An earlier session measured 6 of 6 on this same VM while a second VM and several builds were competing for RAM; running alone it is 11 of 13. That is consistent with a GC-window race rather than evidence against it — and it is why the comparison above uses a twelve-run baseline instead of the handful that first looked deterministic.

The screenshot each surviving run produced is the application's onboarding screen, rendered through the GTK4/GSK path, not an empty window — so this is the app working, not merely a process that failed to die.

Unrelated to this fix and noted for elsewhere: the button label wraps to two lines on Windows where macOS and Linux keep it on one.

PR #1476, opened 2026-09-02
feat(node-gi): bundle the http(s) source and its TLS

JumpLink:
The runtime bundles ship `playback` — playbin3 and uridecodebin3, elements whose entire job is to take a URI — while the only source behind them was `filesrc`, riding in on `coreelements`. So the bundles advertised URI playback and could open exactly one scheme.

Measured on the published `darwin-x64` and `win32-x64` bundles:

| | |
|---|---|
| `Gst.ElementFactory.make('playbin3' \| 'decodebin3' \| 'filesrc')` | an element |
| `Gst.ElementFactory.make('souphttpsrc')` | **null** |
| `Gio.tls_backend_get_default().supports_tls()` | **false** |

A desktop app played its bundled podcast episode and found no source element for its live stream, on the two platforms where the bundle **is** the runtime.

## The half that would have looked finished

Fixing only the first row ships a broken feature that reads as working. `GTlsConnection` has no implementation inside GIO: glib-networking ships one as a module GIO `g_module_open`s out of a directory whose compiled-in default is the **build machine's** prefix. The bundles ship their own `libgio` and shipped no module, so every TLS request in a bundle-activated process got the dummy backend — and `souphttpsrc` reports that as `Internal data stream error`, a network error to read and a missing module in fact. `@gjsify/{tls,http2,ws}` fail the same way one layer up.

Neither payload is reachable by a link walk **on darwin**, so both are seeded explicitly — the third instance of the shape librsvg established. There the plugin `g_module_open`s libsoup by leaf name (so `otool -L` reports glib and gstreamer and nothing else), and nothing links a GIO module at all. On **win32** gst-plugins-good takes the `host_system == 'windows'` branch and *links* libsoup — `gstsoup.dll`'s import table names `soup-3.0-0.dll` — so the seed there is load-bearing only for `Soup-3.0.typelib`'s backing, which is what its comment now says.

## Gates, so absence fails the build and not the user

- The TLS backend is a declared `tls-backend` runtime data set, so an empty module dir fails the build *and* `verify-bundle-manifest.mjs` before publish.
- `GST_REQUIRED_PLUGINS` (`app`, `playback`, `soup`) names the plugins whose absence is a build failure rather than a counted skip, checked against the set the builders actually copied. The existing zero-plugin check passed while the one element an app asked for was missing.
- Windows gains two gvsbuild projects (`glib-networking`, `libsoup3`) and a second invocation, because gst-plugins-good's `soup` feature is meson `auto` and **silently** produces no plugin when libsoup is absent from the prefix. The prefix assertion now names `gstsoup.dll` instead of counting plugins.

## Cost

Measured per platform as the file-set difference between the published `0.45.0` bundle and this branch's own CI artifact — the two are **not** the same order of magnitude:

| | added | bundle | |
|---|---|---|---|
| darwin-x64 | **9.36 MiB** | 72.7 MiB | +13 % |
| win32-x64 | **47.80 MiB** | 77.5 MiB | **+62 %** |

The darwin total matches what this PR first claimed; its *base* did not (85 MiB is the bundle **after** the addition). win32 was never measured, and 34 MiB of its 47.8 is **ICU** — `icudt78.dll` alone is 31.6 MiB — pulled in transitively because gvsbuild builds libpsl against ICU where Homebrew builds it against libidn2 + libunistring. Nothing about the decision needs ICU; reducing it is gvsbuild-side follow-up, recorded in ADR 0037 rather than treated as solved.

The licence characterisation was also wrong. The payload is **mixed**, not all LGPL: LGPL-2.1+/LGPL-3+ (libsoup, glib-networking, gnutls and its closure), **Apache-2.0** (OpenSSL 3 — gvsbuild's TLS backend is `gioopenssl`, not `giognutls`), **MIT** (libpsl, nghttp2, MIT Kerberos), BSD (p11-kit), Unicode (ICU), public domain (SQLite). The substantive claim survives: no codec and no patent-encumbered decoder enters, which is the category the allowlist header keeps out.

## Scope

The curation rule is not widened by argument. What was in question is where the audio path *begins*: reading a URI is the same operation as `filesrc`, which already ships. Encoders, video decode, capture, and streaming **sinks**, servers and adaptive-streaming demuxers stay out. `appsrc` remains the right answer for bytes you already hold, and nothing here changes it.

## Notes for review

- ~~This has not been proven on a built bundle yet.~~ **It has.** This PR's run assembled the widened bundle on all three targets from this checkout — `macos-gtk-windowing-runtime` (both arches) and `windows-gtk-windowing-runtime` (45m37s, the cache-key bump forced a real gvsbuild rebuild) — and every downstream job consumes that run's `actions/upload-artifact`, not a published tarball. `gst-elements.test.mjs` passed 4/4 on the staged darwin-x64 bundle, `souphttpsrc` and the TLS backend included; the win32 prefix assertion found `gstsoup.dll` and 2 GIO modules. The caveat is resolved. (The separate `GTK suites on the shipped closure` legs *do* npm-install the **published** bundle, so their green says nothing about this change.)
- ADR is numbered **0037**: it was written as 0034 against an older `main`, which has since taken 0034 through 0036. No collision; `docs/adr/README.md` is consistent.
- ADR status stays **`Accepted`**. That is the repo's convention for a decision landing with its implementation (30 of 37 ADRs are `Accepted`), and this one ships builders, gates, tests and a green build of the artifact. The three `Proposed` ADRs above it are proposals whose implementation has not landed — a different situation, not a precedent.

## Review corrections landed on this branch

- **A real defect, fixed:** the darwin closure walk copied all 24 GStreamer plugins into flat `lib/` as well, +3.4 MiB of duplicates. `otool -L` prints a dylib's own `LC_ID_DYLIB` among its dependencies and `otoolDeps` dropped one line as "the id" when line 0 is the `<file>:` header. Inert until `resolveBrewDep` began resolving whole references for keg-only formulas, at which point the id resolved to the plugin itself. Verified on the artifact: 24/24 plugins duplicated, 0/1 GIO modules (a loadable bundle has no id) — the control that identifies the mechanism. Gated for the class via `duplicatedModuleLeaves()` in both builders, with unit tests.
- **Still open, and it should block publish:** `libcrypto-3-x64.dll` and `libssl-3-x64.dll` ship in the win32 bundle with **no licence text** (45 components attributed, none is OpenSSL). The win32 licence step attributes by *prefix*, and `assertLicenseCoverage` runs its per-binary checks only under `per-binary` attribution, so an unlicensed family is invisible to every gate. Apache-2.0 § 4 requires the licence to travel with the binary. **INTRODUCED by this PR, not merely exposed** — corrected after measuring the published tarball rather than reasoning about the gate. `npm pack @gjsify/gtk-runtime-win32-x64@0.45.0` contains **zero** files matching `libcrypto`/`libssl`, and 45 licence texts of which none is OpenSSL. The DLLs arrive because this change adds `glib-networking` and `libsoup3` to the gvsbuild build and gvsbuild's TLS backend is `gioopenssl`. What IS pre-existing is the gate weakness that lets it through: prefix attribution plus `assertLicenseCoverage` running its per-binary checks only under `per-binary`. So the hole is old and the payload falling into it is new, which is why this blocks publish rather than being filed as debt.

### DIFF
--- a/.github/workflows/gtk-os-suites.yml
+++ b/.github/workflows/gtk-os-suites.yml
@@ -364,8 +364,17 @@ jobs:
       # bundle's own recorded platform to THIS runner, so an artifact mixup cannot
       # run the arm64 closure on Intel and surface four steps later as an
       # inscrutable dlopen error.
+      #
+      # `--allow-legacy-license-record` is what tells the shared gate this is the
+      # PUBLISHED-closure role and not the pre-publish one. `licenses.binariesCovered`
+      # postdates the 0.45.0 tarballs, so requiring it here turned all three of these
+      # legs red at once over a property no already-published artifact can acquire —
+      # only the next release writes it. The allowance is narrow (an absent record
+      # only; a recorded ZERO still fails here) and self-retiring (the step prints that
+      # the flag is droppable the first time a published bundle carries the field).
+      # release.yml passes NO flag: there the bundle was just built.
       - name: Verify the published bundle's manifest
-        run: node packages/node-gi/scripts/verify-bundle-manifest.mjs --bundle "packages/node-gi/node-gi/prebuilds/$TARGET/gtk" --expect-host-target darwin
+        run: node packages/node-gi/scripts/verify-bundle-manifest.mjs --bundle "packages/node-gi/node-gi/prebuilds/$TARGET/gtk" --expect-host-target darwin --allow-legacy-license-record
 
       # Which two halves this run paired, in the log rather than in a reader's head.
       - name: Report the pairing this run measured
@@ -509,8 +518,10 @@ jobs:
       - name: Stage the SHIPPED closure (published bundle + published addon)
         run: node packages\node-gi\scripts\stage-published-gtk-runtime.mjs --link-into packages\framework\gtk-host --link-into packages\framework\react-native
 
+      # `--allow-legacy-license-record` for the same reason as the darwin leg above:
+      # the PUBLISHED closure, whose manifest predates `licenses.binariesCovered`.
       - name: Verify the published bundle's manifest
-        run: node packages\node-gi\scripts\verify-bundle-manifest.mjs --bundle "packages\node-gi\node-gi\prebuilds\%TARGET%\gtk" --expect-host-target win32
+        run: node packages\node-gi\scripts\verify-bundle-manifest.mjs --bundle "packages\node-gi\node-gi\prebuilds\%TARGET%\gtk" --expect-host-target win32 --allow-legacy-license-record
 
       - name: Report the pairing this run measured
         run: |
--- a/.github/workflows/node-gi.yml
+++ b/.github/workflows/node-gi.yml
@@ -1596,8 +1596,13 @@ jobs:
       HOMEBREW_NO_AUTO_UPDATE: '1'
       HOMEBREW_NO_INSTALL_CLEANUP: '1'
     steps:
+      # `glib-networking` is NAMED although brew's `gstreamer` already depends on it.
+      # It supplies lib/gio/modules/libgiognutls.so — the GIO TLS backend the bundle
+      # ships as its `tls-backend` data set, without which souphttpsrc resolves and then
+      # fails on every https URL. A transitive dep is a fact about today's formula; this
+      # is the one the builder asserts.
       - name: Install GTK4 + Adwaita + GtkSource + icons + GObject-Introspection (build-time closure source)
-        run: brew install gtk4 libadwaita gtksourceview5 adwaita-icon-theme gobject-introspection cairo pango gdk-pixbuf graphene pkg-config gstreamer
+        run: brew install gtk4 libadwaita gtksourceview5 adwaita-icon-theme gobject-introspection cairo pango gdk-pixbuf graphene pkg-config gstreamer glib-networking
 
       - name: Use Node.js 24
         uses: actions/setup-node@v4
@@ -2452,7 +2457,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: C:\gtk-build\gtk\x64\release
-          key: gvsbuild-gtk4-gst-gi-${{ env.GVSBUILD_VERSION }}-x64-v2
+          key: gvsbuild-gtk4-gst-soup-gi-${{ env.GVSBUILD_VERSION }}-x64-v3
 
       # Build-time closure SOURCE only. Verbatim from the `windows` job.
       - name: Download + extract gvsbuild GTK4 (build-time closure source)
@@ -2544,6 +2549,19 @@ jobs:
           # caught it, at the prefix, instead of three jobs later as "no element
           # decodebin" — typelibs are the only reason we build this from source
           # rather than lifting a binary.
+          # TWO INVOCATIONS, AND THE ORDER IS THE POINT. gst-plugins-good's `soup`
+          # feature is meson `auto` and its ext/soup/meson.build calls `subdir_done()`
+          # with a MESSAGE, not an error, when neither libsoup2 nor libsoup3 is found —
+          # so building -good against a prefix without libsoup produces a complete,
+          # green build with no gstsoup.dll in it, which is exactly the silent gap
+          # gst-plugins.mjs § soup was measured on. gst-plugins-good does not DECLARE
+          # libsoup3 as a gvsbuild dependency, so nothing orders the two for us.
+          # `glib-networking` is named although libsoup3 already depends on it: it is
+          # the GIO TLS backend the bundle ships as its `tls-backend` data set, and on
+          # Windows the soup plugin LINKS libsoup (ext/soup/meson.build takes the
+          # `host_system == 'windows'` branch instead of the dlopen loader), so an
+          # https URL would still fail on the missing module alone.
+          gvsbuild build --configuration release --vs-ver $vsVer --enable-gi --build-dir C:\gtk-build glib-networking libsoup3
           gvsbuild build --configuration release --vs-ver $vsVer --enable-gi --skip libvpx --build-dir C:\gtk-build gstreamer gst-plugins-base gst-plugins-good
 
       - name: Assert GStreamer landed in the prefix (typelibs + plugins)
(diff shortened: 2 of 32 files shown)


JumpLink:
## The publish blocker is closed — and it was a gate that could not fail, not one payload

Rebased onto `origin/main` first (17 commits, no conflicts), then fixed the licence blocker this PR recorded.

### What the blocker actually was

`libcrypto-3-x64.dll` / `libssl-3-x64.dll` shipping with no licence text was the symptom. The cause is that `assertLicenseCoverage` ran its per-binary rules **only** under `per-binary` attribution, and the win32 call passes `prefix` — so what it asserted was *"at least one licence text was recovered"*. **A corpus of one file would have passed it.**

Measured on this branch's own `gtk-runtime-win32-x64-windowing` artifact rather than reasoned about:

| | |
|---|---|
| binaries shipped (`bin/` + `lib/**`) | **89** |
| projects the gvsbuild prefix documents | **45** |
| binaries whose project is documented **nowhere** | **14** |

The eight undocumented projects: `glib` (six DLLs — `libglib`, `libgobject`, `libgio`, `gmodule`, both `girepository`; LGPL-2.1-or-later), `freetype`, `graphene`, `libtiff`, `libxml2`, `zlib`, `sqlite`, `openssl`. **Only OpenSSL is new to this PR** — GLib has been unlicensed in every published win32 tarball, and the notice's "over-inclusive rather than silent" claim was true in one direction and empty in the other (it documents `cairomm`, `gtkmm`, `pycairo`, `protobuf` … which the bundle does not carry).

### How it is fixed

**Not** by moving win32 to `per-binary`: per-DLL attribution still is not recoverable from a flat gvsbuild prefix, so claiming one in the shipped notice would be inventing it. What was wrong is not that claim — it is that prefix attribution was *unfalsifiable*.

- **`WIN32_LICENSE_FAMILIES`** declares which project each bundled leaf belongs to. A **name map, never a statement of terms** (the terms still come from the build prefix), so it cannot make a false licence claim; and unlike a hand-listed licence *set* it cannot drift silently, because the gate walks the binaries the builder actually copied — a leaf the table does not know fails the build **by name**.
- **`assertLicenseCoverage` refuses a binary whose family the corpus documents no text for, in either mode**, and refuses `prefix` attribution offered with no family table at all. One gate, one registry — no second mechanism.
- The binaries checked are now **every** binary the tarball carries — GIO modules, GStreamer plugins, pixbuf loaders — matching what darwin already did.
- Where gvsbuild documents nothing (no install step for some; OpenSSL's project installs `LICENSE` while OpenSSL 3 ships `LICENSE.txt`), the text comes from the **upstream release the prefix pins**, vendored under `gtk-runtime-win32-x64/licenses-not-in-prefix/` with per-file provenance. The prefix stays authoritative and is never overridden, and a text enters only for a project some binary in *that variant* needs — the display-free bundle ships no OpenSSL terms.
- `manifest.licenses.binariesCovered` is recorded on **both** platforms and `verify-bundle-manifest.mjs` refuses a bundle without it, so a tarball from a pre-gate builder cannot publish either.

### The gate is red without the fix — shown, not assumed

Replaying the artifact's own binary set against its own corpus:

```
binaries shipped : 89
components in prefix corpus: 45
OLD gate problems: 0   <-- the published bundle passed
NEW gate problems: 14
    libcrypto-3-x64.dll is openssl, and no license text ships for openssl
    libssl-3-x64.dll is openssl, and no license text ships for openssl
    gio-2.0-0.dll is glib, and no license text ships for glib
    …
```

Second control: deleting one vendored directory (`licenses-not-in-prefix/openssl`) turns the unit suite red and names exactly those two DLLs and nothing else. `gtk-runtime-bundle-gates.test.mjs` 43/43, `release-bundle-gate` 40/40, `oxfmt --check` and `oxlint` clean (the latter verified with a discriminator).

ADR 0037's consequence is rewritten from "open, must not publish" to what actually happened, and `gst-plugins.mjs`'s pointer to it no longer says the OpenSSL text is missing.

**Bundle cost is unchanged** by this commit; the numbers in the PR body still stand: darwin-x64 **+9.36 MiB** (72.7 MiB total, +13 %), win32-x64 **+47.80 MiB** (77.5 MiB total, **+62 %**), of which 34 MiB is ICU (`icudt78.dll` alone 31.6 MiB) pulled in because gvsbuild builds libpsl against ICU. The vendored licence texts add ~70 KiB.

JumpLink:
## The three red `GTK suites on the shipped closure` legs — cause and fix

Short answer to the three questions: **it was mine, it was not the check being too strict, and it could not be fixed by rebuilding a manifest.** Pushed as `fix(node-gi): scope the coverage record to publish`.

### 1. Is the check too strict?

No — and specifically not in the way suspected. `binariesModified: true`, `install_name_tool` and the ad-hoc `codesign` live in `licenses.modifications` and are read by nothing in the coverage path. Coverage has never demanded a licence file *per binary*: on darwin every dylib from one keg shares that keg's texts, and on win32 a whole family shares its project's. The failing message names one thing only — the manifest has no `binariesCovered` field.

### 2. Can it be fixed by rebuilding the manifests?

Not from this PR. That step gates **the tarball a consumer already receives**: `stage-published-gtk-runtime.mjs` npm-installs `@gjsify/gtk-runtime-*@0.45.0` and the gate reads *that* manifest. It was written before the field existed; only the **next release** can produce one carrying it. So I had made a shipped artifact fail a check it can never pass — a permanently red leg, not a gate.

`verify-bundle-manifest.mjs` has two roles and I had only thought about one:

| role | call site | bundle under test |
|---|---|---|
| pre-publish | `release.yml` (both legs) | built in the same job → **must** carry the record |
| shipped closure | `gtk-os-suites.yml` ×3 | published 0.45.0 → predates the record |

The requirement now belongs to the first role. The two shipped-closure sites pass `--allow-legacy-license-record`; `release.yml` passes nothing.

**This does not relax what the gate measures.** The two cases are deliberately split:

- an **absent** record = a manifest older than the record → excusable, published-closure role only;
- a **recorded zero** = a builder saying it covered nothing → still fatal in *every* role.

And it is self-retiring in both directions: used, the step prints what it let through; unused, it prints *"--allow-legacy-license-record was not needed … Drop the flag from the call site"* — which is the first run after a release ships a manifest with the field. Three e2e cases pin both edges plus the recorded-zero case (`release-bundle-gate` 43/43, `gtk-runtime-bundle-gates` 43/43, format + lint clean).

### 3. Is `main` red on the same legs?

**No.** `gtk-os-suites.yml` on `main` is green on the last four commits (`e523066c6`, `3356ac7c6`, `ca4980ff5`, `d8483ebcb`). The field does not exist there, so the check does not exist either. The breakage was introduced by my previous commit and is fixed by this one — not a pre-existing fracture, and nothing to carry into the PR body as inherited debt.

### Where the win32 licence gate stands meanwhile

`Windows GTK runtime bundle (build)` went **green on the real gvsbuild prefix** with the new coverage gate in place — the first independent confirmation that the family table plus the vendored corpus covers a real bundle rather than only the artifact I measured. The `--windowing` legs were still running when this went up.

JumpLink:
### CI after the scoping fix — the three red legs are resolving green

| leg | before | now |
|---|---|---|
| GTK suites on the shipped closure (**win32-x64**) | fail | **pass** |
| GTK suites on the shipped closure (**darwin-x64**) | fail | **pass** |
| GTK suites on the shipped closure (**darwin-arm64**) | fail | queued for a macOS runner |

**34 pass, 0 fail**, 2 pending (both only waiting on macOS runners), 2 skipping. Also green on the way through, and the ones that matter for this change:

- **`Windows FULL-windowing GTK runtime bundle (build)`** — the decisive one. The `--windowing` win32 bundle assembled with `gstsoup.dll`, the GIO modules and `libcrypto`/`libssl` in it, **through the new coverage gate over all 89 shipped binaries**. So the family table plus the vendored corpus holds on the real gvsbuild prefix, not just on the artifact I measured it against.
- `Windows GTK runtime bundle (build)` — the display-free variant, same gate, different prefix (the plain GTK4 zip, 37 DLLs).

Not merging — over to you.

PR #1491, opened 2026-09-02
feat(cli): ship an application's own font faces

JumpLink:
## The measured gap

`@gjsify/react-native`'s support table has told consumers, in the `useFonts` row and in
`surfaces/expo-font.ts`'s prose, to

> Ship the font with the application (`gjsify ship` installs it where fontconfig looks)

**There was no such mechanism.** Measured on `origin/main` at `3356ac7c66`:

| probe | result |
|---|---|
| `git grep -i font packages/infra/cli/src/utils/ship/ packages/infra/cli/src/commands/ship.ts docs/ship-formats.md` | **0 hits** |
| `SHARE` — the closed set of directories a payload installs into | `applications`, `icons`, `schemas`, `mime`, `metainfo`, `locale`. No `fonts` |
| `FcConfigAppFontAddFile` / `AddFontResourceEx` / `CTFontManagerRegisterFontsForURL` outside `refs/` | 1 hit, and it is a comment saying the first "is not in any typelib here" |
| the only font-registration call in the tree | `PangoCairo.font_map_get_default().add_font_file()` in `dom-elements/src/font-face.ts` — Canvas `FontFace`, the app registering at runtime, nothing to do with packaging |

The failure mode is the expensive kind. Pango does not report a missing family, it
**substitutes** — the window renders, every test passes, the app merely looks wrong at
exit 0. On a GNOME box the developer never sees it, because the family is usually
installed there anyway.

## The ADR

[`docs/adr/0038-shipped-application-fonts.md`](docs/adr/0038-shipped-application-fonts.md).

It was **revised by its own measurements**: the first draft assumed the Linux mechanism
generalised to all three OSes, on a plausible argument from
`build-gtk-runtime-darwin.mjs`'s seed list (it pulls fontconfig in transitively). The
library being in the closure is not the backend being in use. That correction is written
into the ADR's *Do not* section, because it is the mistake a reader is most likely to
repeat.

Linux was **run on this machine**. macOS and Windows were researched against vendor
documentation and upstream source, and are marked as such throughout — never as
measurements.

## What happens per OS

`gjsify.ship.fonts` (a file or a directory, defaulting to `data/fonts`) stages faces into
`share/fonts/<appId>/` on the **one** prefix-relative plan, mapped per layout like every
other `share/` entry. No second stager, no per-OS payload. What *reaches* them differs in
kind:

| | mechanism | evidence |
|---|---|---|
| **linux**, `.deb`/`.rpm` | the stock `fonts.conf`'s unconditional `<dir>/usr/share/fonts</dir>` | measured, fontconfig 2.17.0 |
| **linux**, Flatpak | `<dir prefix="xdg">fonts</dir>` expanded over `XDG_DATA_DIRS` — which `fonts-conf(5)` does **not** document | measured in **eight** independently built fontconfigs (2.14.1 · 2.15.0 · 2.17.0 · 2.17.1 · 2.18.3), in any list position, recursively, with a cold cache and a negative control |
| **darwin** | `ATSApplicationFontsPath` in `Info.plist`, relative to `Contents/Resources` — the OS activates it for this app **at launch** | Apple's key reference; GTK4 on macOS is not linked against fontconfig at all (`fontconfig_dep = []  # only used in x11 backend`) and Pango there is CoreText |
| **windows** | **nothing declarative exists.** The launcher exports `GJSIFY_FONT_DIR`; the app calls `PangoCairo.FontMap.get_default().add_font_file()` | `pango_win32_font_map_init()` populates from DirectWrite alone; fontconfig is linked and never selected; `FcConfigXdgDataDirs` splits on a hardcoded `:`; `AddFontResourceEx` is GDI |

Three consequences worth calling out:

- **The Linux launcher is unchanged.** `XDG_DATA_DIRS` has been exported at the staged
  `share/` since ADR 0024 § 3 for icons and schemas; the finding is that fontconfig reads
  it too. What changed is that the line is now load-bearing for a reader the man page does
  not list, so the comment above it says so and all three launcher forms assert it.
- **No `fc-cache` row in `cacheRefreshCommands`** — measured unnecessary (a cold
  `XDG_CACHE_HOME` resolves the face on the first call) and both distributions run it from
  their own fontconfig triggers. This is the one place the design *deletes* a step the
  other five `share/` directories have. `/usr/share/fonts` does join the rpm packer's
  system-owned set (`rpm -qf` → `fonts-filesystem`).
- **`Layout.metadata` now takes the payload.** A manifest key naming a directory must not
  be written over a bundle that carries none — the same seam `CFBundleDocumentTypes` will
  need.

Web-font wrappers (`.woff`, `.woff2`, `.eot`) are **refused by name** rather than filtered
away with the strays: FreeType's WOFF support is a build option
(`FT_CONFIG_OPTION_USE_BROTLI`), so one resolves on the packaging host — measured,
`fc-query` reads a `.woff2` fine here — and may not in the FreeType inside
`@gjsify/gtk-runtime-<target>`.

## Tests

`tests/e2e/ship-layout` is **extended, not duplicated**. The fixture carries a real 11 KB
face borrowed from this repo's own showcase assets, the way it already borrows a real
darwin dylib.

**Without the implementation the suite fails 7 tests** — verified by reverting `src/`,
rebuilding and re-running: the Linux payload list, all three launcher forms, the
`Info.plist`, and both gap notices.

Also: `utils/ship/discover-fonts.spec.ts` (10 cases, mostly refusals) and a `plist.spec.ts`
case asserting the key's presence **and** its absence.

## What stays open

Named in `status/open-todos.md`, not left as a comment:

- **The Windows call has no home in `@gjsify/*`.** One line in a GTK app's startup
  registers every face, and it belongs in the GTK host layer (ADR 0027) rather than
  copy-pasted per app — one call covers Windows *and* Linux, since `pango_fc_font_map`
  implements the same vfunc. macOS stays the odd one out (CoreText returns
  `G_IO_ERROR_NOT_SUPPORTED`), which is exactly why the plist key is the darwin mechanism.
- **The win32 GTK bundle ships fontconfig config nothing reads.** `etc/fonts` plus the
  `FONTCONFIG_PATH`/`FONTCONFIG_FILE` wiring is inert, and its comment implies fontconfig
  is sometimes in play — the same claim that made this ADR's first draft wrong. That is
  `packages/node-gi`'s tree, outside this workspace, and it is a *removal* of shipped
  bundle content, so it wants a Windows run behind it.

## Honestly unverified

- **That macOS activates the directory and Pango's CoreText map then holds the family.**
  The plist key is declared and the staged path is asserted against the tree; the
  activation is not, and no leg here runs a `.app`. `Layout.fontGap` prints it.
- **Every Windows claim.** Source and vendor docs only — nothing was executed on Windows.
- `tests/e2e/ship-msi` fails on this host for a pre-existing reason (`wixl` not on PATH),
  unrelated to this change.

## Review pass (second pair of eyes)

Re-measured and refactored on top of the commits above.

**The undocumented Linux claim was re-run independently, and it is stronger than
first written.** `<dir prefix="xdg">fonts</dir>` expanding over `XDG_DATA_DIRS` holds
in **eight independently built fontconfigs across five releases** — 2.14.1, 2.15.0,
2.17.0, 2.17.1, 2.18.3 (Fedora 44 host, `org.gnome.Platform//43` … `//master`,
`org.freedesktop.Platform//24.08` and `//25.08`, `org.fedoraproject.Platform//f44`),
each with a negative control (`XDG_DATA_DIRS=/usr/share` → fallback) and a cold
`XDG_CACHE_HOME`. So it is not a Fedora patch, not recent and not a regression waiting
to be reverted. The table is in the ADR's § *What was measured*, 5.

**The A/B was re-run and the suite discriminates per leg**, not only as a whole:
reverting `packages/infra/cli/src` and rebuilding fails exactly **7** tests, and two
single-mutation probes each go red alone — dropping only the `ATSApplicationFontsPath`
emission fails 1 test, making `fontGap` unconditional fails 2.

**Refactor:** `stagedFontDir()` replaces three hand-spelled `<data dir>/fonts/<appId>`
literals in the launcher (the leaf now comes off `SHARE.fonts`, so the launcher cannot
drift from `planFonts` while both suites stay green); the win32 comment beside it is
cut to the rule plus a pointer, because `LAYOUTS.windows.fontGap` already carries the
three dead candidates in full. `planFonts` had no unit test at all — its app-id nesting
and its duplicate-basename refusal now have one each.

**One claim was hedged:** `expo-font.ts` stated the macOS activation as fact while the
ADR and `Layout.fontGap` both mark it unverified. Three places disagreeing about one
fact, with the confident one the most read, is the shape this PR exists to pay off.

**Still open, named rather than fixed:** the Windows call has no `@gjsify/*` home. The
concrete one is `@gjsify/adwaita-app` beside `initLocale` — `resolveLocaleDir` already
reads `GJSIFY_LOCALE_DIR` off this same launcher, which is the precedent
`GJSIFY_FONT_DIR` was chosen on. Verifiability is *not* the blocker: measured here (GJS,
Pango 1.57.1, `PangoCairoFcFontMap`) a family absent from `list_families()` is present
after `add_font_file()`. What is open is the API shape, plus the Windows confirmation.
`status/open-todos.md` now says so, and names `@gjsify/react-native`'s `expo-font` as
the second caller.

**Renumbered 0037 → 0038**, because #1476 claims 0037 against the same `origin/main`.

## Checks run locally

`gjsify lint` (0 errors) · `gjsify format --check` · `gjsify workspace @gjsify/cli run
check` + `run test` (3519) · `@gjsify/react-native run check` · `gjsify run build` ·
`node scripts/audit-runtimes.mjs --check --strict` · `check-agent-context-size --check` ·
`check-comment-budget` · `check-doc-fences` · `check-source-control-bytes` · e2e `ship`,
`ship-layout`, `ship-from-stage`, `ship-macos`, `ship-windows`, `ship-signing`,
`ship-flatpak`, `ship-declaration`.

https://claude.ai/code/session_01Fg7cvsEJVF84chQMyj5A9x


JumpLink:
## The Windows half is measured now, so the ADR stops hedging it

This PR shipped the Windows rows as *"primary sources; not run on Windows"*. That was the honest state; it is no longer the state. Both candidate mechanisms were put to a face the host does not have, on a **Windows 11 test VM (build 26200)** — Node v24.18.1, `@gjsify/gtk-runtime-win32-x64@0.45.0`, GTK 4.22.4, GLib 2.88.1, Adw 1.9.1, GStreamer 1.28.4.

**Probe:** `BuilderBlocks.ttf` out of the runtime bundle itself — 500 bytes, valid sfnt, family `BuilderBlocks`, one Mac name record, not installed on the host. **Baseline:** `PangoCairo.FontMap.get_default().list_families()` = **82 families**, all DirectWrite, no `Builder*`.

### W1 — fontconfig does not reach the map that counts, in either direction

| probe | families | `BuilderBlocks` |
|---|---|---|
| baseline | 82 | no |
| own `FONTCONFIG_FILE`, probe dir ADDED beside `WINDOWSFONTDIR` | 82 | no |
| own `FONTCONFIG_FILE`, probe dir EXCLUSIVELY | **82** | no |

The exclusive row is the one that settles it: a config naming *only* the probe directory would leave a fontconfig-backed map with a handful of families. It does not move at all — so the default map is not reading a fontconfig that was configured wrong, it is not reading fontconfig.

### W2 — the control, so W1 is a finding and not a broken probe

An explicitly fontconfig-backed `PangoFT2.FontMap`, same process, same configuration: **76 → 77 with `BuilderBlocks`** additively, and **5** exclusively (`BuilderBlocks | Monospace | Sans | Serif | System-ui`). fontconfig parses the config, walks the directory and sees the face; the map GTK renders through never asks it. The API shape agrees without running anything — the default map exposes no `config_changed` and no `set_config`, the FT2 one exposes `config_changed`.

**W3.** The bundle ships no `fc-cache.exe`, so the cache could not be built by the ordinary route either.

### W4 — `add_font_file()` works, and on GTK's OWN map

| probe | result |
|---|---|
| `PangoCairo.FontMap.get_default().add_font_file(path)` | `true` |
| `list_families()` after | 82 → **83, with `BuilderBlocks`** |
| in a real window: `label.get_pango_context().get_font_map() === PangoCairo.FontMap.get_default()` | **`true`** |
| `add_font_file` on that map, then `ctx.load_font` | `BuilderBlocks 40`, layout **54×54 px** |

### W5 — the discriminator, because a substitution is what "resolved" looks like

An invented family `ZzzNoSuchFamilyQx` resolves to **`Tahoma 40`, 63×66 px, Pango warning** — and `BuilderBlocks` behaved *identically* BEFORE `add_font_file`. Afterwards: `BuilderBlocks`, 54×54, no warning. Different family, different metrics, no warning: the green is not a call that returned `true`.

## What changed here

- **ADR 0038** — the Windows section is now `### Windows, RUN on hardware`, with W1–W5 as tables in the § *What was measured* style, dated and with the environment. The source bullets stay, reframed as the MECHANISM the measurement is the OUTCOME of. Decision 3's table row and Decision 6's windows note move from *primary sources* to *measured*; the "not decided" entry about the inert win32 fontconfig plumbing records that the run it wanted now exists.
- **`Layout.fontGap`** (windows row) says it measured rather than inferred, with the reason: the default font map does not read fontconfig — not even an exclusive configuration — and has no `config_changed`/`set_config` to make it, while `add_font_file` on the same map a `Gtk.Label` renders through does register the family. The `fontGap` doc comment's "reached identically on all three layouts" claim is corrected: the payload is STAGED identically, what READS it is not.
- **`share-dirs.ts`**, **`docs/ship-formats.md`**, **`tests/e2e/ship-layout`** comments follow.
- **`expo-font.ts`** — the Windows sentence is now marked measured, in both directions. **macOS is deliberately untouched**: its activation is still unverified, no leg here runs a `.app`, and nothing in this measurement touches it. The `useFonts` limit in `support-table.ts` gains the same marker (it does not reach `SUPPORT.md`, which renders `reason` only — so the generated doc is unchanged, verified by re-running `generate-exports.mjs`).
- **`status/open-todos.md`** — the `initFonts()` entry no longer says the Windows confirmation "no leg here can produce"; it exists, so what is left there is only the API shape.

**Nothing new is claimed.** No leg here builds a Windows program directory and runs it, and no `@gjsify/*` package makes the call the launcher hands `GJSIFY_FONT_DIR` to — both still written down, in the ADR and in `Layout.fontGap`.

## Gates

`gjsify lint` (0 errors) · `gjsify format --check` · `check-rn-surface` · `check-agent-context-size --check` · `gjsify workspace @gjsify/cli run check` + `run test` (3522 pass) · `gjsify workspace @gjsify/react-native run check` + `run test` (1222 pass) · `node --test tests/e2e/ship-layout/run.mjs` (38 pass).

---

Separately, the same VM turned up a node-gi marshalling defect that has nothing to do with fonts — `GstMessage.parse_error()` refuses every `GError` OUT parameter (`OUT type tag 20`). Filed as #1495, not in this branch.

PR #1530, opened 2026-09-03
fix(e2e): skip create-app in the publish window

### LINKED ISSUE #1523: create-app E2E waits 30 min for a publish that measurably takes 43-70
## What

`create-app E2E` waits 30 minutes for the release's packages to appear on npm. Measured across
the last three releases, they take 43–70 minutes. The budget has never been long enough, so for
roughly an hour after every release every CI run that reaches this test fails — and the failure
text reads like a publishing incident rather than a mis-set deadline.

## The budget

`tests/e2e/helpers.mjs:157`

```js
{ label = 'e2e', deadlineMs = 30 * 60 * 1000, intervalMs = 20 * 1000 } = {},
```

and on expiry (`helpers.mjs:172`):

> `awaitRegistryResolvable: the registry still cannot resolve @gjsify/node-gi@^0.47.0 after 30 min. On a release commit that means the publish did not land; otherwise the range is wrong.`

## The measurement

Time from the release tag being published to `@gjsify/node-gi@<version>` being resolvable on
registry.npmjs.org (`.time[version]` from the packument):

| release | tag published | node-gi resolvable | gap |
|---|---|---|---|
| v0.44.0 | 2026-08-27T14:40:39Z | 2026-08-27T15:51:06Z | **70 min** |
| v0.45.0 | 2026-08-31T03:47:45Z | 2026-08-31T04:31:07Z | **43 min** |
| v0.46.0 | 2026-09-03T08:55:13Z | 2026-09-03T09:41:37Z | **46 min** |

Three for three over budget. `@gjsify/node-gi` is one of the ten platform packages published by
their own payload jobs, which run after the ~199-package serial sweep — so it sets the tail, and
the tail is what this test waits on.

## Observed now

PR #1520, based on `main` at 0.47.0, failed [E2E 4/4 Fedora
44](https://github.com/gjsify/gjsify/actions/runs/33806553308/job/100827264088):

```
✖ create-app E2E (2024945.311249ms)
  Error: awaitRegistryResolvable: the registry still cannot resolve @gjsify/node-gi@^0.47.0 after 30 min.
```

which failed `CI gate (GJS)` and left the PR `BLOCKED`. Nothing is wrong with the PR: the v0.47.0
publish was still running. Any PR based on the release commit hits this for the same ~hour.

## Interaction with #1522

#1522 measures ~9 minutes of duplicate workspace build on the publish job's critical path. That
delay lands entirely inside this window, so the two compound: fixing #1522 shrinks the gap by
roughly 9 minutes — from ~46 to ~37 — which is still above 30. **Fixing #1522 alone does not make
this test pass.** Both need addressing, and this one needs a decision, not just a bigger number.

## Options, none of them obviously right

1. **Raise the deadline** past the measured tail (say 90 min). Simplest, but parks a runner for
   up to 90 minutes on any PR unlucky enough to run during a release, and re-breaks the moment
   the package count grows.
2. **Recognise the state instead of timing out.** When the workspace version is not on the
   registry *and* a release run for that version is in progress, the honest result is "not yet
   publishable", not "failed" — closer to this repo's `it.failing(..., { when })` convention than
   to a timeout.
3. **Resolve against the local build** when the registry lacks the version. Fastest, but it
   discards the property the test exists for: that a real consumer `npm install` works.

Option 2 looks right and option 3 looks wrong, but this changes what the E2E guarantees, so it
wants a deliberate decision rather than a quiet edit.

## Not claimed here

That 43–70 min is the whole distribution — three releases is what the registry timestamps and
the retained run logs cover. The direction is unambiguous (every measured gap exceeds the budget);
the exact ceiling is not.

JumpLink:
## The problem

`create-app E2E` waited 30 minutes for the release's packages to reach npm. Measured from the
registry's own `.time[version]`, the publish takes **43–70 minutes**:

| release | tag published | `@gjsify/node-gi` resolvable | gap |
|---|---|---|---|
| v0.44.0 | 2026-08-27T14:40:39Z | 2026-08-27T15:51:06Z | 70 min |
| v0.45.0 | 2026-08-31T03:47:45Z | 2026-08-31T04:31:07Z | 43 min |
| v0.46.0 | 2026-09-03T08:55:13Z | 2026-09-03T09:41:37Z | 46 min |

A deadline shorter than the thing it waits on does not wait, it fails. So for about an hour
after every release, every CI run reaching this suite went red with text that read like a
publishing incident — which is what left #1520 `BLOCKED` on a healthy PR.

## What changed

**`create-app` now skips, on the vocabulary this repo already had for the state.** The two
Yarn-PnP suites answer the same question with `describe(name, { skip }, fn)` and a printed
reason; `create-app` now does too, sharing the probe and the opening clause of the reason.

Three states, one glance apart in the log:

```
✔ create-app E2E                 passed
﹣ create-app E2E # the workspace version is not on npm yet — 1 registry-bound
                   template dependency/-ies 404: @gjsify/node-gi@0.47.0. …
✖ create-app E2E                 failed
```

### `awaitRegistryResolvable` is deleted, not shortened

It had one caller. A retained short wait would be a second answer to "is this version
published yet", which is the thing not to add. Propagation lag is absorbed where it can
actually bite — `npmInstallWithRetry`, at the install, minutes after any pre-flight probe
would have finished. And what the deadline *also* did, name a publish that never landed,
was never this suite's job: `verify-published-closure.mjs --phase post-release` enumerates
`packages/node-gi/*` and `packages/napi/*` explicitly and reports every `notLive` name at
the end of `release.yml`.

### The ordering constraint, resolved rather than moved

`registryGjsifyRanges` needed the tarball map to know which ranges are really registry-bound,
which forced the question after `createTestEnvironment()` had packed the workspace — too late
for a `{ skip }` that `describe` reads at definition time. But the tarball map is not an
independent source of truth: it is a materialization of `packableWorkspaces()`, since
`pack.mjs` iterates exactly that list and exits non-zero if an entry yields no tarball. So the
same function pack.mjs calls answers it with nothing packed. A new first test re-checks that
equality against the map actually produced, so a drift is a named failure rather than a skip
on the wrong package.

### Fail-closed, three times over

A detector that errs toward "release in progress" converts a permanently-red hour into a
permanently-green test. So:

1. Only a range spelled the way `process-template.mjs` writes one (`^X.Y.Z` / `~X.Y.Z` / exact)
   is a candidate — no hand-rolled semver, and any other spelling can never disarm the suite.
2. Only a version the registry **answered about** and whose **package it serves** counts as
   missing. `unpublishedRegistryDependencies` used to read a bare 404 as "not published yet";
   a registry that 404s everything (proxy down, auth failure, mistyped `GJSIFY_E2E_REGISTRY`)
   therefore disarmed **both PnP suites** silently. It now needs the packument as confirmation.
3. Only **this release train's** version can skip. `^99.0.0` unresolvable is a wrong range, not
   a window, and still fails the suite. The train version is the root manifest's — the same
   anchor `verify-published-closure.mjs` probes every candidate at.

## How both branches were proven

A stand-in registry (written for this, because `tests/e2e/mock-registry.mjs` serves only the
packument route and npm also asks `/<name>/<version>`), five states, `GJSIFY_E2E_REGISTRY`
pointed at it, running the real suite files:

| suite | registry state | create-app E2E |
|---|---|---|
| create-app | version 404, packument 200 (**the window**) | **skipped, reason printed** |
| create-app | version 200 (published) | ran → red |
| create-app | 404 on everything | ran → red |
| create-app | 503 | ran → red |
| create-app | nothing listening | ran → red |
| create-app | `^99.0.0` off the train | ran → red, **not even probed** |
| cli-only-pnp / pnp-zip-static-reads | the window | skipped, wording unchanged, exit 0 |
| cli-only-pnp / pnp-zip-static-reads | 404 on everything | ran → red (**was**: skipped) |

Old vs new probe, same registry, A/B:

```
all404   (registry 404s everything)  old ⇒ SKIP   new ⇒ run
window   (version 404, packument 200) old ⇒ SKIP   new ⇒ SKIP
published(version 200)                old ⇒ run    new ⇒ run
```

Exactly one cell moves, and it moves towards running.

`oxfmt --check` and `oxlint` over the changed files: both exit 0.

Closes #1523

https://claude.ai/code/session_01TGTu3xiJDXtxVdB7vkzcyC


JumpLink:
---

## Added after review of my own proof: `tests/e2e/release-window-skip/`

The proof above lived in a scratch harness that dies with the session — nothing in the repo
would have caught a regression in the detector, and a detector that can disarm a suite is
exactly the thing that must not be trusted untested. So the five registry states are now a
suite: a local server, `createAppRegistryGapSkipReason` and `unpublishedRegistryDependencies`
driven against it, 8 tests, under a second, no build required.

Each of the three fail-closed guards is verified by **removing it**:

| mutation | result |
|---|---|
| read a bare 404 as "not published yet" (the pre-#1523 probe) | 2 tests red |
| drop the release-train check | the off-train `^99.0.0` case red |
| replace the pinned-floor reader with a lax "strip non-digits" parse | the unrecognised-spelling case red |

`>=<train>` is the input for that last case on purpose: it *contains* the train version, so a
lax reading recovers it. An obviously-foreign spelling would have been rejected by the
train-version guard instead, and the test would have agreed for the wrong reason.

JumpLink:
Follow-up: `check-e2e-harness-duplication` correctly rejected the new suite for standing up its own `node:http` registry. Reworked onto `startMockRegistry`, with npm's single-version route `/<name>/<version>` — which the shared harness does not model and which is exactly what this detector turns on — added through the documented `onRequest` seam. All three mutation results re-measured on the shared harness and unchanged (2 red / 1 red / 1 red).

PR #1537, opened 2026-09-04
fix(react-native): see through child wrappers

### LINKED ISSUE #1451: An absolutely positioned child wrapped in a Fragment never makes its parent an overlay
`View` becomes a `Gtk.Overlay` instead of a `Gtk.Box` when one of its children is absolutely positioned — the element cannot answer for itself, so the parent has to. The check reads each child element's own props.

**A `<>…</>` breaks it.** A Fragment answers for itself, not for what it contains, so a parent whose absolutely positioned child is wrapped in one never becomes an overlay, and the child's own resolution then fails:

```
<View> absolute — positions this element on top of its parent,
so the PARENT has to be a Gtk.Overlay — and it is not.
```

Measured on a card component whose `overlay` prop is a Fragment (`overlay={<>…</>}`): every video thumbnail on two of the application's five tabs.

## Why this is worth more than a special case

The error is *correct* and its advice is unactionable: the author did not write a plain `View` around an absolute child, they wrote a Fragment, and there is no prop on a Fragment to fix. Wrapping it in a real `View` changes the widget tree to work around a layer defect.

Two ways out, and the choice is a design question rather than a bug fix:

- **Resolve children through Fragments** when gathering the facts a parent needs — a Fragment is transparent to layout in React, and this layer's parent-context carrier is the place that can make it transparent here too.
- **Refuse a Fragment in that position by name**, saying that an absolutely positioned child must be reachable as an element. Worse for the author, but honest, and better than an error that names the parent for something the child's wrapper did.

The first is what React means by a Fragment. The second is at least actionable. What is there now is neither.

JumpLink:
A parent decides whether it becomes a `Gtk.Overlay` from its children's **descriptors** — React renders the parent before the children exist — and two shapes an application writes every day came back unreadable. Both were answered as *"not absolutely positioned"*, which is a perfectly plausible answer that nothing anywhere reports.

## The two holes, and they are independent

**1 — a Fragment answers for itself.** Its props are `{ children }`: no `style`, no `className`. And `Children.toArray` does **not** descend into one (measured, react 19: a Fragment survives as an element whose type is `Symbol(react.fragment)`). So `<>…</>` around an absolutely positioned child hid it, the parent stayed a `Gtk.Box`, and the child then got L2's refusal — which names the **parent** for something the child's wrapper did, with no prop on a Fragment to fix. That is #1451 as reported.

**2 — an `Animated.Value` in a style is exactly what L2 refuses on a plain element.** So a parent reading an `<Animated.View className="absolute" style={{ opacity }}>`'s raw props **threw**, and the swallowing `catch` in `isAbsoluteChild` turned the throw into `false`. This needs no Fragment at all: the matrix row *"an Animated.View with an absolute className, authored directly"* fails on the pre-fix reader. It is what the CORRECTIV desktop app measured — swapping a plain `<View className="absolute …">` for the phone's `<Animated.View>` was the only change needed to lose the overlay, and it is why two finished features (`Animated` since 0.46, `absolute`) did not compose.

The same two blindnesses also hid `text` (a text run inside a Fragment, which competes with a prop for a widget's text sink) and `count` — which nothing in L2 reads yet, and which `justify-between`'s refusal names as the fact it does not have, so it was wrong before a reader could exist. Neither had been reported.

## What changed

- **`src/child-facts.ts`** — new, and it owns the parent's whole question, asked once for the count that decides the overlay and for the placement that fills its `add_overlay` slot. Fragments are expanded away with keys **composed** behind their Fragment's (`.$group` + `:` + `.$badge`), so no sibling is re-keyed and nothing remounts.
- **`splitAnimatedStyle` moved into L2** (`src/primitives/style.ts`), so `Animated.View` and its parent read *one* copy of the split. Two copies were the defect. Dropping the animated half in the parent's read hides nothing: the element itself is still refused by name if it is not an `Animated.View`.
- The `catch` stays, scoped to what it was always for: a **foreign** composite child (`<MyCard className="card-lg">`) may carry any vocabulary, and a parent must not throw for one it cannot route. What is no longer allowed is a shape *this package can render* coming back unreadable.

The issue offered two ways out. This is the first one — resolve through Fragments — because it is what React means by a Fragment, and because the Solid binding already behaved that way: its children stamp themselves and `resolveChildren` flattens arrays, so the class was invisible there. The two frameworks genuinely produced different trees for one authored tree.

## The mechanism, proven red before green

`src/child-facts.spec.ts` does not test the two shapes. It enumerates **every fact `childFacts` produces × every wrapper a child can arrive in** (7 wrappers: bare, Fragment, keyed Fragment, nested Fragments, Fragment with a sibling, array, Fragment in an array), and the **fact list is derived from the record itself** — so a fourth field on `ChildFacts` has no row until someone adds one, and the coverage case says so. Each flag fact carries a **negative** case through the same wrappers; without one, a reader that always answered yes would pass the entire positive matrix.

Measured on gjs 1.88.1 / GTK 4.22.4, same specs, only the reader swapped:

| | result |
|---|---|
| pre-fix reader (`Children.toArray`, raw props) | **42 of 2242 failed**, exit 1 |
| with the fix | **2257 of 2257**, exit 0, zero GTK diagnostics |

The 42 include the two new real-tree mount vectors in `primitives/widgets.spec.ts` (which read the actual `Gtk.Overlay`, its main child and its overlay slot, not a plan) and the new React/Solid parity vector in `solid/solid.spec.ts` — a transparent wrapper, spelled as a Fragment in React and as an array in Solid, which used to build two different widget trees.

Also verified: `gjsify tsc --noEmit` clean, repo-wide `oxlint` exit 0, `oxfmt --check` clean, and `check-rn-surface` / `check-vocabulary-alignment` / `check-agent-context-size` / `check-rn-icon-targets` all exit 0.

## One GTK measurement that fell out

`Gtk.Widget:opacity` is a `gdouble` in the ParamSpec and **8-bit internally**: an animated `0.25` reads back as `64/255`. The vector asserts the quantised value, which is also the discriminator — a misspelled property name becomes a silent JS expando and would read back exactly `0.25`.

## Deliberately not done

- **No new ADR.** The design this restores is ADR 0032 § 6's; nothing about the layer boundary changed.
- **`packages/framework/AGENTS.md` untouched.** It sits at 32752 of the 32768-byte cap (16 bytes of headroom), and its budget-ledger line is a counter two concurrent PRs would collide on. The rule and its incident are in the package README and in `child-facts.ts`' header instead.
- **`docs/release-notes/next.md` untouched**, per its own header.
- **No Solid-side change.** Solid was already transparent here; the parity vector is what says so rather than an assertion.
- **The node leg (`test:gjs-on-node`) could not run locally** — `@gjsify/node-gi` is not installed in this worktree. In CI that leg is a `continue-on-error` probe in `gtk-os-suites.yml`; the gating gjs leg is what is measured above.

Closes #1451

---

## Review follow-ups (2 commits on top)

**`fix(react-native): make composed keys injective`.** Concatenating the two halves is not
injective, and the collision is ordinary authoring: `<Fragment key="row">` around an unkeyed
child produced `.$row.0`, and so did a plain sibling authored `key="row.0"` — the composite key
a `.map()` over grouped data writes. MEASURED (react 19): React answers the duplicate by
matching those two children POSITIONALLY — on a reorder they swap their content and keep each
other's widgets — so a cursor, a scroll position or an `Animated.Value` binding follows the
wrong element, and React's duplicate-key warning goes to a console GJS does not have. A colon
separates the halves and is the one character that can: `Children.toArray` escapes `:` in an
authored key to `=2` (and `=` to `=0`), gives an unkeyed child `.<index base 36>`, and touches
neither `.` nor `/`.

Two spec rows came with it, both red first: the collisions above, and a strengthened
foreign-composite case — the old assertion (`absolute` is 0) stayed green whether or not the
`catch` had a throw path, so it now asserts the read really THROWS, plus the pair that makes
swallowing safe (the parent's read is a SUBSET of the child's own resolution, so every throw
it eats is raised again naming the child's own token).

**`test(react-native): read identities, not key spelling`.** "Nothing remounts" was held by
asserting key STRINGS. The real tree says it now: render, `flushSync` a second render, read the
labels off the SAME GObject references — a rebuild leaves them holding the old widgets.
Verified red by composing an unstable key.

2269 of 2269 on gjs 1.88.1 / GTK 4.22.4, exit 0; `oxfmt --check`, `oxlint`, `gjsify tsc
--noEmit`, `check-rn-surface`, `check-vocabulary-alignment`, `check-agent-context-size`,
`check-rn-icon-targets` and `check-comment-budget` all clean.

Two findings left OUT of this PR, for their own issues: a foreign composite that *renders* an
absolute child still produces #1451's unactionable refusal (it names the parent, and every
branch of its explanation is wrong for that tree), and `<Pressable style={({pressed}) => …}>`
— React Native's documented API — renders a `GtkButton` with the style silently dropped.

### DIFF
--- a/packages/framework/react-native/README.md
+++ b/packages/framework/react-native/README.md
@@ -97,6 +97,34 @@ the table rather than only here.
 L2 is exported (`import { primitives } from '@gjsify/react-native'`) so any binding
 can render the same vocabulary without going through the React components.
 
+### What a parent asks about its children, and what stays transparent
+
+A few decisions cannot be taken by the element that authored them. `position: absolute`
+is the clearest: it positions the element on top of its parent, so the **parent** has
+to be a `Gtk.Overlay` — and a `View` becomes one as soon as one of its children asks.
+The same shape decides whether a text child competes with a prop for a widget's text
+sink, and `justify-between` is refused on a child count.
+
+React's parent renders before its children exist, so it reads their **descriptors**.
+Three things are therefore transparent to that read, and each is transparent because
+an application writes it without meaning to change anything:
+
+- **a Fragment.** `<>…</>` is transparent to layout in React, so it is transparent
+  here: it is expanded away before anything is counted, its children keeping their own
+  keys composed behind its own. A card whose `overlay={<>…</>}` holds the absolutely
+  positioned child gets the overlay it asked for.
+- **an array**, which is the same statement in a different spelling.
+- **an `<Animated.View>`.** Its style carries `Animated.Value`s, which is exactly what
+  the style partition refuses on a plain element — so a parent reads it through the
+  same split the component itself uses. An animated fade and an `absolute` compose.
+
+What is NOT transparent is a foreign composite component: `<MyCard className="…">` may
+carry any vocabulary at all, and a parent must not throw for one it cannot route, so it
+answers "no" and the child — if it ever reaches L2 — is refused by name. That boundary
+is enumerated in `src/child-facts.spec.ts` as every fact a parent reads × every wrapper
+it can arrive in, with the fact list derived from the record itself: a new fact has no
+row until someone adds one, and the suite says so.
+
 ### The split is measured, not asserted
 
 `@gjsify/react-native/solid` exports the same seven primitives as SolidJS components
--- a/packages/framework/react-native/src/child-facts.spec.ts
+++ b/packages/framework/react-native/src/child-facts.spec.ts
@@ -32,11 +32,11 @@ import { describe, expect, it } from '@gjsify/unit';
 import { MINIMAL_TOKENS, type StyleTokens } from '@gjsify/gtk-host/style';
 import { createElement, Fragment, type ReactNode } from 'react';
 
-import { childFacts, childNodes } from './child-facts.js';
+import { childFacts, childNodes, readableProps } from './child-facts.js';
 import { AnimatedValue } from './animated/value.js';
 // `AnimatedView` is the exported name; `Animated.View` is what an application writes.
 import { AnimatedView, Text, View } from './components.js';
-import type { ChildFacts } from './primitives/resolve.js';
+import { declaresAbsolute, type ChildFacts } from './primitives/resolve.js';
 
 const TOKENS: StyleTokens = { ...MINIMAL_TOKENS, spacing: { ...MINIMAL_TOKENS.spacing, '2': '8px' } };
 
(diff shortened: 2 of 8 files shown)


JumpLink:
**Korrektur einer Zahl in meinem Merge-Text, und der Grund dahinter ist nützlicher als die Zahl.**

Der Merge-Body nennt für `@gjsify/react-native` `2242` / `2257` Assertions. Nachgemessen auf dem gemergten Stand `6ba60c4c74` selbst, mit `gjsify workspace @gjsify/react-native run test`: **2329 completed, 0 skipped, exit 0**. Die 2257 stammen aus einem anderen Aufruf oder Scope und reproduzieren nicht. Das `test`-Skript des Pakets ist GJS-only (`build:gjsify && build:test && test:gjs`), eine über einen anderen Pfad erzeugte Zahl ist also nicht vergleichbar.

**Warum das nachgemessen wurde:** nach #1541 fiel dieselbe Zahl auf 2215, und über einen PR, der ~500 Zeilen Spezifikation hinzufügt, sieht ein Rückgang wie eine verlorene Suite aus. Meine Vermutung war, der dort reparierte `beforeEach`-Slot habe Blöcke wieder scharf gemacht, die nun mangels Vorbedingung **überspringen**. Widerlegt, direkt beobachtet statt geschlossen:

| | `6ba60c4c74` (nach #1537) | `54f7f72028` (nach #1541) |
|---|---|---|
| completed | **2329** | **2215** |
| `(skipped`-Zeilen | **0** | **0** |
| ausgeführte Testnamen | 550 | 575 — 0 verschwunden, 25 neu |
| `expect(`-Aufrufstellen | — | +58, keine Datei verliert welche |

**Die Zahl zählt ausgeführte ASSERTIONS, nicht Tests**, und `primitives.spec.ts` assertiert in vierfach geschachtelten datengetriebenen Schleifen über `PRIMITIVES` × Specs × Props × Routes. #1541 hat `primitives/table.ts` umstrukturiert (252 → 257 Einträge), womit sich die innere Iterationszahl unabhängig von Tabellengröße und geschriebenen Assertions bewegt. `PROPS.md` bleibt bei 382 Zeilen, die generierte Fläche ist also nicht geschrumpft.

Also: **formabhängig, nicht deckungsabhängig — in keiner Richtung über Commits vergleichbar.** Ein Anstieg kann eine breitere Tabelle heißen, ein Rückgang eine aufgeräumtere, und beides sagt nichts darüber, was geprüft wird. Wer künftig eine solche Zahl in einen Commit-Text schreibt (ich eingeschlossen): sie belegt einen Lauf, keinen Vergleich.

PR #1563, opened 2026-09-04
feat(gtk-host): a portal seam, and React Native's Modal on it

JumpLink:
`@gjsify/gtk-host`'s element model has one placement question: *how does this parent adopt a child?* `ChildPolicy` answers it in eight kinds, and every one of them ends in a call on the PARENT. That presumes an answer to a question nobody had to ask, because for every widget in the table the answer was yes: **does this node go into its parent at all?**

For an `Adw.Dialog` it does not, and that is why `<Modal>` was `planned` and why the expo-router `dismiss` family still throws *"needs a modal stack this layer has no portal seam for"*.

## The defect, and the precondition that hides it

Measured here on gjs 1.88.1 / GTK 4.22.4 / libadwaita 1.9.3, one process per case, exit codes recorded:

| case | result |
|---|---|
| `box.append(label)`, box in an `Adw.Window` — control | exit 0 |
| **`box.append(dialog)`, box in an `Adw.Window`** | `Adwaita-ERROR **: Trying to add AdwDialog … to GtkBox …` — **exit 134 (SIGABRT), core dump** |
| `box.append(dialog)`, box **detached** | **exit 0, silent**, `dialog.get_parent()` is the box |

The abort is `g_error()` from `adw_dialog_root()` — the ROOT vfunc — so it fires only once the widget joins a tree whose root is a `GtkWindow`. A re-test on a bare box "disproves" it at exit 0, which is why every vector in `portal.spec.ts` roots its parent in a real `Adw.Window` and why the fixture is the only one in the file.

`g_error()` is not an exception: a `try` in the reconciler does not help and a diagnostics gate cannot count it. The host's usual answer — a named `GtkHostError` — was not available either, because the host would have to know it is holding a dialog.

## The seam

`WidgetDescriptor` grows a second, orthogonal declaration. `children` says how a parent adopts a child; `placement` says whether it adopts one at all.

```ts
{ gtype: 'AdwDialog', ctor: () => Adw.Dialog,
  children: { kind: 'single', set: 'set_child' },
  placement: { kind: 'portal', present: 'present', close: 'force_close' } }
```

**It is framework-agnostic by construction** (ADR 0027): the whole seam is three functions in `src/policies.ts`, below all three adapters. `adapters/{react,solid,vue}.ts` are byte-unchanged and none of them mentions a dialog.

Four more measurements decide the rest of it, and each was a silent wrong before it was a design:

- **the close is `force_close`, not `close`.** With `can-close: false` — which is how `onRequestClose` is honoured — `close()` returns **false**, emits `close-attempt` and **leaves the dialog on screen**. And `close()` on a node that was never presented is `Adwaita-CRITICAL **: Trying to close … that's not presented` where `force_close()` is **silent**, so the host also needs no "is it up?" probe — the one it could not write without knowing what a dialog is.
- **the present WAITS on the parent's `notify::root`.** Presenting against a parent that is in no window opens a **separate `GtkWindow`**, at exit 0, with nothing logged: `adw_dialog_present` finds no `AdwDialogHost` among the ancestors and takes its `present_as_window` branch. That is the ORDINARY case at insert time — every framework builds bottom-up. `notify::root` fires on a GRANDCHILD box when the toplevel takes the subtree, and again on unroot.
- **the subscription is kept, not one-shot.** Unrooting the parent leaves an already-presented dialog in the OLD window's host (`w1.visibleDialog` still set after `w1.set_content(null)`), and a bare re-present is `Adwaita-CRITICAL **: Cannot present … as it's already presented for …` plus a `Gtk-WARNING` **with the move not happening**. Close-then-present is the sequence that works.
- **a portal counts for nothing in the parent's child list** — the sibling index, the tail rotation, both slot readers and `addressesOf`. Counting one shifts every later child by one, and rotating one calls the parent's adder on it, which is the abort above.

`attached` is written by the *present*, not by the insert: a deferred portal is claimed by the host and not yet taken by GTK, which is exactly the distinction that flag exists to keep.

**Membership is measured, and a mechanism keeps it complete.** Registration is exact — `lookupWidget('AdwAlertDialog')` answers the GENERATED row, not `AdwDialog`'s — so an inherited placement would never be looked up and `<adw-alert-dialog>` under a rooted box would abort the process. All five rows (`AdwDialog`, `AdwAboutDialog`, `AdwAlertDialog`, `AdwPreferencesDialog`, `AdwShortcutsDialog`; `AdwMessageDialog` is a `GtkWindow` and is not one) are named, and a spec walks `GObject.type_is_a(…, Adw.Dialog)` over the registered table, so libadwaita's sixth dialog subclass fails the suite instead of aborting somebody's application. `descriptorProblems()` checks the declared methods exist and that `present` takes **one** argument — the arity is what separates a portal from a toplevel (`adw_dialog_present` 1, `gtk_window_present` 0).

## `<Modal>`, on it

`partial` rather than `planned`, and an ordinary primitive row: `AdwDialog` outer, an implicit `GtkBox` content node (an `Adw.Dialog` holds ONE child, so two children under a `<Modal>` would be an assignment that silently evicts the first), `can-close: false`, `onRequestClose` → `close-attempt`, `onShow` → `map`.

`can-close: false` is **unconditional**, and it is what makes the two models agree: React Native owns a modal's visibility through `visible` and nothing else, so a dialog that closed itself on Escape would leave the element mounted with `visible` still true and nothing on screen. `onShow` is `Gtk.Widget::map` and not the present call — measured, `present()` against a window that has not been shown yet emits nothing and the emission arrives on the window's own `present()`, which is React Native's *"called once the modal has been shown"* exactly.

Refused by name, each with what to do instead: `onDismiss` (it would never fire — the host disconnects a node's handlers before it retracts the node), `animationType`, `transparent`, `backdropColor`, `presentationStyle`, `allowSwipeDismissal`, `supportedOrientations`, `onOrientationChange`, `modalRef`.

## One check was asking a proxy question, and CI found it

`check-generated-website-data.mjs` decides whether a gallery block's `uncurated-placement` refusal has gone stale by asking whether the gtype appears in `descriptors/*.ts`. That was a proxy for *can a child go inside this widget*, and it held only while every curated row was curated for its CHILDREN. The four dialog subclasses are curated for their PLACEMENT and keep `children: { kind: 'uncurated' }`, so the check called a live refusal stale and turned `Detect runtime-triplet drift` red.

It asks the child policy now. Run as a discriminator both ways: giving `AdwPreferencesDialog` a real `ordered` policy brings the failure straight back, so the narrower predicate is not a dead one.

## Documented for the people who will use it

- **`website/.../frameworks/index.mdx`** — *"A dialog is presented, not adopted"*, beside the child-placement table it is the exception to. The audience there is someone building an app, so it says what they can write, that the dialog takes no place in the box, that rendering opens it and unrendering closes it, and that it waits for a window — not the incident.
- **`website/.../frameworks/react-native.mdx`** — `Modal` in the renders table plus a worked `<Modal visible onRequestClose>` example, why `onRequestClose` is required in practice, and every refused prop with what to use instead. Kept inside its own section, clear of the routing half.
- `react.mdx`, `solid.mdx` and `vue.mdx` are unchanged on purpose: none of them carries the placement contract — that lives once on the frameworks index, and the seam is the host's, so one page is the right number.
- `SUPPORT.md` and `PROPS.md` are regenerated from the table (`gjsify workspace @gjsify/react-native run generate`), and `check-rn-surface.mjs` holds them.

**And "below the adapters" is now measured rather than reasoned, on ALL THREE.** It is a claim about a call graph, and a call graph reads true when it is not — so an `<adw-dialog>` is mounted through React's reconciler, through Solid's compile-time renderer and through Vue's `RendererOptions`, and all three land in the same window with the same empty parent box, all three closing on unmount. Two measured and a third assumed would have been the asymmetry that later reads as coverage.

## Measured here

- `gjsify run test --include @gjsify/gtk-host --include @gjsify/react-native` → **exit 0**; gtk-host 2 649 assertions, react-native 2 305, every vector under `installDiagnosticsGate()` — and that gate is measured live rather than assumed: injecting one `Adwaita-CRITICAL` into a portal case turns it red at exit 1
- `node scripts/check-generated-website-data.mjs` — the arm above, plus the discriminator
- **3 adapter vectors** (React + Solid + Vue) on top of the host ones
- **21 new host vectors** in `packages/framework/gtk-host/src/portal.spec.ts` and **5 new render vectors** in `react-native`'s `widgets.spec.ts`, the latter through a `mountedInWindow` fixture — the existing `mounted` helper mounts into a BARE box, which for a portal measures neither half
- the abort itself is deliberately not triggered in-process (it would take the runner with it); what is asserted is that the parent's child list stays empty while `window.visibleDialog` is set
- `node scripts/audit-runtimes.mjs --check`, `check-adr-index`, `check-agent-context-size --check`, `check-rn-surface`, `check-vocabulary-alignment`, `check-type-surfaces`, `check-doc-fences`, `gjsify lint` (0 errors), `gjsify format`
- **`test:gjs-on-node` was NOT measured**: it fails to build in this tree on `@girs/gtk-4.0/vocabulary`'s named exports being emptied by the node target, reached from `generator.spec.ts` — a file this branch does not touch, and the import predates it.

## Two things named rather than assumed

- **`packages/framework/AGENTS.md` was 50 bytes under the 32 KiB `project_doc_max_bytes` cap**, so rule (10) could not be added without moving something out. The darwin GL 4.1 paragraph went to `packages/framework/webgl/README.md` — the file it already cited for the construct-by-construct measurement — leaving the rule plus one link, which is what the check's own message prescribes. `status/agent-context-budget.json` is updated with it.
- **Two specs used `Modal` as a literal fixture for "a planned name"** and would have failed for the right reason. Both now DERIVE the fixture from the support table, because that literal had already gone stale once before (`Animated`), and each asserts the fixture was found so an empty table cannot pass it vacuously.

## Three hunks in `support-table.ts`, not one

PR #1561 moves this file's tracked surface to react-native 0.87.1, so the diff is kept as narrow as the truth allows: the `Modal` entry, one line of the `Alert` entry (its reason said *"buildable where Modal is not"*, which ships in the generated `SUPPORT.md` and stopped being true), and one line of an `isImportable` docblock that used `Modal` as its example of a planned name. Nothing near the version/snapshot half, and none of the three keys #1561 drops.

## Left open, deliberately

- **The expo-router `dismiss` family is unblocked and not implemented.** Its refusal names this seam as what it was waiting for and a modal stack is now expressible; what `dismissTo` means over an `Adw.NavigationView` is a routing decision, and the router is not this PR's.
- **A `Gtk.Window` appended to a rooted `Gtk.Box` is accepted silently** (measured: exit 0, `win.get_parent()` is the box). It is the quieter half of the same class and is **not** fixed by this placement: `gtk_window_present()` takes no argument, so a window is not presented *against* anything. Filed as a `status/open-todos.md` entry with the measurement, for whoever owns window chrome.
- **`Modal` has no Solid L3**, like `Image`, `Button` and the Touchable family before it. The seam is below the adapters, so a Solid `<Modal>` is a component wrapper and not a second implementation.

## Review round (independent reviewer)

Three commits on top, after the load-bearing measurements were reproduced from scratch on gjs 1.88.1 / GTK 4.22.4 / libadwaita 1.9.3:

- **`fix(gtk-host): close a portal that loses its window`** — `placeAgainst` returned false when the anchor had no toplevel, which is right for a deferred insert and wrong for an unroot: GTK does not take the dialog down by itself. Measured, after `set_content(null)` the dialog is still presented (`visibleDialog` set, `get_root()` unchanged) while the host had written `attached = false` — the conflation § 4 exists to prevent. Only a re-root repaired it, and a merely detached subtree never re-roots. The rule is symmetric now; new vector red without the fix, green with it.
- **`docs(webgl): drop a win32 claim this file disproves`** — the darwin GL 4.1 paragraph moved into the one file whose § Platform coverage had already measured half of it false (`win32` **is** rewritten, because GDK asks for a core profile). Kept what was only there — the `nm -u`/`dlsym` reproduction and the GLES 3.0 draw-time gaps — and deferred the rest. The `AGENTS.md` correction is **byte-neutral at 32692**, so no budget line moves.
- **`test(gtk-host): measure the portal on Vue too`** — the third adapter, closing the one gap where this PR argued instead of measuring.

Both discriminators re-run independently: reverting `check-generated-website-data.mjs`'s predicate reproduces the original CI failure, and a real `ordered` policy on `AdwPreferencesDialog` brings it straight back. `test:gjs-on-node`'s failure is confirmed pre-existing — every file in that import chain is byte-identical to `main`.


JumpLink:
Independent review. I reproduced the load-bearing measurements myself before reading the code, ran both discriminators, and pushed two fixes.

## Reproduced, on this machine

gjs 1.88.1 / GTK 4.22.4 / libadwaita 1.9.3, native Wayland (no xvfb needed), one gjs process per case:

| case | result |
|---|---|
| `box.append(label)`, box in an `Adw.Window` | exit 0 |
| `box.append(dialog)`, box in an `Adw.Window` | `Adwaita-ERROR **: Trying to add AdwDialog … to GtkBox …` — **exit 134, core dump written** |
| `box.append(dialog)`, box detached | **exit 0, silent**, parent is the box |

The trap row is real: the detached case does look like a disproof. Also reproduced: `close()` under `can-close: false` returns false, emits `close-attempt`, leaves `visibleDialog` set, while `force_close()` clears it and emits `closed`; `close()` on an unpresented node is `Adwaita-CRITICAL` where `force_close()` is silent; `present()` against an unrooted parent puts the dialog's root and parent on a **separate `GtkWindow`** at exit 0; `notify::root` fires on a grandchild box on root and again on unroot; the bare re-present is `Adwaita-CRITICAL` + `Gtk-WARNING` **with `get_root()` still the first window**; `adw_dialog_present` arity 1 vs `gtk_window_present` 0; two `set_child` calls silently unparent the first child; and case K — a `Gtk.Window` into a rooted `Gtk.Box` — is exit 0, parent `GtkBox`. Every table row I checked holds.

## Fixed: a portal that loses its window stays on screen

`placeAgainst` returned `false` when the anchor had no toplevel. Correct for a deferred insert; wrong for an unroot, because GTK does not take the dialog down by itself. Measured through the host: after `window.set_content(null)` the dialog is **still presented** — `visibleDialog` is the dialog, `get_root()` is still that window — while the host had just written `attached = false`.

So the sheet kept showing in a window its own subtree had left, and `attached` said the opposite of what GTK held — the exact conflation ADR 0045 § 4 exists to prevent. Only a re-root repaired it, and a merely detached subtree never re-roots. The seam's rule was stated in one direction and enforced in one direction; it is symmetric now, and the retract is unconditional for the reason `retractPortal` is. New vector is red without the added line and green with it; nothing else moves.

## Fixed: the moved darwin paragraph contradicts its destination

Rule (10) needed room, so the darwin GL 4.1 paragraph went to `webgl/README.md`. That file's § Platform coverage already says, measured on the win11-gjsify VM: *"The same rewrite applies on Windows, which is not what this file used to predict … GDK asks for a CORE profile."* The paragraph appended ten lines below asserted *"Mesa's 4.6 (Compatibility Profile) on win32 has it and is NOT rewritten"*, and the condensed AGENTS.md rule repeated it. One file, two answers, the disproven one under its own heading. Its other two thirds restated the section above less precisely, and its "construct-by-construct measurement: `packages/framework/webgl/README.md`" pointer had become a self-reference.

Kept what was only there — the `nm -u`/`dlsym` reproduction for 217-of-219, and the GLES 3.0 spellings that fail at **draw** time rather than compile time. The `#the-darwin-gl-41-ceiling` anchor is unchanged. **The AGENTS.md correction is byte-neutral: 32692 exactly, so no budget line moves.**

## Verified, not taken on trust

- **The diagnostics gate is live for these cases.** All 21 `portal.spec.ts` cases sit inside `gated()`, which registers the hooks inside the `describe` — the #1560 defect is already fixed on `main`. Discriminator: injecting one `Adwaita-CRITICAL` into a portal case turned it red — *"GTK reported 1 diagnostic(s) that would have passed at exit 0"*, exit 1.
- **The `check-generated-website-data.mjs` discriminator works both ways.** Reverting the one-line predicate reproduces the exact CI failure (`"Adw.PreferencesDialog" is refused as uncurated-placement, but … HAS a curated descriptor now`); giving `AdwPreferencesDialog` a real `ordered` policy with the fix in place brings the same failure straight back. The narrower predicate is not a dead one.
- **Framework-agnostic is measured, not reasoned** — `<adw-dialog>` mounts through the React reconciler and Solid's compile-time renderer, the three adapters are byte-unchanged, and the whole seam is in `policies.ts`. **Vue is not covered by a portal vector**, only by the shared code path; that is a fair reading of "below the adapters" but it is inference for Vue, not measurement. The missing Solid L3 `Modal` genuinely follows `Image`/`Button`/Touchable*.
- **Scope claims hold exactly.** `support-table.ts` is 3 hunks (the `Modal` entry, one line of `Alert`'s reason, one line of the `isImportable` docblock), nothing near the version/snapshot half; exactly one router file, `router/navigation.ts`, 4/3.
- **`test:gjs-on-node` attribution is honest.** It fails on `MISSING_EXPORT` for `DECLS`/`ENUM_NICKS`/`OWN_PROPS`/`OWN_SIGNALS`/`PROVENANCE`/`SINCE` out of the node target's emptied `@girs/gtk-4.0/vocabulary`. Every file in that import chain is **byte-identical to `main`**, and the generator predates the branch (#1449).
- **User docs read as user docs.** No incident, no gate, no measurement on either page. Every className in the new snippets (`flex-1 p-m`, `p-l gap-m`, `text-lg`) matches the pages' own `xs/s/m/l/xl` convention and `MINIMAL_TOKENS.fontSize.lg`; all four imported names are `partial`; `check-doc-fences` gates the className fences and passes.

## Left alone, deliberately

The three ledgered omissions each survive the root-cause rule: the `dismiss` family is a **throwing** refusal that names the undecided routing question (and #1567 owns those files); case K is ledgered with its measurement and cannot be fixed by this placement — `gtk_window_present()` takes no argument, which `descriptorProblems()` now refuses as a portal, and there is a vector for that arity; `test:gjs-on-node` is pre-existing, as above.

One nit not worth a commit: the body says gtk-host 2 629 assertions, I measure 2 637 before my change and 2 645 after.

## Verification

`gjsify --version` 0.42.0, shim resolving entirely inside this worktree. `gjsify run test --include @gjsify/gtk-host --include @gjsify/react-native` → **exit 0** (gtk-host 2 645, react-native 2 305). `audit-runtimes --check`, `check-adr-index`, `check-agent-context-size --check`, `check-rn-surface`, `check-generated-website-data`, `check-doc-fences`, `check-vocabulary-alignment`, `gjsify lint` (0 errors), `gjsify format --check` — all exit 0.

JumpLink:
Follow-up to my review above, which left one claim of mine standing that is now false.

## Vue is measured, not inferred

I wrote that Vue was covered only by the shared code path. That asymmetry is closed — two adapters proving a claim about a call graph and a third assumed is exactly what later reads as coverage, so `portal.spec.ts` now mounts `<adw-dialog>` through **all three**:

| adapter | shape | result |
|---|---|---|
| React | reconciler | presents into the window, parent's child list empty, closes on unmount |
| Solid | compile-time renderer, no VDOM | same |
| **Vue** | `createRenderer`'s `RendererOptions` | **same** |

Vue mounts a portal fine, with the same kebab spelling the other two use — so the registry lookup being measured is the same one, not a second path that happens to work. There was a real chance this would come back a refusal instead; it did not.

The vector binds rather than passing vacuously: asserting a wrong label turns it red. Its *absence* would not fail the vector but abort the runner, which is why the file's header says the abort is deliberately never triggered in-process.

## The assertion count is corrected, and it was drifting

The body said **2 629**. The author measured 2 637, I measured 2 645 after my fix, and it is **2 649** with the Vue vector — a hand-copied number that had already drifted twice before anyone looked at it. The body now carries the measured figure.

Worth naming rather than just fixing: **this number is derivable and prose is the wrong place for it.** The runner prints it on every run; every copy in a document is a copy that goes stale silently, which is the same reasoning that took three live counts out of `packages/framework/AGENTS.md`. I checked the rest of the tree — `2 629` appears in no file, no ADR, and not in `packages/framework/AGENTS.md`; it lived only in the PR body. My own commit message on the unroot fix quotes 2 645, which was accurate for that commit and I have not rewritten pushed history to chase it.

## Unchanged on purpose

`status/agent-context-budget.json` is untouched and `packages/framework/AGENTS.md` is still **32692 bytes** — the Vue vector needed no edit there.

`gjsify run test --include @gjsify/gtk-host --include @gjsify/react-native` → exit 0, gtk-host **2 649**, react-native 2 305; `gjsify lint` and `gjsify format --check` both exit 0.

Issue #1549, opened 2026-09-04
provideWindowChrome has no vector at its only shipping call site


JumpLink:
`packages/framework/react-native/src/app-registry.ts` calls `provideWindowChrome(chrome, …)` in exactly one place — `runApplication`'s render call — and that is the only place a shipping application ever gets the window's chrome published to its tree. Nothing tests it there.

## What that costs, measured

Delete the call, drop `chrome` from the `buildWindow` destructuring, delete the now-dead import. That is not a hypothetical: it is precisely what a rebase resolution produces when the branch side of the conflicted hunk is taken and the resulting `TS6133` is "fixed" by removing the unused import. Measured on `main` + #1538:

| gate | result with the call deleted |
|---|---|
| `oxfmt --check .` | clean, 3694 files |
| `oxlint` | 0 errors |
| `gjsify workspace @gjsify/react-native run build` (tsc) | 0 |
| `@gjsify/react-native` unit suite | **2345 completed, 0 failed** |
| #1540's ten window-chrome vectors | **all green** |
| the running application, `DumpTree` over a routed navigator | **2 mapped `AdwHeaderBar`** |

Two bars is #1460 exactly: two sets of window controls, and only one of the two close buttons closes the window.

## Why the existing vectors cannot see it

`src/router/router.spec.ts` composes the shell by hand — `buildWindowShell()` then `provideWindowChrome(shell.chrome, element)` at lines 411 and 1329 — and never runs `AppRegistry.runApplication`. So it measures the composition it wrote, not the one that ships.

`window-chrome.ts`'s own header already warns about this shape, one function earlier:

> A function rather than four lines inside `buildWindow`, because the router's own vectors have to measure THIS composition — a spec that rebuilt it by hand would pass while the shipping shell drifted.

That reasoning extracted `buildWindowShell`. The other half of the same composition, `provideWindowChrome` at the bootstrap's call site, was left with the drift the sentence describes.

## Status

#1538 adds the first vector that can see it: a third `routed` vector in `tests/e2e/react-native-devtools`, which renders `<RouterRoot>` through `registerRootComponent` and counts mapped `AdwHeaderBar`s from another process over DBus. Red before green — `got 2` — while `env` and `option` stay green, because a non-routed tree never asks for the chrome.

That closes the observability gap but not the CI gap: the suite is ledgered in `scripts/e2e-unlisted-suites.mjs` and runs in no job. See the companion issue on giving it a host.

## The pattern, which is the actionable part

[...]

PR #1564, opened 2026-09-04
test(e2e): a host that promised the environment must say so

### LINKED ISSUE #1550: Give react-native-devtools a CI host, and a skip-is-fatal switch
`tests/e2e/react-native-devtools` (ADR 0043) is the only external observer of `AppRegistry.runApplication`'s body. It runs in no CI job.

## What is unguarded because of it

Three lines in `runApplication` are observable only by running the loop, so no unit vector can reach any of them:

- `registerBuiltinWidgets()` — without it the first React commit throws `GtkHostError: No descriptor registered for <GtkBox>` inside a GObject handler, GJS swallows it, and the process stays up with an unmapped window answering `Screenshot` with `(@ay [],)`.
- the `toShellOptions` passthrough — `app-registry.spec.ts` holds the *value*, which is why that one has a unit vector, but not that `runApplication` uses it.
- `provideWindowChrome()` — see #1549; deleting it leaves every in-process gate green and the application drawing two header bars.

## Where it can run

Confirmed by reading `.github/workflows/main.yml` and `.github/actions/gjsify-setup/action.yml`:

- **`examples` job** — has the environment (`xvfb-run … dbus-run-session`) and every precondition the SKIP gate checks: the build-output cache path list is `packages/*/*/lib`, so all four framework libs are restored; `bootstrap-bundles: 'true'` restores `dist/cli.gjs.mjs`; the `@gjsify/rolldown-native-linux-x64` typelib is committed, not built. **But its trigger is `@gjsify/example-*` in the closure**, so a change confined to `packages/framework/react-native/**` skips it — silent for exactly the PRs the suite exists to catch.
- **`e2e` job** — right trigger, no display.
- **`test` job** — has a display (its whole sweep runs under `xvfb-run`), no session bus.

So: widen the `examples` trigger, or give the suite its own step with `xvfb-run … dbus-run-session`, in the shape the `test` job already uses for `Test WebGL conformance`. Add `GTK_A11Y=none GSK_RENDERER=cairo GDK_BACKEND=x11 LIBGL_ALWAYS_SOFTWARE=1` — the same set the `examples` job uses; without `GTK_A11Y` the fixture logs `Unable to acquire the address of the accessibility bus … Permission denied` (measured, warning only).

## The part that must not be skipped

**A skip-is-fatal switch, or the step buys nothing.** The suite's SKIP gate has nine preconditions and a CI host that misses any one of them goes quiet rather than red — the same green-that-checked-nothing the ledger exists to prevent, just relocated into a workflow where it looks like coverage. An env var (`GJSIFY_E2E_REQUIRE=1`) that turns the gate's `skip` into a failure naming the missing precondition is the shape; it generalises to `devtools-export` and `terminal-native`, the other two ledgered suites.

Ledger entry and reasoning: `scripts/e2e-unlisted-suites.mjs`, `status/open-todos.md`.

JumpLink:
Closes #1550.

`tests/e2e/react-native-devtools` is the only external observer of `AppRegistry.runApplication`'s body — three lines no unit vector can reach, one of which was deleted in a measurement and left every in-process gate green while the application drew two header bars (#1549). **It ran in no CI job.** Neither did `devtools-export`.

## Where it runs now

The `examples` job has the environment and the wrong trigger: gated on `@gjsify/example-*` being in the closure, so a change confined to `packages/framework/react-native/**` skips it — silent for exactly the PRs the suite exists to catch. The `e2e` job has the right trigger and no display, which is **one step to add** rather than a trigger to widen.

```yaml
- name: Devtools E2E on a display + session bus (shard 1)
  if: ${{ matrix.shard == 1 }}
  env:
    GJSIFY_E2E_REQUIRE: '1'
  run: |
    su testuser -c '… GTK_A11Y=none GSK_RENDERER=cairo GDK_BACKEND=x11 LIBGL_ALWAYS_SOFTWARE=1 \
      xvfb-run -a --server-args="-screen 0 1280x720x24" \
      node --test tests/e2e/react-native-devtools/run.mjs tests/e2e/devtools-export/run.mjs'
```

`xorg-x11-server-Xvfb`, `dbus-daemon` and `dbus-x11` are all in `.docker/ci-fedora.Dockerfile`.

## The switch is what makes the step worth having

Both suites guard themselves with a dozen preconditions, and a runner missing **one** goes quiet rather than red. A silent suite inside a workflow step reads exactly like coverage — the green-that-checked-nothing class, relocated into a place where it looks like progress.

`GJSIFY_E2E_REQUIRE=1` turns a skip into a failure that **names the missing precondition**. That is why the gates became named preconditions instead of a chain of `&&`: a boolean could only report that something was absent, on a host nobody can log into.

`requireEnvironment(label, preconditions)` lives in `tests/e2e/helpers.mjs` and generalises to the third ledgered suite whenever `terminal-native` gets a home.

## Measured on this workstation, which has a display and a bus

```
node --test …/react-native-devtools …/devtools-export        → 4 pass, 0 fail
GJSIFY_E2E_REQUIRE=1 (same)                                  → 4 pass, 0 fail
DISPLAY= WAYLAND_DISPLAY= (no switch)                        → 0 pass, 0 fail, 0 skipped — silence
DISPLAY= WAYLAND_DISPLAY= GJSIFY_E2E_REQUIRE=1               → 1 fail:
  "a display, a session bus and a built workspace — missing: DISPLAY or WAYLAND_DISPLAY"
```

The third line is the state this closes: a suite that reports nothing at all, in a job that has a green tick.

Both suites stay in `scripts/e2e-unlisted-suites.mjs` — `test:e2e` still cannot carry them, and the shared batch has no display. The ledger entries say what changed rather than being deleted.

Also verified: every `audit-runtimes.yml` script, `check-e2e-suite-coverage`, `check-workflow-run-syntax`, `gjsify format --check`, and `main.yml` re-parsed as YAML.


JumpLink:
Closing in favour of #1565, which does the same job and does two things this PR gets wrong.

**The claim this PR rests on is false, and I measured it after it was challenged.** The body says the `e2e` job "has the right trigger" — I took that from #1550's own text instead of measuring it. Against the classifier on `main`:

```
$ printf 'packages/framework/react-native/src/primitives/widgets.ts\n' \
    | node packages/infra/cli/lib/index.js affected --changed-from-stdin
  global:          false
  run-e2e:         false
```

`main.yml:1302` gates the job on `run-e2e == 'true' || global == 'true' || global == ''`, so a change confined to `packages/framework/react-native/**` skips it. This PR would have added the step to a job that does not run for exactly the PRs the suite exists to catch — the step would report green having measured nothing, which is the class it was written to end. #1565 carries the trigger fix and the classifier defect behind it.

**And the switch here is too blunt.** `GJSIFY_E2E_REQUIRE=1` demands every ledgered suite, including `devtools-export`, whose loss of its DBus name in the containerised runner is a measured and still unexplained open question in `status/open-todos.md`. That would have turned an open question into a red gate on arrival. #1565's variable takes a list of suite names, so a host asserts only what it has been shown to provide.

The one thing this branch has that #1565 may not is the per-precondition naming inside each suite's gate (`{ need, ok }` pairs rather than a chain of `&&`) — #1565's `e2eSkipReason` takes the same shape, so nothing is lost.

Issue #1529, opened 2026-09-03
check-refs-citations verifies the file, never the line — wrong citations ship


JumpLink:
## What

`scripts/check-refs-citations.mjs` validates that a cited `refs/` **file** exists. It does not
check that the cited **line** says what the citing text claims. So a citation can be wrong by
hundreds of lines and the gate stays green — and at least one wrong citation currently ships
inside a published package.

Found during an independent review of #1528.

## Measured

Three citations in that PR point at the wrong lines in `refs/gtk/gtk/gtkmenutrackeritem.c`. The
**semantics they describe are correct** — the reviewer read every site — but the addresses are not:

| claim | cited | actually at |
|---|---|---|
| `self->sensitive = enabled` | c:333 | **c:332** |
| role RADIO / CHECK | c:551-562 | **c:336-346** — c:551-562 is a block of local declarations, ~215 lines away |
| an actionless item is sensitive | c:588-591 | branch at **c:591-595**, assignment at **c:594** |

`node scripts/check-refs-citations.mjs` exits **0** on all three.

Five other citations in the same PR are correct as cited (c:271-272, c:267-268, c:534-535,
c:541-543, c:582-586), so this is not a systematically shifted file — it is per-citation drift
that nothing catches.

## Why it matters more than a typo

The `c:551-562` string is inside `@gjsify/adwaita-core`'s **conformance vectors**, so it is
published to npm. A reader who follows it lands on variable declarations and concludes the
vector's rationale is wrong.

More generally: citations into `refs/` are how this repo grounds behaviour in upstream source
rather than in recollection. A citation that cannot be trusted to point at the right line is
the same failure as a check that cannot be trusted to fail — it converts "verified against
upstream" into "asserted, and the assertion was never read".

`refs/` submodules also move. A citation correct when written drifts silently on the next bump,
and today nothing reports that.

## Suggested shape

Require a citation to carry an anchor the checker can verify at the cited line — an identifier,
a signature fragment, or a short quoted substring — and fail when the line does not contain it.
That also makes a `refs/` bump report its own damage instead of leaving stale addresses behind.

Whatever the mechanism, the point is that the check should verify what the citation claims,
not merely that a file by that name exists.

JumpLink:
Fixed on `main` in faf2d1491 (#1565), commit on `check-refs-citations`.

[...]

Issue #1516, opened 2026-09-03
Nothing holds a prose code fence in the reference docs against reality


JumpLink:
Two pages shipped the same snippet that cannot run, for an unknown length of time, and no check could see either.

`website/src/content/docs/ship/windows.md` and `website/src/content/docs/cli-reference.md` both handed the reader a font-registration loop calling `listFontFiles(dir)` — **a function defined nowhere on either page**, with `GLib` unimported in one of them. Copy it, get a `ReferenceError`. Both are fixed now (#1512, #1515), but the fix is the instance, not the class.

## Why nothing caught it

`scripts/check-doc-fences.mjs` covers blueprint / ts / scss / nativescript fences. It does **not** cover the `js` and `ts` fences in `cli-reference.md`, and there is no other reader of prose code fences on the site.

And `cli-reference.md` is **hand-written** — verified: no generator (`grep -rn cli-reference scripts/ website/scripts/ .github/` finds only the sidebar entry), no AUTO-GENERATED banner, last touched by ordinary feature commits. That is precisely how two hand-maintained copies of one snippet drifted apart and then both went stale: nobody was wrong on purpose, and nothing was watching.

The same passage also carried a live count — "the default font map stays at its 82 system families" — which is the baseline of one Windows 11 VM rather than a property of anything. The finding it was reaching for is that the number moves by **zero**. `AGENTS.md` already names live counts in prose as a drift source; this is one, on the user-facing site.

## What would actually hold

Worth deciding rather than assuming — a fence check has a real false-positive problem, since documentation snippets legitimately elide imports and reference project-specific names.

Options, roughly in order of cost:

1. **Extract-and-typecheck the fences that claim to be complete programs.** Mark those explicitly (a meta on the fence), extract them to a scratch project, run `tsc --noEmit` against the real `@gjsify/*` types. Catches the exact defect: an undefined identifier. Requires the fences to declare their own completeness, which is a small authoring burden and a large payoff.
2. **Reference-check identifiers against the packages a fence imports.** Cheaper, no build, but a genuine false-positive risk on deliberately abbreviated examples.
3. **At minimum, extend `check-doc-fences` to SEE `js`/`ts` fences** in the reference docs, even if it only enforces something narrow at first. It currently cannot report on them at all, which is the part that surprised us.

Do not build option 2 without measuring its false-positive rate first — this repository has a recorded history of guards that flag more noise than signal being switched off and then proving nothing.

## Two smaller items found in the same pass

[...]

Issue #1446, opened 2026-08-31
The widget table offers classes the running GTK does not have, because it is generated on one OS


JumpLink:
The first GTK-bearing Windows leg ran `@gjsify/gtk-host`'s suite on Windows for the first time. It fails **6 of 2264** there, against **2274 of 2274** on both darwin arms, and all six have one cause:

```
GtkPageSetupUnixDialog: Cannot read properties of undefined (reading 'list_properties')
GtkPrintUnixDialog:     Cannot read properties of undefined (reading 'list_properties')
```

Both are `…UnixDialog`. **GTK does not build them on Windows.** The generated widget table offers them as though it were platform-neutral, because it is generated from the GIR on a **Linux** host and bakes Linux-only classes in.

This is the failure class a one-OS pipeline *cannot* get wrong: under Linux the class is always there, so nothing that only ever ran there could have caught it.

## It is not the bundle and not the operating system

Ruled out before concluding anything:

- The published `@gjsify/gtk-runtime-win32-x64` verified clean — 33 library-backed typelibs, Gtk **and** Adw among them, `windowing=true`, decode probe through the bundle's own GTK.
- `check-batteries.mjs` was green before the suite started.
- `refuses a type whose namespace it cannot import` is **green**, so no namespace is missing — a class inside a present one is.
- Both published 0.45.0 closures were pulled and their typelib sets diffed: win32 45, darwin 47, and the difference is platform-appropriate on all but three entries (`AppStream-1.0`, `Xmlb-2.0`, `GDesktopEnums-3.0` absent on win32; `GIRepository-2.0` absent on **darwin**). Nothing a widget table needs.

## It is a platform omission, not a version skew — and the difference decides the repair

The tempting diagnosis is that gvsbuild's GTK is simply older than Homebrew's, and the
byte sizes support it: `Gtk-4.0.typelib` is **559,728** bytes in the win32 bundle against
**567,784** in the darwin one. That story is wrong, and it is worth stating why, because
**it sends someone to a GIR bump that fixes nothing.**

The full identifier diff between the two typelibs is **12 names, and they are one
subsystem**:

```
GtkPageSetupUnixDialog  GtkPrintUnixDialog  GtkPrinter  GtkPrintJob
GtkPrintCapabilities    PrintBackend        PrinterFunc PrintJobCompleteFunc
```

GTK's Unix print stack, complete, and nothing else. An older GTK would be missing classes
[...]

PR #1565, opened 2026-09-04
fix: the four places where a green result meant nothing

### LINKED ISSUE #1556: Two Image vectors are red on win32-x64 only, invisible behind a continue-on-error probe
Two `@gjsify/react-native` vectors have been failing on win32-x64 and nowhere else, and nothing has
reported it because they sit behind a `continue-on-error` probe whose step conclusion is `success`
either way (#1552).

Read out of `GTK OS suites` run `33864086251`, job log, on the win32-x64 leg:

```
❌ [Node.js 24.19.0] 2 of 2211 tests failed  (5.04s)
  ✖ the P2 widgets, in a real tree › renders an Image as a Gtk.Picture that FILLS,
    with the file GTK will open                      — Expected values to match using ===
  ✖ the P2 widgets, in a real tree › puts an ImageBackground's picture BEHIND its
    children, not over them                          — Expected values to match using ===
```

**Both pass on the other legs of the same run** — the identical two names appear with `✔` twice
later in the same log, so this is not a broken vector, it is a platform difference. `2 of 2211` is
the residual after #1541 fixed the six `GTK_A11Y=none` failures that were in the same probe; before
that the win32 leg read `8 of 2038` and the two were invisible inside the larger number.

`Expected values to match using ===` on a `Gtk.Picture`'s file and on an `ImageBackground`'s child
order both point at the same shape — what the widget ends up holding differs from what the layer set
— but which half differs is not in the log, and I have not reproduced it on a Windows machine. So
this issue is the report, not the diagnosis.

## A third one in the same run, probably already known

The gtk-host probe on the same job reads:

```
❌ [Node.js 24.19.0] 1 of 2579 tests failed  (6.05s)
  ✖ generated surface vs installed typelib › offers no property the installed GTK does not
    have as writable                      — Expected values to be deeply strictly equal
```

That reads like the symptom of #1446 — the widget surface generated on one OS offering what the
running GTK does not have — rather than a separate defect. Worth confirming against that issue
rather than filing twice. It is mentioned here because it is the second silently-red probe on the
same leg, which matters for the count question below.

## Why these are worth a ticket rather than a note

They were found while measuring something else, and the reason nobody found them earlier is
structural rather than careless: a `continue-on-error` step reports `success`, so the only way to
see these is to open a job log for a leg that is passing. Once #1552's reporting half lands, three
red assertions appear at once on win32 and the natural reading is "those were always red" — which
is how a known-red count becomes permanent.

So the useful thing is to have them named BEFORE that happens, with the numbers they had:

| leg | probe | failing |
|---|---|---|
| win32-x64 | `@gjsify/react-native` | 2 of 2211 — the two above |
| win32-x64 | `@gjsify/gtk-host` | 1 of 2579 — likely #1446 |
| darwin-arm64 / darwin-x64 | `@gjsify/react-native` | 0 of 2215 |

That table is also the ledger #1552's second half asks for: an expected-count entry is only
honest once someone has decided whether the expected count is 2 or 0.

Measured on `48d751f125`, run `33864086251`, 2026-09-04.

JumpLink:
**Stacked on #1560** (`fix: instruments that state what they measured`) — base is
`fix/measurement-apparatus`, not `main`. #1560 landed the reader that turns a
`continue-on-error` step's `outcome` into a job-summary row; this PR is the four things
that reader would otherwise have reported as "always red".

Closes #1556, #1550, #1516, #1529.

**#1564 covered this PR's #1550 half** and has been closed in favour of this one — both
sides independently measured that the `e2e` job does not run for a change confined to
`packages/framework/react-native/**` (`run-e2e=false`). What survives from that comparison
is in this PR: the `SCRIPT_COUPLINGS` entry that fixes the trigger, `GJSIFY_E2E_REQUIRE`
scoped per suite rather than all-suites, `terminal-native` routed through the same switch,
and `devtools-export` deliberately NOT demanded in CI — it loses its well-known bus name in
the containerised runner for a reason nobody has found, and a gate on an open question is a
red gate.

Every one of these is the same defect wearing a different hat: something reported success
without having measured anything. So each section says what was measured, on what, and
what is deliberately still open.

---

## #1556 — two Image vectors, red on win32 only

**I cannot run win32 from here and did not.** The primary evidence is the job log the
issue names, re-read at `48d751f125`, run `33864086251`, job `GTK suites on the shipped
closure (win32-x64 / node)`:

```
❌ [Node.js 24.19.0] 2 of 2211 tests failed
  Expected: /nonexistent-gjsify-vector.png (string)
  Actual:   \nonexistent-gjsify-vector.png (string)
```

Both vectors asserted `picture.file?.get_path()` came back the POSIX literal they wrote.
`Gio.File.new_for_path` does not store the string it is given — it CANONICALISES and joins
with `G_DIR_SEPARATOR`. **Measured locally on linux-x64, gjs 1.88.1**, which is the half I
*can* run: `/a//b/./c.png` comes back `/a/b/c.png` and `/a/b/../c.png` comes back
`/a/c.png`, so `get_path()` is the canonicaliser's output rather than the input echoed;
`GLib.DIR_SEPARATOR_S` is `/` here and `\` on Windows. That is the whole mechanism, and it
matches the log character for character.

The fix asserts in GIO's vocabulary instead of POSIX's — `Gio.File.equal` compares
canonical paths, which is exactly the claim ("the picture holds the file the source
named"), plus the basename so that *some other* file cannot pass. Not a weakening: the
literal was only ever true where the separator happened to match.

**A real win32 defect fell out of the same function.** `normaliseSource` classifies a
source by scheme with `^([a-zA-Z][a-zA-Z0-9+.-]*):`, and RFC 3986 lets a scheme be a
single ALPHA — so `{ uri: 'C:\images\logo.png' }` was refused as "a `c:` URI, and this
layer only opens what GTK can read synchronously", on the one OS that spells absolute
paths that way. No vector reached it because no vector ever wrote a Windows path. `*` → `+`
is the fix (no registered scheme is one letter; WHATWG's URL parser carves the same
exception), unconditional so one source string cannot mean two things. Three new vectors:
both separators are paths, and `ws:` still reads as a scheme.

After: `@gjsify/react-native` **2236 completed / 0 failed** on GJS 1.88.1 locally, `tsc`
clean. The win32 leg is what has to confirm the two, and it is a probe on that workflow by
design — the readout will be on the run.

`gjsify.osNotes.win32` named these two as a residual and made promotion conditional on
them; that clause now records the fix. The third silently-red probe on that leg is #1446
(`packages/framework/gtk-host`, not mine) and is untouched.

## #1550 — a CI host, and a switch that makes a skip fatal

`tests/e2e/react-native-devtools` (ADR 0043) ran in **no** job. It is the only external
observer of three lines in `runApplication` that no unit vector can reach.

`e2eSkipReason(suite, [[what it needs, is it here], …])` replaces the boolean SKIP chains
in all three ledgered suites. It returns the reason to skip, or **throws** naming the
missing preconditions when `GJSIFY_E2E_REQUIRE` names the suite. **A/B, measured, with the
display taken away:**

| | exit | output |
|---|---|---|
| no switch | **0** | `tests 0 … fail 0` — the defect |
| `GJSIFY_E2E_REQUIRE=react-native-devtools` | **1** | `host is missing 1 of 11 precondition(s) — a display (DISPLAY or WAYLAND_DISPLAY)` |
| switch + a display | **0** | `tests 3 … pass 3 … skipped 0` |

A skip-is-fatal switch that was never seen to fail is the same defect one level up, so
that middle row is the point of the section.

The step goes in `main.yml`'s `e2e` job, shard 1, under `xvfb-run … dbus-run-session` with
the env the `examples` job already uses. `examples` was the other candidate and is the
wrong host: it is gated on `@gjsify/example-*` being in the closure, so a
`packages/framework/react-native/**`-only change skips it — silent for exactly the PRs the
suite exists to catch. **No fourth required check**; three still gate `main`.

That trigger is fixed with a `SCRIPT_COUPLINGS` entry, the shape `.githooks/**` already
needed. Writing its vectors found a real classifier bug: the test-only shortcut returned
`runE2E: touchedE2E` and dropped `couplingRunE2E`, so a coupling's reach depended on the
*shape* of the changed path. `.githooks/pre-commit` survived only because a hook is not a
spec file; `…/react-native/src/*.spec.ts` fell through and turned the tier back off. The
vector that says so was RED when it was written. `tests/e2e/affected-classifier-bundle`:
14/14, and the committed `dist/affected.gjs.mjs` is rebuilt.

Only `react-native-devtools` is named in CI. The switch generalises to the other two, but
`devtools-export` loses its bus name in that container for a reason nobody has found, and
nothing in CI stages `terminal-native`'s prebuild — demanding either would turn an open
question into a red gate.

## #1516 — a doc fence that shows its imports must run

**#1516 asks for the false-positive rate before the guard. Measured, on every js/ts/jsx/tsx
fence in the docs tree: the cheap rule (option 2 — flag every called-but-unbound identifier)
flags a large minority of them, and reading every one, essentially none is a defect.** They are
class methods (`vfunc_startup`, a `constructor`), GJS ambient globals (`print`),
deliberately elided namespace imports in `patterns/` (`GObject`, `Gtk`, `Gio`),
compiler-output samples (`_$createElement`), and helpers the prose says live elsewhere.
That is the switched-off-and-proves-nothing guard this repo has paid for before, so it is
not built.

What is built is option 1 without the authoring burden: a fence is checked **iff it shows
at least one `import … from`** — such a fence is claiming to be a program a reader can
copy; one that elides its imports is an excerpt, and the arm prints how many qualify on every run. They are handed to
the repo's own `tsc` and only `TS2304`/`TS2552` ("Cannot find name") are read — an
unresolved module is a separate `TS2307`, which is what lets this arm run in `tree-checks`
before anything is built.

**Two holes in my own detector were found by its own fixture matrix, and this is the part
worth reading.** My first measurement said *0 flagged*. That number was itself a
green-that-checked-nothing: the scratch `tsconfig` named the GJS ambient types as a `types`
entry, which cannot resolve a subpath, so tsc stopped at `TS2688` before checking a single
fence and the arm reported a clean corpus it had never read. Then the control fixture
found a second hole: **`TS2552` is the same diagnostic with a suggestion attached**
(`Cannot find name 'logError'. Did you mean 'onerror'?`), so reading `TS2304` alone would
miss every unbound name one edit from a DOM global. Both are now fatal-by-fixture: the fixture matrix runs before the docs, a wrong verdict stops the arm, and a tsc diagnostic with no
file in front of it is an error rather than a clean corpus.

With both fixed the arm found **8 live instances of the #1516 defect** — e.g.
`packages/dom.md` writes `new Adw.ApplicationWindow({ application: app })` with `app`
created nowhere on the page. All 8 are fixed here, not exempted.

**And it has a measured blind spot, which found a 9th instance by hand.** The scratch
project compiles with `lib: ["ES2024","DOM"]`, so `window`, `document`, `location` and ten
more count as BOUND in a fence that imports `gi://` and will have none of them at runtime.
Measured: `confirmDialog(window, …)` reports nothing under `["ES2024","DOM"]` and
`TS2304: Cannot find name 'window'` under `["ES2024"]`. There was a live instance —
`guides/native-adwaita-app.md` passed a bare `window` to four `@gjsify/adwaita-app` calls
as the parent `Adw.ApplicationWindow` — and it is fixed. Dropping the DOM lib is not the
repair: `console` is declared by neither `lib.es2024` nor `@girs/gjs`, so every GJS fence
would flag it, and the canvas/iframe fences legitimately use `document` and
`ResizeObserver`. That wants per-fence lib selection plus a curated GJS global set;
ledgered rather than guessed at.

Scope is per PAGE, not per fence: a guide legitimately builds one program across several
fences (`patterns/bridges.mdx` opens an `app` and uses it in five more), the same rule the
TOKENS arm already reads. It does not soften the incident — `listFontFiles` was defined by
no fence on either page.

It is an arm of `check-doc-fences.mjs` rather than a new script **because the simplicity
rule asks what a guard lets you delete, and the honest answer here is "nothing"** — so it
buys the smallest thing it can: no new CI step, no second exemption ledger, no second
failure reporter, no fence-meta convention to agree first.

## #1529 — verify the line, not just the file

**The heuristic first, because it is the one the issue nearly asks for and it is dead.**
Deriving the anchor from the citing prose — take the backticked tokens near a coordinate,
require one at the cited lines — **flags 14 of the 40 line citations that have a backticked
token nearby**, and reading them they are overwhelmingly prose that *mentions* a name
rather than quoting the cited source (`SHORTCUT_LABEL_VECTORS` is the citing file's own
constant; `.rpm` is a sentence). That is the quote-versus-mention wall
`status/sections/priorities.md` records at 42 of 98 for the ledger, hit independently here.
Not built.

Two zero-false-positive arms instead:

- **RANGE** — the cited lines must exist, be ordered, and not be blank. This is the half
  that makes a `refs/` bump report its own damage.
- **ANCHOR** — `…/wrap.js:1305-1306#getFinished`: the text after `#` must appear inside the
  cited range. The only arm that catches a citation wrong *at birth*, which RANGE cannot —
  `c:551-562` exists, it simply says something else.

`node scripts/check-refs-citations.mjs` now reports, per run, how many coordinates name a
line, how many were read against the file, how many `#anchor`s the tree carries and how many
of them this host verified, how many are skipped as not checked out, and how many self-test
shapes ran ahead of all of it. The detector is asserted before it is believed, and a corpus of zero —
or an anchored corpus of zero — is a failure, so the arm cannot be silently emptied.

Live negative controls on the real tree, not just fixtures:

```
refs/node/lib/internal/tls/wrap.js:1305-1306 carries the anchor `getSharedSigalgs`, and
  lines 1305-1306 does not contain it. What is there: 'getFinished', / 'getPeerFinished',
refs/node/lib/internal/tls/wrap.js:1305-99999 names line 99999, and the file has 1922 …
```

Note the first message prints **what is actually at the cited lines** — that is the
property that would have caught #1529's instance in review.

It also found a live one. `refs/node/lib/_tls_wrap.js` no longer exists (upstream moved it
to `lib/internal/tls/wrap.js`); the EXISTING file arm was already red on it here and CI
never sees it, because CI checks out only `refs/libadwaita`. Fixed and anchored, which is
also what seeds the ANCHOR arm.

#1529's own three `gtkmenutrackeritem.c` coordinates were already corrected in #1528 before
the merge — `551-562`, `c:333` and `588-591` are not in the tree — so no `adwaita-core` file
is touched here.

---

## User docs vs contributor docs

The audience for `website/src/content/docs/**` is someone building an app WITH gjsify, so
the incidents, gates and measurements in this PR live in commit bodies and
`status/open-todos.md`, never on a page.

- **#1556 and #1550 are contributor-only on the machinery side** — a win32 assertion, a CI
  host, a classifier tier. Nothing a reader does changes, so no user page moves for them…
- **…with one exception, which #1550 turned up.** `guides/devtools.md` gave two ways to turn
  devtools on, both requiring an `Adw.Application` you wrote. A `@gjsify/react-native` reader
  has neither — `registerRootComponent` IS the bootstrap — so the page left that audience with
  no answer. The suite is what establishes the answer, so the page now carries it. My first
  draft of that snippet was wrong (`AppRegistry.runApplication` from `react-native`; it is
  `registerRootComponent` from `@gjsify/react-native`), caught against the suite's fixture —
  and my second draft was caught by **this branch's own fence checker**, for an unbound `App`.
  That is the first evidence the arm works on fresh writing rather than on the corpus it was
  tuned against.
- **#1516 is entirely user-facing** — those fences *are* the documentation, and a wrong one is
  a bug the reader hits by copying it. 9 fences across 9 pages are fixed, in the pages' own
  voice; nothing about the checker or the incident appears on any of them.
- **No `contributing/` page is touched**, and no contributor material moved onto a user page.

**Was any fence wrong in a way that needed a CODE change?** No — and the reason is structural
rather than lucky. TS2304/TS2552 report unbound NAMES only; a fence importing something a
package does not export is TS2305, which this arm does not read. I measured that class
anyway, and the result is in `open-todos.md`: **there are no findings, because the probe
cannot currently produce trustworthy ones.** The per-PAGE unit that makes the TS2304 half
correct makes an exports check wrong — two fences importing `Adw` from `gi://Adw` and from
`@gjsify/adwaita-nativescript` collide as `TS2300`, and the cascade reports a FALSE
`TS2305: … has no exported member 'Adw'` (`src/index.ts:192` is `export * as Adw`).

That probe is also worth one line here, because it repeated this PR's own subject: it
reported `0 diagnostics` **twice** while tsc had never run — first `TS2688`, then `TS5101
baseUrl is deprecated` — and only an injected control (an import of a deliberately
non-existent export) exposed it. The shipped arm treats a file-less `error TS…` as fatal
precisely for this.

## Files touched outside my scope, deliberately and minimally

- `tests/integration/tls-session/src/channel-binding.spec.ts` — one citation line. Required:
  the new arm cannot be green while a cited file does not exist.
- `website/src/content/docs/**` (8 pages carrying the SNIPPETS arm's own findings, plus
  `guides/devtools.md`) — Landing the arm red is not an option, and exempting real
  defects is not either.
- `packages/infra/cli/**` (classifier + its spec + the committed bundle) — this is where the
  e2e tier for `packages/framework/react-native/**` is decided.
- `docs/ci-selective.md` — its SCRIPT_COUPLINGS paragraph said "the entry that exists"
  (singular) and described only seed-bearing couplings.
- `packages/web/adwaita-core/src/accent.ts` and `tests/AGENTS.md` (+ its budget line) — added
  by the review pass below; the reasons are in its commits.

## Deliberately left open — all in `status/open-todos.md`

Two entries are DELETED (the image vectors, and `react-native-devtools needs a display`).
Four are added:

- **Nearly every line citation carries no `#anchor`, and `refs/gtk` is checked out by no job.**
  ANCHOR is opt-in and a handful have opted in. Retrofitting is not mechanical work — each needs
  a human to open the cited lines — and doing it automatically would vouch for a wrong
  address as readily as a right one. Also named there: the `c:332` shorthand in
  `adwaita-core/src/menu.ts` is invisible to the reader, so some of the most load-bearing
  citations are not in the gate's corpus at all.
- **The doc fences that show no imports are unread**, on purpose — that selector is what
  makes the arm zero-false-positive. Closing it needs a fence-completeness convention first.
  The same entry carries the API-shape class above: why it wants a different job and a
  different unit rather than a wider regex on this one.
- **The DOM-lib blind spot above**, with its A/B measurement and why the obvious fix is wrong.
- **The two reference-doc gaps from #1516** (no "how do I test my app?" page; `gjsify
  showcase` lists no `webrtc-video` while the sidebar does).

I did NOT add a ledger guard — `priorities.md` item 3 says three were measured and all three
are worse than nothing.

## Verified

`gjsify --version` 0.42.0, shim resolving inside this worktree ·
`audit-runtimes --check` · `check-probe-outcomes-read` (18 probes, every outcome read) ·
`check-workflow-run-syntax` · `check-e2e-suite-coverage` · `check-refs-citations` ·
`check-doc-fences` · `gjsify lint` 0 errors · `gjsify format --check` clean ·
`@gjsify/react-native` 2236/2236 + `tsc` clean · `tests/e2e/affected-classifier-bundle` 14/14 ·
`tests/e2e/react-native-devtools` 3/3. All re-run after the rebase onto `4753f4c2c`.

`@gjsify/cli`'s suite has 4 pre-existing failures on this host (`buildPublishHeaders`,
`probeTrustState`, two node-gi `system-gi.js` agreement vectors). Confirmed pre-existing by
stashing this branch's cli changes and rebuilding: **4 of 3705 without, 4 of 3709 with, the
same four names.**

---

## Independent review pass (7 further commits)

Reviewed by a second agent that did not write this, against the standard in the sections
above. Findings and their before/after are in the review comment on this PR; the substantive
ones:

- `tests/e2e/devtools-export/run.mjs` lost `DRIVER` and `parse()` in the SKIP refactor while
  both are still called — `ReferenceError` before the suite reached dbus, invisible because
  no job runs that suite. Restored; the suite now passes 1/1 end to end.
- The ANCHOR arm counted anchors it had READ and reported the number as a property of the
  tree, so `audit-runtimes.yml` (which checks out `refs/libadwaita` and nothing else) found
  both anchors inside `refs/node`, counted zero and turned two required checks red saying the
  tree had none. Tree-level and host-level are separate facts now, and an anchored
  `refs/libadwaita` citation keeps the arm exercised on the job that gates `main`.
- The SNIPPETS arm called `print` its load-bearing control; measured, `lib.dom` binds `print`
  regardless. Control moved to `printerr`, and a tsc killed by a signal is now fatal rather
  than a clean corpus.


JumpLink:
## Independent review — fixed on the branch, 7 commits

I did not write this PR. Everything below was broken on purpose first and watched go red;
where I state a number I re-derived it rather than copying it.

### 1. The third instance of the class, and it is in this PR (`2b58350`)

`tests/e2e/devtools-export/run.mjs` — the SKIP-chain refactor deleted `const DRIVER` and
`function parse()` and left both call sites standing.

```
✖ exports org.gjsify.Devtools … — ReferenceError: DRIVER is not defined
    at TestContext.<anonymous> (run.mjs:128:70)
```

Nothing could see it: the suite is ledgered as unlisted so no job runs it, `oxlint` has no
`no-undef`, and `{ skip: SKIP }` means the body never executes on a host missing a
precondition. Restored from the base commit. After: the suite passes **1/1** end to end
here, `GJSIFY_E2E_REQUIRE=devtools-export`, on a real session.

That the switch this PR adds is what made the suite *runnable enough to catch it* is a point
in its favour, not against it.

### 2. Why the two required checks were red (`99b5e10`) — and the decision I took

`Detect runtime-triplet drift` and `Manifest checks (Windows)`, both on
`node scripts/check-refs-citations.mjs`:

```
- no line citation in the tree carries an `#anchor`, so the arm that verifies WHAT is at a
  cited line ran over nothing. Anchor at least one …
```

The tree had two. `anchored` was incremented **inside the read loop**, so it counted anchors
this host could OPEN, and `audit-runtimes.yml` checks out `refs/libadwaita` and nothing else
(~13 MB against ~150 GB for the pool — that step's own comment says it "holds what that one
contains and no more"). Both anchors are in `refs/node`. The gate then told the author to add
a thing that was already there: a gate misreporting its own reason, which is this PR's
subject.

**Two facts, now separate — this is the design call, say so if you disagree:**

| | source | zero means | verdict |
|---|---|---|---|
| anchors **in the tracked tree** | `git ls-files` — same answer in every job | nobody opted in; the arm can fire nowhere | **FATAL** |
| anchors **verified here** | which `refs/` submodules are on disk | this host could not read them | **named line in the summary** |

Not "skip when empty": the fatal half no longer depends on the host at all, and the readable
half is kept non-empty **by construction** — `packages/web/adwaita-core/src/accent.ts` now
anchors `refs/libadwaita/src/stylesheet/_colors.scss:146-170#--accent-bg-color`, and
libadwaita is the one submodule that job does check out. I opened those lines and picked a
token that is actually there, per the ledger's own rule.

Measured three ways: every anchor removed → fatal, new wording · anchors present, none
readable → `NOTE — none of the 3 anchored citation(s) was readable here; they live in
libadwaita, node` and exit 0 · the libadwaita anchor pointed at a token that is not there →
red, printing what is at 146-170.

The alternative — make the job check out what it reads — stays open and is the better
long-term answer for `refs/gtk`; it is a CI-cost decision and is ledgered.

### 3. A right rule with an invented reason (`722c38d`)

The SNIPPETS header: *"The `print` fixture is the load-bearing control — without `@girs/gjs`'s
ambient declarations every GJS fence would flag `print`."* Measured under the arm's own
tsconfig:

| lib | `print` | `printerr` | `logError` |
|---|---|---|---|
| `["ES2024","DOM"]` | **bound** (`Window.print()`) | TS2304 | TS2552 |
| `["ES2024"]` | TS2304 | TS2304 | TS2304 |

`print` is bound by the DOM lib whatever `@girs/gjs` does — the blind spot this PR ledgers
one section down. The control is now `printerr`, and is strictly stronger: with the ambient
reference broken it reports `[logError, printerr]` where it used to report `[logError]` alone,
so the TS2304 path is covered as well as the TS2552 one.

### 4. Two more holes in the same arm (`5c17829`)

- **A tsc killed by a signal printed nothing and read as a clean corpus.** `spawnSync` leaves
  `error` unset; the TS2688 guard reads what tsc *prints*; the fixture matrix is a separate
  run that has already returned. Replacing the spawn with `kill -9 $$`: clean before, `tsc was
  killed by SIGKILL … so no fence was checked` after.
- **The failure named the page, not the fence.** A unit is a whole page, so a maintainer got
  "something here" and diffed by hand — which is what the nine fixes in this PR already cost.
  The unit already opens each fence with `// --- <page>:<line> ---`; the message now carries
  it. Injected an unbound name: reports `native-adwaita-app.md:161`.

### 5. Counts asserted where a script derives them (`1ff37fa`, `32ca4be`, `722c38d`)

Every one of these was **already wrong on this branch**:

- ledger: *"119 coordinates name a line, 79 readable, exactly ONE anchored"* — the gate reads
  120 / 80 / 3. The entry's own prose example is a valid coordinate, so writing the entry
  moved the number it states.
- ledger + PR body: *"36 of 203 doc fences"*, *"167 of 203 qualify"* — the arm prints 168 of
  204. I re-derived the corpus with an independent walk of `website/src/content/docs`:
  **204 js/ts fences, 168 import-bearing, 47 pages** — the arm reads the right set.
- the precondition list is counted in four places at two numbers: "nine" in `helpers.mjs` and
  the suite, "eleven" in the workflow and the ledger. It holds eleven, and the helper's own
  message computes it (`missing 1 of 11`). None of the four states a number now.

### 6. A doc claiming a command the workflow does not run (`32ca4be`)

`docs/ci-selective.md`, the workflow comment, the ledger entry and the step NAME all said the
step runs under `xvfb-run … dbus-run-session`. It runs under `xvfb-run` only — deliberately:
the suite opens a fresh `dbus-run-session` **per vector** so the fixture never inherits a stale
owner of its well-known name. Corrected in all four, with the reason, so nobody "fixes" it by
adding an outer bus.

### 7. Governance (`e58f57b`)

No AGENTS.md was touched. `e2eSkipReason` + `GJSIFY_E2E_REQUIRE` is a new repo-wide
convention for how an e2e suite reports a missing precondition, and `tests/AGENTS.md` § E2E
already enumerates `helpers.mjs`'s members by name — that is where the next suite author
looks. One sentence added, `status/agent-context-budget.json` updated with it (10782 → 11390).

Otherwise clean: the new step is inside the existing `e2e` job, which `gate` already `needs`,
so **no fourth required check**; four `###` entries added and two RESOLVED entries deleted
rather than struck; no new ledger guard.

### What I checked and did NOT change

- **The classifier fix is real.** Reverting `runE2E: touchedE2E || couplingRunE2E` in the built
  lib: `…/react-native/src/widgets.spec.ts → runE2E=false`; with it, `true`. The vector goes
  red without the fix. `affected-classifier-bundle` 14/14; the committed bundle carries the new
  coupling and is held by `verify-committed-bundles.mjs`.
- **The win32 reasoning survives without a win32 host.** Both sides of `expectFileIsSource`
  go through the same `Gio.File.new_for_path`, so `equal()` holds in every separator and
  `get_basename()` splits on the native one — the claim does not depend on which separator
  GLib picks. `normaliseSource`'s `*` → `+` is one character wide and the `ws:` vector proves
  it is not a hole. `@gjsify/react-native` 2236/2236 on GJS 1.88.1 here.
- **Not building the two heuristics was right.** Both flag-rate measurements reproduce as
  noise, and `priorities.md` item 3 forbids exactly that shape. What *was* built earns its
  place: SNIPPETS is an arm of an existing script (no CI step, no second ledger) and found 9
  live defects; RANGE found a live dead-file citation nothing else could see.
- **The website edits read as user docs.** No `contributing/` page touched, no incident or gate
  on any of them, and I checked the claims rather than the prose: `devtools: true` really is on
  `RunApplicationOptions` (via `AdwaitaAppOptions`), `registerRootComponent` and `Text` really
  are exported, `registerRootComponent` really returns a Promise so the `await` is right, and
  `packages/dom.md`'s "the WebGL example below reuses this one" is true of the fence below it.
  One nit left alone: `frameworks/index.mdx` spells `default_width` where every other page
  spells `defaultWidth` — the new lines match that page's existing line 207, so fixing it means
  touching a line this PR did not.

- **The one "fixture reads what an earlier step wrote" question this raises, checked and
  clear.** The new step is the first CI-run e2e suite that needs four framework `lib/` trees,
  and for exactly the diff the coupling exists to catch the classifier returns
  `global=false, runE2E=true` with a closure of **one** workspace — measured. That would be a
  red required check for an environment reason if the other three were not built. They are:
  the build runs `gjsify foreach build $INCLUDE_ARGS --with-dependencies`, whose expansion is
  a transitive `affectedClosure`, and the step's `BUILT` list is precisely
  `@gjsify/react-native` → `@gjsify/adwaita-app` → `@gjsify/devtools`, plus
  `@gjsify/gtk-host`. Worth knowing before anyone edits either list: they have to stay the
  same set.

### What I could not measure

`ghcr.io/gjsify/ci-fedora:44` is a 403 from here, so I could not run the new step in its own
container. That step is now gating through `CI gate (GJS)`, and by design it goes RED rather
than quiet if the container misses a precondition — **the first run on this PR is that
measurement**, and it is the one thing left to watch before merging.

### Everything re-run after the last commit

`gjsify --version` 0.42.0, shim resolving inside this worktree · `gjsify lint` **0 errors** ·
`gjsify format --check` clean over 3704 files · `audit-runtimes --check` · `check-doc-fences`
(168 of 204, 8.0 s) · `check-refs-citations` · `check-probe-outcomes-read` ·
`check-workflow-run-syntax` · `check-e2e-suite-coverage` · `check-agent-context-size --check` ·
`@gjsify/react-native` 2236/2236 · `affected-classifier-bundle` 14/14 ·
`react-native-devtools` 3/3 with the switch on · `devtools-export` 1/1 with the switch on.

Issue #1072, opened 2026-08-08
16 conformance tables no renderer suite drives


JumpLink:
Found while wiring `COMBO_SELECTION_VECTORS` to every suite in #1071. A census of `packages/web/adwaita-core/src/conformance/`: **zero** tables have no consumer at all, but **44 have exactly one** — and 43 of those are core-only, i.e. the derivation is asserted against itself and no renderer is held to it.

That is not automatically wrong. The triage:

### Legitimately core-only (19)

Internal steps whose *composed* result is genuinely renderer-driven — the search pipeline feeding `PREFERENCES_SEARCH`, the clamp trio feeding `CLAMP_ALLOCATE`, the carousel arithmetic feeding `CAROUSEL_PAGE_LIST`, and so on. Driving the steps separately from a renderer would assert the same thing twice.

Two exceptions inside this group: `NATURAL_SIDEBAR_WIDTH` and `TAGS_CONFLICT` feed into `SPLIT_VIEW_MEASURE` and `NAVIGATION_ACTION`, which are themselves undriven — so those chains terminate nowhere.

### Reason already recorded (8)

But three of them — `TOOLBAR_VIEW_ALLOCATE`, `TOOLBAR_VIEW_MEASURE`, `TOOLBAR_VIEW_CONTENT_FOR_SIZE` — state the reason in *both renderers'* `adw-toolbar-view.ts` and not in `conformance/chrome.ts`, so the table itself is still silent to anyone reading it. Same pattern for the NativeScript half of the carousel wheel. Moving one sentence each fixes it.

### Silent gaps (16)

**Trivial to wire** — a loop in an existing spec, the element already calls the function:

- `AVATAR_MODE` (web)
- `BUTTON_CONTENT_DEFAULT` (both)
- `CAROUSEL_WHEEL` (web)
- `POPOVER_KEY` (web — `adw-popover.ts` already calls `resolvePopoverKey` on keydown)
- `TAB_PAGE_DESCRIPTOR` (both; one row's `rule` string literally names a NativeScript bug)

**Needs element capability first:**

- the five `OVERLAY_SWIPE_*` tables — no gesture code exists in either renderer (tracked in #1068)
- `NAVIGATION_ACTION` — no push/pop surface (also #1068)
- `CAROUSEL_WHEEL_LOCKOUT` — needs a real `Date.now`
- `ENTRY_ROW_GUARD` — the web element discards the setter's boolean
[...]

PR #1566, opened 2026-09-04
feat(adwaita)!: one model on every list widget, and only there

### LINKED ISSUE #1524: A portable list model is worth building narrower than it looks: 12 of 41 ledger entries are list-shaped, but GTK has a model property for only 5 of the widgets
The widget vocabulary is unified. The DATA those widgets consume is not. This is the measurement of the list half of that gap, taken at `origin/main` (`0ef1d7c137`), and the recommendation is **narrower than "a portable `Gio.ListModel`"** — because the measurement says one abstraction cannot span what is actually two different mechanisms.

## The short version

- List divergence is real and it is **~4x the menu divergence** by every yardstick that justifies the menu work: 12 of 41 vocabulary-ledger entries vs 3; 5 of 17 gallery refusals vs 2; 17 NativeScript collection properties vs 2.
- But it is **not the same job**. `Gio.MenuModel` is a writable property on **13** GTK widget interfaces. A list `model` property exists on **5** — and for the widgets that actually drive the ledger and the refusals (Sidebar, TabView, ToggleGroup, ViewStack, Carousel) GTK has **no list property at all**; the collection is `adw_sidebar_append()`.
- So: build a shared **item vocabulary + a positional change signal** in `@gjsify/adwaita-core` for the model-shaped widgets. Do **not** invent a model type for the method-built ones — those are a `descriptors/` gap, and the gallery already says so by name (`uncurated-placement`).
- There is a cheaper fix that needs no portable model at all, and it should go first: `<adw-combo-row>` and `<gtk-drop-down>` are the same list widget sharing one `ComboState`, and on adwaita-web only one of them can have its items set after connect.

## 1. The list-driven widget set, per surface

Not "which mention ListModel" — which take a collection of items.

| surface | list-driven | how it is fed |
|---|---:|---|
| `@gjsify/gtk-host` (GJS) | **5** with a real model prop; **8** curated child-collection descriptors; **1** neutral seam (`ListController`) | GObject model property, or descriptor `insert`/`add`/`append` calls |
| `@gjsify/adwaita-web` | **23** elements | 11 writable typed array props, 10 read-only projections, 4 JSON attributes, 11 DOM children |
| `@gjsify/adwaita-nativescript` | **17** collection properties | plain array assignment, every one a snapshot |
| `@gjsify/adwaita-react-native` | **7** of 28 widgets (+4 list components in `@gjsify/react-native`, +3 router surfaces) | `readonly T[]` props or `ReactNode` children |

The GJS number under-counts on purpose, and that is the finding: GTK's own answer for half the set is a method, not a property.

### The five widgets where GTK really has a model

`packages/framework/gtk-host/src/generated/props.ts` (provenance `Gtk-4.0/4.23.3 Adw-1/1.10.0`):

```ts
:1141  AdwComboRowProps      model?: Gio.ListModel | null;
:3220  GtkDropDownProps      model?: Gio.ListModel | null;
:3055  GtkColumnViewProps    model?: Gtk.SelectionModel | null;
:3832  GtkGridViewProps      model?: Gtk.SelectionModel | null;
:4279  GtkListViewProps      model?: Gtk.SelectionModel | null;
```

Those are real acceptances at runtime — `packages/framework/gtk-host/src/props.ts:196` falls through to `return value;` for a GObject-valued prop, which is how `packages/framework/adwaita-react-native/src/widgets/combo-row.gtk.tsx` builds a live `Gtk.StringList` (`:70`) and passes it as `model={strings}` (`:82`).

`GtkListView`, `GtkGridView`, `GtkColumnView`, `GtkDropDown`, `AdwComboRow`, `AdwTabView`, `AdwCarousel` are all **uncurated** in `packages/framework/gtk-host/src/descriptors/` — 42 curated gtypes out of 169 generated tags. Inserting a child into any of them raises a named error.

### Combo row / drop down — one widget, four spellings

| surface | signature |
|---|---|
| GJS | `model?: Gio.ListModel \| null;` — `generated/props.ts:1141` |
| adwaita-web (`gtk-drop-down`) | `set options(value: ReadonlyArray<AdwComboOptionInput>)` — `packages/web/adwaita-web/src/elements/gtk-drop-down.ts:93` |
| adwaita-web (`adw-combo-row`) | **no setter at all** — `JSON.parse(this.getAttribute('items') \|\| '[]')` once in `connectedCallback`, `packages/web/adwaita-web/src/elements/adw-combo-row.ts:43` |
| nativescript | `set options(value: AdwComboOption[])` — `packages/nativescript-bridge/adwaita/src/widgets/adw-combo-row.ts:141` |
| react-native | `model?: readonly AdwComboOptionInput[];` — `packages/framework/adwaita-react-native/src/props.ts:561` |

`AdwComboOptionInput` is the one thing three of the four already share — `packages/web/adwaita-core/src/rows.ts:97`:

```ts
export type AdwComboOptionInput = string | { value?: unknown; label?: unknown };
```

And React Native says exactly why it takes the name but not the type, `packages/framework/adwaita-react-native/src/props.ts:531-540`:

> `model` IS THE LIBADWAITA NAME AND NOT THE LIBADWAITA TYPE, which is the one liberty this surface takes and the reason it can exist at all. `AdwComboRow:model` is a `Gio.ListModel`; a props file that may import neither `gi://Gio` nor `react-native` cannot name that type […]

### The widgets GTK builds with methods

`AdwSidebarProps` (`generated/props.ts:1910-1943`) has `dropPreload`, `filter`, `menuModel`, `mode`, `placeholder`, `prefix`, `selected`, `suffix` — **no item list**. `AdwTabViewProps` (`:2217-2241`) has `defaultIcon`, `menuModel`, `selectedPage`, `shortcuts` — **no page list**. `AdwToggleGroupProps` (`:2252-2268`) has `active`, `activeName`, `canShrink`, `homogeneous` — **no toggle list**.

The ports each invented their own container, and the ledger records all of it as `own` — `scripts/check-vocabulary-alignment.mjs:562`:

> Items are ADDED on GTK — `adw_sidebar_append()` — and AdwSidebarProps carries no item list […]. The port takes the labels as a flat array because its list view is index-driven.

### The gallery: 5 of 17 refusals are list-shaped; 2 are menu-shaped

`website/src/data/adwaita-framework-snippets.ts` — measured, **23 blocks with a snippet + 17 refusals = 40**. (If you are carrying "19 refusals" from an earlier note, it is 17.)

| cause | count | blocks |
|---|---:|---|
| list-shaped, **sole** cause | **2** | `Adw.ComboRow` ("its items are a Gio.ListModel"), `Gtk.DropDown` ("its options are a Gtk.StringList model") |
| list-shaped **+ uncurated-placement** | **3** | `Adw.Sidebar`, `Adw.TabView`, `Adw.ToggleGroup` — each "its items/pages/toggles are … GObjects, which have no tag either" |
| menu-shaped | 2 | `Adw.SplitButton`, `Gtk.MenuButton` |
| widget reference | 3 | `Adw.ViewSwitcher`, `Adw.ViewSwitcherBar`, `Adw.InlineViewSwitcher` |
| child policy only | 3 | `Adw.PreferencesDialog`, `Adw.BottomSheet`, `Adw.Carousel` |
| not a widget / imperative | 3 | `Adw.Toast`, `Adw.AlertDialog`, `Adw.AboutDialog` |
| `Gtk.Adjustment` | 1 | `Adw.SpinRow` |

The three compound ones matter for scoping: a portable model does **not** unblock them on its own. `uncurated-placement` is gtk-host's descriptor refusal, and `scripts/adwaita-gallery-trees.mjs:557-561` says so — "a property of the descriptor table on `main`, not of the widget".

### The vocabulary ledger: 12 of 41 entries are list-shaped

`NS_PROPERTY_ALIGNMENT` in `scripts/check-vocabulary-alignment.mjs:448-589` — 41 entries, 14 `gir` (should converge) + 27 `own` (no counterpart key).

**List-shaped — 12:**
- `gir` (4): `adw-combo-row.options`→`model`, `gtk-drop-down.options`→`model`, `adw-tab-view.selected`→`selectedPage`, `adw-view-stack.visibleChildIndex`→`visibleChildName`
- `own`, collection is method-built (5): `adw-sidebar.items`, `adw-sidebar.sections`, `adw-tab-view.views`, `adw-tab-view.tabs`, `adw-toggle-group.options`
- `own`, selection/position read-back is read-only on GTK (3): `adw-combo-row.selectedValue`, `gtk-drop-down.selectedValue`, `adw-carousel.position`

**Menu-shaped — 3:** `gtk-menu-button.menuItems`→`menuModel`, `adw-split-button.menu`→`menuModel`, `gtk-menu-button.menuTitle` (`own`).

**Icon colour — 5, and it is NOT a data-model problem.** `adw-button-content.iconColor`, `adw-button-row.startIconColor`, `adw-icon.iconColor`, `adw-image-button.iconColor`, `adw-status-page.iconColor` are one fact five times: GTK colours a symbolic through the stylesheet, so no GIR type has the property. Nothing to converge towards, and no model would change it.

<sub>Incidental, since this file gates its own hand-kept counts elsewhere: there are **five** icon-colour entries, and two of them both claim to be the fourth — `adw-image-button.iconColor` says "Same as the other three" and `adw-status-page.iconColor` says "The fourth instance of the same fact".</sub>

### The sharpest cost is inside one surface, and needs no model to fix

`<adw-combo-row>` and `<gtk-drop-down>` are the same list widget on GTK (both `model?: Gio.ListModel | null`) and share one `ComboState` on the web. They do not share an authoring surface:

```ts
// packages/web/adwaita-web/src/elements/gtk-drop-down.ts:85
return ['options', 'items', 'selected', 'enable-search', 'disabled'];
// … with `set options(value: ReadonlyArray<AdwComboOptionInput>)` at :93 and `set items` at :104

// packages/web/adwaita-web/src/elements/adw-combo-row.ts:28
return ['title', 'subtitle', 'selected'];
// … no `options`/`items` accessor anywhere in the file; the list is read once, at :43
```

A caller who sets `.items` on an `<adw-combo-row>` after connect gets silence. That is a bug, and it is the concrete thing a user feels today.

### Carry

**Items — but there is no shared item vocabulary yet.** The web surface alone has **six** spellings of "an item": a named interface (`AdwSidebarItemSpec`, `AdwComboOption`, `AdwViewStackPageSpec`), a bare `string` (`ToggleGroupState.labels`, `packages/web/adwaita-core/src/rows.ts:451`), a union (`AdwComboOptionInput`), an index-signature bag (`AdwDataGridRow`, `packages/web/adwaita-core/src/data-grid.ts:69`), an opaque generic compared by pointer identity (`NavigationViewState<P>`), and a structural shape (`readonly { readonly visible: boolean }[]`, `packages/web/adwaita-core/src/view-switcher.ts:815`).

**A positional change signal.** Exactly one exists today, and it is the template — `packages/web/adwaita-core/src/tab-view.ts:163-172`:

```ts
export type TabViewPagesChangeKind = 'attached' | 'detached' | 'reordered' | 'pinned' | 'updated';
export interface TabViewPagesChange {
    kind: TabViewPagesChangeKind;
    id: string;
    position: number;          // index AFTER the change
    previousPosition: number;  // index BEFORE; -1 for 'attached'
}
```

Its doc says the point out loud: "a renderer subscribes to this to insert, remove or move exactly ONE tab instead of rebuilding the bar." Every other collection in adwaita-core answers a mutation with a full `_rebuild()` / `_render()` / `replaceChildren()`, and every NativeScript collection setter is a silent snapshot.

**Selection.** It cannot be left out, for three measured reasons: three ledger entries are selection read-back; the conformance suite already carries the cross-renderer rule (`packages/web/adwaita-core/src/conformance/rows.ts:219` — "a selection past the end of the new model falls back to 0 — GtkSingleSelection autoselects upstream […] this row is the ports' shared answer"); and on GJS `ListController` had to wrap its store in `Gtk.NoSelection` because `Gtk.ListView:model` is a `Gtk.SelectionModel` (`packages/framework/gtk-host/src/list/controller.ts:176-186`).

**Length / empty**, because the sidebar placeholder is computed from the filtered count.

### Leave out

**Sorting. Zero demand, measured.** Across all four surfaces there is no sorter implementation. The only hits are two generated GIR nicks (`GtkSorterChangeNick`, `GtkSorterOrderNick`, `generated/props.ts:465-466`) and one comment about DOM focus order.

**Filtering, as a model stage.** It exists on one widget, and its semantics are peculiar enough that generalising them taxes everything else — `packages/web/adwaita-core/src/conformance/sidebar.ts:470-473`: `adw_sidebar_set_filter` feeds a `GtkFilterListModel` that only the row list and the placeholder see, "the selection index space stays on the UNFILTERED `items_model`". Two index spaces is a sidebar rule, not a list rule.

**Virtualisation and item factories.** `ListController` already owns that on GJS, and it is a published subpath.

## 5. Cost and published-API risk

**Every surface in this survey is published on npm** (`0.46.0` live, `0.47.0` in tree): `@gjsify/adwaita-web`, `@gjsify/adwaita-core`, `@gjsify/adwaita-nativescript`, `@gjsify/adwaita-react-native`, `@gjsify/gtk-host`. Every accepted type quoted above is public API; none can be removed, only added beside.

Two specific risks:

1. **`@gjsify/gtk-host/list` is a published subpath**, and `ListController.setRows(rows: readonly Row[])` (`packages/framework/gtk-host/src/list/controller.ts:197`) is the one framework-neutral seam where a collection crosses into GTK. `@gjsify/react-native` consumes it (`packages/framework/react-native/src/lists/controller.ts:23`). Widening it is fine; changing what it accepts is not.
2. **The ledger already has the precedent for handling this**: `adw-password-entry-row.peeking` is kept, not deleted, "so existing NativeScript callers keep working while `revealed` is the shared name" (`check-vocabulary-alignment.mjs`). Add the convergent name, keep the alias, declare it.

In-repo call sites touching list properties: adwaita-core 69 (mostly conformance vectors), adwaita-web 26, showcases 22, nativescript 9, stories 3. Out-of-repo consumers cannot be counted from here.

**One structural obstacle to name rather than assume:** `@gjsify/gtk-host` does **not** depend on `@gjsify/adwaita-core` (its deps are `@girs/*` only) and imports it zero times, while adwaita-web, adwaita-nativescript and adwaita-react-native all do. Whether gtk-host should acquire that edge, or whether the shared vocabulary should live somewhere both can reach, is the first design decision — not something to settle by starting to type.

Related: the conformance gate knows exactly two renderers by construction — `scripts/check-adwaita-conformance-drivers.mjs:136-139` lists `adwaita-web` and `nativescript`. gtk-host and adwaita-react-native are outside it, so any shared vector table added here starts covering half the surfaces.

## 6. Proposed scope

**Do:**
1. Fix `<adw-combo-row>`'s items being write-once on adwaita-web. This is a bug with a same-surface control (`<gtk-drop-down>`), and it needs none of the rest.
2. Define the shared item vocabulary in `@gjsify/adwaita-core`, starting from `AdwComboOptionInput`, which three surfaces already accept.
3. Generalise `TabViewPagesChange` into the positional change signal the other collections lack, and give it conformance vectors — it is the only working instance of the thing and it is already shaped right.

**Do not:**
4. Invent a model type for Sidebar / TabView / ToggleGroup / ViewStack / Carousel. GTK has no list property for them; the gallery refusal is `uncurated-placement`, and the fix is a curated descriptor in `packages/framework/gtk-host/src/descriptors/` using the `ChildPolicy` kinds that already exist (`single`/`ordered`/`indexed`/`keyed`/`slotted`, 42 curated gtypes).
5. Carry sorting. Carry filtering only if a second widget ever needs it.

## Appendix: list vs menu, same yardstick

| axis | list | menu |
|---|---:|---:|
| vocabulary-ledger entries | **12** / 41 | 3 / 41 |
| gallery refusals (sole cause) | **5** (2) / 17 | 2 (2) / 17 |
| GTK widget interfaces with a writable model prop | 5 | **13** |
| NativeScript collection properties | **17** | 2 |
| surfaces carrying the widget | **4** | 3 — `adwaita-react-native` ships no menu widget at all |

The menu model wins on one axis, GTK breadth — and that is the axis that makes it cheap, not the one that makes it valuable.

<sub>Measured at `0ef1d7c137` in a read-only worktree. Every path:line above was read.</sub>

JumpLink:
Closes #1524.

`Gio.MenuModel` is a writable property on **13** GTK widget interfaces. A list `model` property exists on **5** — `AdwComboRow`, `GtkDropDown`, `GtkColumnView`, `GtkGridView`, `GtkListView`. For the widgets that carry most of the list divergence — Sidebar, TabView, ToggleGroup, ViewStack, Carousel — GTK has **no item property at all**: the collection is `adw_sidebar_append()`. So ADR 0046 is the sibling of 0042 and is deliberately narrower: one abstraction cannot span both, and a model type for the method-built half would put a GTK word on a value GTK does not have. That half is filed in `status/open-todos.md` with its measurement, not half-built.

## What the divergence actually was

Two things, and only one of them was the item shape.

**The item vocabulary was already shared** for the model-shaped widgets. `AdwComboOptionInput` is accepted by `@gjsify/adwaita-web`, `@gjsify/adwaita-nativescript` and `@gjsify/adwaita-react-native`; #1524's "six spellings of an item" are five spellings belonging to the METHOD-BUILT widgets this ADR does not cover. So it keeps its published names and moves to `list.ts` — a `AdwListItem` beside it would have been the seventh. What was missing is a name for the LIST those items form: `AdwListModel` / `AdwListModelInput`, which is what a `model` property holds and accepts.

**What no collection in this package had was a positional change signal.** `ComboState.setOptions` reported only that the SELECTION may have moved, so both browser selectors answered every model assignment by dropping every option node and building them again. The portable form is not ours to invent — `Gio.ListModel::items-changed (position, removed, added)`, with `g_list_store_splice`'s own stated reason: it "only emits items-changed once for the change". `listItemsChanged(previous, next)` computes that one minimal splice.

`TabViewPagesChange` — the sole existing instance of the shape — keeps its published type and gains a **mapping** onto it. `AdwTabView` is method-built, so it stays out of the model half; what it contributes is the evidence, and the mapping is what turns "it is the same signal" from a sentence into a test.

## Measured

**The splice is real, and here is the discriminator.** Replace `_applyItemsChanged` in both browser selectors with a full rebuild, rebuild the browser bundle, run it in Firefox:

```
adwaita-web: total=4900 passed=4891 failed=9
  FAIL [gtk-drop-down model] items-changed — an APPEND touches one position…
  FAIL [gtk-drop-down model] items-changed — a PREPEND is one insertion at 0…
  FAIL [gtk-drop-down model] items-changed — a removal from the middle…
  FAIL [gtk-drop-down model] items-changed — a RELABELLED item at an unchanged value…
  FAIL [gtk-drop-down model] a row that a splice shifted still selects itself
  FAIL [adw-combo-row …] items-changed — an APPEND touches one position…
  FAIL [adw-combo-row …] items-changed — a PREPEND is one insertion at 0…
  FAIL [adw-combo-row …] items-changed — a removal from the middle…
  FAIL [adw-combo-row …] items-changed — a RELABELLED item at an unchanged value…
```

Restored: `adwaita-web: total=4901 passed=4901 failed=0`. Those rows compare `<option>` / row-button node **identity** across an assignment, which is the only thing that separates a splice from a rebuild — no assertion about the model, the labels or the selection can see it. `LIST_ITEMS_CHANGED_VECTORS` carries a `survivors` figure per row for exactly that.

**The `<adw-combo-row>` write-once bug is NOT this PR's, and I am not claiming it.** It landed in #1525 (`cd6c9ab48`) with its own three-red-on-untouched-main proof, along with `check-adwaita-collection-reactivity`. Issue #1524 was measured at `0ef1d7c137`, before it. This PR inherits that fix and renames its door.

**Distance to one vocabulary**, printed by `scripts/check-vocabulary-alignment.mjs`, before → after:

```
… 145 settable propert(y|ies) between them — 104 already agree …, 41 do not (12 should converge, 29 declared own, 0 undecided).
Distance to one vocabulary: 2 widget name(s) and 12 property name(s)

… 145 settable propert(y|ies) between them — 106 already agree …, 39 do not (10 should converge, 29 declared own, 0 undecided).
Distance to one vocabulary: 2 widget name(s) and 10 property name(s)
```

`adw-combo-row.options` and `gtk-drop-down.options` are deleted from the ledger, not re-declared: `options`/`items` are **not** kept as aliases, because one of the three spellings is the GTK one and an entry that stays in the ledger is a distance that never closes.

**Test counts, from the runs I did** (`N completed` on this branch's CLI counts assertions, not tests — #1560 splits them):

| suite | before | after |
|---|---:|---:|
| `@gjsify/adwaita-core` (Node + GJS, each) | 12 886 | 12 946 |
| `@gjsify/adwaita-nativescript` (Node + GJS, each) | 3 537 | 3 551 |
| `@gjsify/adwaita-web` (Firefox, assertions) | 4 811 | 4 901 |

All green. `@gjsify/adwaita-react-native`: Node leg 478 green; the GJS leg is **13 of 298 failing, and it fails identically on the untouched baseline** — I reverted every file in this branch, rebuilt core and re-ran to check. All 13 are `uncurated-placement` refusals for `AdwToastOverlay`, `AdwExpanderRow` and `AdwWrapBox`, i.e. the `descriptors/` gap already ledgered as "138 of 164 generated widgets have no measured placement rule". Not mine, not fixed here.

**Every table is driven where I say it is, and only there.** `LIST_NORMALIZE_VECTORS`, `LIST_PARSE_VECTORS`, `LIST_ITEMS_CHANGED_VECTORS` and `LIST_SELECTION_CLAMP_VECTORS` run against the real `<adw-combo-row>` and `<gtk-drop-down>`, every row, no filter. They are **not** driven on NativeScript and this PR does not claim they are: that suite drives `ComboState` rather than the widget, for the reason its own header gives (importing the widget evaluates a bare `@nativescript/core` specifier, unresolvable on GJS and Node). One row is narrower than its vector and says so in place — `<adw-combo-row>` cannot HOLD a non-finite index, so the clamp's `NaN` row reaches the same outcome by another route there and is driven at the function in core.

`TAB_PAGES_ITEMS_CHANGED_VECTORS` is `CORE-ONLY:` with its chain named: both renderer tab-view suites replay the `TabViewPagesChange` stream **their real widget emitted** through `replayTabPagesAsSplices` and compare the result against the page order that widget ended on.

## The gate that went blind, and what it cost to notice

`check-adwaita-collection-reactivity.mjs` carried a KNOWN LIMIT in its own header: the attribute reader looked for a literal `JSON.parse(` inside a class member, so a parser that MOVED into core took the attribute out of its sight. **That limit came true in this PR** — moving both selectors' parsing into `parseListModel` made `attr:model` invisible at the exact moment rules 2 and 3 were meant to hold it. Fixing the census instead of the reader would have recorded the blind spot as a fact about the widgets.

`coreListParsers` DERIVES core's string→list functions (a `string` in, a list out through the existing one-alias hop) rather than listing them, because a hand-kept list fails the same way one rename later. Two more defects came out with it:

- a call on a **non-core receiver** was credited to the single core class declaring that method name — `this._menuView.setModel(…)` had `<gtk-menu-button>` reported as carrying `SplitButtonState.setMenuModel`, with a fix instructing it to publish a setter for a collection it does not own. Receiver resolution now keeps non-core fields, with **one forwarding hop** so `NsNavigationStack.replaceWithTags` (which really does reach a core state) is not thrown out with `PopoverMenuView` (which does not).
- rule 3's list-setter test was a bare bracket match, so every setter whose parameter is a NAMED alias — `AdwMenuInput`, `AdwListModelInput` — read as "list setters: [none]".

Net: both menu buttons are under rules 2 and 3 for the first time, `<gtk-menu-button>` has its first census entry, and the KNOWN LIMIT is retired rather than restated. `check-adwaita-collection-reactivity: OK — 7 core collection setter(s) over 8 collection(s); 15 on browser, 8 on NativeScript, across 19 widget(s)` (was 18).

## BREAKING

| published as | in | becomes |
|---|---|---|
| `ComboState.setOptions` / `.options` | `@gjsify/adwaita-core` | `setModel` / `model` |
| `{ op: 'setOptions', options }` | `@gjsify/adwaita-core/conformance` | `{ op: 'setModel', model }` |
| `options` / `items` (property **and** attribute) | `@gjsify/adwaita-web` `<adw-combo-row>`, `<gtk-drop-down>` | `model` |
| `options` (property) | `@gjsify/adwaita-nativescript` `AdwComboRow`, `GtkDropDown` | `model` |
| `AdwComboRowProps.model?: readonly AdwComboOptionInput[]` | `@gjsify/adwaita-react-native` | `AdwListModelInput` — assignment-compatible, no consumer edit |

`AdwComboOption` and `AdwComboOptionInput` keep their names and shape; only their module moved, and `rows.ts` re-exports both.

Two more fixes fell out and are in the diff rather than deferred: each drop-down row's click handler closed over the index it was BUILT at (harmless under a rebuild, wrong under a splice), and the NativeScript `model` setter took descriptors only — `model = ['a','b']` stored strings whose labels read back `undefined`. That NativeScript fix is held by core's vectors and by no widget test, for the import reason above; it is filed.

## Left open, deliberately

- **The GTK half.** ADR 0042 § 8 gave `coerce` a `GMenuModel` branch; the list needs the twin — an array into a `Gtk.StringList`. It is **not** here: `packages/framework/gtk-host/**` is being reworked by two other changes and a speculative edit would land as a conflict, not a feature. So `Adw.ComboRow` and `Gtk.DropDown` stay refused in the gallery, with half their reason gone — the refusals now name the seam ("nothing turns the portable list form into one at the ParamSpec seam") instead of the value. Filed as *A portable list model reaches every renderer except GTK*, with the three things that close it.
- **The method-built widgets.** Filed with the `generated/props.ts` evidence — no item list on `AdwSidebarProps`, `AdwTabViewProps` or `AdwToggleGroupProps` — and with why a curated `descriptors/` entry is the fix rather than a model type.
- **The NativeScript widget door**, per above.

## Files outside my scope

**None in `packages/framework/gtk-host/**` or `packages/framework/react-native/**`** — no descriptor, no `coerce`, no `ListController`. The only `packages/framework/**` files touched are `adwaita-react-native/src/{props.ts,widgets/combo-row.native.tsx,widgets/preferences.native.spec.tsx}`, which is my scope.

## Verified from the worktree root

`gjsify --version` 0.42.0 (the published bootstrap, ADR 0002), shim resolving inside this worktree. `gjsify run test` over the four packages; `tests/browser` Playwright/Firefox 6/6; `gjsify lint` (0 errors; no new warning in my four packages) and `gjsify format --check` clean; `node scripts/audit-runtimes.mjs --check`, `check-vocabulary-alignment`, `check-adwaita-collection-reactivity`, `check-adwaita-conformance-drivers`, `check-adwaita-element-properties`, `check-storybook-control-parity`, `check-generated-website-data`, `check-doc-fences`, `check-agent-context-size --check` all green. `check-refs-citations`, `check-type-surfaces`, `check-source-visibility` and `check-lint-visibility` fail on this machine **before** my first commit — an unchecked-out `refs/node`, `@girs/gtk-4.0` drift against the generated table, and an `oxlint --debug=files` probe that exits 1 here.

## Reader-facing docs

`adwaita/boxed-lists.mdx` and `gtk/controls.mdx` gained what an app author needs and nothing else: the two shapes the array takes, that a new model can be assigned at any time, that adding an item leaves the rest of the list alone, that the same array works on the drop-down, on NativeScript and in React Native — and, in their terms, that sidebar, tab view, toggle group, view stack and carousel have no `model`, with what to use on each instead.


JumpLink:
## Independent review — reproduced, then fixed on the branch (6 commits, `1a4c275..8d6e4f7`)

I did not write this PR. Everything below was measured in a worktree on this branch.

### The two things I was asked to reproduce

**The splice discriminator reproduces exactly.** Replacing `_applyItemsChanged` in both browser
selectors with a full rebuild, rebuilding `dist/test.browser.mjs` and running Firefox:
`adwaita-web: total=4900 passed=4891 failed=9` — the same nine rows the body lists, five on
`gtk-drop-down` and four on `adw-combo-row`. Restored: `4901/4901`. The splice path is
load-bearing and the `survivors` column is what sees it.

**The vocabulary numbers are derived, not asserted.** `check-vocabulary-alignment.mjs` prints
`106 already agree …, 39 do not (10 should converge, 29 declared own, 0 undecided)` and
`Distance to one vocabulary: 2 widget name(s) and 10 property name(s)` — the body's after-figures,
verbatim. `ADWAITA_ATTRIBUTE_COUNT = 276` is regenerated, not hand-edited
(`check-generated-website-data`: "output current"). No prose in the tree restates one.

### Defect 1 — the breaking rename missed **eight** call sites, and one of them is your red CI

`fix: write model on every list call site`

Grepping the old names across the whole tree (`.ts/.tsx/.mjs/.mdx/.astro/.xml/.md`, refs and
build output excluded) finds eight callers still on `options`/`items`:

| where | shape | consequence |
|---|---|---|
| `packages/nativescript-bridge/storybook/src/controls.ts:130` | `row.options =` | **TS2339 — this is the `Node-pillar suites (win32-x64)` failure on #1566** |
| `showcases/dom/adwaita-storybook-nativescript/src/{feedback/preferences-dialog,overview/widgets,rows/preferences-group}.ns.ts` | `x.options =` | silent; that showcase is excluded from the npm workspaces (its own `typescript ~5.4`), so nothing type-checks it |
| `packages/web/adwaita-storybook/src/controls.ts:112` | `setAttribute('items', …)` | the storybook **select control** renders an empty combo row |
| `showcases/gtk/adwaita-storybook/src/browser/{feedback/preferences-dialog,rows/preferences-group}.web.ts` | `setAttribute('items', …)` | empty combo row in two stories |
| `showcases/gtk/adwaita-storybook/src/browser/overview/widgets.web.ts:62` | `el('adw-combo-row', { items })` | empty combo row in the overview |

`observedAttributes` is now `['title','subtitle','model','selected']`, so `items` reaches nothing —
no error, no type, just an empty widget. The four attribute sites are exactly the class the brief
warned about.

On the win32 leg: **the `spawnSync git ENOENT` in that log is not the failure.** It is
`scripts/install-git-hooks.mjs` reading `core.hooksPath`, the installer tolerates it
("git hooks may not be active"), and it appears identically in the **green** *Windows suites* run
on `main` (`33901958564`). The job continued for eight more minutes and died on
`src/controls.ts:130 — Property 'options' does not exist on type 'AdwComboRow'`.

Why it was invisible locally: `packages/nativescript-bridge/adwaita/lib/types/` was a stale build
output still declaring `get options()`. The type check only failed here after
`gjsify workspace @gjsify/adwaita-nativescript run build`. The mechanism worked; the artefact lied.

### Defect 2 — a false coverage claim in the tree, of the exact shape #1072 is about

`fix(adwaita-core): retire a false coverage claim`

`LIST_NORMALIZE_VECTORS`' docblock said every surface taking a `model` "must answer these rows:
the browser elements through their `model` property, **the NativeScript widgets through theirs**".
No NativeScript spec drives it — that suite reaches `ComboState` and never the widget class, which
this branch's own open-todos entry states. Every claim arm of `check-adwaita-conformance-drivers`
begins at a `*_VECTORS` citation, and that sentence named no table, so nothing looked at it.

The three list tables now name the suite that drives them, which puts the claims **under arm 9**:
swapping "browser" for "NativeScript" there fails with
`credits the nativescript suite with LIST_NORMALIZE_VECTORS, which it does not drive`.
Single-suite claims resolved against the tree: **18 → 21**.

Two related corrections to the body, which overstates coverage in the same direction: of the four
tables, only `LIST_NORMALIZE_VECTORS` and `LIST_ITEMS_CHANGED_VECTORS` ran on **both** selectors.
`LIST_PARSE_VECTORS` and `LIST_SELECTION_CLAMP_VECTORS` ran on `<adw-combo-row>` alone.

- `test(adwaita-web): drive the parse table on both` — the parse table now runs on
  `<gtk-drop-down>` too. Both read the attribute through one `parseListModel`, but the half that
  breaks is each element's own `attributeChangedCallback` branch, which is precisely what the row
  did not have (#1525). Firefox: **4901 → 4906**, green.
- The clamp table stays on the row, and the header now says why: `<gtk-drop-down>` refuses an
  out-of-range `selected` set outright, so the starting index those rows replace cannot be put on
  that element at all.

### Defect 3 — the class had no reader, so I gave it one where it was cheap

`test: fail on a preview attribute nothing observes`

`check-website-attr-samples` walks the *observed* names and looks for each value in the fence — so
an attribute the fence writes that the element does **not** observe is compared to nothing. The
incident was already written down one file away, in `AdwWidgetWindow.astro`: the boxed-lists
`<adw-combo-row>` preview wrote `items`, which the row never observed and its `connectedCallback`
read anyway.

Arm 3 is that direction. Two exceptions are the whole false-positive surface — the platform-global
names, and the six elements that observe nothing because a parent reads their attributes
(`<adw-view-switcher-page>`). **Zero findings over the 39 shipped blocks**; restoring
`items='…'` in `boxed-lists.mdx` fails it by name. Two fixtures hold the rule itself.

I did **not** build the reader for the imperative shape that four of the eight misses take. Scoped
to two directories it finds all four with zero false positives; run tree-wide it is 384 findings of
which about ten are real, because `gtk-host` JSX and the `*.gtk.tsx` widgets spell the same tags
with a different vocabulary. A gate whose scope is a hand-kept directory list goes blind the first
time a fifth showcase appears — the failure mode `coreListParsers` was just rewritten to avoid.
Filed with the measurement in `status/open-todos.md`
(`docs(status): file the attribute-rename reader gap`).

### Defect 4 — four quotations the rename invalidated

`docs: refresh text the model rename invalidated`

ADR 0034 and `adwaita-react-native/src/props.ts` each quote a gallery refusal string **verbatim**,
and this branch rewrote both strings; the adwaita-web README still called the state an "options
list"; a comment in `gtk-drop-down.ts` still named `setOptions`. A quote that no longer matches its
source is what a reader greps for and does not find.

### What I checked and found sound

- **The gate repair is a repair, not an accommodation.** `coreListParsers` derives the string→list
  functions from core instead of listing them, which is what makes the retired KNOWN LIMIT stay
  retired. Receiver resolution keeping non-core fields with one forwarding hop is right and both
  answers are in the tree (`NsNavigationStack.replaceWithTags` vs `PopoverMenuView`). Rule 3's
  alias-aware setter test is real: `AdwMenuInput`/`AdwListModelInput` read as "[none]" before.
  Census is exact-matched, so a lost collection fails rather than passes:
  `7 core collection setter(s) over 8 collection(s); 15 on browser, 8 on NativeScript, across 19 widget(s)`.
- **Scope claim verified.** `packages/framework/gtk-host/**` and `packages/framework/react-native/**`
  are untouched; the only `packages/framework/**` diff is the three `adwaita-react-native` files
  named in the body.
- **Governance.** ADR 0046 precedes the implementation; three deferred items are `### ` entries in
  `status/open-todos.md`; nothing was struck through instead of deleted; `packages/web/AGENTS.md`
  updated with its budget line; `status/status.json` carries no claim the rename touched.
- **The react-native GJS leg (13 of 298) is baseline.** All 13 are `uncurated-placement` refusals
  for `AdwToastOverlay`/`AdwExpanderRow`/`AdwWrapBox`, nothing this PR touches.
- **Docs read as a user.** Every snippet on the two changed pages runs: `appendPage()` and
  `sections` exist on the elements the note sends readers to, `selectedValue = 'Teal'` resolves
  through the attribute, and the append line really does leave the two standing options alone
  (that is the `survivors: 3` row).

### Left, deliberately, and one for the author

- **The three omissions are genuinely separate decisions.** The GTK `coerce` branch is a
  `gtk-host` change two other PRs are moving underneath; the method-built widgets are a
  `descriptors/` gap that no model type unblocks; the NativeScript widget door is closed by
  module-evaluability, not by this ADR. All three are ledgered with their measurement.
- **`ComboState.model` still hands out the internal array, and the splice makes that sharper.**
  `const m = row.model; m.push(x); row.model = m;` used to work under the rebuild and is now a
  silent no-op — `listItemsChanged` compares the mutated array with a copy of itself and finds no
  change. The docs correctly say to assign a new array, and every in-tree caller does. Fixing it
  properly means copying in the getter, which changes a published accessor's identity and
  allocation behaviour — **your call, not mine**, so I have not touched it.

### Verified from the worktree root, after the fixes

`gjsify --version` 0.42.0, shim resolving inside this worktree
(`/home/jumplink/Projekte/gjsify-wp-list/packages/infra/cli/…`).
`gjsify run test` — adwaita-core 12 946 (Node + GJS), adwaita-nativescript 3 551 (Node + GJS), all
green. `tests/browser` Firefox 6/6, adwaita-web **4906/4906**, adwaita-core 12 821/12 821.
`gjsify lint` 0 errors, `gjsify format --check` clean, `audit-runtimes --check` ok, and green from
`check-website-attr-samples` (13 fixtures, 190 cells, 39 blocks, 0 unobserved),
`check-adwaita-conformance-drivers`, `check-adwaita-collection-reactivity`,
`check-vocabulary-alignment`, `check-generated-website-data`, `check-adwaita-element-properties`,
`check-website-adwaita-gallery`, `check-storybook-{control-parity,widget-coverage,story-parity}`,
`check-doc-fences`, `check-adr-index`, `check-agent-context-size`, `check-adwaita-upgrade-order`,
`check-nativescript-xml-doors`, `check-{browser,node}-test-registration`,
`check-changelog-references`. The four scripts the body names as failing on this machine
(`check-refs-citations`, `check-type-surfaces`, `check-source-visibility`, `check-lint-visibility`)
fail here for the reasons it gives and are unrelated to the win32 red.

JumpLink:
## Follow-up: the model accessor is fixed here after all (`30d2571`)

I flagged this as the author's call and was told it is this PR's to fix — correctly, and the
reasoning was better than mine: it is a **regression this PR introduces** (the pattern worked
under the full rebuild), the PR is already `!`, and copying does not document the hazard, it
removes it. So it is in.

### What changed

`ComboState.model` hands back a copy, **items included**, and `setModel` stores one — the class
shares nothing across its boundary in either direction. The `model` getters on
`@gjsify/adwaita-web` and `@gjsify/adwaita-nativescript` forward to it, so they inherit it.

Fresh items and not just a fresh array, because the same failure sits one level down:
`m[0].label = 'x'` through a shared descriptor mutates the stored one too, and
`listItemsChanged` compares by **both** halves. Freezing the items would have made the mutation
loud; the copy makes the round trip correct, which is what was asked for.

### Held by vectors, A/B per copy

`LIST_MODEL_OWNERSHIP_VECTORS` — five rows, driven at `ComboState` **and** at both real browser
elements. The DOM assertions are what make the element rows worth having: without the copy the
*model* still reads correctly on the mutate-then-assign rows (it is `_options` that was mutated),
and only the `<option>` / row-button list shows that no splice was ever emitted.

Each copy removed **alone**, so neither masks the other:

| removed | red |
|---|---|
| the **getter** copy | 4 of 5 rows, on all three surfaces — core 4, `<adw-combo-row>` 4, `<gtk-drop-down>` 4 |
| the **setter** copy | the 5th row, **at core only** — the element setters run `normalizeComboOptions` and close that door themselves |

That asymmetry is the fifth row's whole point, and it is why the row exists rather than being
folded into the other four.

### Cost — measured, not assumed

Node 24.19, `{value,label}` records, `ComboState.model` read in a loop:

| n | read | assignment beside it |
|---:|---:|---:|
| 5 | 24 ns | 81 ns |
| 20 | 79 ns | 116 ns |
| 100 | 365 ns | 427 ns |
| 1 000 | 4.3 µs | 6.3 µs |
| 10 000 | 42 µs | 67 µs |

Linear, about 4 ns per item, and cheaper than the assignment it accompanies at every size. **No
caller reads it per frame.** Three read it per *item*, and those are now hoisted:
`_applyItemsChanged` in each browser selector (a splice of n items was about to become O(n²)) and
`_filter` in `<gtk-drop-down>`, which runs per row **per keystroke** of the search field — that
one was already a wasted property read per item before this change. A model large enough for the
linear term to matter is a `Gtk.ListView`, which is `ListController`'s virtualisation and ADR
0046 § 7 keeps out of this class on purpose.

### Recorded

- **ADR 0046 § 3** now states the ownership rule and why deciding by comparison forces it, with
  the A/B above.
- **The BREAKING table** has its fifth row: *"`ComboState.model`, and every `model` getter
  forwarding to it, returned the STORED array → returns a COPY, items included. Same type, same
  contents; what changes is IDENTITY."*
- **Consequences** carries the "why here and not in a follow-up" argument, in the form you gave
  it: a release with the splice and without this is one where the round trip compiles, runs and
  does nothing.
- The commit is `fix(adwaita)!:` with a `BREAKING CHANGE:` trailer, so it reaches the changelog
  on its own rather than only through the PR title.

### The website pages

Nothing there taught "assign a new array" as a *workaround* — it is the API, and the page says so
as style. So no sentence had to go. One was **added**, because the copy changes what a reader can
expect and it is theirs to know:

> Reading `row.model` hands you a copy of the list, so you can build the next one out of it —
> push, filter, rename an item — and nothing on screen moves until you assign it back.

### Not done, and why

I did **not** add the rule to `packages/web/AGENTS.md`. That file's `adwaita-core` row already
points at ADR 0046 by number on the very line that describes `list.ts`, the rule is five machine-
checked rows away from anyone who tries to optimise the copy out, and the file's own budget
section says a second copy of a rule is the one that drifts. Say the word if you want it there
and I will add it with the `agent-context-budget.json` line in the same commit.

### Verified after the change

adwaita-core **12 956** ×2 (Node + GJS), adwaita-nativescript **3 551** ×2, adwaita-react-native
Node 478 green with the GJS leg still the untouched 13/298 `uncurated-placement` baseline (same
three widgets). `tests/browser` Firefox **6/6**, adwaita-web **4936/4936** (4906 → 4936, the 30
new ownership assertions), adwaita-core browser 12 831/12 831. `gjsify lint` 0 errors,
`format --check` clean, and green from `check-adwaita-conformance-drivers` (**172** tables now,
22 single-suite claims resolved), `check-adwaita-collection-reactivity`,
`check-vocabulary-alignment`, `check-generated-website-data`, `check-website-attr-samples`,
`check-adr-index`, `check-agent-context-size`, `check-doc-fences`.

One measurement worth stating rather than leaving as a surprise: `check-comment-budget` exits 1,
and it does so on **`main`** for nine trees. Measured in a throwaway worktree at `main`, at the
author's tip `059033868` and here: `showcases` is 276 over and `website` 161 over at **all three**,
identical, and `scripts` went 78 (main) → 116 (author's tip) → **114** (now), so this review round
took two lines off it rather than adding any. The gate is advisory by the root AGENTS.md's own
description ("a score, so it advises"), and nothing here moved it.

Issue #1555, opened 2026-09-04
react-native: PROPS.md omits the per-value refusals


JumpLink:
`PROPS.md` is generated from `PRIMITIVES` through `primitives/answers.ts`, and ADR 0039
§ 2's whole point is that the answer a reader gets and the answer the resolver throws are
the same string. For `PropertyRoute.refuses` — the per-value refusal added in #1541 —
they are not the same string, because one of them does not exist.

## What the generated document says

```
| `accessibilityRole` | property | `accessible-role` | — |
```

## What the layer actually says

```
<View accessibilityRole="keyboardkey">
→ PrimitiveError: prop "accessibilityRole" = "keyboardkey" — describes a key of an
  on-screen keyboard, and neither GTK nor ARIA has a role for one — a desktop keyboard
  is the input method, not a widget in your tree. If you are building a keypad, the keys
  are `accessibilityRole="button"`; the container needs no role, because React Native has
  no spelling for ARIA's `group` and a plain `<View>` is already the `generic` a screen
  reader walks through
```

Seven role names are answered like that — `keyboardkey`, `summary`, `scrollview`,
`horizontalscrollview`, `drawerlayout`, `slidingdrawer`, `pager` — and **none of them is
reachable except by rendering one and catching the throw**. That is the state ADR 0039
§ 1 exists to end ("a consumer could not ask *does this layer take `onPress` on `Text`?*
without mounting a tree and catching a throw"), reintroduced for the newest and most
carefully written refusals in the table.

## Cause

`answerForRoute()` in `primitives/answers.ts` returns `why: ''` for
`case 'property'`. That was correct while a property route had nothing to say; a route
carrying `refuses` does.

## Suggested shape

`PropAnswer` grows the per-value refusals — as a field rather than folded into `why`, so
`explainProp` can be asked about one value and `PROPS.md` can render them as a nested
list under the row. The `why` column for a mapped prop is then the place the seven
sentences live, and the generator needs no new source: it already imports the table.

Worth stating in the same change: the row above claims `accessibilityRole` is answered,
which is true for 33 of React Native's 40 role names and false for 7. A reader of
`PROPS.md` today cannot tell which.

Same treatment applies to `AccessibleMembersRoute.refuses` (`accessibilityState`'s
per-key refusals), which is empty today and will not stay empty.

JumpLink:
Fixed on `main` in 349d1fda2 (#1567), commit `feat(react-native): publish the per-value refusals`.

[...]

Issue #1551, opened 2026-09-04
AppRegistry's live handle assumes exactly one runApplication, and never says so


JumpLink:
`packages/framework/react-native/src/app-registry.ts` keeps the running application in a module-scoped `live`, written in `createWindow` and cleared on the way out:

```ts
const code = await runAdwaitaApp(toShellOptions(options, (app) => { … live = { app, window, content, root }; … }));
const mounted = live;
live = null;
mounted?.root.unmount();
```

The comment says "this function is the only writer of `live`, so it is also the one that clears it". That is true of the *function* and not of a *call*, and the difference is two reachable defects.

## 1. A second concurrent call cross-unmounts

Two `runApplication` calls in one process — different keys, different application ids — both write `live`. Whichever resolves first takes `mounted = live`, which by then is the *other* call's handle, and unmounts a React root belonging to an application that is still running. The window stays up and empty.

Not exotic: `runAdwaitaApp` documents the single-instance handoff path explicitly ("brought the existing instance to the front instead of opening a second window"), so a second call returning promptly without ever having built a window is an expected shape, and it lands in the same clear.

## 2. A rejected launch leaks the handle

If `runAdwaitaApp` rejects after `createWindow` ran, `live` is never cleared, and `getApplication()`/`getWindow()` keep answering with a dead application — the state the code's own comment calls "worse than answering `null`".

I could not construct a throw path for (2) that survives GJS swallowing exceptions inside GObject handlers, so this half is reasoned, not measured. It is recorded because the fix for (1) covers it anyway.

## Shape of the fix

Own the handle per call rather than per module: capture the handle the closure produced, and clear only if `live` is still that handle.

```ts
let mine: LiveRoot | null = null;
try {
    const code = await runAdwaitaApp(toShellOptions(options, (app) => { … live = mine = { … }; … }));
    return code;
} finally {
    if (live === mine) live = null;
    mine?.root.unmount();
}
```

Whether a second concurrent `runApplication` should instead be *refused* by name — the way `RouterRoot` refuses a second mount with `RouterError` — is the design question worth settling first. A refusal is probably the better answer: two React Native applications in one GJS process have no defined meaning here, and a named error beats a silently correct-looking clear.

Found while reviewing #1538, which introduced the accessors. Not a regression it caused — there was no handle to leak before — and not fixed there because it needs a vector, and `runApplication` is only observable by running the loop (see #1550).

JumpLink:
[...]

Issue #1547, opened 2026-09-04
Two per-route caches in the stack navigator are never pruned


JumpLink:
Two maps in `packages/framework/react-native/src/router/` are keyed by route key, written
during render, and never cleaned up. They share one shape, which is the useful part: a
cache keyed on something the router creates a new value of on every navigation, with no
edge back to the release path the pages already have.

**`useTitleSlots` — `slots.current`** (`src/router/chrome.ts`, added in #1540).
`slotFor(key)` caches one `ChromeSlot` object per route key and the map is only ever
written. The caching is deliberate and correct — an unstable slot identity would re-run
every contributor's effect on each commit, which for that consumer means clearing and
re-setting the contribution in a loop — but nothing ever deletes an entry.

**`StackView` — `optionsRef.current`** (`src/router/stack.ts`, pre-existing).
`for (const descriptor of pages) optionsRef.current.set(descriptor.route.key, ...)`. Held
across renders on purpose: `reconcilePopped` runs from a signal handler, outside any
render, and needs the options of a page that may already have left `descriptors`. Same
absence of a delete.

## Size

Small and bounded by distinct route keys, not by navigations to the same route — but route
keys are not the route *names*: React Navigation mints a fresh key per pushed instance, so a
long-lived application that pushes `detail/[id]` a few thousand times holds a few thousand
entries in each. One small object and one options record apiece. Not urgent; it is a leak
with a slope, not a wall.

## The shape, and why one issue rather than two

`usePageBindings` already solves exactly this for its own per-key cache: the pages have a
release path (`release(key)`, driven by the page's `hidden` signal plus the unmapped
sweep), and the fix for both maps is to hang a delete off it rather than to invent a second
lifecycle. Doing them together is what makes that visible; doing them separately would
produce two ad-hoc prunes.

Found while reviewing #1540. Neither is a defect that shows on screen.

JumpLink:
Fixed on `main` in 349d1fda2 (#1567), commit `fix(react-native): prune the router's per-route caches`.

One `PerRouteCache` lifecycle in `router/per-route-cache.ts` replaces the three ad-hoc maps, pruned from a single layout effect over the page list and cleared on unmount — driven off the page list rather than `release(key)`, so a route removed from the middle of the stack is pruned too. `perRouteCacheEntries()` is the instrument: measured with the three `retain` calls removed, 27 entries still held after popping back to a one-page stack, against 3.

[...]

PR #1567, opened 2026-09-04
fix(react-native): deep links, per-route caches, and the bootstrap's vectors

### LINKED ISSUE #1453: Deep-linking to a non-index tab loops between React and the ViewStack
`Tabs` renders `Adw.ViewStack` behind an `Adw.ViewSwitcher`, and the mapping is the right one — the switcher is driven by the stack's own page model, so a route file adds a button with no tab-bar bookkeeping. It works when the application starts at the tab group's index route. It does not survive being deep-linked to a non-index tab.

## Measured

An application with five tabs, entered by URL:

| entry | result |
|---|---|
| the index route | all five tabs mount, **zero** errors |
| one non-index tab (`/entdecken`) | deep-links cleanly |
| three other non-index tabs | **loop** |

The loop is React error **#185** (a state update in an unbounded cycle) together with, roughly 30 times a second:

```
gtk_widget_set_child_visible: assertion 'GTK_IS_WIDGET (widget)' failed
```

432 occurrences in 15 seconds. The screens themselves are fine — the same three render correctly when reached by tapping from the index route. What does not survive is the page selection in the `ViewStack` when the initially focused page is not the first one.

The GTK assertion says the widget being made visible is not a widget yet, which points at ordering: the stack is asked to select a page before that page's child has been inserted, React sees the resulting state disagree with its own, and the two drive each other.

## Why the shape matters more than the count

Entering at a non-index tab is what a **deep link** is, and a desktop application gets them from the same places a phone does — a URI handler, a notification, a restored session. So this is not an edge case reachable only from a test; it is the ordinary path for every entry point that is not the launcher.

That two of five tabs *do* work is the useful part of the report: whatever it is, it is not "the switcher cannot be entered from outside".

JumpLink:
Closes #1453, #1547, #1546, #1549, #1551, #1555.

Six open defects in `@gjsify/react-native`, four of them the same shape from different angles: a mechanism that was correct, and pointed at nothing that ships.

## A deep link did not survive a non-index tab (#1453)

`<Tabs>` read every `notify::visible-child-name` as a user press. `Adw.ViewStack` emits that one signal for three different events and only one of them is a click: the user's, the layer's own `set_visible_child_name`, and libadwaita's pick of a first page while the reconciler is still INSERTING them (`add_page` selects a page whenever the stack has none). The only thing standing between the third and a dispatched navigation was the React ref not being attached yet.

The echo guard asked the CONTAINER — "does this name already equal the focused route" — and `BaseNavigationContainer` publishes its state from an effect, so during the commit that shows the deep-linked page it still holds the route the render has already left.

**Measured**, on a five-tab `(tabs)` group under a root stack, entered by URL at each tab:

| entry | `JUMP_TO` actions the container was asked for, before → after |
|---|---|
| `/t0` (the index tab) | 0 → 0 |
| `/t1`, `/t2`, `/t3`, `/t4` | **1 → 0**, each |

`4 of 2254 tests failed` before the fix, `0 of 2254` after. That the invented jump named the *right* tab in-process is timing: the same lag with the stack still on `add_page`'s pick dispatches the FIRST page instead, which is the deep link the reporter watched disappear.

The guard is now this layer's own record of what it asked for, written *before* the call because the notify is emitted from inside it. A real switcher press still produces exactly one `JUMP_TO` — asserted in the same block, because a guard that closes the bridge would pass the first assertion.

**Left open, with what was measured:** the `Gtk-CRITICAL`/SIGSEGV half of #1453 is a libadwaita defect this layer can trigger and it is NOT closed. Read against `refs/libadwaita/src/adw-view-stack.c`, `stack_remove` clears a page's widget without clearing `self->last_visible_child`, so a later `adw_animation_skip` calls `gtk_widget_set_child_visible` on a NULL or freed widget — which is the reporter's backtrace exactly. It did not reproduce in-process: these vectors never remove a `ViewStack` page one at a time, because React unmounts a deleted subtree from its top. Also measured against the issue's own suggestion: `Adw.ViewStack:enable-transitions` is **not** the escape — `false` sets the animation duration to 0 and `adw_animation_play` still schedules a tick, so the in-flight window survives. Recorded in `status/open-todos.md` with what is left to settle.

## Two per-route caches were three, and none was pruned (#1547)

The title slot per header bar, the screen options per page, and — the one the issue did not name — the ref/`hidden` binding pair per page. All keyed by route key, and a route key is not a route NAME: React Navigation mints a fresh one per pushed instance, so pushing `detail/[id]` a few thousand times held a few thousand entries in each.

One lifecycle rather than three ad-hoc deletes: `PerRouteCache`, pruned from a single layout effect over the page list the closing-page bookkeeping already maintains, and cleared when the navigator unmounts. Driving it off the page list rather than off `release(key)` is deliberate and wider — a route removed from the MIDDLE of the stack was never on screen, so no `hidden` ever arrives for it.

**Measured** with the three `retain` calls removed, on the new vector: **27** entries still held after pushing eight detail pages and popping back to a one-page stack, against **3** with them. `perRouteCacheEntries()` is the instrument, for the reason `announce.ts` keeps one: a pruned cache and an unpruned one render identically.

## `provideWindowChrome` had no vector where it ships, and `windowChromeProblems()` had no shipping caller (#1549, #1546)

Two sides of one thing, fixed as one. `runApplication` cannot be entered from a spec — **measured**: a nested `g_application_run` inside the test runner's own main loop never returns (`g_application_run: assertion '!application->priv->must_quit_now' failed`, case timed out at 5 s). So the bootstrap is split where a vector can reach it, and both halves are exported: `mountApplicationRoot` and `ownTheApplication(appKey, launch)`, **whose launcher is a PARAMETER** — this package's own rule — so the vector drives the shipping code rather than a copy of it.

**Measured** by deleting `provideWindowChrome(chrome, …)` from the render call, the shape a careless rebase produces:

| | with the call | without it |
|---|---|---|
| `windowChromeCensus().headerBars` | 1 | **2** |
| `windowChromeProblems()` | `[]` | `["2 sets of window controls draw at the end … 1 of those close buttons close nothing the user is looking at (2 mapped header bar(s))"]` |
| the suite | 2542 completed, 0 failed | **6 of 2533 failed** |

Four of those six are collateral and worth naming, because they are the reason the vectors destroy their window in a `finally`: `Gtk.Window.get_toplevels()` is GTK's own live list, and a window a failing assertion leaves behind fails four `Dimensions` cases in `apis.spec.ts` instead of the case that caused it.

**The #1546 decision is option A, and it is written down in ADR 0043's amendment**: `AppRegistry` runs the check on its own window, once, on the idle after `map`, ungated, with the result readable as `lastWindowChromeProblems()`. Once and not per commit because the invariant is about the RESTING composition — `Adw.NavigationView` keeps a departing page mapped while the arriving one slides in, so a per-commit check would report a defect for every push. At map time no transition is in flight, and it is when the defect it was built for surfaces: #1460 was an application that OPENED with two close buttons. Option B — a driveable `@gjsify/devtools` method for a composition that breaks *later* — is not taken and not dead; it is in `status/open-todos.md` with what A leaves open.

## `AppRegistry` assumed exactly one `runApplication` (#1551)

The comment said "this function is the only writer of `live`, so it is also the one that clears it". True of the FUNCTION, not of a CALL. A second concurrent call is now refused BY NAME, which is the design question the issue left open: this layer CREATES the application and a process has one, so two React Native applications in one process have no defined meaning and a named error beats a clear that only looks correct. `ownTheApplication` claims the slot before the await and releases everything in a `finally`, so the rejected-launch half — reasoned in the issue, unreachable through `runAdwaitaApp` because GJS swallows exceptions inside GObject handlers — is one line with the launcher as a parameter, and has a vector.

## `PROPS.md` omitted the per-value refusals (#1555)

`accessibilityRole` rendered as `| property | accessible-role | — |`, which reads as "every value lands". Seven of React Native's forty role names are answered by name instead, each with a sentence saying what to write, and none was reachable except by rendering one and catching the throw — the state ADR 0039 § 1 exists to end, reintroduced for the newest refusals in the table.

`PropAnswer` grows `refuses` (a field, not a fold into `why` — the prop is accepted and only some values are not), keyed by the VALUE for a mapped property route and by the KEY for an accessible record, so `AccessibleMembersRoute.refuses` is carried too rather than waiting for its first entry. `prop-table` gains `acceptsPropValue`, `explainPropValue` and `propRefusedValues`; the subject is built there because `answers.ts` may hold no relative value import, so `describeValue` moved to `errors.ts` and both the throw and the static answer call it.

**Measured**: `grep -c keyboardkey PROPS.md` is **0** before and **37** after; the document gained 205 lines. The generator is the fix, not the document — reverting `generate-exports.mjs` alone makes `node scripts/check-rn-surface.mjs` red on the byte comparison.

## Measured here

Every number above came from a run on this branch. Every fix was verified RED before it was verified green, by reverting the fix and keeping the vector.

- `@gjsify/react-native` suite: **807 tests / 2542 assertions, 0 failed** (baseline on `main`: 756 tests / 2231 assertions)
- `node --test tests/e2e/react-native-gate/run.mjs` — 9 pass, 0 fail
- `node scripts/check-rn-surface.mjs`, `node scripts/audit-runtimes.mjs --check`, `node scripts/check-vocabulary-alignment.mjs`, `node scripts/check-doc-fences.mjs`, `node scripts/check-agent-context-size.mjs --check` — all exit 0
- `gjsify format --check` (3700 files) and `gjsify lint` — 0 errors
- `gjsify workspace @gjsify/react-native run check` (tsc) — 0

## Not measured, and said plainly

- The GTK crash class in #1453, above. Reasoned from the libadwaita source and the reporter's core dump; not reproduced by these vectors.
- `registerBuiltinWidgets()` inside `runApplication` is still guarded by nothing CI runs, and so is everything past `toShellOptions` — those need the loop, which needs the ledgered `react-native-devtools` suite to have a host. That entry in `status/open-todos.md` is updated rather than deleted, because the CI half is still open.
- `packages/framework/gtk-host` fails `build:types` in this worktree on pre-existing `@girs/*` skew in files this branch does not touch (`generated/props.ts` wanting `Gtk.CellEditable.SignalSignatures`). Not investigated; it is not this branch's and not this branch's scope.

## Scope

Nothing outside `packages/framework/react-native/**` was touched except the four places a decision has to be recorded: ADR 0043 (amended), `packages/framework/AGENTS.md` (its react-native row), `status/open-todos.md` and `status/agent-context-budget.json`. `support-table.ts` is untouched, so #1561's move to react-native 0.87.1 does not conflict here; the `Modal` entry and everything under `packages/framework/gtk-host/**` were left to the portal-seam branch. `website/src/content/docs/frameworks/react-native.mdx` is edited in its routing, AppRegistry and prop-table sections only.

**One collision to expect:** `packages/framework/AGENTS.md` was 50 bytes under the 32 KiB cap, so recording the two decisions meant paying for them — three live counts in the same row are gone, which the repo's own rule asks for ("write what the number establishes, not the number"). `status/agent-context-budget.json` moves 32718 → 32763, and #1560 edits that file too.


JumpLink:
Independent review — I did not write this branch, and I have pushed six commits onto it.

## What I reproduced, and it holds

Every number I spot-checked was reproducible by reverting the fix and keeping the vector.

- **#1453** — removed the `shownRef` guard line alone: `/t0` produces 0 `JUMP_TO`, `/t1`–`/t4`
  exactly 1 each, `4 of 2542 failed`. And the discriminator discriminates: replacing the guard
  with an unconditional `return` — a bridge that is *closed* rather than guarded — turns all five
  deep-link cases green and takes both `still follows a REAL press` and `follows the USER when the
  switcher changes the visible child` red. Both halves are load-bearing.
- **#1547** — removed the three `retain` calls: the vector goes red, and `pushed - rested === 24`
  still passes. That assertion is what proves the instrument counts all THREE maps (8 pages × 3)
  rather than two, so the ref/`hidden` binding pair really is in the count.
- **#1555** — reverted `generate-exports.mjs` alone: `check-rn-surface.mjs` red on the byte
  comparison, so the fix is in the generator. Regenerating from a clean tree reproduces `PROPS.md`
  byte for byte (identical md5, empty `git status`). `grep -c keyboardkey` is 0 → 37. One number
  to correct: the document is +205/−20, i.e. net **+185** lines, not 205.
- **#1549** — deleted `provideWindowChrome(chrome, …)`: `headerBars` 1 → 2 and
  `windowChromeProblems()` returns the sentence the body quotes, word for word. 2 failures here
  rather than 6 — the four `Dimensions` collateral is gone, because the vectors destroy their
  window in a `finally`.
- **libadwaita** — read `refs/libadwaita/src/adw-view-stack.c` myself rather than taking it.
  `stack_remove` clears `visible_child` and never `last_visible_child`; `transition_done_cb`
  dereferences `last_visible_child->widget`; `play()` adds a tick callback for everything except a
  globally-disabled or unmapped widget, so `enable-transitions: false` (duration 0) really does
  leave the in-flight window open. Both readings hold.
- **Baseline** — `packages/framework/gtk-host` fails `build:types` here with 15 errors in three
  files (`generated/props.ts`, `generated/widgets.ts`, `generator.spec.ts`), none of which this
  branch touches. Not attributable.

## The one defect, and it was inside the vector written to close a defect of the same class

`runs the chrome check ITSELF, and reports what it found (#1546)` asserted that
`lastWindowChromeProblems()` is `[]`. `[]` is the clean answer AND the never-ran answer, so the
case **passed with `checkWindowChrome(window)` deleted** from `mountApplicationRoot` — measured.
ADR 0043's amendment names `announce.ts` as the precedent for publishing the answer, and
`announce.ts` publishes a COUNT for exactly this reason.

`windowChromeChecks()` is that count. The vector now waits for it to rise before reading the
list, which fails loudly when nothing checks and also stops it reading the answer a previous case
left behind. Red with the call deleted, green with it. — 1909548

## Three more, fixed

**`lastWindowChromeProblems()` was not importable from anywhere.** ADR 0043 publishes it and the
frameworks page tells a consumer to assert it, but it was exported from `app-registry.ts` only,
and `src/index.ts` is the module a bundler aliases `react-native` to. Put it on the root surface
beside `AppRegistry`, regenerated the derived own-export list (19 → 20 names), gave the doc the
import line it was missing. — e6f41ef

**A doc snippet that does not run.** The one-application-per-process example was
`const code = await …` followed by `return process.exit(code)` at the top level of a module:
`SyntaxError: Illegal return statement`, confirmed with `node --check`. A reader copying it gets a
build failure, not an application. Wrapped in an async `main()`, the shape the
self-executing-package guide already uses. I syntax-checked all 18 fences on the page while I was
there; that was the only one. — 176dd7c

**The libadwaita crash was ledgered in the wrong file.** Root `AGENTS.md` exception (b) routes an
upstream GNOME gap to `status/upstream-patch-candidates.md` — which is also the 4-column table
`STATUS.md` renders for readers — and the entry existed only in `open-todos.md`, so the one file
that answers "what do we maintain that upstream could delete" did not know about it, and the entry
never said what a consumer should do meanwhile. Added the row with the source reading, the
reporter's backtrace, the `enable-transitions` measurement and the upstream ask.

The same source reading answers the question the entry left open as "to settle":
`update_child_visible` clears `last_visible_child` whenever the page it names stops being visible
(`adw-view-stack.c:1073`), so hiding the child and letting that notify land before the remove takes
the freed page out of play through public API. That belongs in `gtk-host`'s `ViewStack` descriptor
with a shrinking-route-set vector holding it — not here, and not as an unmeasured guard — so
`open-todos.md` now keeps only that half and points at the upstream row for the mechanism. One
copy, not two. — d72ef44

## Two smaller ones

- Three live counts (`33 of 40 role names`, `refuses 7`, `seven of forty`) were added to comments
  by the same PR that removed three live counts from `packages/framework/AGENTS.md` for being live
  counts. They drift on the next table row — the PR itself says one of the two maps "will not stay
  empty" — and `propRefusedValues()`, published here, answers the number. — d8ba306
- `accepts the 33 role names that DO map, value by value` asserted one name, `button`. The title
  claims the set; the case measured a sample of it. Now it loops the table's own map with a
  non-vacuity assertion: 2542 → 2607 assertions. — f115172

## Checked and clean

**Scope.** `src/support-table.ts` and `packages/framework/gtk-host/**` do not appear in
`git diff main...HEAD --name-only`; the `Modal` entry lives in `support-table.ts`, so it is
untouched by construction. `react-native.mdx` has exactly three hunks — the AppRegistry, routing
and prop-table sections — and my edits stayed inside two of them. No `InteractionManager`
reference anywhere in the diff (the tracked mentions are in `support-table.ts`, `SUPPORT.md`,
`react-native-surface.json` and `animated/timing.ts`, none of them changed).

**Derivation.** The three counts dropped from the AGENTS.md row do not reappear as prose anywhere
in the tree — only in a CHANGELOG line, which is history. `check-rn-surface.mjs` reads a non-empty
input set and prints it: 18 surfaces by name, 20 own exports, 92 React Native names, 48 self-test
vectors, and out loud which mode it ran in.

**Governance.** Both new `status/open-todos.md` entries are `### <title>`; nothing is struck
through; the owning `packages/framework/AGENTS.md` row is updated in this PR.

**`packages/framework/AGENTS.md`: zero bytes added by me.** Still 32763, so the 32737 the merge
with #1563 will need is untouched.

## Left for the maintainer

`acceptsPropValue` / `explainPropValue` / `propRefusedValues` are on the `./prop-table` subpath
only, while their siblings `acceptsProp` / `explainProp` / `propAnswer` / `propNames` are ALSO
re-exported from `src/index.ts`. The docs route readers to the subpath, so nothing is broken —
whether the root alias surface should carry the per-value half too is a published-contract call,
and yours rather than mine.

## Verification, from the worktree root, at the last commit

The checked-in `packages/infra/cli/dist/cli.gjs.mjs` was stale (0.42.0, built 25 Aug) and the shim
prefers it, so it was rebuilt before anything was measured.

- `./node_modules/.bin/gjsify --version` → 0.47.0, shim resolving inside this worktree
- `gjsify run test --include "@gjsify/react-native"` → 2607 completed, 0 failed
- `node --test tests/e2e/react-native-gate/run.mjs` → 9 pass, 0 fail
- `check-rn-surface.mjs`, `audit-runtimes.mjs --check`, `check-agent-context-size.mjs --check`,
  `check-doc-fences.mjs`, `check-vocabulary-alignment.mjs`, `generate-status.mjs` → all 0
- `gjsify lint` → 0 errors · `gjsify format --check` → 3700 files clean ·
  `gjsify workspace @gjsify/react-native run check` → 0

Issue #1548, opened 2026-09-04
The superseded-run cancel is posted, reported, and does not stop the run — force-cancel does


JumpLink:
`cancel-pr-runs.yml` selects the right runs, posts the cancel, reports success — and the run stays
`queued`, holding the `main-pr` concurrency group. Measured today, twice, on two different PRs.

## What happened

Five PRs open, the runner pool saturated. Two runs were superseded by later pushes:

| run | workflow | branch | created | superseded by |
|---|---|---|---|---|
| `33857738939` | `GJS` | `fix/one-header-bar-per-routed-window` | 09:19:07 | `e0985951b2` |
| `33857585236` | `GJS` | `fix/overlay-on-absolute-animated` | 09:17:14 | `84986d6f6e` |

**The workflow did its job.** Both cancel jobs ran and named the run in their own output:

```
cancelled run 33857738939
::notice::cancelled 4 run(s) superseded by synchronize on PR #1540.

cancelled run 33857585236
::notice::cancelled 1 run
```

The selection was right, the branch narrowing was right, the `event=pull_request` listing found them
(I checked: both carry `event=pull_request`, so the filter is not the cause). Nothing in
`select-superseded-runs.mjs` is at fault.

**Twenty minutes later both were still alive.** At 09:35 the API reported:

```
run 33857738939  GJS  status=queued  conclusion=null  updated=2026-09-04T09:35:50Z
run 33857585236  GJS  status=queued  conclusion=null  updated=2026-09-04T09:33:55Z
```

I then cancelled both by hand — the documented escape hatch — with `gh run cancel`. Accepted, "✓
Request to cancel workflow … submitted", **and both stayed `queued`.** So each run had its cancel
requested twice, from two different callers, and did not stop.

`POST /actions/runs/{id}/force-cancel` on the same two ids ended both within a minute:

```
33857738939  status=completed  conclusion=cancelled
33857585236  status=completed  conclusion=cancelled
```

Unfinished runs across the repository went from 24 to 20 at that moment.

## Why this matters more than the four runner-minutes

The header of `cancel-pr-runs.yml` describes the deadlock this workflow exists to break:

> a force-push during a backlog leaves the stale run alive, the stale run keeps the queue saturated,
> and the queue keeps the new run from starting: a deadlock whose only exit is a human cancelling
> runs by hand.

**The escape hatch is the same call.** A human running `gh run cancel` posts exactly the endpoint the
[...]

Issue #1533, opened 2026-09-04
Release-window skip: an untested guard, a too-coarse scope, and no expiry


JumpLink:
Follow-ups from the independent review of #1530, which introduced the release-window skip for
`create-app E2E`. The detector itself was verified sound — every unreachable, timing-out, 5xx,
401/403, HTML-proxy and 404-everything state runs rather than skips, and each of its three
guards was proven load-bearing by removal. These are the edges that survived.

## 1. One guard is untested: removing it leaves the suite 8/8 green

```
$ # delete `if (exact !== 404) return 'unverified';` from tests/e2e/helpers.mjs
$ node --test tests/e2e/release-window-skip/run.mjs   → exit 0, pass 8, fail 0
```

That line is what stops a 401 or 403 on the version URL from being read as the release window.
The `answers 5xx` test passes for the wrong reason: the mock returns 503 to **both** requests, so
the packument confirmation withdraws the claim whether or not the early return exists.

A case where `/{name}/{version}` answers 401 or 403 while `/{name}` answers 200 would pin it.
Worth fixing because `tests/e2e/release-window-skip/` exists precisely to prove each guard by
removal, and this is the one guard that does not answer.

## 2. The skip is coarser than the condition — 49 tests suppressed, ~21 of them needlessly

Only 4 of 7 templates carry a registry-bound `@gjsify/*` edge. From `templates/*/package.json`:
`gtk-minimal`, `adw-canvas2d`, `adw-webgl` and `adw-game` declare `@gjsify/node-gi` (a `file:`
edge that `process-template.mjs:105` rewrites to `^<version>`); `cli`, `web-server-hono` and
`web-server-express` declare only `workspace:^` names, every one of them in `PACKED_NAMES`.

`{ skip }` sits on the outer describe (`tests/e2e/create-app/run.mjs:520`), so those three
templates — 21 tests that would pass — are suppressed for the whole window too.
`registryGjsifyRanges()` already walks per template, so a per-template skip is a small change.

## 3. The skip has no expiry, and cannot tell "not yet" from "never"

`registryVersionState` is stateless: version 404 + packument 2xx ⇒ `missing`, whatever the cause.
If a release's `publish-node-gi` job fails permanently, `main` sits at a version npm does not
have, and **every** subsequent PR and main run reports `create-app E2E` green-by-skip until the
next version bump. Under the deleted wait, that state was red on every run.

The mitigation is real and was verified rather than assumed: `release.yml:1871` runs
[...]

Issue #1531, opened 2026-09-03
foreach --exec silently drops numeric args; the release publishes with an argv nobody wrote


JumpLink:
## What

`gjsify foreach --exec -- <cmd> …` **silently drops every numeric argument** after the `--`
separator. yargs types a bare number in its `populate--` array as a `number`, and `foreach.ts`
filters that array to strings only, so the value disappears without a diagnostic and the option
before it swallows the next flag.

This is live in the release pipeline today.

## Proof

`packages/infra/cli/src/commands/foreach.ts:322` (script mode) and `:330` (exec mode):

```ts
const fromDoubleDash = (((args as Record<string, unknown>)['--'] as unknown[]) ?? []).filter(
    (v): v is string => typeof v === 'string',
);
```

Run against the repo's own yargs with `parserConfiguration({ 'populate--': true })`:

```
raw   : ["gjsify","publish","--verify-timeout",5,"--tag","latest"]
types : string,string,string,number,string,string
after string-filter: ["gjsify","publish","--verify-timeout","--tag","latest"]
```

The filter is not narrowing a union — it is discarding data.

## It is happening in production

`package.json`'s `npm:publish` declares:

```
gjsify foreach -v --topological --no-private --exclude "@girs/*" --exec -- gjsify publish \
  --tolerate-republish --tolerate-untrusted-new --verify-defer --verify-timeout 5 \
  --tag latest --access public
```

`foreach.ts:848` echoes `argv.join(' ')`, i.e. the real argv. Release run
[33804937881](https://github.com/gjsify/gjsify/actions/runs/33804937881), 209 times over:

```
[@gjsify/ws] $ gjsify publish --tolerate-republish --tolerate-untrusted-new --verify-defer --verify-timeout --tag latest --access public
```

No `5`. So `gjsify publish` is invoked with `--verify-timeout` taking `--tag` as its value,
`latest` left as a stray positional, and **`--tag latest` gone**.

## Why nobody noticed

npm's default dist-tag is `latest`, so dropping `--tag latest` changes nothing observable — the
packages land on `latest` either way. And the read-back probe still reported
`confirmed after 1 probe(s)`, so `--verify-timeout`'s garbage value evidently fell back to a
default rather than erroring.

That is luck, not design. The next `foreach --exec` that passes a number where the default is
not what you want will be wrong silently, and this one is already publishing 209 packages with
an argv nobody wrote.

## Blast radius

[...]

PR #1568, opened 2026-09-04
fix: four defects on the release path, and the .deb changelog

### LINKED ISSUE #1522: release.yml builds the workspace twice, ~9 min on every release's critical path
## What

`release.yml` builds the workspace, then `npm:publish` builds it a second time before
uploading anything. The rebuild sits on the critical path of every release, inside the
exact window that `release.yml`'s own comment records as a measured two-hour outage.

## The duplication

`.github/workflows/release.yml:107-111` runs, as its own steps:

```yaml
- name: Build packages
  run: PATH="$PWD/node_modules/.bin:$PATH" gjsify run build
- name: Build examples
  run: PATH="$PWD/node_modules/.bin:$PATH" gjsify run build:examples
```

`release.yml:171-173` then runs `gjsify run npm:publish`, and that script (`package.json`)
begins with the same two commands:

```
gjsify run build && gjsify run build:examples && node scripts/verify-package-outputs.mjs --scope examples && gjsify foreach -v --topological --no-private --exclude "@girs/*" --exec -- gjsify publish ...
```

Same job, same checkout, same container — so the second build starts from the tree the
first one just produced.

## Measured cost

Run [33735989472](https://github.com/gjsify/gjsify/actions/runs/33735989472) (v0.46.0, the last
release that completed):

| moment | time |
|---|---|
| `gjsify run build` (workflow step) | 08:59:09 → 09:04:25 (5m16s) |
| `gjsify run build:examples` (workflow step) | 09:04:25 → 09:07:29 (3m04s) |
| `gjsify run npm:publish` starts | 09:07:29 |
| **first package actually published** (`@gjsify/adwaita-core@0.46.0`, `PUT registry.npmjs.org`) | **09:16:10** |
| publish step ends | 09:28:31 |

**8m41s** elapse inside `npm:publish` before the first `PUT` — 41% of a 21m02s publish step,
and roughly a quarter of the whole 33-minute job.

The in-flight v0.47.0 run [33804937881](https://github.com/gjsify/gjsify/actions/runs/33804937881)
reproduces it: publish step started 21:55:45Z, and 18 minutes later a registry sweep of all 209
publishable packages found 208 still at 0.46.0 (the one exception, `@gjsify/adwaita-core@0.47.0`,
is stamped 21:27:38Z — it came from the earlier attempt that npm 503'd, not from this step).

## Why it matters beyond the minutes

The comment at `release.yml:177-191` documents the incident this lengthens:

> this phase runs after the ~130-package serial npm publish, and the gap between "release
> published" and "assets present" was a MEASURED two-hour window in which
> `releases/latest/download/install.mjs` 404'd — breaking the documented install path and all
> of ts-for-gir's CI.

`release-cut.yml` was changed to upload the assets early precisely to shrink that window. The
duplicate build widens the sibling window — release tag exists, packages do not — by ~9 minutes
on every release, for no output.

## Constraint on the fix

`npm:publish` must keep building when run by hand: locally there is no preceding workflow step,
and a publish from a stale `dist/` is worse than a slow one. So the fix is not "delete the build
from the script" — it is to give CI an entry point that publishes what the job already built,
while `npm:publish` stays whole for human use.

## Not claimed here

Whether the second build is a full rebuild or a partial one — the 8m41s is measured end to end
and also covers `verify-package-outputs.mjs --scope examples`. Splitting that number is worth
doing before choosing between "separate CI script" and "make the build a no-op when warm".

JumpLink:
Closes #1522, #1531, #1533, #1548.

Four defects on the release/ship path, plus the last error-severity lintian tag the `.deb` writer carried. Three of the four are the same shape one layer apart: a mechanism that RUNS, reports success, and does not have the effect its report claims.

## `foreach --exec` published with an argv nobody wrote (#1531)

yargs types a bare number in its `populate--` array as a `number`, and `foreach.ts` (twice) and `run.ts` narrowed that array with `.filter((v): v is string => typeof v === 'string')`. That reads like a type narrowing and is a deletion. Reproduced against this repo's own yargs 18.0.0:

```
in    : gjsify publish --verify-timeout 5 --tag latest
raw   : ["gjsify","publish","--verify-timeout",5,"--tag","latest"]
filter: ["gjsify","publish","--verify-timeout","--tag","latest"]
```

It was live: release run 33804937881 echoed `--verify-timeout --tag latest` **209 times** — `--verify-timeout` taking `--tag` as its value, `latest` a stray positional, `--tag latest` gone. Invisible because npm's default dist-tag *is* `latest`. `cli-reference.md` documented `gjsify run ./server.mjs -- --port 8080`, an argv that reached the server as `--port` with no value.

Fixed at the **parse**: every command setting `populate--` now also sets `parse-positional-numbers: false`, so the `--` tail stays the text the caller typed while declared `type: 'number'` options are still parsed (measured: `--jobs 4` stays `4`). A `String(v)` in the reader would recover the value and lose its spelling — measured against yargs 18.0.0, `1.0` comes back as 1, `0.10` as 0.1, `0x10` as 16 and `1e3` as 1000. (`007` does NOT: it stays the string it was given. That was written down here as the example and is the one shape that never broke; corrected in the review round below.)

Three readers of one array disagreed (`foreach` ×2 and `run` filtered, `dlx` coerced); they are one `doubleDashArgs()` now, which refuses a non-string instead of dropping it. The spec drives the **real** builders — a `parserConfiguration` assertion still passes over a builder that sets the key and a reader that drops the value — and snapshots inside the handler, because yargs reuses that object and `--` is gone by the time `parseSync()` returns.

## `release.yml` built the workspace twice (#1522)

Its own two build steps, then `gjsify run npm:publish`, whose first two commands are the same two builds — same job, same checkout, same container. Re-read from the API's step timestamps for run [33735989472](https://github.com/gjsify/gjsify/actions/runs/33735989472) (v0.46.0, the last release that completed):

| step | window | duration |
|---|---|---|
| `Build packages` | 08:59:09 → 09:04:25 | 5m16s |
| `Build examples` | 09:04:25 → 09:07:29 | 3m04s |
| `Publish packages to npm` | 09:07:29 → 09:28:31 | 21m02s |
| first registry `PUT` | 09:16:10 | **8m41s into the step** |

That is 41% of the publish step, inside a 33m02s job, redoing the build over the tree it had just written — and it widens the sibling of the window the asset-upload comment records as a measured two-hour outage.

`npm:publish` stays whole for a human (`build && build:examples && npm:publish:prebuilt`) and `release.yml` runs the second half, whose precondition is the two steps immediately above it. **No new check** — the answer to "what does this let you delete" is the 8m41s above, less the `verify-package-outputs --scope examples` that stays. #1522 left those two unsplit; the same job log splits them (last example build 09:16:09.626, the check's two lines 09:16:09.724, the sweep's first child 09:16:10.084), so the check is under half a second and the deleted work is the two builds.

*Not claimed as measured:* the after. I cannot time a release run. What is shown above is the before, and the deleted work is exactly the second invocation of those two scripts.

## Release-window skip: an untested guard, a too-coarse scope, no expiry (#1533)

**The guard that answered nothing.** Deleting `if (exact !== 404) return 'unverified'` left the suite **8/8 green** — `answers 5xx` passed for the wrong reason, its mock returning 503 to both requests. The new cases put the registry in the one state only that line can decide: `/{name}/{version}` answers 401 (and 403) while `/{name}` answers 200. **Verified by removal: with the guard deleted the suite is 14/16 and the two failures are exactly those cases.** The request log is asserted too — the guard returns before the second request, so a right answer with two requests logged would be right for the wrong reason.

**The scope.** Only 4 of 7 templates carry a registry-bound edge; `{ skip }` on the outer describe suppressed the other three too. The decision is per template now, probed concurrently, and a template with no such edge never reaches the network. The outer suite still skips when the window covers every template, because its `before()` packs the whole workspace.

**The expiry.** `registryVersionState` is stateless, so a permanently failed publish job leaves every later PR green-by-skip until the next version bump. The clock is `CHANGELOG.md` — written by the same `release-it` run that bumps the manifests, in the same commit. The registry dates a different release; git is out (depth-1 checkout, no history, no tags); the GitHub release record would put an authenticated, rate-limited call inside a test helper.

The grace is one full day, and the number is measured. A publish takes 43-70 min, so any bound over an hour would do — what forces a whole day is that a cut can straddle midnight: v0.47.0's changelog entry is dated 2026-09-03 and `@gjsify/node-gi@0.47.0` reached npm at **2026-09-04T00:49:56Z**. "Yesterday is too old" would have called that healthy release a failed one for 49 minutes. Both directions need positive evidence: the skip disarms a suite and demands proof, so the expiry that re-arms one demands proof too, and an undatable version changes nothing.

## The superseded-run cancel had no effect (#1548)

Not taken on faith. Re-read from the API while writing this:

| run | created | cancel posted by | terminal | gap |
|---|---|---|---|---|
| `33857585236` | 09:17:14Z | 09:20:29Z (cancel job 33857795159) | 09:36:53Z `cancelled` | **16m24s** |
| `33857738939` | 09:19:07Z | 09:27:47Z (cancel job 33858431251) | 09:36:52Z `cancelled` | **9m05s** |

Both cancel jobs ran, named their runs and emitted `::notice::cancelled N run(s)`. At 09:35 both were still `status=queued conclusion=null`. A human's `gh run cancel` — the documented escape hatch, the same endpoint — was accepted and changed nothing. `force-cancel` ended both within a minute; those `updated_at` stamps are its.

The selection was never at fault. The `run:` block counted its own POSTs and called them cancellations, so that half moves out of the workflow into `scripts/cancel-superseded-runs.mjs`, under the same fixtures as the selection.

**The discriminator is `queued`, not a clock.** A graceful cancel marks intent and waits for the job to ACKNOWLEDGE it; a run whose jobs were never assigned a runner has nothing to deliver that intent to — which is exactly the saturated-queue deadlock this workflow exists for, and its only case. So: `completed` after the settle needs nothing; `queued` is force-cancelled (no step ever ran, so there is no `always()` cleanup to lose); `in_progress` is left alone and reported, because it has a runner, received the cancel and is unwinding. The issue's suggested "force-cancel anything not completed" would take that cleanup away from the runs that were behaving.

The notice now counts runs that reached `completed`, not requests that were accepted — the count that would have surfaced this the first time.

## The `.deb` had no changelog — Policy § 4.4 (ledger item 10)

`E: gjsify: no-changelog usr/share/doc/gjsify/changelog.Debian.gz (non-native package)`, the only error-severity tag left once the copyright landed. `dpkg -i` installs such a package, `apt` says nothing, and only a real lintian objects.

**What the entry says** — the open question the ledger named. ONE entry, for the version being packaged: § 4.4's file is the *packaging* changelog and `gjsify ship` packages a version once. Its body is the project's own `CHANGELOG.md` (or `CHANGELOG`, `CHANGELOG.txt`, `NEWS.md`, `NEWS`), searched up to the repository root exactly as the licence is, overridable with `gjsify.ship.changelogFile`; the GitHub release body says the same thing and is unreachable from an offline packing host, while CHANGELOG.md is in the tree the stage is assembled from. Only the packaged version is dated, from `mtime` — the build stamp every other header shares — because dating the older ones would mean fabricating a time. `rfc822()` is the formatter this tree did not have.

It is gzipped **at assembly**, which is what the schema 5 → 6 bump is for: `--from-stage` rehydrates the overlay out of the sidecar, and a gzip on the packing host would put a second compressor's output into an artifact whose other bytes came from the first. `StageOverlayFile` carries either `text` or `base64`; `text` stays preferred where the content is text, because the sidecar is the one artifact in this handoff a human reads.

**Measured with lintian 2.117.0ubuntu1.5 in `ubuntu:24.04`**, against the `.deb` that actually shipped as v0.47.0 and the one this branch builds:

```
BASELINE   E: gjsify: no-changelog usr/share/doc/gjsify/changelog.Debian.gz (non-native package)
           + 12 warnings

AFTER      no error-severity tag at all
           the same 12 warnings, minus none, plus the two below
```

Two new warnings, both on the new file, both stated rather than hidden:

- `initial-upload-closes-no-bugs` — expects the first upload of a source package to the Debian **archive** to close its ITP bug. A package built by `gjsify ship` is never uploaded there and there is no bug number that would be true to write. Inapplicable, not unmet.
- `changelog-not-compressed-with-max-compression` — Policy asks for `gzip -9`, and neither `CompressionStream` nor `@gjsify/zlib` takes a level (`gzipSync`'s options parameter is literally named `_options`). **Ledgered in `status/open-todos.md`** as the core capability it is. Stamping the gzip header's XFL byte to claim max compression is not the fix: the mtime and OS bytes `gzipDeterministic` normalises are facts about the build environment, XFL is a claim about the compression performed.

`debian-changelog-line-too-long` fired on the first artifact and is closed by wrapping at 80 columns.

## Documentation

- `website/src/content/docs/ship/linux-packages.md` — a new section on the Debian changelog: that every `.deb` now carries one, where the prose comes from, what the entry looks like, `gjsify.ship.changelogFile`, and the two lintian warnings not worth acting on. This is a user-visible behaviour change: `E: no-changelog` is something *their* package gets flagged for.
- `website/src/content/docs/cli-reference.md` — `run` and `foreach` now state that everything after `--` reaches the target as typed, numbers included. The page's own `-- --port 8080` example was one of the argvs that did not work.
- `docs/publishing.md` — the two publish entry points and the measurement that split them.
- `status/open-todos.md` — ship-roadmap item 10 deleted; **item 5's lead corrected**: every container stages 4/5 promised has shipped (`macos-app`, `macos-app-zip`, `macos-app-dmg`, `windows-dir`, `windows-dir-zip`, `msi`, behind `tests/e2e/{ship-macos,ship-msi,ship-windows,ship-signing}`) and the entry read as though none had. What is genuinely open there is the mime/icon/desktop-entry half and the loose `.typelib`.

#1522, #1533 and #1548 are CI-internal and change nothing a user of gjsify can see.

## Measured here

- `@gjsify/cli` unit suite **3752/3752**, `tests/e2e/ship` 44/44, `ship-from-stage` 20/20, `release-window-skip` 16/16, `ci-cancel-superseded-runs` 27/27, `foreach-command` 22/22, `run-command` 16/16
- release-window guard proven **by removal** (8/8 green before, 14/16 with the guard deleted)
- `lintian` before/after in `ubuntu:24.04`, on the artifact v0.47.0 actually shipped and on this branch's
- `audit-runtimes --check`, `check-workflow-run-syntax`, `check-workflow-release-globs`, `check-agent-context-size --check`, `gjsify format --check`, `gjsify lint` (0 errors)

## Outside the stated scope, named

`.github/ship-oracle/verify-deb.sh` — the changelog path assertion, four changelog tags in its named gate, a read-back through gzip after the install, and its file count now derives the overlay half from the sidecar rather than the literal `+ 1` that broke the moment a second overlay file existed. `packages/infra/cli/src/commands/publish.ts` — its private `base64Encode` was the second copy of `@gjsify/buffer`'s and is now the import. **That is not the only new edge into that package**: `utils/ship/stage-manifest.ts` imports `base64Encode` *and* `base64Decode` from it for the gzipped changelog's trip through the sidecar, and it is the file the build-order gate names. Both put `@gjsify/buffer` into the CLI's own tsc — see the review round below.

---

## Independent review round

Reviewed on the branch, defects fixed here rather than reported. Five commits.

**`build: type @gjsify/buffer before the CLI`** — `Detect runtime-triplet drift`, a REQUIRED check, was red: `@gjsify/cli build` is clause 12 of `build:infra` and nothing emits `@gjsify/buffer`'s declarations until clause 21. Invisible locally, because a warm tree has the previous build's `lib/types/`. `build:types` before the CLI is the whole fix. The dedup itself is byte-verified against the removed private function — 3182 cases (empty, all 256 byte values, every pad class, the old 0x8000 chunk boundary, 3 MB, UTF-8 non-ASCII), 0 mismatches, and every one also equals Node's `Buffer.toString('base64')`.

**`test(e2e): find the publish sweep after the split`** — `tests/e2e/published-closure` was red. Its prevention half reads `scripts['npm:publish']` and asserts the sweep still carries `--topological`; the split left it looking at three `gjsify run` calls and no sweep, and it failed with the message written for exactly that ("verified NOTHING about publish order"). Its subject is derived from `release.yml` now. Reds on a dropped `--topological` and on a workflow pointing at a script the manifest does not declare; stays green if the workflow points back at the whole `npm:publish`.

**`fix(ci): close five gaps in the cancel it replaced`** (#1548) — the fix would have reddened every open PR: scripts come from `base.sha`, the YAML from the merge ref, and the unconditional `node scripts/cancel-superseded-runs.mjs` on a base without it exits 1 under `set -e` (reproduced against `git ls-tree 0498711fe -- scripts/`). Also: `pending` runs are selected and were never force-cancelled (this repo's last 300 runs: 263 completed / 25 queued / 9 in_progress / **3 pending**); a `fetch` throw failed a cost-control job the old `|| true` could not; "0 of 1 stopped" was a green job with a `::notice::`; and every survivor was logged as "the cancel was accepted" including ones just logged as refused. Two assertions verified nothing — the sparse-checkout regex spanned the whole file (deleting the entry left it 27/27) and the `::notice::` is emitted from a CLI block no test ran (the incident's own wrong string could be pasted back, green). 36/36, and seven mutations each red exactly one test.

**`fix(e2e): stop the release-window skip telling the time`** (#1533) — the suite would have gone red at 2026-09-05T00:00Z with no code change: one case asked for a skip with no `window` oracle, so it was decided by the wall clock against the real CHANGELOG. Measured under a clock-shifting preload: pre-fix +30 days → 15/16; post-fix → 22/22. The grace was pinned from below (0 reds) but not from above (2 was green). And "43-70 min … any bound above an hour would do" is contradicted by the release the same paragraph cites — re-measured from each run's start to `npm view @gjsify/node-gi time`: 70 / 43 / 46 / **236** min for v0.44.0-v0.47.0. The conclusion survives, the argument did not, and it was also in the runtime message. Finally, item 2 of the issue — one decision per template — was correct and **unprovable**: the rule was module-private in a suite whose module scope packs the workspace, so reverting the narrowing reddened nothing. It lives in `helpers.mjs` now with fixtures.

**`docs: split the 8m41s #1522 left unsplit`** — see above.

**Verified and left alone.** The `.deb` changelog, rerun end to end in `ubuntu:24.04` on lintian 2.117.0ubuntu1.5 against a `.deb` this branch builds: **no error-severity tag at all**, exit 0, with `changelog-not-compressed-with-max-compression` and `initial-upload-closes-no-bugs` the only additions — the two the PR names. Baseline (the same package with the changelog removed and md5sums adjusted): `E: no-changelog … (non-native package)`, exit 2. `dpkg-parsechangelog` parses the file (exit 0, right Version/Distribution/Urgency/Date), and on a project that HAS a `CHANGELOG.md` the bullets path renders links flattened, bold scopes stripped, the next version's entries excluded and the longest line at 79 columns with `debian-changelog-line-too-long` silent. Both accepted warnings are the right calls, and refusing to stamp the gzip XFL byte is correct — the mtime and OS bytes `gzipDeterministic` normalises are facts about the environment; XFL is a claim about compression. The `@gjsify/zlib` gap is ledgered with the core path named (`gzipSync`'s parameter is indeed literally `_options`). The `#1522` trace is clean: nothing else on the release path consumed the deleted invocation's output — the two workflow build steps carry no `if:`, the only step between them and the publish is mutually exclusive with it, `verify-package-outputs --scope examples` stayed in `npm:publish:prebuilt`, `verify-tarball-outputs` never ran on this path, and every `needs: publish` job does its own checkout and build.


JumpLink:
## Independent review — six defects found, all fixed on the branch

I reviewed this against the release path's own standard ("does the check fail when it should?") and fixed what I found rather than reporting it. Five commits on top of yours, plus one that corrects a measurement. Two of the six would have shipped; two were already red.

### Red before I started

**1. A REQUIRED check was failing.** `Detect runtime-triplet drift` → `check-build-infra-order.mjs`: `@gjsify/cli build` is clause 12 of `build:infra`, and nothing emits `@gjsify/buffer`'s declarations until clause 21. Two files put that edge into the CLI's tsc — `commands/publish.ts` (the dedup you named) and `utils/ship/stage-manifest.ts` (`base64Encode` **and** `base64Decode`, for the changelog's trip through the sidecar), which is the one the gate names and which the PR body did not mention. Reverting the dedup would therefore not have fixed it; the changelog needs the edge. Fixed with a `build:types` clause before the CLI. Worth writing down why nobody saw it locally: a warm tree has the *previous* build's `lib/types/`, so it only fails cold.

Before deciding to keep the dedup I checked the replacement byte-for-byte against the function it removed: 3182 cases — empty input, all 256 byte values, every padding class, the old `0x8000` chunk boundary, 3 MB, UTF-8 non-ASCII — 0 mismatches, and every one also equals Node's `Buffer.toString('base64')`.

**2. `tests/e2e/published-closure` was red.** Its prevention half reads `scripts['npm:publish']` and asserts the sweep still carries `--topological`; splitting that script left it looking at three `gjsify run` calls and no sweep. It failed with the message written for exactly this — *"verified NOTHING about publish order"* — which is the design working, and it was not acted on. Its subject is derived from `release.yml` now (which script does the release actually run?) and expanded across `gjsify run`, so naming a script here cannot reopen the hole one rename later.

### Would have shipped

**3. #1548's fix reds every open PR the moment it merges.** The scripts are checked out from `pull_request.base.sha`; the YAML comes from the merge ref. So the first event on a PR based before this merges runs the new, now-**unconditional** `node scripts/cancel-superseded-runs.mjs` against a `scripts/` that has only the selection — `Cannot find module`, exit 1 under `set -euo pipefail`. Reproduced against `git ls-tree 0498711fe -- scripts/`. This is #1340's asymmetry from the other side, and deleting the `[ -z "${ids}" ]` early exit removed the accident that used to shield it. Guarded on the file existing, with the guard under a fixture.

**4. `tests/e2e/release-window-skip` goes red tonight, on a repository nobody touches.** `skips, naming the package and version, …` calls `createAppRegistryGapSkipReason` with **no `window` oracle**, so a case about the mechanism is decided by the wall clock against the real CHANGELOG. This checkout: cut 2026-09-03, grace 1, so `days` flips 1 → 2 at 2026-09-05T00:00Z and the skip it asserts stops existing. Measured under a clock-shifting `--import` preload: pre-fix at +30 days → 15/16; post-fix → 22/22. Every case pins its clock now; the one that binds to the real file is the one that reads the date out of it.

### Not defects in the outcome, defects in the evidence

**5. Three numbers are asserted rather than derived**, all of which I re-measured:

- *"yargs parses `007` to 7"* — it does not; `007` stays the string it was given. That claim is in `double-dash-args.ts`, its spec and `cli-reference.md`, and it is the one shape in the argument that never broke. The **decision is right** — `1.0`→1, `0.10`→0.1, `0x10`→16, `1e3`→1000 all lose their spelling under a coercion — so the example moved to those, with `007` kept and labelled as the control.
- *"43-70 min … any bound above an hour would do"* — contradicted by the release the same paragraph cites. From each run's start to `npm view @gjsify/node-gi time`: **70 / 43 / 46 / 236 min** for v0.44.0-v0.47.0. An hour would have called v0.47.0 a failed publish for nearly three hours. The conclusion (a day) survives; the argument did not, and it was also in the runtime message a maintainer reads on a real expiry. `ONE FULL DAY` went with it: `floor(days) <= 1` is 24-48 h depending on the hour of the cut.
- *"cancelled 4 run(s) while four runs kept their concurrency groups"* — the 4 was a POST count and **two** runs were measured. Copying that confusion into the file that exists to prevent it.

Also corrected: the here-string comment's stated reason ("a pipeline's RHS runs in a subshell, so its exit would not fail the step") is measurably false in both `set -e` and `set -eo pipefail`. The here-string is still right; SIGPIPE is the reason.

**6. Assertions that verified nothing**, each proven by removal:

| what | with it broken |
|---|---|
| `sparse-checkout:[\s\S]*?cancel-superseded-runs\.mjs` — spans the whole file, matched down to the *invocation* 40 lines below | entry deleted → **27/27 green** over the #1340 regression it names |
| the `::notice::` is emitted from a CLI block no test runs | the incident's own `cancelled ${posted} run(s)` pasted back → **27/27 green** |
| `SETTLE_MS` / `POLL_MS` — every fixture injects both | `SETTLE_MS = 0`, and a deleted `await sleep()` → **green** |
| the per-template narrowing (#1533 item 2) is module-private in a suite whose module scope packs the workspace | reverting it → caught by **nothing** |

All four now fail when they should. `registryGjsifyRanges` moved to `helpers.mjs` beside the rest of the release-window decision.

### Behaviour I changed, and why

- **`pending` is force-cancelled too.** The selection takes `status !== 'completed'`; the force-cancel keyed on the literal `queued`. This repo's last 300 runs: 263 completed, 25 queued, 9 in_progress, **3 pending** — and a pending run has no runner by your own argument, so three live runs were being reported and none freed. `NO_RUNNER` is enumerated, so an unrecognised future status is reported rather than force-cancelled. Your call on `in_progress` I agree with and left alone.
- **A survivor is a `::warning::`, not a `::notice::`.** "0 of 1 stopped" — the deadlock this workflow exists to break — was a green job with a line that does not reach the checks list, which is the incident's shape one level out: it now measures correctly and still did not escalate. Gated on `posted > 0` so a fork PR's read-only token stays quiet.
- **A `fetch` throw no longer fails the job.** The `run:` block this replaced was `gh api … || true` per run; a transient now put a red X on a cost-control job.
- **The survivor line says what was observed.** It read "the cancel was accepted and has not taken yet" for every survivor, including ones whose POST had just been logged as refused.

### Verified and left alone

**The `.deb` changelog is right.** Rerun end to end in `ubuntu:24.04`, lintian **2.117.0ubuntu1.5**, against a `.deb` this branch builds:

```
AFTER
W: ship-demo: changelog-not-compressed-with-max-compression [usr/share/doc/ship-demo/changelog.Debian.gz]
W: ship-demo: copyright-without-copyright-notice
W: ship-demo: initial-upload-closes-no-bugs [usr/share/doc/ship-demo/changelog.Debian.gz:1]
W: ship-demo: no-manual-page [usr/bin/ship-demo]
--- exit 0 ---          error-severity tags: 0

BASELINE (same package, changelog removed, md5sums adjusted)
E: ship-demo: no-changelog usr/share/doc/ship-demo/changelog.Debian.gz (non-native package)
--- exit 2 ---
```

No error-severity tag at all, and the only two additions are the two you name. `dpkg-parsechangelog` — an independent parser you did not claim — accepts the file at exit 0 with the right `Version: 1.2.3-1`, `Distribution: unstable`, `Urgency: medium`, `Date: Fri, 04 Sep 2026 …`, so the trailer spacing is right. On a project that *has* a `CHANGELOG.md` the bullets path renders links flattened, bold scopes stripped, the next version's entries excluded, and the longest line at **79** columns with `debian-changelog-line-too-long` silent. All sixteen tag names `verify-deb.sh` asserts exist in that lintian's `--list-tags`.

Both accepted warnings are the right calls, and **not** stamping the gzip XFL byte is correct: the mtime and OS bytes `gzipDeterministic` normalises are facts about the build environment, XFL is a claim about the compression performed. The `@gjsify/zlib` gap is properly ledgered with the core path named — `gzipSync`'s parameter is indeed literally `_options`.

**#1522's trace is clean.** Nothing on the release path consumed the deleted invocation's output: the two workflow build steps carry no `if:`, the only step between them and the publish is mutually exclusive with it, `verify-package-outputs --scope examples` stayed in `npm:publish:prebuilt`, `verify-tarball-outputs` never ran on this path, and every `needs: publish` job does its own checkout and build. The four step durations re-read exactly from the API. I also split the one term the issue explicitly left unsplit — from the same job log, `verify-package-outputs --scope examples` is **under half a second** of the 8m41s (last example build 09:16:09.626, its two lines 09:16:09.724, the sweep's first child 09:16:10.084), so the deleted work is the two builds and the choice of shape was not resting on an unmeasured term. The PR body said "8m20s of duplicated build", which is the *first* build's two step durations applied to the second; corrected.

**#1533's central proof reproduces.** With `if (exact !== 404) return 'unverified'` deleted the suite is 15/17 and the two failures are exactly the 401 and 403 cases. The midnight argument holds and is a real measurement: `@gjsify/node-gi@0.47.0` reached npm at 2026-09-04T00:49:56Z against a CHANGELOG entry dated 2026-09-03. Every other removal in `registryVersionState` leaves the suite green because those lines are provably redundant, not because the suite is blind.

**`status/open-todos.md` item 5's correction is accurate** — all six containers exist in `FORMATS` and all four suites exist on disk.

### Nothing here needs a decision that is not mine

The one judgement I would flag rather than claim: the `::warning::` escalation is a policy choice about how loud this job should be. I took it because #1548's own framing is "nothing in the log looks wrong", and a notice is not in the checks list. Revert that one line if you disagree — it is `const level = …` in the CLI block, and a fixture covers both branches.

### Verified on the branch

`gjsify run build:infra`, then: `check-build-infra-order`, `audit-runtimes --check`, `check-workflow-run-syntax`, `check-workflow-release-globs`, `check-changelog-references`, `check-agent-context-size --check`, `gjsify lint`, `gjsify format --check` — all exit 0. e2e `ship` 44/44, `ship-from-stage` 20/20, `release-window-skip` 22/22, `published-closure` 35/35, `ci-cancel-superseded-runs` 36/36, plus the `@gjsify/cli` unit suite. Every fix above was driven red before it was driven green; the mutation matrices are in the commit messages.

Issue #1546, opened 2026-09-04
windowChromeProblems() watches one caller: the router's own vectors


JumpLink:
`windowChromeProblems()` / `windowChromeCensus()` landed with #1540 in
`@gjsify/gtk-host/conformance`. They are correct, they have a measured discriminator, and
they name the failure in exactly the right words. **Nothing in a shipping code path ever
runs them.**

## The measurement

Grepped every `src/` tree in the repo (`packages`, `showcases`, `examples`, `tests`;
excluding `lib/` bundles). The reader is reached from **two files, both specs**:

- `packages/framework/gtk-host/src/conformance.spec.ts`
- `packages/framework/react-native/src/router/router.spec.ts`

`packages/framework/react-native/src/router/chrome.ts` and
`packages/framework/react-native/src/window-chrome.ts` name it only in comments — they
import `useWindowChrome`, never the census.

## Why that matters, with the case that made it concrete

While reviewing #1540 a composition was found that the guard describes perfectly and no
vector reached: a `headerShown: false` screen pushed on top of a bar-ful root stack — a
documented option, and what a full-bleed reader route uses. Measured:

```
before push:  headerBars=1  start=0  end=1
after  push:  headerBars=0  start=0  end=0
windowChromeProblems(window):
  ["no window control draws anywhere in this AdwWindow, so its chrome offers no way to
    close or move it (0 mapped header bar(s)) — an Adwaita window carries no titlebar of
    its own, so a header bar has to"]
```

A window with no close, no maximise and nothing to drag it by. The guard produced the
right sentence the moment it was pointed at the shape; it was never pointed at it, because
the only thing that points it anywhere is a hand-written vector. Fixed in #1540, but the
class is not: the router is one composer of windows, and every consumer that builds its own
window, nests navigators in an order the vectors do not enumerate, or mixes a hand-written
`Adw.HeaderBar` into a routed tree, is outside everything this reader watches.

This is the shape the repo keeps paying for: a check that runs, is green, and was never
aimed at the case it exists for.

## Two ways to aim it, and the trade between them

**A — a dev-mode assertion in `AppRegistry`.** After the root commit, walk the window's
content and report every problem. Catches the defect in the application the developer is
running, with no extra step, and it is the only option that fires for a consumer who never
writes a vector. Costs a tree walk per commit, so it has to be gated (`NODE_ENV`, or an
explicit opt-in) — and a gate is one more thing that can be off in the configuration where
[...]

PR #1572, opened 2026-09-05
ci: a keyword per issue, or the list closes one

### LINKED ISSUE #1453: Deep-linking to a non-index tab loops between React and the ViewStack

### LINKED ISSUE #1556: Two Image vectors are red on win32-x64 only, invisible behind a continue-on-error probe

JumpLink:
## What

A closing keyword closes ONE issue. GitHub repeats the keyword before every reference; a comma-separated list after a single keyword closes the first and leaves the rest open — while reading, to the author and the reviewer and the merge, as though it closed all of them.

**The incident: three bodies written inside twenty minutes on 2026-09-04 (18:22–18:42Z), all three merged into 0.48.0.** #1567 and #1565 each opened their body with one keyword and a list:

```text
Closes #1453, #1547, #1546, #1549, #1551, #1555.
Closes #1556, #1550, #1516, #1529.
```

Each merge closed the first reference and no other — #1453 at 2026-09-04T20:26Z, #1556 at 2026-09-05T10:40Z — with one exception that is the whole reason both surfaces are read: **#1549 closed too**, because a commit body inside #1567 happened to say `That closes` and then name it, mid-sentence. **Seven issues stayed open** — open while the work that fixed them was on `main`, in a release, and being read by whoever went looking for what was left to do. They were closed by hand on 2026-09-05, 14:47–14:48Z, each with the commit that had already fixed it.

**The control is #1568**, the third of those bodies, same squash mechanics, whose four issues all closed at 2026-09-05T09:46Z. Its body carried the identical broken list — this check flags it, measured below — and what saved it was the *other* surface: a `Closes #N.` line per issue in its commit messages. So the variable is the SPELLING and not the merge, and a PR is only ever one surface away from the defect. Both surfaces are read.

## Where, and why this is not a new gate

`commitlint.yml` already owns the strings that land on `main` and is already one of the three required checks, so this is a fourth STEP there rather than a fourth blocking workflow (AGENTS.md § Governance: a required check that does not run blocks the PR forever, and every guard's cost is the sum).

It is the same asymmetry `check-pr-title-length.mjs` and `check-pr-title-types.mjs` were written for — a string that becomes history and that nothing looked at — one surface further on. The difference that earns the step: a bad title turns `main` red and is visible; a bad body changes the state of OTHER records, and reports success while doing it.

## What it cost itself

The first detector had a false positive its own vectors could not see. Fenced code is blanked before the scan (so prose *about* this defect — this PR body included — does not trip it), and the fence terminator was a bare `$` under `/m`. With a lazy body that matches the FIRST line end, so a three-line fence had one line blanked and the rest read as prose: the check failed on line 3 of its own example. The single-line vector passed throughout. There is a multi-line vector now, plus the unterminated and indented shapes.

The independent review found three more, all fixed on the branch:

- A colon after the keyword was not matched. GitHub's own page says the keywords "can be followed by colons or in uppercase" — so a colon form closed its first reference and this check called the body clean, a false negative on a spelling documented by the very page the keyword list mirrors. `&` and `;` were missed for the plainer reason that no separator claimed them.
- `\s*` on either side of the separator let a chain cross a blank line, so a `Closes #a` paragraph followed by one opening `And #b is a follow-up.` was reported as a broken list — a **false positive on a required check**, which is the one failure mode this cannot have. A separator now spans at most one line break, and a bare `and` has to stay on the reference's own line.
- With `PR_BASE_SHA`/`PR_HEAD_SHA` empty the check skipped the commit walk, read the body alone, exited 0 and printed a success line naming the body it had read. It now refuses, the way `check-pr-title-length.mjs` refuses an empty `PR_TITLE`. An empty `PR_BODY` stays legitimate — GitHub sends null for a PR with no description.

## Measured, through the entry point, against the real corpus

| PR body | verdict |
|---|---|
| #1567 | flagged — 5 refs named, not closed by that keyword |
| #1565 | flagged — 3 |
| #1568 | flagged — 3 (closed anyway, by its commit bodies) |
| #1570 | clean |
| #1566 | clean |

35 self-test vectors run before any real data is read, in **both** directions: a detector that flags everything passes a one-sided suite as easily as one that detects nothing, and this one sits on a required check, where a false positive blocks a correct merge. The vectors name issue numbers outside this repository's range on purpose — they end up in a commit message, and a vector naming a real issue would ask GitHub to close it.

Every vector is load-bearing, measured by mutation: reverting the fence terminator to a bare `$`, mutating the colon away, narrowing the separator class back to a comma, widening the gap to `\s*`, dropping the newline out of it, letting `and` cross a line, making the `#` optional in a reference, unblanking code, and shifting the line counter each turn the self-test red.

An empty commit range fails loudly rather than reporting clean, and so does a missing one.

## Honest limits

- It reads the PR body and the PR's commits. A merger who **edits** the squash body in the GitHub UI at merge time is not covered by anything here; nothing in this repo reads that string.
- `Refs #a, #b` is untouched by design — it claims nothing.
- The keyword list is mirrored from GitHub's documentation, the way `check-pr-title-length.mjs` mirrors `header-max-length`. A keyword GitHub adds later fails open on exactly the spelling it forgot.
- Fenced blocks and inline spans are exempt; a four-space indented block and an `<!-- -->` comment are not. That is deliberate rather than missed — a commit message is not markdown, and a commit body is the surface #1549 actually closed on, so blanking those shapes would trade a rare false positive for a blind spot on the surface that does the closing.
- `**Closes** #a, #b`, where emphasis splits the keyword off its space, and a bare issue URL after a keyword are unmatched. GitHub's documented syntax lists only `#N` and `OWNER/REPO#N` as closing references.

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -78,10 +78,11 @@ jobs:
       # reads on merge, and the only one that changes the state of OTHER records:
       # a `Closes #a, #b, #c` list closes #a and leaves the rest open while
       # reporting the intent to close all three. #1567 and #1565 left seven issues
-      # open that way on 2026-09-04 — they were closed by hand a day later, each
-      # with the commit that had already fixed it. This reads the body AND the PR's
-      # commit messages, because a squash body is composed from those: that is the
-      # surface which rescued #1568, whose own body carried the same broken list.
+      # open that way across the 0.48.0 release — they were closed by hand on
+      # 2026-09-05, each with the commit that had already fixed it. This reads the
+      # body AND the PR's commit messages, because a squash body is composed from
+      # those: that is the surface which rescued #1568, whose own body carried the
+      # same broken list.
       - name: Check that every closing keyword closes what it names
         if: github.event_name == 'pull_request'
         env:
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -73,3 +73,19 @@ jobs:
           PR_TITLE: ${{ github.event.pull_request.title }}
           PR_NUMBER: ${{ github.event.pull_request.number }}
         run: node scripts/check-pr-title-length.mjs
+
+      # The two steps above cover the title. The BODY is the third string GitHub
+      # reads on merge, and the only one that changes the state of OTHER records:
+      # a `Closes #a, #b, #c` list closes #a and leaves the rest open while
+      # reporting the intent to close all three. #1567 and #1565 left seven issues
+      # open that way on 2026-09-04 — they were closed by hand a day later, each
+      # with the commit that had already fixed it. This reads the body AND the PR's
+      # commit messages, because a squash body is composed from those: that is the
+      # surface which rescued #1568, whose own body carried the same broken list.
+      - name: Check that every closing keyword closes what it names
+        if: github.event_name == 'pull_request'
+        env:
+          PR_BODY: ${{ github.event.pull_request.body }}
+          PR_BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          PR_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: node scripts/check-closing-keywords.mjs
--- a/scripts/check-closing-keywords.mjs
+++ b/scripts/check-closing-keywords.mjs
@@ -1,0 +1,291 @@
+#!/usr/bin/env node
+// A closing keyword closes ONE issue. A comma after it closes nothing.
+//
+// THE INCIDENT, twice on 2026-09-04 and both released in 0.48.0. #1567's body
+// opened with `Closes #1453, #1547, #1546, #1551, #1555.` and #1565's with
+// `Closes #1556, #1550, #1516, #1529.`. GitHub closed the FIRST ref of each and
+// left seven issues open — open while the work that fixed them was on `main`, in
+// a release, and being read by whoever went looking for what was left to do.
+// They were closed by hand a day later, each with the commit that had already
+// fixed it.
+//
+// THE CONTROL, same day, same repository, same squash mechanics: #1568 closed all
+// four of its issues. Its BODY carried the identical broken list — measured here,
+// this check flags it — and what saved it was the other surface: four separate
+// `Closes #N.` lines in its commit messages. So the variable is the SPELLING and
+// not the merge, which is what makes this checkable at all; and a PR is only ever
+// one surface away from the defect, which is why both are read below. GitHub's own
+// rule: the keyword must be repeated before every reference.
+//
+// WHY THIS SHAPE OF DEFECT IS WORSE THAN A BROKEN LINK. It reports success. The
+// body SAYS five issues are closed, the merge SAYS it succeeded, and the four
+// that stayed open look — to anyone reading the issue list — like work nobody has
+// done. That is the same class `check-pr-title-length.mjs` and
+// `check-pr-title-types.mjs` were written for: a string that becomes history and
+// that nothing looked at. Those two cover the PR title. This is the third string
+// GitHub reads, and the only one that changes the state of OTHER records.
+//
+// WHERE IT RUNS, and why no new gate. `commitlint.yml` already owns "the strings
+// that land on `main`" and is already one of the three required checks, so this
+// is a step there rather than a fourth blocking workflow (AGENTS.md § Governance:
+// a required check that does not RUN blocks the PR forever, and every guard's
+// cost is the SUM).
+//
+// WHAT IT READS. The PR body — the surface both incidents were on — and every
+// commit message in the PR, because a squash body is composed from those and
+// #1568 proves a `Closes` line there does close. Code is exempt: fenced blocks
+// and inline spans are blanked before the scan, so prose ABOUT this defect (this
+// header included) can spell the broken form without tripping the check, and so
+// a body quoting a log line is not read as an intent to close.
+//
+// Usage (in a pull_request job):
+//   PR_BODY=… PR_BASE_SHA=… PR_HEAD_SHA=… node scripts/check-closing-keywords.mjs
+
+import { execFileSync } from 'node:child_process';
+
+/**
+ * GitHub's closing keywords, verbatim from its documentation on linking a pull
+ * request to an issue. Mirrored here for the reason `check-pr-title-length.mjs`
+ * mirrors `header-max-length`: this list IS the claim being checked, there is no
+ * resolved config to read it from, and a missing entry fails open on exactly the
+ * spelling it forgot.
+ */
+const CLOSING_KEYWORDS = ['close', 'closes', 'closed', 'fix', 'fixes', 'fixed', 'resolve', 'resolves', 'resolved'];
+
+/** `#123` or `owner/repo#123` — the two forms GitHub resolves. */
+const REF = String.raw`(?:[A-Za-z0-9._-]+\/[A-Za-z0-9._-]+)?#\d+`;
+
+/**
+ * What separates a dangling reference from the one before it. Both spellings the
+ * two incidents used, plus the `and` form, which reads even more like a list that
+ * works: `Fixes #10 and #11`.
+ */
+const SEPARATOR = String.raw`(?:\s*,\s*(?:and\s+)?|\s+and\s+)`;
+
+const CHAIN = new RegExp(String.raw`\b(${CLOSING_KEYWORDS.join('|')})\s+(${REF})((?:${SEPARATOR}${REF})+)`, 'gi');
+
+/** Every reference inside an already-matched tail. Each is keyword-less by construction. */
+const TAIL_REF = new RegExp(REF, 'g');
+
+/**
+ * Blank out code, PRESERVING LENGTH so byte offsets still name the right line.
+ *
+ * Fenced blocks first, then inline spans — the other order would let a lone
+ * backtick inside a fence eat the rest of the document. Only the CONTENT is
+ * blanked; newlines and the delimiters themselves survive, so a fence is still a
+ * fence to the line counter.
+ *
+ * The terminator is `\2` or END OF INPUT, spelled `$(?![\s\S])` and not `$`.
+ * Under `m` a bare `$` matches every line end, and with a lazy body that is the
+ * FIRST one — so a three-line fence had its first line blanked and the rest read
+ * as prose. Measured while writing this: a fence containing the broken form
+ * failed the check on line 3 of its own example. The single-line vector below
+ * passed throughout, which is why there is now a multi-line one.
+ */
+function blankCode(text) {
+    const blank = (match) => match.replace(/[^\n]/g, ' ');
+    return text
+        .replace(/^([ \t]*)(`{3,}|~{3,})[^\n]*\n[\s\S]*?(?:^[ \t]*\2[^\n]*$|$(?![\s\S]))/gm, blank)
+        .replace(/`+[^`\n]*`+/g, blank);
+}
+
+/** 1-based line number of `index` in `text`. */
+function lineOf(text, index) {
+    let line = 1;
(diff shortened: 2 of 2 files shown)


JumpLink:
## Independent review

Reviewed on a separate worktree, `git log -1 -p` plus the branch as pushed. Every number below was re-derived here, not copied from the body. Three findings fixed on the branch in `6610e0d`, `47b2210`, `0b181d5`; the body prose is corrected too.

### What I broke on purpose

41 adversarial vectors against `danglingClosingRefs`, each with an expectation derived from GitHub's documented rule ("KEYWORD #ISSUE-NUMBER" or "KEYWORD OWNER/REPOSITORY#ISSUE-NUMBER", "use full syntax for each issue", and — the one that mattered — "The keywords can be followed by colons or in uppercase. For example: `Closes: #10`, `CLOSES #10`, or `CLOSES: #10`"). **15 of 41 disagreed with that rule. After the fixes, 7 — all of them deliberate, listed at the end.**

Plus 21 mutations of the detector, run against the self-test.

### Confirmed, fixed

**1. A colon after the keyword was invisible (false negative, on a documented spelling).** `CHAIN` required `\s+` directly after the keyword, so `Closes: #a, #b` closed `#a` and this check called the body clean. The header calls the keyword list "verbatim from its documentation" — the same paragraph of that documentation says the colon is part of the keyword's grammar. `&` and `;` were missed for the plainer reason that no separator claimed them. Fixed: `:?` in `CHAIN`, `[,;&]` in `SEPARATOR`. `/` and `+` were NOT added — neither appears in either incident and both widen the false-positive surface for nothing.

**2. A chain could cross a blank line (false positive, on a REQUIRED check).** `\s*` on both sides of the separator matched paragraph breaks, so

```text
Closes #99001

And #99002 is a follow-up.
```

was reported as a broken list and would have blocked a correct merge. Same for a next line opening `and …`. That is the one failure mode this check cannot have, and it was in the shipped version. Fixed: a new `GAP` spans at most one line break, and a bare `and` has to stay on the reference's own line. `Closes #99001,\n#99002` — a real wrap — still flags.

**3. A missing commit range reported clean (green that read nothing).** The header says "An empty commit range fails loudly rather than reporting clean". An *absent* one did not: `if (!baseSha || !headSha) return null` skipped the walk, and the run exited 0 with `Read the PR body (0 characters)`. Measured, on the shipped version with no `PR_*` environment at all:

```text
closing-keywords: self-test green — 25 vector(s). Read the PR body (0 characters); no closing keyword carries a reference it does not close.
exit=0
```

The absent range is the likelier accident — a renamed step, a dropped `env:` block, the workflow copied without it — and it is the surface #1568 closed all four of its issues from. It now refuses, the way `check-pr-title-length.mjs` refuses an empty `PR_TITLE`. An empty `PR_BODY` stays legitimate (GitHub sends null). Dropping the `null` case also deleted the `commits !== null` branch, the mutable `sources` array and the two-way success line.

Two smaller ones went in with it: `export function danglingClosingRefs` is a name no caller can reach (the module runs its CLI at import and exits) — dropped; and a single dangling reference read `#99002 are not closed by it`.

**4. Three claims in the header did not survive re-measurement.**

- #1567's body names **six** references, not five: the quoted line omits #1549. The real first line is `Closes #1453, #1547, #1546, #1549, #1551, #1555.` The table's "flagged — 5" is right; the quote and the "five issues" two paragraphs down were not.
- **#1549 closed anyway**, in the same second as #1453 (`2026-09-04T20:26:58Z`, `closed` event, commit `349d1fd`) — because #1567's commit body says `That closes #1549:` mid-sentence, at column 0. That is the same second surface that rescued #1568, occurring *inside the PR the header calls the failure*. It strengthens the argument; it is now in the header.
- "twice on 2026-09-04" and "same day" dated the **merges**, and the merges were `2026-09-04T20:26:57Z` (#1567), `2026-09-05T10:40:36Z` (#1565), `2026-09-05T09:46:29Z` (#1568). What happened on one day is the *writing*: all three bodies opened 18:22–18:42Z on 2026-09-04, all three in 0.48.0 (cut `2026-09-05T13:14:22Z`). The hand-closes were 2026-09-05, 14:47–14:48Z.

**Seven is right and stays** — four of #1567's six, three of #1565's four. Issue states and timestamps read off `/issues/N/timeline`.

### Measured, re-derived

Corpus, through the entry point, after the fixes — unchanged from the table:

| PR body | dangling refs |
|---|---|
| #1567 | 5 — #1547, #1546, #1549, #1551, #1555 |
| #1565 | 3 — #1550, #1516, #1529 |
| #1568 | 3 — #1531, #1533, #1548 |
| #1570 | clean |
| #1566 | clean |

**The self-test does hold the regression it was written for.** I reverted the fence terminator to a bare `$` and the suite went red on `inside a MULTI-LINE fence — the shape a one-line vector cannot see`. My first attempt at that mutation reported GREEN — the `perl -0pi` substitution had silently matched nothing. Worth saying out loud on a PR about checks that report success without doing anything.

17 distinct mutations. The 15 below each turn the self-test red: detector returns nothing · keyword becomes any word · `Refs` treated as closing · colon tolerance dropped · colon made mandatory · `[,;&]` narrowed to comma · gap widened to `\s*` · gap loses its newline · bare `and` may cross a newline · `#` anchor dropped from `REF` · code not blanked · inline spans not blanked · fence terminator bare `$` · `lineOf` off by one · `repaired()` joiner dropped. Two mutations of mine were no-ops and are reported as such rather than as coverage: `if (dangling.length === 0) continue` is unreachable (`match[3]` matched `(?:SEP REF)+`, so it holds at least one reference) — harmless, left alone — and `if (false)` on the range guard does not restore the old behaviour, so that one was measured directly on the shipped file instead.

Nine new vectors, 25 → 35, six flagged / three not. The `#` anchor and the blank-line gap each had **no** vector holding them; they do now.

**Workflow wiring, checked against the real run.** `base.sha`/`head.sha` are reachable under `fetch-depth: 0` — job `101325601238` on this PR logged `Read the PR body (3801 characters) and 1 commit message(s)`, so the step ran and the range resolved. `if: github.event_name == 'pull_request'` correctly skips the `push: [main]` trigger, which has no PR body. Locally, against real history: valid range → reads 1 commit; `base == head` → exits 1 loudly; unknown SHA → exits 1 with the `fetch-depth` hint; empty `PR_BODY` with a valid range → clean, correctly. No ReDoS — pathological inputs (4 000 backticks, 500 unterminated fences, 500 references, a 108 KB body) all complete in under 1 ms.

`oxfmt --check` and `oxlint` clean on every one of the three commits. `check-comment-budget` on `scripts/` was already over its ceiling before this PR (−326) and is advisory (`--warn`); this lands at −364.

### Deliberately left alone

- **Four-space indented blocks and `<!-- -->` comments are read as prose**, so a broken list in either is flagged. Blanking them would trade a rare false positive for a blind spot on the surface that actually does the closing: a commit message is not markdown, and #1549 closed on ordinary commit-body prose. Now stated in the header rather than left as an unmeasured edge.
- **`**Closes** #a, #b`** (emphasis splitting the keyword off its space) and a **bare issue URL** after a keyword. GitHub's documented syntax lists only `#N` and `OWNER/REPO#N` as closing references, so the URL case is not evidence-backed in either direction; both are named as limits.
- **`/` and `+` separators.** Invented by me, in neither incident, and each one widens what a required check can refuse.
- **`GH-123`.** Not listed as a closing reference on the page the keyword list mirrors; the detector ignores it and that matches the documentation.
- **No AGENTS.md change.** § Governance still holds: `main` is blocked by exactly three checks, and this is a step inside one of them, not a fourth. The architectural-decision list (package boundaries, API patterns, build, deps) does not reach a CI step, and the repo has no CONTRIBUTING.md for a "one keyword per issue" rule to live in.

### For the author

The **first commit's message** (`d1bc371`) still carries the wrong quote, "five", and "twice on 2026-09-04" — I fixed the header and the PR body but did not rewrite your commit. On a squash merge both texts land in the same body. Amend `d1bc371` before merging, or accept that the composed body corrects itself two commits later.

PR #1573, opened 2026-09-05
feat(adwaita)!: a page is chosen, not counted

JumpLink:
## What

[ADR 0048](https://github.com/gjsify/gjsify/blob/feat/portable-page-selection/docs/adr/0048-page-selection-by-identity.md). The three sibling ADRs each gave one thing a portable VALUE — a menu (0042), a list (0046), a range (0047). This one is different in kind: **there is no portable selection value to introduce.** The value already exists on every surface. What diverged is which of GTK's own shapes each port picked.

GTK offers three and chooses per widget. Measured against the in-repo GIR surface `packages/framework/gtk-host/src/generated/props.ts`:

| the shape GTK gives the selection | the types that have it |
|---|---|
| identity, by NAME — `visible-child-name` | `AdwLeaflet`, `AdwViewStack`, `GtkStack` |
| identity, by OBJECT — `selected-page` | `AdwTabView` |
| ORDINAL — `selected: number` | `AdwComboRow`, `AdwSidebar`, `GtkDropDown` |

A `GtkDropDown` selects a row of a `GListModel`, where a position IS the identity the model exposes; a view stack selects one of a set of NAMED pages. **The port took the ordinal in the two places GTK names an identity**, on both renderers. React Native — which has a view stack and no ordinal door — is the control: the converged shape was already shipping there.

The three ordinal widgets (`adw-sidebar`, `gtk-drop-down`, `adw-combo-row`) are deliberately untouched. The rule is *take the counterpart's shape*, not "identity always wins".

## Breaking

| published as | becomes |
|---|---|
| `visibleChildIndex` (setter) | `selectNthPage(n)`, or `visibleChildName` |
| `AdwTabView.selected` (settable ordinal) | `selectedPage` — the page on the web, the page **id** on NativeScript |
| `<adw-tab-view selected="1">` | `<adw-tab-view selected-page="<page id>">` |

The `visibleChildIndex` / `selectedIndex` getters stay. A position is what a switcher bar renders, and `notify::visible-child` already carries `index`.

**One migration does fail silently, and the first version of this body denied it.** A TypeScript error catches an unmigrated PROPERTY write. It does not catch `setAttribute('selected', …)`, and it does not catch NativeScript XML. Both bit this branch — see below.

## Why an ordinal is safe to read and unsafe to hold

`ViewStackState` carries the code that proves it: an insertion before the visible page increments the stored index (`view-stack.ts:296`), a reorder rewrites it (`:322`), a removal decrements it (`:348-349`). Every one of those lines exists so a number a caller HOLDS does not come to mean another page.

`selectNthPage(n)` is new on both view stacks: a call resolves against the page list as it is at that moment and keeps nothing. It is **not** an ordinal libadwaita refuses — `adw_tab_view_get_nth_page` (`adw-tab-view.h:192`) is public and `select_nth_page_cb` (`adw-tab-view.c:2137`) is exactly `get_nth_page` + `set_selected_page`, private only because a shortcut handler is not API. What is port-owned is putting it in the public surface.

It is also the only door onto two page shapes the specs already pin: a page whose name is `''` cannot be selected by name (`view-stack.spec.ts:226`), and where two pages share a name the lookup resolves to the FIRST (`:252`). A switcher bar renders button *n* over whatever the stack holds, blanks and duplicates included.

## What the independent review found (all fixed on the branch)

1. **The branch was red.** 2 of 1647 `@gjsify/adwaita-web` specs, plus a third passing vacuously — a `describe` block left on the retired attribute.
2. **The website wrote the attribute.** `CommandTabs.astro` and `AdwWidget.astro` both did `view.setAttribute('selected', …)`; the docs' command tabs and every gallery block's implementation tabs had gone dead. A grep for the property names cannot see it. The mechanism now lives in `check-website-attr-samples.mjs` — which already owned this rule and its incident (ADR 0046's `items` → `model`) and whose arms only read markup, so a scripted `setAttribute` was outside all of them. Arm 4 follows element-selector bindings to their attribute calls.
3. **`selectedPage` selected the wrong page** — a defect this branch introduced. Ids are unique within ONE view (`_nextId` counts per view), so `viewA.selectedPage = viewB.pages[2]` moved viewA to index 2 where C asserts `page_belongs_to_this_view`. **Fixed in the core, not in the renderers**: `TabViewState.setSelectedPage` takes the page or an id, checks identity against its own page list, and records the assertion C raises — a renderer-side guard was two copies with a silent refusal.
4. **NativeScript XML lost its only tab selection**, because a property holding an object is not an XML door. The setter takes the ID there, which is what a page handle already is on that port (`isClosing`, `closePage`, `setPagePinned` all take one). The gate's missing kind — a door whose string half NAMES something rather than serialising it — is written down in `status/open-todos.md` rather than worked around.
5. **Four ADR claims re-measured**, including the two corrected above.

## Measured

| | |
|---|---|
| `@gjsify/adwaita-core` | **2099 tests, 13049 assertions**, including the new vectors: a foreign page is refused with C's own assertion, this view's OWN page of the same id is accepted, and `interactive` survives both arms of the core's overload |
| `@gjsify/adwaita-web` | **1649 tests, 4980 assertions** — 1647 was the pre-fix total, two of them failing |
| `@gjsify/adwaita-nativescript` | 1244 tests, 3587 assertions |
| `check-vocabulary-alignment` | distance **7 → 5** property names; settable 143 → 142, agreeing 107 → 108 |
| `check-nativescript-xml-doors` | coercing / typed / un-XML-able: **64 / 71 / 19**. That triple IS the proof for finding 4 — an object-typed setter moved it to 64/70/20, and arm 1 used to `continue` past a `string`-classified door without counting it, so "classified" and "not read" printed identically |
| `check-adwaita-element-properties`, `check-generated-website-data`, `check-refs-citations` | green |
| `tsc` | `adwaita-core` emits, `adwaita-web` and `adwaita-nativescript` clean |

## Left open, in `status/open-todos.md`

- `isClosing` / `closePage` / `setPagePinned` still take an id where libadwaita takes an `AdwTabPage *`. Settle them together or not at all.
- A tab page with no `page-id` cannot be named from markup, and the reflection then writes a generated id into the DOM.
- The XML-door gate has no kind for "an object, or the key that names it". **`selectedPage` takes the page on the web and the id on NativeScript because of that gap** — the reasons given for the NativeScript shape stand on their own (an id already is the page handle there, and markup can carry nothing else), but the narrowing was triggered by a classifier this repo agrees is wrong, and should be re-taken on its own merits when the fourth kind lands.
- The property's getter returns the page on both surfaces while the NativeScript setter takes an id, so it does not round-trip there. Deliberate, and the first thing a consumer meets.

### DIFF
--- a/docs/adr/0048-page-selection-by-identity.md
+++ b/docs/adr/0048-page-selection-by-identity.md
@@ -1,0 +1,158 @@
+# 48. A selection takes the shape its counterpart gives it: a name, a page, or an ordinal
+
+- Status: **Proposed**
+- Date: 2026-09-05
+- Deciders: Pascal Garber
+- Related: [ADR 0004 (headless Adwaita core)](0004-headless-adwaita-core.md), [ADR 0027 (GTK host layer)](0027-gtk-host-layer.md), [ADR 0034 (widget vocabulary convergence)](0034-widget-vocabulary-convergence.md), [ADR 0042 (portable menu model)](0042-portable-menu-model.md), [ADR 0046 (portable list model)](0046-portable-list-model.md), [ADR 0047 (portable adjustment)](0047-portable-adjustment.md)
+
+## Context
+
+ADRs 0042, 0046 and 0047 each gave one thing a portable VALUE — a menu, a list, a range.
+This one is different in kind and it is worth saying so first, because the obvious reading
+of it is wrong: **there is no portable selection value to introduce.** The value already
+exists on every surface. What diverged is which of GTK's own shapes each port picked.
+
+GTK offers three, and it picks between them per widget. Measured against
+`packages/framework/gtk-host/src/generated/props.ts`, the in-repo GIR-derived surface:
+
+| the shape GTK gives the selection | the types that have it |
+|---|---|
+| identity, by NAME — `visible-child-name` | `AdwLeaflet`, `AdwViewStack`, `GtkStack` |
+| identity, by OBJECT — `selected-page` | `AdwTabView` |
+| ORDINAL — `selected: number` | `AdwComboRow`, `AdwSidebar`, `GtkDropDown` |
+
+The split is not an accident of history. A `GtkDropDown` selects a row of a `GListModel`,
+where a position IS the identity the model exposes; an `AdwViewStack` selects one of a set
+of NAMED pages, and a `GtkStack` page's name outlives an insertion before it. So an ordinal
+is right where the model is a list, and wrong where the pages have names of their own.
+
+**The port took the ordinal in the two places GTK names an identity**, and both renderers
+took it the same way:
+
+| widget | `@gjsify/adwaita-nativescript` | `@gjsify/adwaita-web` | `@gjsify/adwaita-react-native` | GIR |
+|---|---|---|---|---|
+| `AdwViewStack` | `visibleChildIndex` (settable) beside `visibleChildName` | the same pair | `visibleChildName` only | `visible-child-name` |
+| `AdwTabView` | `selected` (settable ordinal) | the same | — | `selected-page` |
+
+Two things follow from that table and both shaped this decision.
+
+**React Native is the control.** It carries a view stack and no ordinal door — its
+`visibleChildName` prop is the only way in, written that way from the start
+(`view-stack.native.tsx:65`). So the converged shape is not a proposal here; it is already
+shipping on a third renderer, and this ADR brings the older two to it rather than inventing
+anything.
+
+**The three ordinal widgets are already right and are deliberately untouched.**
+`adw-sidebar.selected`, `gtk-drop-down.selected` and `adw-combo-row.selected` name the same
+slot their counterpart names, in the same shape. A rule reading "identity always wins"
+would have broken all three, which is how a convergence programme starts producing
+divergence — the rule is *take the counterpart's shape*, and this ADR is one application
+of it.
+
+### What an ordinal DOOR costs, beyond disagreeing with the GIR
+
+The ordinal is derived, and `ViewStackState` already contains the code that proves it:
+inserting a page before the visible one increments the stored index
+(`view-stack.ts:296`), a reorder rewrites it (`:322`), a removal decrements it (`:348-349`).
+Every one of those lines exists so that a NUMBER a caller is holding does not silently come
+to mean another page. A name needs none of them — it survives all three operations
+unchanged. An ordinal is therefore safe to READ, at a moment, and unsafe to hold; a
+settable property invites exactly the holding.
+
+## Decision
+
+**An ordinal is a report. Identity is the door.**
+
+### 1. `AdwViewStack`: the settable ordinal becomes a call
+
+`visibleChildIndex` loses its setter on `@gjsify/adwaita-nativescript` and
+`@gjsify/adwaita-web`. The GETTER stays on both, unchanged: "which position is selected
+right now" is a legitimate question, it is what a switcher bar renders, and
+`notify::visible-child` already carries `index` in its payload for the same reason.
+
+Selecting is `visibleChildName`, which both renderers already have and which is
+`Adw.ViewStack`'s own property — **or `selectNthPage(n)`, which is new.** A method rather
+than a property because the difference is the holding: a call resolves against the page
+list as it is at that moment and keeps nothing, while a property invites a caller to hold
+a number that later means another page.
+
+**It is port-owned, and the reference says so**: `refs/libadwaita/src/adw-view-stack.h`
+declares no ordinal API at all, and neither does `adw-tab-view.h` — `select_previous_page`
+and `select_next_page` are the only relative moves there. So this is not GTK's shape being
+restored; it is the port's own, named for the `selectNthPage` the tab view has carried all
+along so that the two widgets ask the same question the same way.
+
+Without it the translation is `stack.pages[n]?.name`, and the first draft of this change
+wrote that line **three times** across the two switcher bars — the copied-arithmetic smell
+[ADR 0047](0047-portable-adjustment.md) found in `AdwSliderRow`, arriving in the same
+change that was supposed to reduce it. One method, three call sites, no copies.
+
+### 2. `AdwTabView`: `selected` becomes `selectedPage`, and it holds the page
+
+`selected` — a settable ordinal on both renderers — is replaced by `selectedPage`, which is
+`Adw.TabView`'s own property name and, as there, holds the PAGE rather than a number. The
+core already models exactly this (`TabViewState.selectedPage`), so the renderers expose a
+value that existed and was not reachable from the outside except through
+`setSelectedPage(id)`.
+
+`selectedIndex` and `selectedId` stay as the getters they already are. Neither is a door.
+
+### 3. On the web, the MARKUP door names the page too
+
+`<adw-tab-view selected="1">` becomes `<adw-tab-view selected-page="<page id>">`, reflected
+on every selection change as the ordinal was.
+
+Markup cannot hold an object, so the attribute holds the page's **id** — the one
+`<adw-tab-page page-id>` already declares, and the reason that attribute exists. A page
+that declares none gets a generated id, which an author cannot predict and therefore cannot
+select by; that is the same bargain `<adw-view-stack visible-child-name>` has always made
+for a page with no name, and it is why the parse rule is *ignore an id no page holds*
+rather than *fall back to the first page*.
+
+`@gjsify/adwaita-nativescript` needs no equivalent: its XML door is the property, and the
+property now holds a page.
+
+### 4. The core keeps its ordinal API, and that is the point of the split
+
+`ViewStackState.setVisibleIndex` and `TabViewState.selectNthPage` are unchanged, and the
+conformance vectors that drive them (`selectIndex` ops in
+`@gjsify/adwaita-core/conformance/view-stack`) are unchanged with them.
+
+A switcher bar IS ordinal: it renders button *n* and must select page *n*. That translation
+belongs where the ordinal is a fact about a rendered list, not on the widget's public
+surface where it is a promise about a page. So the two switcher bars translate through
+`pages[n].name`, which is one lookup and is what makes the promise true.
+
+## Consequences
+
+- **Breaking**, on two published packages. `stack.visibleChildIndex = 2` becomes
+  `stack.selectNthPage(2)` (or `stack.visibleChildName = '…'`), and `view.selected = 2`
+  becomes `view.selectNthPage(2)` — a method both tab views already had — or
+  `view.selectedPage = page`. Every migration is one line and mechanical, and none has a
+  silent failure mode: the removed setters were the only writable properties of their name,
+  so an unmigrated write is a TypeScript error and, at runtime, a no-op on a getter.
+- The vocabulary ledger loses two `should converge` property entries. That is the metric
+  moving because the thing it measures moved: `check-vocabulary-alignment.mjs` counts
+  SETTABLE properties, because a door is what a portable authored tree writes.
+- No XML door changes. `visibleChildIndex` and `selected` were reachable from NativeScript
+  XML as numbers; `visibleChildName` is a string, which is what an XML attribute is anyway,
+  and `selectedPage` holds an object, which XML cannot express and therefore never offered.
+- Nothing in `website/`, `showcases/`, `packages/framework/stories` or the generated
+  NativeScript templates writes either property — measured before writing this, which is
+  why this ADR proposes a removal rather than a deprecation.
+
+## What this does not decide
+
+- **`visible-child` (the widget) is not added.** `Adw.ViewStack` also selects by CHILD
+  WIDGET, and neither renderer offers that. It is a third door onto the same slot and
+  nothing has asked for it; the ADR that adds it should say what it is for.
+- **`AdwLeaflet` and `GtkStack` are out of scope** — neither renderer ports them today.
+  When one does, the table above is the rule it inherits.
+- **Ordinal REPORTS elsewhere are untouched**, including every `selected` on the three
+  widgets whose counterpart is ordinal.
+- **`setSelectedPage(id)` keeps its id parameter**, so the two renderers now offer the page
+  through a property and its id through a method. libadwaita's own
+  `adw_tab_view_set_selected_page()` takes the PAGE, so the method diverges from its
+  counterpart in shape — but a method is not what a portable authored tree writes, the
+  vocabulary ledger does not read one, and changing it is a second breaking change with no
+  measurement behind it yet. It is written down in `status/open-todos.md` instead.
--- a/docs/adr/0048-page-selection-by-identity.md
+++ b/docs/adr/0048-page-selection-by-identity.md
@@ -76,16 +76,43 @@ than a property because the difference is the holding: a call resolves against t
 list as it is at that moment and keeps nothing, while a property invites a caller to hold
 a number that later means another page.
 
-**It is port-owned, and the reference says so**: `refs/libadwaita/src/adw-view-stack.h`
-declares no ordinal API at all, and neither does `adw-tab-view.h` — `select_previous_page`
-and `select_next_page` are the only relative moves there. So this is not GTK's shape being
-restored; it is the port's own, named for the `selectNthPage` the tab view has carried all
-along so that the two widgets ask the same question the same way.
+**It is port-owned as a SETTER, and the reference is more precise than that**: neither
+`refs/libadwaita/src/adw-view-stack.h` nor `adw-tab-view.h` declares a function that
+SELECTS by position — `select_previous_page` / `select_next_page` are the only relative
+moves. An ordinal is not absent from either, though, and pretending it is would make the
+weaker argument:
+
+- `adw_tab_view_get_nth_page (AdwTabView *self, int position)`
+  (`refs/libadwaita/src/adw-tab-view.h:192#adw_tab_view_get_nth_page`) is a public ordinal
+  ACCESSOR.
+- `select_nth_page_cb` (`refs/libadwaita/src/adw-tab-view.c:2137#select_nth_page_cb`) is
+  libadwaita's OWN ordinal selection — the
+  Alt+1…Alt+9 / Alt+0 shortcut — and it is exactly this composition:
+  `adw_tab_view_get_nth_page()` then `adw_tab_view_set_selected_page()`. Private, because a
+  shortcut handler is not API.
+- both widgets hand out a `GtkSelectionModel`
+  (`refs/libadwaita/src/adw-view-stack.h:166#adw_view_stack_get_pages`;
+  `refs/libadwaita/src/adw-tab-view.h:268#adw_tab_view_get_pages`), and selection on one is
+  positional.
+
+So `selectNthPage` is not an ordinal libadwaita refuses; it is the ordinal libadwaita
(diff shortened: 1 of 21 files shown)


JumpLink:
## Independent review

Reviewed as someone who did not write this, in a worktree at `feat/portable-page-selection`.
Five commits pushed. **The branch was red when I started** and the PR's "Measured" table does
not list a test run.

### What I broke on purpose

| broke | to prove | result |
|---|---|---|
| reinstated `CommandTabs.astro`'s `setAttribute('selected', …)` | the new gate is not decorative | fails, naming file, tag and the four attributes the element *does* observe |
| removed the `(?<![\w$.])` lookbehind from the arm-4 receiver regex | its fixtures are not decorative | the "member access that merely ENDS in the binding name" fixture fires |
| wrote the cross-view selection test before the guard | the defect is real, not reasoned | `viewA.selectedPage = viewB.pages[2]` moved viewA to index 2 |
| a wrong method + wrong return type on the `Adw.TabView`-typed website receiver | the type resolves rather than falling back to `any` | `TS2339` + `TS2322`, so `astro check` covers the new call |

### What I re-derived and it held

- **The three-shape table is exact.** `packages/framework/gtk-host/src/generated/props.ts`: `visibleChildName` on `AdwLeaflet`(1515) `AdwViewStack`(2349) `GtkStack`(5342); `selectedPage` only on `AdwTabView`(2226); `selected?: number` only on `AdwComboRow`(1146) `AdwSidebar`(1927) `GtkDropDown`(3225). No fourth spelling anywhere in the file.
- **The `ViewStackState` citations land.** `:296` is the insert increment, `:322` the reorder rewrite, `:348-349` the removal — read, not trusted.
- **React Native really has no ordinal door.** `AdwViewStackProps` (`props.ts:920`) declares `pages` + `visibleChildName` and nothing else; `view-stack.native.tsx:65` is the `visibleChildName` parameter; RN carries no tab view at all.
- **The vocabulary numbers are exact.** Ran the gate against `HEAD~1`'s files and then HEAD: 143 → 142 settable, 107 → 108 agreeing, distance 7 → 5. Both entries left `should converge` for the reason claimed.
- **The `open-todos` anchor is right.** `refs/libadwaita/src/adw-tab-view.h:150` really is `adw_tab_view_set_selected_page`.
- **`showcases/`, `packages/framework/stories` and the NS templates write neither property.** Confirmed.
- **`interactive` is preserved.** Old web attribute path → `selectNthPage(n)` → `setVisibleIndex(n, true)`. New → `setSelectedPage(id)` → `interactive = true` default → `setVisibleIndex(pos, true)`. Same on the property setters.
- **The conformance vectors are untouched by construction.** `applyTabViewOp` already drove `setSelectedPage` / `selectNthPage`; neither changed. The `if (vector.diagnostics)` assertion at `tab-view.spec.ts:135` runs against a view mounted EMPTY, so no attribute path reaches it.
- **The switcher-bar short-circuit is intact**, and `xmlNumber` is still used by 12 other widgets.

### Test and gate counts (measured here, not claimed)

| | |
|---|---|
| `@gjsify/adwaita-web` (firefox, tests/browser) | **was 2 of 1647 failing** → now **1649 / 1649, 4976 assertions** |
| `@gjsify/adwaita-nativescript` (node) | 1244 / 1244, 3587 assertions |
| `@gjsify/adwaita-core` (node) | 2097 / 2097, 13041 assertions |
| gates | `vocabulary-alignment`, `adwaita-element-properties`, `generated-website-data`, `storybook-widget-coverage`, `nativescript-xml-doors`, `adwaita-conformance-drivers`, `website-attr-samples`, `adr-index`, `refs-citations`, `agent-context-size` — all green |
| oxfmt / oxlint | clean on every changed file |

Not run: the GJS leg of the NS suite and the whole-workspace build. `gjsify run build` fails in `@gjsify/gtk-host` on a `@girs/gtk-4.0` mismatch in this worktree's `node_modules` (`Gtk.SvgWidget`, `SignalSignatures`), which predates this branch and left `@gjsify/web-streams` unbuilt, which is what the `--app gjs` test bundle cannot resolve. Nothing about that touches these edits, and it is why the NS `tsc` residual is all "no exported member from `@gjsify/adwaita-core`" — `packages/web/adwaita-core/lib/types/` does not exist here.

### Five findings

**1 — the website wrote the attribute, and it went silently dead. `fix(website): the tab door the element still reads`**

> "Nothing in `website/`, `showcases/`, `packages/framework/stories` or the generated NativeScript templates writes either property — measured before writing this"

Two files did:

- `website/src/components/CommandTabs.astro:169` — the npm / yarn / gjsify runtime tabs, restored from `localStorage` on load and mirrored across every command window on the page.
- `website/src/components/AdwWidget.astro:596` — the implementation tabs of every block in the Adwaita gallery.

Both spelled it `view.setAttribute('selected', String(index))`, which is why a grep for the PROPERTY names missed them. `selected` is no longer observed, so both writes became no-ops: the stored preference stopped applying and stopped mirroring, with no error anywhere. That is also the counterexample to

> "none has a silent failure mode: the removed setters were the only writable properties of their name, so an unmigrated write is a TypeScript error"

— true of the property, false of the attribute, and false of NativeScript XML. Migrated to `selectNthPage(index)`, which is the ADR's own prescription, with the receiver typed `Adw.TabView` so the call is under `astro check` where the attribute never was.

**2 — the branch was red. `test(adwaita-web): the attribute it observes`**

`describe('adw-tab-view selected attribute')` in `packages/web/adwaita-web/src/tab-view.spec.ts` was left on the old attribute while the element moved. Measured in firefox: 2 of 1647 failing — `selects the declared index instead of the auto-pick` (expected 2, got 0, because `selected="2"` is not read and the auto-pick stands) and `IGNORES an out-of-range or unparseable index` (expected 1, got 0). The third case passed VACUOUSLY: it asserts `hasAttribute('selected')` is false on an emptied view, and that is now false at every moment of the element's life.

Migrated, and given the case the change needed and did not have — see finding 4.

**3 — the mechanism, in the file that already owns the rule. `feat(scripts): gate the scripted attribute writes`**

`check-website-attr-samples.mjs` arm 3 already carries this exact rule *and* its incident: ADR 0046 renamed `<adw-combo-row items>` to `model` and four call sites kept writing the dead name. Arm 3 reads MARKUP, so `el.setAttribute('x', …)` was outside it, and ADR 0048 walked into the same hole one ADR later.

Arm 4 follows a binding taken from a custom-element selector (`querySelector`, `querySelectorAll`, `createElement`, `for … of` included) to its `set/get/has/removeAttribute` calls and holds each name against the same `observedAttributes` list. No new script, no new CI step — it is already wired at `audit-runtimes.yml:751`. Deliberately shallow, so it reports how many receivers it RESOLVED and fails at zero rather than reporting a clean site: `18 fixtures and 188 cells across 39 gallery blocks, 0 unobserved attribute(s) written; 0 scripted attribute name(s) on 3 resolved element binding(s)`.

**4 — `selectedPage` selects the wrong page. `fix(adwaita): a page belongs to ONE view`**

The property holds the page; both setters pass its **id**. An id is unique within a view and never across two — a declared `page-id` is the author's to repeat, and the web element's `_nextId` counts per element, so two views both mint `tab-1`. Measured before the guard:

```
viewA.selectedPage = viewB.pages[2]   →  viewA.selectedIndex === 2
```

where `adw_tab_view_set_selected_page` asserts `page_belongs_to_this_view` and refuses. This is a defect the branch INTRODUCES: before it, the tab views had no page-taking door, so the mistake was unreachable. Both setters now refuse by object identity, pinned by a browser vector. It is also, exactly, the measurement `status/open-todos.md` said the id-vs-page question lacked — see "what you must decide".

Separately: an unknown `selected-page` id and `selectedPage = null` with pages present now push an `assertionDiagnostic` where the ordinal path pushed none. Correct (it is where C's assertions live), unmentioned by the ADR, and now asserted rather than left to be discovered.

**5 — four claims in the ADR. `docs(adr): four claims in 0048, re-measured`** (the PR body carries the same four)

- *"`refs/libadwaita` gives neither widget an ordinal API at all"* — an ordinal SELECTOR, yes. But `adw_tab_view_get_nth_page` (`adw-tab-view.h:192`) is a public ordinal accessor, `select_nth_page_cb` (`adw-tab-view.c:2137`) is libadwaita's own ordinal selection and is literally `get_nth_page` + `set_selected_page`, and both widgets hand out a positional `GtkSelectionModel`. The corrected reading is the **stronger** argument: `selectNthPage` is not an ordinal libadwaita refuses, it is the one it performs, promoted out of a private shortcut handler. The four new coordinates are anchored, so CI reads them at the cited text (anchored citations 4 → 8).
- *"No XML door changes."* — `check-nativescript-xml-doors`, before → after: coercing non-string setters **66 → 64**, setters XML cannot carry **19 → 20**. Two XML doors were removed and none replaced them; `<AdwTabView selected="1">` worked through `xmlNumber` and is now silently dead. See "what you must decide".
- *"none has a silent failure mode"* — finding 1.
- *"Nothing in `website/` …"* — finding 1.

I also replaced the `selectNthPage` justification with a stronger one. The "three copies of `pages[n]?.name`" argument is about tidiness; the load-bearing fact is that `pages[n].name` is **not a translation of `selectNthPage(n)` at all** for two page shapes your own specs already pin: a page whose name is `''` (`view-stack.spec.ts:226`) and two pages sharing a name, where a lookup resolves to the first (`:252`, written against exactly that bounce). A switcher bar renders button *n* over whatever the stack holds. The method is not a convenience over the name; it is the only door onto the pages the name cannot reach. That answers "is this a second door the ADR argues against" — no: it is the *first* door for those pages.

### What I deliberately left, and what you must decide

- **NativeScript XML lost its only tab selection.** After this change `AdwTabView` has no string-typed selection setter at all, and `selectedPage` holds an object XML cannot carry — so an NS XML tree can no longer declare which tab starts selected, while the web kept `selected-page`. Recorded in `status/open-todos.md`. The candidate is a `selectedPageId` string setter mirroring the web attribute, which is the same question as the next bullet, so I did not add it unilaterally.
- **The id-vs-page shape belongs in the core, and now has its measurement.** My guard is two copies of a one-line identity check in two renderers, and its refusal is SILENT where C `g_critical`s — because the assertion lives in `TabViewState.setSelectedPage` behind an id that a colliding page never gets past. Letting `setSelectedPage` take the PAGE beside the id puts the check and its diagnostic in the one place both renderers share and **breaks no caller** (every existing one passes a string), which is a materially different proposition from the "second breaking change with no measurement" the todo describes. Your call, and it is the one bullet I would not merge without.
- **The NativeScript half of finding 4 is untested**, structurally: no spec in that package can import a widget module, because `extends GridLayout` evaluates the bare `@nativescript/core` specifier at module-eval (`tab-view.spec.ts` header says so). The web twin carries the vector for both. Not something to fix in this PR.
- **The `_resolveStack` cast in the web switcher bar** (`document.getElementById(value) as AdwViewStack`) turns a mis-targeted `stack="…"` from a silent no-op into a `TypeError` on click. I left it: a foreign element yields `pages → []`, so zero buttons render and no click handler exists to throw. Worth knowing, not worth code.
- **I did not merge.**

JumpLink:
## Independent review — round 2 (`6c5a1cfbd` + `35a0fe5d8`)

Reviewed only the two new commits. Two more pushed on top; nothing merged.

### 1. The recursion is right — all three sub-questions measured

Probed `TabViewState` directly (`lib/esm/tab-view.js`), not reasoned:

| | result |
|---|---|
| `interactive` via the PAGE overload, `false` | notified `interactive: false` ✔ |
| `interactive` via the PAGE overload, default | `true` ✔ |
| `interactive` via the ID arm, `false` | `false` ✔ |
| foreign page, colliding id | refused, **1** diagnostic, selection unmoved ✔ |
| unknown id | refused, **1** ✔ |
| own page | accepted, **0** ✔ |
| `null` with pages | refused, **1** ✔ |
| `undefined` / a page-shaped object literal | refused, **1** each ✔ |

**No double-push anywhere**, and the reason is structural: the page arm `return`s before it recurses, so the id arm never runs for a page the view does not hold. The one path that reaches the id arm's diagnostic through the page arm is a page whose `id` was mutated after insert — `pages.includes` passes, `indexOfName` fails, exactly one `page_belongs_to_this_view`, selection unmoved. `AdwTabPageState.id` is `readonly`, so that is unreachable in typed code and correct when it is not.

I also checked the assumption the whole design rests on — that a held page keeps its identity. `TabViewState.pages` memoises a frozen array over `content` objects the stack owns, and every `_pages` mutation nulls the memo. Measured across `setPageTitle` / `setPageLoading` / `setPagePinned` on the held page, an unrelated `appendPage`, a `reorderPage` of the held page and a `closePage` of another: identity kept in all seven, `setSelectedPage(held)` accepted in all seven, 0 diagnostics. (Control: the title and pinned flag really did change on the held object — the records are mutated in place, so `readonly` is annotation only. That is what makes a held page safe, and it is worth knowing it is *implementation*, not type.)

**Two things the vectors did not hold, now fixed** — `d677f78c5`:

- **`interactive` had no witness at all.** A recursion that dropped the flag reads identically at the call site and would notify `interactive: true` for a model-driven selection. Verified by breaking it: removing the argument fails exactly the new case, 1 of 2099.
- **`expect(last?.includes('page_belongs_to_this_view')).toBe(true)`** is satisfied by a double push as happily as a single one — and a double push is *the* plausible failure of this recursion. Now the exact text and exact length, before and after, plus the accepted path recording nothing.

### 2. The NS door is back, and now the gate can say so — `fef38ffa7`

`AdwTabView.selectedPage` is classified **`string`**, measured by running `settersOf` + `attributeKind` over the source: `string | null` → parts `['string']` → `string`. Same kind as `defaultIcon`, an existing door. Not exempt by any rule.

But the gate could not tell you that. Arm 1 counted the coercing setters and the uncarryable ones and skipped `string` at a bare `continue` — so "classified `string`" and "the reader stopped resolving `string`" printed the same summary. That is the one bucket that did not add up, in a file whose own header argues at length that they must ("the totals have to add up to the setters that exist, or *all of them coerce* is a claim about the parser"). Counted now, with a floor:

| | coercing non-string | string, as-is | XML cannot carry |
|---|---|---|---|
| before this ADR | 66 | 70 | 19 |
| with your first object-typed setter | 64 | 70 | **20** |
| as it stands | 64 | **71** | 19 |

Verified by breaking it: retyping the setter `AdwTabPage | null` moves 71 → 70 and 19 → 20 in the same run. `selected` and `visibleChildIndex` left the numeric bucket, `selectedPage` joined the string one — two doors removed, one added, nothing left in the bucket an attribute cannot reach.

On "does the string reach the setter as an attribute would": the setter takes the raw string and hands it to `setSelectedPage`, which is a lookup — no coercion to get wrong, which is exactly why the gate has no rule for string doors. What no gate anywhere does is *exercise* a string door at runtime; that is uniform across all 71 and not something this PR should change.

### 3. The specs still mean what they say — and one meant less than it did

`RECORDS the refusal, where the ordinal door recorded nothing` is untouched by your change: the attribute path was always `setSelectedPage(string)`, and `selectedPage = null` never enters the page arm. Still green, still asserting the exact list.

`REFUSES a page belonging to ANOTHER view` **did** quietly stop measuring the interesting half. I wrote it against a silent renderer guard, so it asserted only that the selection had not moved — which is now the weaker of the two facts. It asserts C's assertion in the refusing view, an empty list in the other one, and that the same view's OWN page of that id is still accepted, which is what makes it a test of identity rather than of the id.

| | |
|---|---|
| `@gjsify/adwaita-web` (firefox) | **1649 / 1649, 4980 assertions** |
| `@gjsify/adwaita-core` (node) | **2099 / 13049** |
| `@gjsify/adwaita-nativescript` (node) | **1244 / 3587** |
| `tsc` | adwaita-core, adwaita-web, adwaita-nativescript — **0 errors each**, with the stale `tmp/.tsbuildinfo` cleared. Negative control: two deliberate errors in the NS setter surface as `TS2339` + `TS2322`, so that zero is real |
| gates | vocabulary-alignment, adwaita-element-properties, generated-website-data, storybook-widget-coverage, nativescript-xml-doors, adwaita-conformance-drivers, website-attr-samples, adr-index, refs-citations, agent-context-size — all green |
| oxfmt / oxlint | clean |

### 4. The asymmetry: right call, wrong citation, and nothing measures it

I do not disagree with the decision. I disagree with what you cited for it, on three counts, all checkable:

- **"converges in NAME, never in shape" is not § 1's rule.** It is the gloss on ONE declaration kind, `{ composes: [...] }` — "the same UI, assembled differently because the platform forces it… `adw-image-button` is this". That kind is about a widget diverging from its **GIR counterpart**, not about two renderers diverging from each other. § 1's actual rule is about widget NAMES and namespace exports.
- **The clause that does apply is § 1 clause 3**: "declares every divergence from 1 and 2, with a reason. An UNDECLARED divergence is what goes red." That is satisfiable and now satisfied — by the comments on both setters and by the table I added to ADR 0048 § 2.
- **Nothing measures it either way.** `check-vocabulary-alignment`'s property half compares a NS property name against `interfaceKeys(...)` — a **key set**. It never reads types. `selectedPage: string | null` agrees with `AdwTabViewProps.selectedPage?: Adw.TabPage | null` purely because the name matches. So 0034's machine reading neither sanctions nor objects; it is blind by construction. I would not build a cross-renderer shape gate — shapes legitimately differ per platform, and clause 3's declaration is the right instrument.

On the merits I would sign off, with the cost stated plainly rather than left for a reader to hit: the two NS accessors now have unrelated types (legal since TS 5.1 — I checked that it compiles, it is not an oversight), so `view.selectedPage = view.selectedPage` is a type error. **The thing I would push back on is the causal order.** `AdwTabPage | string | null` is what the core takes, makes the round-trip legal, and is what your own first commit had; you narrowed it because `attributeKind` files `<object> | string` as `json` and the door goes red. That is a published API shape chosen by a classifier you agree is wrong — and the commit message's "not a concession to the gate" sits uneasily beside the open-todos entry that says the annotation "turns the door red". Your independent reasons are good, so nothing needs undoing; but when the fourth kind lands, this decision should be re-taken on those reasons alone rather than inherited.

### 5. Two stale claims and two wrong numbers

**In the tree** (fixed in `fef38ffa7`): ADR 0048 **§ 3 still said NativeScript has no XML door**, cited `19 → 20`, and pointed at the `status/open-todos.md` entry you deleted. All three were written before the setter took the id. § 2 still said the property "holds the PAGE" full stop, with no mention of the asymmetry. Both now carry the measured table and the declared divergence.

**In the PR body** — two numbers do not match the tree, both understatements:

| says | should say |
|---|---|
| `@gjsify/adwaita-web` — 1647 specs, green | **1649 / 1649, 4980 assertions**. 1647 was the total *before* the fixes (2 of which were failing); the migrated block and two new vectors took it to 1649 |
| `@gjsify/adwaita-core` — 2098 tests, 13046 assertions | **2099 / 13049** |
| `check-nativescript-xml-doors` — OK, 45 elements | worth the numbers instead: **64 coercing / 71 string / 19 uncarryable**, since that triple *is* the proof that finding 4 is fixed |

Everything else in the body re-derives: the three-shape table, `view-stack.ts:296/:322/:348-349`, `adw-tab-view.h:192` and `adw-tab-view.c:2137` (read against `refs/libadwaita` in a checkout that has it), `view-stack.spec.ts:226/:252`, "2 of 1647 plus a third passing vacuously", and 7 → 5 / 143 → 142 / 107 → 108. I did not edit the body myself — editing it re-queues `Lint commit messages`, which then blocks the merge while `gh pr checks` still reads green.

One gap: the body's "Left open" does not mention the asymmetry or the unrelated get/set types. It is the decision a consumer meets first.

### Verdict

Both commits are better than what they replace, and the core one is the right place for the check. Nothing blocking. **Not merged.**

PR #1574, opened 2026-09-05
feat(adwaita-nativescript)!: the sheet is just open

JumpLink:
## What

`AdwBottomSheet.openState` was **the one entry in the property ledger the convergence rule did not explain**, and ADR 0034 § *What the remainder is* said why:

> both sides hold a BOOLEAN, so the rule says converge. What stops it is a collision inside JavaScript — `open` is taken by the class's own `open()` method, and libadwaita gives the type no method names to rename those to, so converging would trade a declared property divergence for an invented, undeclared method one.

The premise is right and the conclusion had one option too few. **Renaming the methods was never the only way out of the collision; deleting them was.**

Measured against `refs/libadwaita/src/adw-bottom-sheet.h`:

```console
$ grep -c 'adw_bottom_sheet_open\|adw_bottom_sheet_close' refs/libadwaita/src/adw-bottom-sheet.h
0
$ grep -n 'adw_bottom_sheet_set_open' refs/libadwaita/src/adw-bottom-sheet.h
50:void     adw_bottom_sheet_set_open (AdwBottomSheet *self,
```

The type has no open/close pair at all — `adw_bottom_sheet_set_open()` is the setter of this very property. So `open()` and `close()` were port-owned conveniences whose entire body was `this.openState = <bool>`, and they were the reason the property could not carry its own GIR name.

## Breaking

| published as | becomes |
|---|---|
| `sheet.openState` | `sheet.open` |
| `sheet.open()` | `sheet.open = true` |
| `sheet.close()` | `sheet.open = false` |
| `<adw:BottomSheet openState="true">` | `open="true"` |

Eleven call sites, all migrated in this PR: the widget, its spec, the NativeScript storybook, the gallery XML and its expected tree, the generator that writes both, the shipped template data, and two docs snippets.

`requestClose(source)` is untouched — it is the `can-close`-respecting dismissal and always was a different thing from the property.

## No new ADR; ADR 0034 gets Amendment 11

The rule was already decided (ADR 0034 clause 1); this is an application of it, the way #1569's `AdwIcon` → `GtkImage` rename was. What the amendment records is not the verdict but **the shape of the mistake**: the reasoning ruled out a convergence by considering exactly one repair for the obstacle and finding it worse than the disease. The passage that said so is left as written.

> A collision with a port-owned method is not a reason a name cannot converge — it is a question about the method, and the question is whether the counterpart has one.

`check-vocabulary-alignment.mjs` carries that sentence in its ledger header now, where the next entry of this shape will be read.

## Measured

| | |
|---|---|
| `check-vocabulary-alignment` | distance **7 → 6** property names on this base |
| `check-nativescript-xml-doors` | OK — 45 elements; the boolean door still goes through `xmlBoolean` |
| `check-generated-website-data` | OK — 28 XML templates byte-identical to the view the probe app inflates |
| `check-refs-citations` | green; the new citation is anchored for the CI run that has `refs/libadwaita` |

## Note on the stack

Independent of #1573 (`a page is chosen, not counted`), which removes two other ledger entries. Both edit `NS_PROPERTY_ALIGNMENT` in different places; whichever lands second may need a trivial rebase. Together they take the distance 7 → 4.


JumpLink:
## Independent review

Reviewed on a worktree of `feat/bottom-sheet-open` at `6bd831f8`, node_modules hardlinked from a
built checkout, deps built in-tree. Everything below is a command I ran, not a reading of the diff.
Three findings fixed and pushed as `957fefda`, `54c596f3`, `479cefd6`.

### The claims re-derived — all of them hold

| claim | how I checked | result |
|---|---|---|
| `grep -c 'adw_bottom_sheet_open\|adw_bottom_sheet_close'` is 0 | ran it on `refs/libadwaita` at the ADR's pin `42f647ff` | **0** |
| `:50` is `adw_bottom_sheet_set_open` | `grep -n` | `50:void     adw_bottom_sheet_set_open (AdwBottomSheet *self,` |
| distance 7 → 6 | ran `check-vocabulary-alignment.mjs` on the base (`68461e4a`) and on the branch | base `7 property name(s)` / `107 agree, 36 do not (7 should converge)`; branch `6` / `108 agree, 35 do not (6 should converge)` |
| the citation is anchored for a CI run that has the submodule | symlinked `refs/libadwaita` in and re-ran `check-refs-citations` | `118 resolved … 2 of the tree's 4 #anchor(s) verified`. A/B: changing `:50` to `:51` turns it red with `line 51 does not contain it` |
| the generator is idempotent | `node scripts/generate-adwaita-nativescript-templates.mjs` then `git status` | `28 template(s), 12 refusal(s)`, working tree clean |
| no call site missed | `grep -rn openState` over the whole tree — `.xml`, `.mdx`, `expected.ts`, the generator, `website/src/data`, every `lib/`-shipped path | 3 files, all deliberate prose (the ADR, the widget doc, the ledger header). Nothing left to migrate |

**"Eleven call sites" is a real number wearing the wrong sentence.** Measured per file, the lines
removed that contained `openState`: spec 2, storybook 3, gallery XML 1, `expected.ts` 1, generator 1,
`adwaita-nativescript-templates.ts` 1, `navigation.mdx` 2 — **11**, and the widget adds 4 more on top.
So eleven is every such line *outside* the widget. The sentence that carries it lists nine places and
counts the widget among them, and 3 of the 11 are comments rather than call sites. Number right,
derivation not reconstructible from the prose. Body-only, nothing in the tree says it.

### What the rename could break — measured, and it breaks nothing

`open` is a common name, so I went looking for the collision rather than trusting that the compiler
would have found it:

- **No `open` on any base class.** Fetched `packages/core/{ui/core/view/index.d.ts,
  ui/core/view-base/index.ts, ui/layouts/layout-base.d.ts, ui/layouts/grid-layout/index.d.ts,
  data/observable/index.ts}` from `NativeScript/NativeScript@main` and grepped: no member named
  `open` anywhere on `Observable → ViewBase → View → LayoutBase → GridLayout`. The in-repo ambient
  `src/ns-core.d.ts` declares none either, and `gjsify tsc --noEmit` on the package is **exit 0**.
- **Not a setter-only base accessor.** `SETTER_ONLY_ON_BASE` is `rows`/`columns`; arm 2 of
  `check-nativescript-xml-doors` reports `2 setter-only base accessor(s) unshadowed`.
- **The XML door still coerces.** `set open(value: boolean | string)` goes through `xmlBoolean`;
  gate reports `66 non-string setter(s) on 47 XML-reachable class(es), all coercing`.
- **The `notify::open` payload was never keyed on the old name** — `NOTIFY_OPEN = 'notify::open'`
  and `NotifyOpenEventData.open` were already the GIR spelling. The property now *matches the event
  it emits*, which it did not before.
- **`@gjsify/adwaita-core` lines up unchanged** — `BottomSheetPresentation.open`/`.setOpen`, which the
  accessor forwards to verbatim.
- **No props-bag collision.** The gallery probe reads declared attributes by name off the instance
  (`gallery-page.ts` `readBack`), so an accessor answers exactly as the old one did; `args.open` in the
  storybook is a plain key on `StoryView.args` and shadows nothing. `check-storybook-control-parity`:
  `291 (control, rendering) pairs — 283 wired`, `261 write(s) … held against that widget's members`.

Two things the PR could have claimed and did not:

- **All three renderers now spell it the same.** `@gjsify/adwaita-web`'s `<adw-bottom-sheet>` already
  had `get open()`/`set open()`, `gtk-host`'s generated `AdwBottomSheetProps` has `open?: boolean`, and
  the GTK story already wrote `sheet.open = !sheet.open`. After this the NS story is line-for-line the
  GTK one. That is a stronger argument than the ledger delta.
- **Upstream names the property as the programmatic path itself.**
  `refs/libadwaita/src/adw-bottom-sheet.c:2071` — "Bottom sheet can still be closed using
  [property@BottomSheet:open]." libadwaita's own answer to "how do I close it past `can-close`" is the
  property, not a method. The spec already cites that line; the ADR amendment does not.

### Findings, fixed on the branch

**1. Amendment 11 leaves the superseded passage unfindable, and the ADR has a convention for this.**
`957fefda`. The amendment says the passage "is left as written" — it is, but with no pointer, so a
reader who lands on § Amendment 7's property-names section reads *"`AdwBottomSheet.openState` is the
single entry the rule does not reach"* as current. This ADR marks every other superseded passage in
place (§ 3 line 618, § Amendment 3 line 1216, § Amendment 7 line 1653, the 2026-08-30 row at line 373),
all in the same `> **Superseded …**` blockquote. That convention is what makes "left as written" and
"findable" the same claim. It now carries one. Two facts inside the amendment were also wrong:

- It cites the passage as `§ *What the remainder is*`. **No section of this ADR has that name** —
  `grep` finds the string only inside Amendment 11 itself (twice). Corrected to § Amendment 7's
  *The property names: eleven of twenty-five, and the rule that split them*, where it actually lives.
- It calls the remainder "an adjustment, a widget, a page object, a class list, a name where the port
  holds an index". **There is no adjustment entry left** — #1570 retired all three on this branch's own
  base. Measured against the table: the six `gir` entries target `cssClasses` (×2), `titleWidget` (×2),
  `selectedPage`, `visibleChildName`.

**2. The ledger header names three keys whose entries are gone — and Amendment 11 copied the error
from it.** `54c596f3`. The paragraph above `NS_PROPERTY_ALIGNMENT` lists the GIR keys the remainder
points at: "a list model (`model`, `menuModel`), an adjustment object (`adjustment`), …". Measured:
`menuModel` entries went in #1528, `model` in #1566, `adjustment` in #1570; at `68461e4a` the sentence
names all three and `grep -c "gir: 'adjustment'\|gir: 'menuModel'\|gir: 'model'"` is **0**. Same for
"`AdwSpinRow`'s three scalars", also gone in #1570. It is the second-copy anti-pattern this repo names,
and the second copy is the one Amendment 11 read. The fix deletes the enumeration rather than
refreshing it — kinds stay, keys go — and records *why* it drifted, which is the part worth keeping:
a shape difference is not permanent, and ADRs 0042/0046/0047 are three cases where giving the port
the shape converged every entry that named the kind.

**3. Four comments still document `open()`/`close()` as the widget's API.** `479cefd6`.

```
adw-bottom-sheet.ts:6   "`open`/`close`" in the surface list
adw-bottom-sheet.ts:17  "`open()`/`close()` stay the PROGRAMMATIC pair"
adw-bottom-sheet.ts:26  "an app … wraps `open`/`close` in view.animate({ translate })"
bottom-sheet.spec.ts    a test NAMED `close() stays ungated while a dismissal is gated`,
                        whose comment says "`close()` is that property"
```

The module header is the first thing a reader of this widget opens and it was promising a method pair
the class no longer has. The test name is worse: it is what a failure prints. The `REGRESSION PIN`
comment that mentions `close()` historically is kept — the incident is the point — but now says the
method is gone.

### Gates and specs

All **exit 0** on the branch after my commits: `check-vocabulary-alignment`,
`check-nativescript-xml-doors`, `check-generated-website-data`, `check-nativescript-theme-classes`,
`check-storybook-widget-coverage`, `check-adwaita-conformance-drivers`, `check-storybook-control-parity`,
`check-storybook-story-parity`, `check-adwaita-element-properties`, `check-adwaita-tag-vs-class`,
`check-doc-fences`, `check-adr-index`, `check-changelog-references`, `check-comment-budget` (advisory;
`packages/nativescript-bridge` keeps 106 spare), `check-refs-citations` (twice — as-is, and with
`refs/libadwaita` present). Whole-tree `oxlint` exit 0 (warnings only, all pre-existing);
`oxfmt --check` clean on every changed file.

**Specs run.** `gjsify workspace @gjsify/adwaita-nativescript run test` → **exit 0 on node and on gjs**,
including the renamed case, and `gjsify tsc --noEmit` → exit 0. What that does *not* cover, stated
plainly: `bottom-sheet.spec.ts` deliberately imports `bottom-sheet-state.ts` and never the widget, so
the suite exercises the shared state machine, not the accessor. The accessor is held by `tsc` and by the
on-device gallery probe, and the probe needs an emulator I do not have here.

### For you to decide

- **I trimmed the property's doc comment** (dropped the rename argument, kept the migration note plus a
  pointer to ADR 0034 § Amendment 11). It was the third copy of one lesson, after the amendment and the
  ledger header. Revert that hunk if you want the argument at the call site.
- **`refs/nativescript` is not checked out anywhere on this machine**, so the "no `open` on any NS base
  class" measurement came from `NativeScript/NativeScript@main` over the network, not from the pinned
  submodule. If you want it pinned: `git submodule update --init --depth 1 refs/nativescript` and re-grep.
- **Pre-existing, not introduced here:** the story's toggle button writes `this._sheet.open` without
  touching `this.args.open`, so the button and the "Open" control disagree until the next `updateArgs`.
  The GTK and web stories have the same shape.
- **#1573 conflict is slightly wider than the note says.** My header rewrite touches the paragraph
  directly above `NS_PROPERTY_ALIGNMENT`, so whichever lands second rebases the comment as well as the
  table.

Verdict: the reasoning is sound and the measurement behind it is real — the deletion is not a
convenience removal, it is what upstream's own documentation points at. Not merged, per instruction.

PR #1575, opened 2026-09-05
feat(adwaita-nativescript)!: a look is a class list

JumpLink:
## What

[ADR 0049](https://github.com/gjsify/gjsify/blob/feat/portable-style-classes/docs/adr/0049-style-classes-are-a-list.md). Adwaita's button looks and the flat header bar are **not properties in GTK**. They are style classes, and a widget carries a LIST of them in `GtkWidget:css-classes`:

| the look | where it is defined |
|---|---|
| `.suggested-action` | `refs/libadwaita/src/stylesheet/widgets/_buttons.scss:220` |
| `.destructive-action` | `:230` |
| `.pill` | `:323` |
| `.flat` on a header bar | `refs/libadwaita/src/stylesheet/widgets/_header-bar.scss:89` |

The port had spelled them as two different kinds of thing, and the property ledger counted both: `GtkButton.variant` — an enum holding ONE look — and `AdwHeaderBar.flat`, a boolean for one look on one widget.

## The convergence is also the feature

The enum's own documentation recorded the cost:

> `pill` is the rounded SHAPE and combines with no other variant here (set the shape OR the accent intent)

On GTK `.pill.suggested-action` is an ordinary button. Through the enum it was unreachable; `cssClasses = 'pill suggested-action'` reaches it. That is the opposite of the usual trade in this series, where convergence costs a spelling and returns portability.

## Breaking

| published as | becomes |
|---|---|
| `button.variant = 'suggested'` | `button.cssClasses = 'suggested-action'` |
| `header.flat = true` | `header.cssClasses = 'flat'` |
| `<gtk:Button variant="pill">` | `<gtk:Button cssClasses="pill">` |
| `AdwButtonVariant` (exported type) | gone — the values are class names now |

Note the names are **class** names, not the web element's attribute names: `suggested-action`, not `suggested`. `gtk_widget_set_css_classes` takes any names and unknown ones match no rule, so the port neither validates nor rewrites. `ADW_BUTTON_STYLE_ALIASES` stays exactly where it was — it is `@gjsify/adwaita-web`'s **attribute** vocabulary (`<gtk-button suggested>`), and an attribute name is not a class name.

## Two decisions worth arguing with

**The base class is not in the list.** `adw-button` / `adw-header-bar` stay on `className` so the port's stylesheet still applies, and `cssClasses` reads back exactly what the caller wrote. That is GTK's own rule rather than a convenience: a widget's CSS name is not a member of `css-classes`, and `gtk_widget_get_css_classes` never returns it.

**The door takes a string; the read-back is a list** — the DOM's own `className`/`classList` split. Both writers here are strings (a NativeScript XML attribute and `View.className`); the getter holds what GTK's property holds.

An array-taking door would need a **fourth attribute kind** in `check-generated-website-data`, which knows `number`, `boolean` and `json` (the last added by ADR 0047's change). Measured while writing this: annotating the setter `string | readonly string[]` files it as `json`, and the gate then demands the gallery block write `cssClasses='["pill"]'`. Inventing a kind to offer a second spelling of a door that already works is the *"second way to say what the table can already say"* that gate's own header refuses.

## Measured

| | |
|---|---|
| `check-vocabulary-alignment` | distance **7 → 5** property names on this base |
| `check-nativescript-xml-doors` | OK — 45 elements |
| `check-generated-website-data` | OK — 28 XML templates byte-identical to the view the probe app inflates, 40 gallery blocks |
| `check-storybook-widget-coverage` | OK — 44 widgets on both renderers, 38 with a story |
| `check-nativescript-theme-classes` | OK — 179 classes emitted, 146 styled |
| `tsc` | the package is clean for these edits (the residual TS2307s are `@gjsify/native-platform`, unbuilt in this worktree and unrelated) |

Migrated with the widgets: four gallery views, the generator that writes them, the shipped template data and thirteen NativeScript storybook pages.

## Deliberately not in scope

- **`@gjsify/adwaita-web` keeps its boolean attributes** on six elements. The property half of ADR 0034's rule is scoped to the NativeScript surface by construction — the gate says so in the line it prints — and the web's attribute vocabulary is a second corpus nothing has read. Its own change, its own migration.
- **`AdwHeaderBar.title` / `.subtitle` stay** — structural, not spelling. `Adw.HeaderBar` has no `title` key at all (measured: `titleWidget` is the only title-shaped key in `AdwHeaderBarProps`), all three renderers carry the same derived-title divergence and `@gjsify/adwaita-core` models it as `HeaderBarRenderState.derivedSubtitle`.

## Stack

Independent of #1573 and #1574; all three remove different entries from `NS_PROPERTY_ALIGNMENT`, so whichever lands last may need a trivial rebase. Together the three take the distance **7 → 2**, and what remains is the two structural entries above plus `AdwImageButton`'s taken name.


JumpLink:
## Independent review

Reviewed `6b6d0f34f` + `037131317` (the docs follow-up landed mid-review). Every number below
is re-derived here, not copied from the PR. Four commits pushed: `6e2c520`, `b0205b9`,
`7aa2f3d`, `e60c9f9`.

**One blocking defect, fixed: `cssClasses` is not a free name on NativeScript, and taking it
kills both widgets in their own constructors.**

---

### 1. The name collision (blocking — fixed in `b0205b9`)

`@nativescript/core@9.1.0-alpha.11` — the version `showcases/dom/adwaita-storybook-nativescript`
pins — declares on `ViewBase`:

```
ui/core/view-base/index.d.ts:366   readonly cssClasses: Set<string>;
ui/core/view-base/index.js:226     this.cssClasses = new Set();
ui/core/view-base/index.js:1140    const cssClasses = view.cssClasses;      // classNameProperty.valueChanged
             :1142                 cssClasses.has(CSSUtils.MODAL_ROOT_VIEW_CSS_CLASS);
             :1144                 cssClasses.clear();  … .add(…)
```

That `valueChanged` runs on **every** `className` write, and `Property.set`
(`ui/core/properties/index.js:157-201`) calls it unconditionally on a change. A subclass
accessor on `GtkButton.prototype` / `AdwHeaderBar.prototype` shadows the constructor's
assignment, so the `Set` never exists — and the first `className` write is the one inside
`GtkButton`'s **own constructor**:

```
TypeError: cssClasses.has is not a function
```

Measured, not reasoned: those two bodies copied verbatim into a repro driven by the PR's own
accessor and by `tsc`'s emitted `gtk-button.js`. `new Gtk.Button()` and `new Adw.HeaderBar()`
both throw. Even without the throw, the CSS engine loses the set it matches selectors against,
so no `.adw-button` rule would ever apply again.

**Why nothing here could see it — worth more than the bug.**

- `packages/nativescript-bridge/adwaita/src/ns-core.d.ts` is an ambient `declare module`, and
  an ambient module declaration **wins over the real package**. Measured: with the real
  `@nativescript/core` symlinked into `node_modules`, `View.cssClasses` still reads
  *"Property 'cssClasses' does not exist on type 'View'"*. The file's own header says the
  opposite ("its own (far richer) types take precedence"); that is false for this package's
  own check.
- No widget test constructs a widget — `extends FlexboxLayout` cannot be imported outside a NS
  runtime, so every suite drives a pure helper.
- The showcase is compiled by nothing: it has no `node_modules`, no `check` script, and no CI
  job. `tsc` there dies on unresolved `@nativescript/core` before it can reach anything.

**The mechanism** (this is the part that outlives the PR): `ns-core.d.ts` now declares the
member, so `gjsify tsc` answers the runtime hazard by name —

```
TS2611: 'cssClasses' is defined as a property in class 'Button',
        but is overridden here in 'GtkButton' as an accessor.
```

An accessor scan across all 48 widget classes against the real `ViewBase`/`View` `.d.ts` says
these two were the **only** collisions in the package.

**The spelling**: `styleClasses` — libadwaita's own word for these, and `grep` finds it nowhere
in `@nativescript/core`. The module is now `widgets/style-classes.ts`; the 13 storybook pages,
3 gallery views, the generator, `expected.ts`, the shipped template data and the 8 doc samples
follow. **The name is yours to change** — the non-negotiable part is only that it must not be
`cssClasses`.

**What this costs the PR**: the ledger entries come back as `should converge` on `cssClasses`,
with the collision as their measured reason, so the distance is **7 property names, not 5**.
The composable class list is still the feature; the convergence is not available on this
surface. ADR § 1 and Consequences now say that.

---

### 2. The header-bar citation is wrong (fixed in `6e2c520`)

`_header-bar.scss:89` is `.titlebar headerbar:not(.flat) {` — a **negation** inside the
window-shadow block, not a definition. Re-derived against `refs/libadwaita` @ 1.10.beta~30:

| claim | verdict |
|---|---|
| `.suggested-action` → `_buttons.scss:220` | ✅ `&.suggested-action {` |
| `.destructive-action` → `:230` | ✅ `&.destructive-action {` |
| `.pill` → `:323` | ✅ `&.pill {` |
| `.flat` on a header bar → `_header-bar.scss:89` | ❌ — the style is `%headerbar-flat` at `_header-bar.scss:112`; the **class** is `headerbar.flat` at `_deprecated.scss:456`, under *"Deprecated: use AdwToolbarView"*. `adw-header-bar.c` mentions `flat` nowhere. |

`check-refs-citations` passed it *because* `flat` is a substring of `:not(.flat)`. I proved the
arm is live by breaking the anchor — it goes red and prints the line it read. Also: a bare
`:230` after a full path is a **continuation the gate does not read**, so three of the four
coordinates were verified by nothing. Every coordinate now carries its own anchor; the tree's
anchored citations go 5 → 13.

And a caveat about the gate itself: **in a worktree without `refs/`, `check-refs-citations`
verifies nothing** — it printed *"0 resolved … the ANCHOR arm verified nothing on this run"*
and exited 0. I hardlinked `refs/libadwaita` in to make it real.

---

### 3. The array-door argument is right, but its detail is wrong (fixed in `6e2c520`)

I annotated the setter `string | readonly string[]` and ran both gates, then restored.
`attributeKind` does file it as `json`, `check-nativescript-xml-doors` demands a
`parseAdjustment()`-shaped door, and `check-generated-website-data` refuses every gallery
block. But the ADR said the gate "demands the gallery block write `cssClasses='["pill"]'`" —
it does not: `literalKind` calls a literal `json` only when it parses to a plain **object**
(`!Array.isArray`), so `'["pill"]'` classifies as `string` and fails identically. No spelling
satisfies it. The conclusion holds, harder than it was written.

### 4. A rule with a measurably wrong reason (fixed in `6e2c520`)

`normalizeCssClasses` accepted an array "because the widget's internals rebuild the list
through it". The only two callers are the two setters, both typed `string | null | undefined`.
The arm was unreachable and its `String(name)` coercion guarded nothing. Removed.

### 5. `AdwButtonVariant` still named as live (fixed in `e60c9f9`)

`scripts/nativescript-xml-doors.mjs` justifies `NS_TYPE_DIRS`' third entry with a type this PR
deleted. The incident stays, dated; the entry is still load-bearing, measured — removing
`packages/web/adwaita-core/src` moves "typed as something no XML attribute can carry" from
19 to 33.

### 6. No test for the new property (added in `7aa2f3d`)

`src/style-classes.spec.ts`, 10 cases / 21 assertions, green on Node and red on a normalizer
that stops de-duplicating and on one that stops guarding `undefined` (both checked). The
expectations are measured against GTK under gjs 1.88.1 / libadwaita 1.9.3:

```
fresh Gtk.Button                  []                             the CSS name is NOT in the list ✅
set ['a','b','a']                 ['a','b']                      de-duplicates ✅
set []                            []                             replaces, never accumulates ✅
set ['pill','suggested-action']   ['suggested-action','pill']    ← order is NOT the caller's
set ['  x  ']                     ['  x  ']                      GTK does not trim; the port does
```

**GTK's read-back order is an artifact, not a contract**: both orders of
`['zzz-one','aaa-two']` answer `['zzz-one','aaa-two']`, because the list is held in GQuark
order — which name was first interned anywhere in the process. So "the getter holds what GTK's
property holds" overstates it; the port keeps the order *written*, which is better, and the
ADR and the suite now say which is which.

---

### Counts and claims re-derived

| claim | measured |
|---|---|
| `gtk_widget_get_css_classes` never returns the CSS name | ✅ fresh `Gtk.Button` → `[]`, css-name `button` |
| `Adw.HeaderBar` has no `title` property | ✅ 43 properties via `list_properties()` on libadwaita 1.9.3; `title-widget`, `show-title`, `show-{start,end}-title-buttons`, no `title` |
| thirteen storybook pages | ✅ 13 |
| **four gallery views** | ❌ **three** — `GtkButton.xml`, `AdwHeaderBar.xml`, `AdwWrapBox.xml`; 3 generator entries, 3 website-data entries, of 28 templates / 28 views. PR body + commit message need the correction. |
| distance 7 → 5 | was true as written; now **7**, see §1 |
| variant→class mapping at every site | ✅ `suggested`→`suggested-action`, `destructive`→`destructive-action`, `pill`/`flat` unchanged; no `'default'` became a literal class; `CLASSES_BY_STYLE[style] ?? ''` correctly yields `[]` |
| nothing reads `variantClasses` / `AdwButtonVariant` | ✅ only the stale comment in §5 |
| no `className === 'adw-header-bar flat'` dependency | ✅ theme selectors are compound (`.adw-header-bar.flat`), and `flat`/`pill`/… were never tracked by `check-nativescript-theme-classes` (`isTracked` = `adw-` prefix), so its 179/146 is unchanged either way |

**The 8 `.mdx` doc samples** were a real miss in `6b6d0f34f` — you found and fixed them in
`037131317` while I was reviewing, via `check-doc-fences`, a gate the PR's MEASURED table
does not list. Same for `check-storybook-control-parity`, which is what actually holds the 13
storybook pages: I injected `button.variant = 'pill'` into `toast.ns.ts` and it went red (261
writes held against each widget's members), restored and green. Worth adding both to the table.

### Gates, at `e60c9f9`

All green: `check-vocabulary-alignment` (distance 1 widget / 7 property), `-nativescript-xml-doors`
(45 elements), `-generated-website-data` (28 templates byte-identical, 40 gallery blocks),
`-storybook-widget-coverage` (44/38), `-nativescript-theme-classes` (179/146),
`-adwaita-conformance-drivers`, `-refs-citations` (13 anchors, 11 read here),
`-storybook-control-parity`, `-storybook-story-parity`, `-storybook-category-order`,
`-doc-fences`, `-adwaita-style-classes`, `-adwaita-tag-vs-class`, `-adr-index`,
`-node-test-registration`, `-test-entry-run`, `-adwaita-element-properties`,
`-website-adwaita-gallery`, `-website-attr-samples`, `-nativescript-accent-rules`,
`-adwaita-collection-reactivity`. `generate-adwaita-nativescript-templates.mjs` idempotent
(no diff). `oxfmt --check` + `oxlint` clean.

`tsc --noEmit` for `packages/nativescript-bridge/adwaita` is **fully clean** — the residual
TS2307s the PR noted disappear once `@gjsify/native-platform`'s declarations are emitted
(`tsc -p packages/nativescript-bridge/platform`); no need for the caveat.

`showcases/dom/adwaita-storybook-nativescript` **could not be typechecked** and I am not
implying coverage: it has no `node_modules`, `@nativescript/core` is nowhere in the workspace,
and `tsc` there produces a cascade of unresolved-module errors. `check-storybook-control-parity`
is what holds those files. `check-type-surfaces` fails here on `@girs/gtk-4.0` drift in this
worktree's `node_modules` (`SignalSignatures`, `SvgWidget`) — pre-existing and unrelated.

### For you to decide

1. **The property name.** `styleClasses` is my choice; anything free works. If you want to keep
   `cssClasses`, the only shape that does not shadow is to drop the accessor entirely and let
   NativeScript's own `Set` be the property — but then the read-back is a `Set` containing
   `adw-button` plus `ns-android`/`ns-phone`/`ns-light`, which is not GTK's contract.
2. **The ADR status.** § 1 now carries a decision the original did not, and the headline
   measurement changed from 7→5 to 7→7. Re-read it before flipping to Accepted.
3. **The PR body**: "four gallery views" → three, and the `check-vocabulary-alignment` row.
4. Whether the deprecation of `headerbar.flat` upstream (AdwToolbarView is the replacement)
   deserves its own follow-up on this port.

Not merged.

PR #1581, opened 2026-09-05
docs(status): free an entry from the header comment

JumpLink:
## What

`### Two XMLHttpRequest implementations, and the docs name the wrong one as the only one` has been sitting **inside `status/open-todos.md`'s own HTML comment**.

The splice is visible once you know where to look. The header opens:

```
<!-- Authored Open-TODO sections — THIS FILE is the tracked source of truth (the
     rendered STATUS.md view is generated and gitignored). One `### <title>` per open item.
```

Someone appended the entry directly after that backtick, so `` `### <title>` `` was cut in half: the first half became the entry's own heading, the second half surfaced 24 lines later as

```
### <title>` per open item.
     A RESOLVED item is DELETED …
```

and everything between them — the entire entry — is inside `<!-- -->`.

## Why it matters more than a typo

- The ledger's **most-read line** — the one that tells every agent and every reader what the file's convention is — reads as garbage.
- The entry documents a **tier-1 published-contract question** (`@gjsify/xmlhttprequest` ships one XHR class, `@gjsify/fetch` ships another, and the routing table sends the scoped name to the second). It renders nowhere: the generated `STATUS.md` view comes from this file, so the entry is missing there too.
- It is reachable by `grep`, which is exactly why nobody noticed. Anyone opening the file at the top cannot see it.

## The fix

The header is back to its one sentence; the entry is the first `###` section after it, **unchanged in wording**. 6 insertions, 5 deletions, and 177 `###` sections where a reader could previously see 176.

## What now catches it (added after review)

A reviewer pointed out the error was completely silent — the entry vanished from the rendered `STATUS.md` with no gate firing. The existing conformance check (reading headings via `/^### /gm`) does not validate extraction completeness, so a heading inside a comment would pass undetected.

Added: `status-data` rule now checks that no `###` heading appears inside the HTML comment block of `status/open-todos.md`. The check fails with the heading text and a message naming where to move it. The generator's regex matches line-start only, and comments are never rendered to STATUS.md, so any heading found inside a comment is certain to be lost — the check fires exactly when the error would occur.

- **Cost**: 16 lines in `scripts/generate-status.mjs`
- **False positives**: zero (no legitimate heading should live in a comment)
- **Caught by negative-test**: yes (breaking the file by moving the entry back into the comment makes the check go red; restoring makes it green)

Found while another branch appended to the tail of the same file.

### DIFF
--- a/scripts/generate-status.mjs
+++ b/scripts/generate-status.mjs
@@ -562,6 +562,22 @@ export function loadStatusData(root, facts) {
             );
         }
     }
+    // A heading inside an HTML comment is silently dropped by the generator (regex /^### /gm
+    // matches line-start only, and comment-wrapped text does not export to STATUS.md). The error
+    // is completely silent — the entry vanishes with no warning. Catch it here.
+    const commentMatch = todosMd.match(/<!--[\s\S]*?-->/);
+    if (commentMatch) {
+        const commentText = commentMatch[0];
+        const inCommentHeadings = [...commentText.matchAll(/###\s+([^\n`<>]+)/g)];
+        for (const match of inCommentHeadings) {
+            const headingText = match[1].trim();
+            failures.push(
+                `status/open-todos.md has a \`### ${headingText}\` heading inside an HTML comment. ` +
+                    'The generator silently drops headings in comments — they do not render in STATUS.md. ' +
+                    'Move the heading and its section outside the comment block.',
+            );
+        }
+    }
 
     // The OTHER direction, and it is the same class as the corpse check above.
     //
--- a/status/open-todos.md
+++ b/status/open-todos.md
@@ -1,5 +1,10 @@
 <!-- Authored Open-TODO sections — THIS FILE is the tracked source of truth (the
-     rendered STATUS.md view is generated and gitignored). One `### Two XMLHttpRequest implementations, and the docs name the wrong one as the only one
+     rendered STATUS.md view is generated and gitignored). One `### <title>` per open item.
+     A RESOLVED item is DELETED (its record is the commit + CHANGELOG that closed
+     it) — the status-data check rejects struck-through / ✓ / "Completed"
+     headings, so the done-log cannot regrow. -->
+
+### Two XMLHttpRequest implementations, and the docs name the wrong one as the only one
 
 `@gjsify/xmlhttprequest` ships a class in `src/index.ts`. `@gjsify/fetch` ships another in
 `src/xhr.ts` (11 269 bytes). `packages/infra/resolve-npm/lib/index.mjs` routes the SCOPED
@@ -25,10 +30,6 @@ constants as module-level `export const`s and may or may not carry the same defe
 either class is a published-contract change (`@gjsify/xmlhttprequest` is tier 1), and merging
 them is a decision about which package OWNS the API, not a refactor. Establish that first.
 
-### <title>` per open item.
-     A RESOLVED item is DELETED (its record is the commit + CHANGELOG that closed
-     it) — the status-data check rejects struck-through / ✓ / "Completed"
-     headings, so the done-log cannot regrow. -->
 
 ### A `Gtk.Window` in a child list is accepted silently, where an `Adw.Dialog` aborts
 


JumpLink:
## Independent review

### Verification (measured)

1. **Byte-identity of entry text**: PASS
   - Extracted entry body from `origin/main` (line 3-26, skipping comment syntax):
     ```bash
     git show origin/main:status/open-todos.md | sed -n '3,26p' > /tmp/original.txt
     ```
   - Extracted entry body from `HEAD` (line 8-31, skipping heading):
     ```bash
     git show HEAD:status/open-todos.md | sed -n '8,31p' > /tmp/branch.txt
     ```
   - Result: `diff -u /tmp/original.txt /tmp/branch.txt` → no output (byte-identical) ✓

2. **Header restoration**: PASS
   - `origin/main` header (lines 1-2):
     ```
     <!-- Authored Open-TODO sections — THIS FILE is the tracked source of truth (the
          rendered STATUS.md view is generated and gitignored). One `### Two XMLHttpRequest...
     ```
     (BROKEN: the sentence is split, with entry heading in the middle)
   
   - `HEAD` header (lines 1-5):
     ```
     <!-- Authored Open-TODO sections — THIS FILE is the tracked source of truth (the
          rendered STATUS.md view is generated and gitignored). One `### <title>` per open item.
          A RESOLVED item is DELETED...
          ...
          headings, so the done-log cannot regrow. -->
     ```
     (FIXED: header is now a complete, sensible sentence) ✓

3. **Section count and ordering**: PASS
   - `origin/main` section headings: 177 total
   - `HEAD` section headings: 177 total
   - First heading on `origin/main`: `<title>` per open item.` (malformed tail of header)
   - First heading on `HEAD`: `Two XMLHttpRequest implementations, and the docs name the wrong one as the only one` (correct)
   - All subsequent sections shifted by +1 line (expected for net +1 line change) ✓

4. **Renderability claim**: VERIFIED TRUE
   - The regex `/^### ([^\n]+)$/gm` (used by `scripts/generate-status.mjs` line 407) matches only lines that START with `###`
   - In `origin/main`, the entry heading appears on line 2 after spaces: `     rendered...One \`###...`
   - Since the line starts with spaces, the regex does NOT match it
   - The entry was therefore invisible in the generated STATUS.md ✓

5. **Conformance checks**: PASS
   - `node scripts/check-refs-citations.mjs` → 0 failures ✓
   - `node scripts/manifest-conformance/rules/status-data.mjs` → 0 failures (exit 0) ✓
   - `node scripts/check-doc-fences.mjs` → TypeScript not installed (environmental, not a PR defect)

### Gate decision analysis

The PR adds no new gate. The stated rationale: "one instance is an accident and a rule invented for a single member is the 'second way to say what the table can already say' this repo's gates refuse."

**Finding**: The existing conformance check does NOT validate heading extraction completeness. The `status-data` rule:
- Validates that extracted headings are not marked "resolved" (struck-through/✓/Completed)
- Does NOT validate that ALL expected headings were extracted
- Would NOT have caught this bug (both `origin/main` and `HEAD` pass with 177 headings each, the difference being which heading is first)

A simple preventive gate could be: "Every `###` heading in `status/open-todos.md` must start at column 0 (line position)." This would:
- Have zero false positives (no legitimate heading should be indented or inside a comment)
- Catch this exact bug if reintroduced
- Be approximately 5 lines of code

**Verdict on no-new-gate**: Defensible but not airtight. The incident was manual editing (not a code defect), and the repo's philosophy appears to avoid narrow preventive gates for single incidents. However, this is a case where the error was completely silent (the entry vanished from the rendered output with no warning), which usually justifies a gate under the "Grüne CI, die nichts geprüft hat" principle. A one-line assertion (`headings.every(h => h.body.trim().startsWith('###'))` after extraction) would close this with negligible cost and false-positive risk.

**Recommendation**: The gate omission is acceptable given the incident's one-time nature and manual origin, but a future similar incident would argue for adding the check retroactively. The header comment's explicit documentation ("One `### <title>` per open item.") is the practical preventive measure here.

### Result

✓ **APPROVE** — the fix is complete, correct, and minimal. The entry text is unchanged, the header is properly restored, all headings render correctly, and conformance checks pass. The no-new-gate decision is justified, though marginal.

JumpLink:
## Reversal: gate added

After initial independent review argued against adding a gate (reasoning: one-time incident, manual editing, repo philosophy), the PR author reconsidered and made the case that:

1. The error was **completely silent** — entry vanished from STATUS.md with no warning
2. The existing `status-data` check does NOT validate extraction completeness  
3. A check that fires exactly when the error occurs is warranted under the "green CI that checked nothing" principle

**Check implemented** (`scripts/generate-status.mjs:565–579`):

- Parses the HTML comment block of `status/open-todos.md`
- Searches for any `### ` heading inside it
- Reports the heading text + location + fix instructions if found

**Properties**:
- **False positives**: zero (no legitimate heading lives in a comment)
- **Negative-tested**: yes (temporarily moved entry back into comment → check failed; restored → check passed)
- **Cost**: 16 lines, one `/###\s+/g` regex pass
- **Triggered by**: the exact condition that makes the generator silently lose the entry

Commit: `272041e10` — "add silent-headings check to status-data rule"

PR body updated to reflect the check; branch force-pushed with new commit on top of the position fix.

PR #1589, opened 2026-09-05
feat(devtools): a tree that says what it was given

JumpLink:
## The gap

`DumpTree` says a widget is mapped and `GetProperty` says its properties are what the code set. Neither distinguishes a label allocated the width it asked for from one allocated less and now showing two thirds of its text — so a **layout** defect was the one class of defect this control plane could see and could not diagnose.

The workaround was `Screenshot(<widget path>)` per widget and a read of the PNG header. Three round trips, and it reports the allocation without the request: it says *that* a widget is the wrong size and never whose arithmetic made it so.

## What lands

Every mapped node in a `DumpTree` result now carries `geometry`:

```jsonc
{
  "type": "GtkLabel",
  "geometry": {
    "width": 170, "height": 23,          // the MARGIN box, see below
    "widthRequest": [67, 170],           // [minimum, natural]
    "heightRequest": [42, 42],           // measured AT THE ALLOCATED WIDTH
    "short": true                        // allocated below the minimum: clipped
  }
}
```

Unconditional rather than behind a flag, because the caller who needs it does not know which widget to suspect — sweeping a dump for `short` finds them. On a real 954-node application window that is 294 measured nodes and 14 findings: two are the defect being chased, two are viewports inside scrolled windows, five are `AdwGizmo` indicator dots.

## Two decisions worth reviewing

**The height is measured at the allocated width.** A wrapping label asked for its height at its *natural* width answers one line; asked at the width it actually got, two. The two numbers agree except where a parent measured one and allocated the other, which is exactly the defect. Measuring at `-1` would report the number that agrees with the allocation and hide the case worth finding.

**GTK has three boxes behind four accessors, and the first version of this mixed them.** Measured on GTK 4.22.4, a `Gtk.Label` with `padding: 3px 5px` over a 68×17 content box:

| | padded | margined |
|---|---|---|
| `get_width()` / `get_height()` | 68×17 | 68×17 |
| `compute_bounds(self)` | 78×23 | 68×17 |
| `measure()` natural | 78×23 | 82×25 |

So `get_width()` is the **content** box, `compute_bounds()` the **border** box, and `measure()` speaks the **margin** box. Asking the padded label's height at 68 answers 40 rather than 23, because 68 read as a margin-box width leaves 58 for text that then wraps. That mistake called **116 of 293** widgets in that same window clipped. Everything here works in the margin box — `compute_bounds()` plus the four margins, which is the rectangle `get_allocation()` returns without calling what GTK 4.12 deprecated.

Position is deliberately absent: x/y would need a `Graphene.Rect` across the bridge, `Screenshot(<path>)` already answers where a widget is by showing it, and the question this exists for is answered without it.

## Cost

Two `gtk_widget_measure()` calls per mapped node. GTK caches a size request per `(orientation, for_size)`, and the pair asked for here is the pair the layout just asked for itself, so it reads that cache rather than provoking a re-layout.

## Tests

Six new specs in `widget-tree.spec.ts`, all on duck-typed shapes so they run on Node and GJS: the pair is reported; the height is measured at the allocated width and not the natural one; the margin box is the one used; `compute_bounds` failing leaves the field out; a mapped widget with no allocation measures at `-1`; an unmapped widget carries no geometry at all — its two zeros beside a request would read as "clipped to nothing" for every page of a stack that is not visible.

`@gjsify/devtools`: 61 passed · 156 assertions.

## What found it

The [CORRECTIV desktop host](https://github.com/faktenforum/correctiv-app/tree/desktop/apps/desktop), a React-Native-on-GTK4 application, had a masthead date allocated 170×23 at every window width from 700 to 1400 while one line of it needs 206. Every readable property was correct — `wrap: true`, `max-width-chars: -1`, `hexpand: false` — and the defect lived entirely in the two numbers that were not readable.

With this, the cause took one dump: the label is allocated its own natural width, and at its own natural width it asks for two lines. That turns out to be a GTK/Pango accounting fault with letter-spacing, characterised separately.

### DIFF
--- a/docs/react-native-gtk-measurements.md
+++ b/docs/react-native-gtk-measurements.md
@@ -98,3 +98,25 @@ Design decisions: [ADR 0032](adr/0032-react-native-on-the-gtk-host.md),
    public getter. `has_role()` is the exception: it takes the role, so the ROLE is a
    real value check. Everything else is AT-SPI's to report, and every GTK CI leg runs
    with `GTK_A11Y=none`.
+
+9. **A wrapping `Gtk.Label` reports a natural width its own text does not fit into,
+   once anything is letter-spaced.** GTK's natural width is `ceil(logical width)`, and
+   Pango's logical extents EXCLUDE the spacing after the final glyph while its
+   line-breaker COUNTS it — so the breaker needs `ceil(logical + spacing)` and is handed
+   one or two pixels less. Measured over four strings and six spacings:
+
+   | spacing | 0.0 | 0.2 | 0.5 | 1.0 | 1.5 | 2.0 |
+   |---|---|---|---|---|---|---|
+   | short by | 0 | 0 | 0–1 | 1 | 1–2 | 2 |
+
+   `<Text>` is a label with `wrap: true`, so the consequence is a label that wraps onto
+   two lines **at its own natural width**, in a parent that measured its height for one
+   — and the second line is clipped. It reaches every flex child that does not expand,
+   which is most text in an application. Reached through CSS `letter-spacing` and
+   through a `Pango.AttrLetterSpacing` alike, unchanged by all three
+   `natural-wrap-mode` values, and not helped by excluding the final character from the
+   attribute (that lowers the natural width by the same amount). There is nothing for
+   this layer to set, so it is an `it.failing` vector in
+   `primitives/text-metrics.spec.ts` that retires itself the day GTK fixes it. Found in
+   a shipped application, in four labels of seventy-six — each one an overline or a
+   masthead date, i.e. exactly the letter-spaced kind.
--- a/packages/framework/devtools-protocol/src/types.ts
+++ b/packages/framework/devtools-protocol/src/types.ts
@@ -33,6 +33,64 @@ export interface ActionList {
 }
 
 /**
+ * What a widget WAS GIVEN beside what it ASKED FOR — the pair that says whether a
+ * layout is right, and the one an introspection tree could not previously answer.
+ *
+ * `mapped` says a widget is on screen and a property read says its properties are what
+ * the code set; neither distinguishes a label allocated the width it asked for from one
+ * allocated less and now showing two thirds of its text. The workaround was to
+ * rasterise the widget through `Screenshot(<path>)` and read the PNG header, which is
+ * three round trips per widget and answers only the allocation — never the request, so
+ * it says THAT a widget is the wrong size and never whose arithmetic made it so.
+ *
+ * The measured case: a `Gtk.Label` in a `flex-row` masthead came out 170x23 at every
+ * window width from 700 to 1400, while one line of its text needs 206. Everything
+ * readable about it was correct — `wrap: true`, `max-width-chars: -1`, `hexpand: false`
+ * — and the defect is entirely in the two numbers that were not readable.
+ *
+ * NO POSITION. `width`/`height` come off `get_width()`/`get_height()`; x/y would need
+ * `compute_bounds()` and a `Graphene.Rect` across the bridge, and the question this
+ * exists for is answered without it. `Screenshot(<path>)` already says where a widget
+ * is by showing it.
+ */
+export interface NodeGeometry {
+    /**
+     * Allocation in logical pixels, in the MARGIN box — content plus CSS padding, border
+     * and margin.
+     *
+     * That box and not another one because it is the box a size request speaks, so these
+     * two numbers and the two below are comparable. GTK has three: `get_width()` answers
+     * the content box, `compute_bounds()` the border box, and `measure()` the margin box,
+     * and mixing them is not a rounding error. Measured on GTK 4.22.4, a `Gtk.Label` with
+     * `padding: 3px 5px` over a 68x17 content box asks 78x23; asked for its height at 68
+     * — its content width read as a margin-box width — it answers 40, because 58 is left
+     * for text that then wraps.
+     */
+    width: number;
+    height: number;
+    /** [minimum, natural] width, measured with no height given. */
+    widthRequest: [number, number];
+    /**
+     * [minimum, natural] height, measured AT THE ALLOCATED WIDTH.
+     *
+     * At the allocated width and not at `-1`, because that is the whole question. A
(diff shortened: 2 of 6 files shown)


JumpLink:
Second commit, and it is what the first one found.

**`test(react-native): the natural width a label cannot fit into`**

`<Text>` is a `Gtk.Label` with `wrap: true`, which puts this layer's text on GTK's height-for-width path. That path has one property React Native's does not: a wrapping label's natural width is the width of its text on one line, so a parent handing a label exactly its natural width is handing it exactly enough. Every flex container does that for a child that does not expand — most text in an application.

GTK does not have that property once anything is letter-spaced. Natural width is `ceil(logical width)`; Pango's logical extents **exclude** the spacing after the final glyph while its line-breaker **counts** it. So the breaker needs `ceil(logical + spacing)` and is handed one or two pixels less. Measured on gjs 1.88.1 / GTK 4.22.4 / Pango 1.57, four strings × six spacings:

| spacing (px) | 0.0 | 0.2 | 0.5 | 1.0 | 1.5 | 2.0 |
|---|---|---|---|---|---|---|
| short by (px) | 0 | 0 | 0–1 | 1 | 1–2 | 2 |

The label then wraps onto two lines **at its own natural width**, inside a parent that measured its height for one — and the second line is clipped.

`it.failing` rather than a note, because there is nothing for this layer to set: it reaches through CSS `letter-spacing` and through `Pango.AttrLetterSpacing` alike, is unchanged by all three `natural-wrap-mode` values, and excluding the final character from the attribute lowers the natural width by the same amount. So the vector asserts the **correct** behaviour, runs, and fails the suite the day GTK satisfies it. Its control — the same assertion with no letter-spacing — passes, so an expected failure cannot be a broken harness.

`@gjsify/react-native`: 626 passed · 2704 assertions · 1 expected failure.

No upstream issue found on 2026-09-05; not filed from here because a report wants a C reproduction rather than a GJS one.

---

**On this PR's red checks:** they are not this branch. `ts-for-gir` released `@girs/*` v4.7.0 at 19:36 UTC and its *Release Types* job failed at the submodule step, so only `@girs/adw-1` and `@girs/cairo-1.0` — alphabetically the first two — reached npm at 4.7.0. `@girs/cairo-1.0@4.7.0` requires `@girs/gjs@^4.7.0`, which does not exist, and `@girs/gjs@4.6.0`'s own `^4.6.0` range for cairo resolves up into it. Every clean install of the closure now dies with `ETARGET No matching version found for @girs/gjs@^4.7.0`, which is what took *GTK OS suites* and *macOS suites* down on `main` at cdddcd3 as well. Fixes are already in flight in `gjsify/types`.

JumpLink:
Third commit, and it is a fix to the first one. An independent review found the core measurement wrong, and the evidence that refutes it was in this PR's own comment table.

## `fix(devtools): the allocation is not the border box`

`compute_bounds()` answers the **border** box, and `get_margin_start()` answers the margin-start **property** — a margin from a stylesheet is invisible to it and lives inside the allocation instead. So the sum was short by exactly the CSS margin. Measured on GTK 4.22.4, one label under five decorations:

| | plain | padding | border | CSS margin | widget margin |
|---|---|---|---|---|---|
| `compute_bounds` + props | 173 | 183 | 177 | **173** | 187 |
| `get_allocation` + props | 173 | 183 | 177 | **187** | 187 |
| `measure()` natural | 173 | 183 | 177 | 187 | 187 |

Only `get_allocation()` plus the four properties tracks `measure()` in every column. On an ordinary Adwaita preferences window the error called **12 of 127** mapped widgets clipped, **11 of them wrongly**; the corrected formula reports one, and that one is a viewport inside a scrolled window, which really is allocated less than it asks for.

`get_allocation()` is deprecated in 4.12 and is taken knowingly — it is the only accessor that answers the box a size request speaks, and the comment says so rather than leaving the next reader to rediscover why the tidier call is wrong.

## And the dump made GTK complain

`measure(VERTICAL, w)` below a widget's minimum width prints `Trying to measure GtkBox … for width of 34, but it needs at least 40`, then clamps and answers anyway. That is exactly the `short` case — so a window with a genuinely clipped widget got one warning per clipped widget **per dump**, and `installDiagnosticsGate()` records at `LEVEL_WARNING`, so a dump inside a test would have failed `assertQuiet()` for finding the bug it was asked to find. Measured: every width from 0 up to the minimum returns the identical answer, so clamping is behaviour-preserving. `short` still compares the **raw** allocation, so no flag is lost.

## Why the specs could not catch it

`@gjsify/devtools`' suite runs with no display by design, so the duck-typed stub **is** the model of GTK — and a stub with no notion of CSS margin cannot fail on a CSS-margin bug. `Sized` now models the boxes as measured; `compute_bounds` stays on the stub returning the border box, so a rewrite reaching for it again is caught. Two new vectors cover the margin and the clamp, and both were checked to fail against the previous implementation.

## Prose corrections

Three claims were false. "NO POSITION … `width` comes off `get_width()`" described a revision that no longer existed; "GTK has three boxes" undercounts by one; and `text-metrics.spec.ts` asserted that a detached widget resolves no CSS — measured, it resolves the display's CSS and answers the identical natural width, stylesheet letter-spacing included. The `4000` magic width there is now `natural + 200`, which cannot stop being true for a longer string.

`short` on a viewport inside a scrolled window is structural rather than rare, and the field's doc said a sweep "finds the defective widgets" — one word too strong. Suppressing it would need ancestor knowledge and would hide real clipping under a scroller, so the key states the fact and the reader judges.

## Unchanged, because it verified

The `it.failing` letter-spacing table was re-measured independently over 8 strings × 6 spacings and is correct in every cell; the mechanism checks out numerically (natural is exactly `ceil(logical)`, the breaker needs exactly `ceil(logical + spacing)`). The caching claim holds — a counting `vfunc_measure` sees one call per distinct `for_size`, and 20 distinct sizes evicted nothing. `docs/react-native-gtk-measurements.md` needed no change.

Gates: `@gjsify/devtools` 62 tests · 162 assertions; `@gjsify/react-native` 626 tests · 2704 assertions · 1 expected failure; `check`, `lint` and `format --check` all clean. End to end against a real Adwaita window: 111 mapped nodes measured, **0 `short`, 0 GTK diagnostics** — was 12 and 3.

PR #1600, opened 2026-09-07
fix(react-native): let a truncating Text break inside a word it cannot break between

JumpLink:
`Gtk.Label` wraps at WORD boundaries only, so a label's MINIMUM width is its longest word — and a token with no break in it pins the layout open at that width, however narrow the window gets.

MEASURED on GTK 4.22.4 with one 48-character filename as the text:

| `wrap-mode` | minimum | natural |
| --- | --- | --- |
| `word` | **394 px** | 394 px |
| `word-char` | **13 px** | 394 px |

So the change costs nothing where the text can already be broken; it only stops one long run from deciding how narrow an application may be. That 394 was a card in a real application that would not shrink with its window, because a live-radio stream announced its track as `20260901_Gamescom_Laberpocast_Sophie_Amelie_final`.

React Native has no such failure mode, and not because its breaker is cleverer: a string it cannot break **overflows** its box rather than widening it. GTK has no spelling for that — a label's minimum IS its contribution to the layout — so `word-char` is the nearest thing that lets the text yield instead of the window.

## It rides on `numberOfLines` and is not a default

That is the second half and it was learned by looking. As a default on every `Text` it is visibly worse: a short label in a tight box breaks mid-word instead of asking for room, and a "Live" badge rendered as **"LIV-"**.

The two cases are told apart by the AUTHOR rather than by the layer — `numberOfLines` is someone saying this text may be truncated, and a text that may be truncated should yield before the window does.

So the ledger row is `kept`, not `normalised`: GTK's default stands, and the route writes the other value where the author asked for truncation. `defaults.spec.ts` is closed in both directions and refuses the other spelling.

https://claude.ai/code/session_01DQMZfw5A1RPVCUZwLvcZhF

### DIFF
--- a/packages/framework/react-native/src/primitives/defaults.ts
+++ b/packages/framework/react-native/src/primitives/defaults.ts
@@ -115,6 +115,15 @@ export const DEFAULT_ROWS: readonly DefaultRow[] = [
     },
     {
         gtype: 'GtkLabel',
+        property: 'wrap-mode',
+        gtk: 'word',
+        reactNative: 'a token that cannot be broken overflows its box, it does not widen it',
+        source: 'React Native lays text out with its own breaker and lets an unbreakable run spill; the box keeps the width the layout gave it.',
+        verdict: 'kept',
+        reason: 'GTK has no spelling for "overflow": a label’s MINIMUM width is its contribution to the layout, and at `word` that minimum is its longest word — MEASURED on GTK 4.22.4 with a 48-character filename, 394 px at `word` against 13 px at `word-char`, natural 394 either way. GTK’s default is KEPT because normalising it globally is visibly worse: a short label in a tight box then breaks mid-word instead of asking for room, and a "Live" badge rendered as "LIV-". `word-char` is written by the `numberOfLines` route instead — that prop is the author saying this text may be truncated, and a text that may be truncated should yield before the window does.',
+    },
+    {
+        gtype: 'GtkLabel',
         property: 'xalign',
         gtk: 0.5,
         reactNative: 'the script’s natural alignment — start, i.e. left in LTR',
--- a/packages/framework/react-native/src/primitives/primitives.spec.ts
+++ b/packages/framework/react-native/src/primitives/primitives.spec.ts
@@ -487,6 +487,11 @@ export default async () => {
             // declared no-op adds nothing to them.
             const p = plan('Text', { allowFontScaling: true, maxFontSizeMultiplier: 2 }).plan;
             expect(p.node.props).toStrictEqual({ wrap: true, xalign: 0, yalign: 0 });
+            // `wrap-mode` is NOT among them: it rides on `numberOfLines`, because a
+            // label that may be truncated should yield before the window does and one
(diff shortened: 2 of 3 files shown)


JumpLink:
Closing this: an independent review falsified the reasoning, and I reproduced every number.

## The trigger does not tell the two cases apart

The premise was that `numberOfLines` is the author saying "this text may be truncated", so a text that may be truncated should yield before the window does. Measured on GTK 4.22.4, `wrap: true` + `ellipsize: end`, minimum width:

| text | `lines` | `word` | `word-char` |
| --- | --- | --- | --- |
| 48-character filename | 1 | 13 | 13 |
| 48-character filename | 2 | 394 | **13** |
| `Bundesverfassungsgericht` | 2 | 187 | **15** |
| `Live` | 1 | 13 | 13 |
| `Live` | 2 | 29 | **14** |

At `numberOfLines={1}` — the commonest value — the change is a **total no-op**: `ellipsize` alone already collapses the minimum to 13 px. At two lines it does the thing it was for *and* the thing it was supposed to avoid: `Live` drops to 14 px and Pango breaks it inside the word, which is the "LIV-" rendering that made me revert the global default in the first place. A two-line card title in a narrow column is the ordinary case, not a corner.

So the real discriminator is `numberOfLines === 1` versus `>= 2`, which is not a statement about truncation at all.

## And the mechanism in the ledger row was the wrong reason

The row claims React Native "lets an unbreakable run spill". Android lays text out with `StaticLayout` at `BREAK_STRATEGY_HIGH_QUALITY`, which breaks an over-long word at character level; `NSLineBreakByWordWrapping` does the same. So RN is closer to `word-char` than to `word`.

What differs is not the breaker — it is that **RN's `flexShrink` defaults to 0**, so a box is never squeezed below its content and the breaker never gets asked. Lowering a GTK label's minimum lets GTK do something RN never has the chance to do. That makes this a `flexShrink` question wearing a `wrap-mode` costume, and there is no GTK property for the real one: a label's minimum IS its contribution to the layout.

## Worth keeping from the review, as follow-ups

* **The ledger has a third direction it does not close.** `tableOverrides()` walks `widgetProps` only, never a route's `also`, so a `normalised` verdict is impossible for a route-written value and — measured — deleting my row entirely leaves all 629 tests green. The file promises "a row that stops being true fails", and for route-written values it cannot keep that promise. A `routed` verdict plus walking `Object.entries(spec.props)` would close it.
* **`also` clobbers by JSX attribute order.** `{ ellipsizeMode: 'middle', numberOfLines: 2 }` writes `ellipsize: 'end'` and drops the author's `middle`; reversing the two attributes gives `middle`. Pre-existing, in the block this PR edited.
* **`numberOfLines={0}` means UNLIMITED in React Native** and renders one ellipsised line here, because `gtk_label_ensure_layout` only applies a line limit for `lines > 0`.
* **`PROPS.md` shows `numberOfLines → lines`** and hides the companions the route really writes, because `answerFor` returns `route.names` without `also`.

I would rather have these four as their own changes than merge a behaviour change whose reason does not survive its own measurement.

PR #1601, opened 2026-09-07
feat(node-gi): name the video plugins the bundles would need, and what each platform lacks

JumpLink:
The runtime bundles ship the AUDIO path and say so: *"video decode, encoding of any kind, capture, and every streaming SINK, server or adaptive-streaming demuxer are out"*. The rule was right — Homebrew's `gstreamer` drags in ffmpeg, x264, x265, faac and fdk-aac, and redistributing those belongs to whoever ships the product, not to a script that reads a directory.

This keeps the rule and gets video anyway, by taking the **decoder from the operating system**: `applemedia` on darwin (VideoToolbox), `d3d11`/`mediafoundation` on win32. That is the trade the audio SINKS already make one element over, and it puts no encumbered library in the payload.

MEASURED on GStreamer 1.28.6 for which plugin each element lives in, because the two names differ for three of the four:

| element | plugin |
| --- | --- |
| `gtk4paintablesink` | `gtk4` |
| `videoconvert` | `videoconvertscale` |
| `h264parse` | `videoparsersbad` |
| `hlsdemux2` | `adaptivedemux2` |

## Staged, not wired

`expectedGstPlugins` asks for none of it, so **no build can go red over a name**. What each platform is missing is measured rather than left for the first red build to find.

**win32**, twice, and both halves agree. The gvsbuild release this pipeline extracts carries no GStreamer at all — 6741 entries read out of its central directory over a range request, not one plugin — which is why the workflow builds `gstreamer gst-plugins-base gst-plugins-good` itself and stops. Of the four, only `videoconvertscale` is in base. And on the win11 VM against the published `@gjsify/gtk-runtime-win32-x64@0.48.0`: GStreamer 1.28.4, 22 plugins, every one audio.

```
Audio-Pfad:  playbin3 JA   decodebin3 JA   souphttpsrc JA   audioconvert JA
Video-Pfad:  gtk4paintablesink nein   videoconvert nein   h264parse nein
             hlsdemux2 nein   avdec_h264 nein   d3d11h264dec nein   mfh264dec nein
```

**darwin** is one plugin short, and it is the one with no formula. Homebrew's `gstreamer` 1.28.6 is the mega-formula, so three of the four plus `applemedia` arrive with the brew line the workflow already runs. `gst-plugins-rs` is **not** a Homebrew formula — the API answers 404 — so `gtk4paintablesink` would have to be built with cargo-c or vendored.

Which makes the paintable sink the whole of the remaining work on both, and it is the piece with no second source: it is the only GStreamer sink that hands out a `GdkPaintable`, and a `Gtk.Picture` takes nothing else.

## Where it came from

Making video work on Linux, where the system GStreamer has all of it: a real PeerTube HLS stream plays through `playbin3` + `gtk4paintablesink`, the position tracking the wall clock and the rendition climbing from 640x360 to 2560x1440 four seconds in.

https://claude.ai/code/session_01DQMZfw5A1RPVCUZwLvcZhF

### DIFF
--- a/packages/node-gi/scripts/gst-plugins.mjs
+++ b/packages/node-gi/scripts/gst-plugins.mjs
@@ -237,6 +237,69 @@ export const GST_PLUGIN_GAPS = {
 };
 
 /**
+ * THE VIDEO PATH, staged and not yet shipped — see the note at the end of this block.
+ *
+ * The audio list above says "video decode, encoding of any kind, capture, and every streaming
+ * SINK, server or adaptive-streaming demuxer are out", and it was right to: Homebrew's
+ * `gstreamer` drags in ffmpeg, x264, x265, faac and fdk-aac, which is a redistribution decision
+ * belonging to whoever ships the product rather than to a script that reads a directory. This
+ * list keeps that rule and gets video anyway, by taking the DECODER from the operating system
+ * instead of from the payload — the same trade the audio sinks already make, one element over.
+ *
+ * MEASURED on GStreamer 1.28.6 for which plugin each element actually lives in, because the
+ * element name and the plugin name differ for three of the four:
+ *
+ *   gtk4paintablesink  -> `gtk4`               (gst-plugins-rs; the only GdkPaintable sink)
+ *   videoconvert       -> `videoconvertscale`  (merged with videoscale since 1.22)
+ *   h264parse          -> `videoparsersbad`
+ *   hlsdemux2          -> `adaptivedemux2`     (`hlsdemux` is the older one, in `hls`)
+ *
+ * NOTHING HERE DECODES, and that is the point. On darwin `applemedia` is VideoToolbox
+ * (`vtdec_h264`), on win32 `d3d11`/`mediafoundation` are the OS's own — no GPL element, no
+ * patent-encumbered library inside the bundle, and hardware decode for free. On Linux the answer
+ * would be `libav` or a VA plugin, and Linux has no bundle: it runs on the system GStreamer,
+ * where this was measured playing a real PeerTube HLS stream.
+ *
+ * STAGED, NOT WIRED — {@link expectedGstPlugins} asks for none of this yet, so no build can go
+ * red over it. What each platform is missing is MEASURED rather than left for the first red
+ * build to find:
+ *
+ * **win32**, twice, and both halves agree. The PREFIX: the gvsbuild release this pipeline
+ * extracts (`GTK4_Gvsbuild_2026.6.0_x64.zip`) carries no GStreamer at all — 6741 entries read
+ * out of its central directory, not one plugin — which is why the workflow builds
+ * `gstreamer gst-plugins-base gst-plugins-good` itself, and stops there. Of the four above only
+ * `videoconvertscale` is in base: `videoparsersbad` is -bad and `gtk4` is gst-plugins-rs. The
+ * PAYLOAD: measured on the win11 VM against the published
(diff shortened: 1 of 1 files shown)


JumpLink:
Closing this. An independent review found the docblock's central conclusion false, and I reproduced the correction. Better to redo it than to merge a comment that is confidently wrong.

## The darwin conclusion was backwards

I wrote that `gst-plugins-rs` is not a Homebrew formula — the API answers 404 — and concluded `gtk4paintablesink` would have to be built from source or vendored, "the piece with no second source".

The 404 is real. The inference is not. `Formula/g/gstreamer.rb` vendors gst-plugins-rs itself:

```ruby
resource "rs" do url ".../gst-plugins-rs-gstreamer-1.28.6.tar.bz2" end
(buildpath/"subprojects/gst-plugins-rs").install resource("rs")
depends_on "cargo-c" => :build
-Drs=enabled
-Dgst-plugins-rs:gtk4=enabled
```

So `libgstgtk4.dylib` is already in the keg the workflow's existing `brew install … gstreamer` line installs, and on darwin the remaining work is only adding the names to what `isBundledGstPlugin` copies. A formula-name lookup was the wrong probe; `gst-inspect-1.0 gtk4paintablesink` is the right one.

## And the list is not the video path

Measured with an isolated registry holding exactly `GST_AUDIO_PLUGINS` + the four proposed names + one H.264 decoder:

| stream | result |
| --- | --- |
| fMP4 HLS | plays |
| TS HLS (two sources) | `No suitable plugins found` |
| the same TS stream plus `mpegtsdemux` | plays |

`tsdemux` is in `mpegtsdemux`. My own measurement passed and hid this because PeerTube packages HLS as fMP4, which `qtdemux` from the already-shipped `isomp4` covers. Classic HLS is TS.

## Four more, each fatal to a sentence

* **`mediafoundation` has no video decoder at all** — its element set is two audio decoders, four encoders and a capture source. And `mfh264dec`, which I listed among the measured nulls, is not a GStreamer element in any configuration, so that probe proved nothing.
* **`vtdec_h264` is not an element in 1.28** either. The factory is `vtdec` (rank secondary) with `vtdec_hw` beside it. In a docblock whose whole point is that element and plugin names differ and must be measured, a stale factory name is the wrong kind of error.
* **`adaptivedemux2` is gst-plugins-good**, which win32 already builds — so two of the four arrive today, not one, and my enumeration silently dropped it.
* **`applemedia` ships seven encoders and a capture source**, against the header's own "encoding of any kind, capture … are out". The licensing argument survives (OS-backed, LGPL, no bundled codec library); the stated rule does not.

## What was right

The element→plugin mapping for all four names, and the gvsbuild measurement — the release really carries no GStreamer at all: 6741 central-directory entries, and the only six containing "gst" are glib headers.

## Why closed rather than fixed in place

Two further review findings say the shape is wrong even once the facts are: 56 comment lines for 5 code lines in a file whose own comment-budget script exists to stop that, and a declared gap with no retirement path in a file that demands one (`GST_PLUGIN_GAPS`' own standard: "printed on every build, and fails the day it stops applying"). Finding 1 is what that costs — the darwin gap was false the day it was written and nothing could say so.

The redo is a `status/open-todos.md` entry with a probe that can go red, plus the element→plugin table here and nothing else. Also needed before any of it is wired: `WIN32_LICENSE_FAMILIES` entries for `gst-plugin-gtk4` (MPL-2.0) and `gst-plugins-bad`, which the current catch-all `/^gst[a-z0-9]+\.dll$/i` would misattribute to three projects they do not belong to.

PR #1602, opened 2026-09-08
fix(cli): warn on musl fallback to glibc prebuilds

JumpLink:
## The problem

`prebuildDirCandidates` already prefers `<target>-musl` on a musl host and falls back to the glibc directory when the package ships none. That fallback is **silent**, and it is not harmless.

musl's loader resolves `libc.so.6` and `libm.so.6` to itself, so a glibc-built prebuild loads and mostly works — until it reaches a symbol musl does not implement.

## Measured

OnePlus 6, postmarketOS v26.06, musl 1.2.6, aarch64, GJS 1.88.1, against the published 0.48.0 train. Nine of ten prebuilds loaded. One did not:

```
Error relocating …/@gjsify/lightningcss-native-linux-arm64/prebuilds/linux-arm64/libgjsify_lightningcss.so:
gnu_get_libc_version: symbol not found
```

**Nothing in the resulting failure names libc.** `gjsify install` prints `System dependencies OK`, and the first symptom arrives much later — the CSS plugin is simply absent from a build, which rolldown reports as:

```
rolldown: Bundler::generate: UNLOADABLE_DEPENDENCY "Could not load src/application.css"
```

Working back from that to "one prebuild wants a glibc symbol" took a while. Hence: report it where the fallback is actually chosen.

## The wider gap this does *not* fix

The root cause is that no `-musl` package exists to fall back *from*. Registry check (`registry.npmjs.org`, all ten native packages):

| target | result |
|---|---|
| `linux-arm64` | **10 / 10 published** |
| `linux-arm64-musl` | **0 / 10** — all 404 |
| `linux-x64-musl` | **0 / 10** — all 404 |

The machinery is otherwise in place: the CLI detects the host libc and globs `linux-arm64-musl` first (with specs), and `prebuilds.yml` carries a `build-prebuilds-musl` leg. But `release.yml` does not mention musl once, and as `prebuilds.yml` notes, `-musl` "is not a `gjsify.platforms` token in the first place" — the leg proves compilation, it does not produce publishable per-target packages.

Closing that means adding `-musl` to the platform grammar, per-target packages, and a publish matrix — release infrastructure I can't validate from here, so it is deliberately out of scope. This PR makes the fallback legible in the meantime.

For contrast, `oxlint` — vendored in this very repo — ships `@oxlint/binding-linux-arm64-musl`.

## The change

- `muslPrebuildFallbacks(libc, packages)` — pure, takes libc as a parameter like the rest of the module, since there is no musl runner in CI and an injected `libc: 'musl'` is the only way that branch executes there. It asks `parsePlatformToken` what a musl prebuild is rather than testing the `-musl` suffix itself: the suffix is honoured on linux ONLY, so `darwin-arm64-musl` is a malformed token and not a musl build, and a suffix test reads it as covered.
- `install.ts` prints the affected packages after the existing prebuild report.

Output on the host above:

```
Warning: musl host — 1 of 10 package(s) ship no `-musl` prebuild, so the
  default (glibc) build was used:
    @gjsify/lightningcss-native-linux-arm64
  Most of these run anyway: a bridge that links only GLib records no libc at all.
  One that links glibc loads too — musl aliases `libc.so.6` to itself — and then
  fails at dlopen on its first glibc-only symbol:
    Error relocating <prebuild>.so: <symbol>: symbol not found
  A bridge lost that way is silently absent from the next build rather than
  reported, so the first symptom can be an unrelated-looking bundler error
  (`Could not load <file>` from rolldown, for the CSS bridge). Nothing to
  install: no `-musl` package is published for any target yet.
```

It leads with the proportion because on a musl host the count is EVERY prebuild in
the tree and nine of ten are fine, quotes both error strings because the reverse
mapping from the bundler one is what took a while to work out, and states that
there is nothing to install — a fact now carried in `status/open-todos.md` and
checked from both sides by `generate-status.mjs`.

## Validation

- 9 spec cases in `detect-native-packages.spec.ts`: the musl fallback, all-musl,
  glibc/null quiet, directory-vs-path, the malformed `darwin-arm64-musl` token,
  and a real `detectNativePackages` walk over the existing fixtures in BOTH
  shapes — the directory fallback and the companion-package fallback, which is
  the one actually measured on the phone and which fires a level earlier.
- The five original assertions were additionally executed against the
  **compiled** `lib/utils/detect-native-packages.js` on the musl host.
- CI is green, including the CLI unit suite this branch could not run locally:
  the host has no Node, which is rather the point of the bug.
- `oxfmt --check` over the tree, `oxlint`, `generate-status.mjs --check` and
  `audit-runtimes.mjs --check` clean.

## Reviewed

Reviewed independently before merge; four defects found and fixed, and what the
review deliberately left alone is recorded with it:
https://github.com/gjsify/gjsify/pull/1602#issuecomment-5582331484

https://claude.ai/code/session_01123EBLUnMaph7EQWiNU8GV

### DIFF
--- a/packages/infra/cli/src/commands/install.ts
+++ b/packages/infra/cli/src/commands/install.ts
@@ -1496,7 +1496,7 @@ async function runPostInstallChecks(args: InstallOptions): Promise<void> {
             );
             for (const name of fellBack) console.warn(`  ! ${name}`);
             console.warn(
-                '\nThese load only as far as musl\'s libc.so.6 alias carries them. A glibc-only\n' +
+                "\nThese load only as far as musl's libc.so.6 alias carries them. A glibc-only\n" +
                     'symbol fails at dlopen with "Error relocating … symbol not found", which can\n' +
                     'surface much later as an unrelated build error.',
             );
--- a/packages/infra/cli/src/commands/install.ts
+++ b/packages/infra/cli/src/commands/install.ts
@@ -25,7 +25,7 @@ import { forceExit } from '../utils/force-exit.js';
 import { discoverWorkspaces } from '@gjsify/workspace';
 import type { Command } from '../types/index.js';
 import { buildInstallCommand, detectPackageManager, runMinimalChecks } from '../utils/check-system-deps.js';
-import { detectNativePackages } from '../utils/detect-native-packages.js';
+import { detectHostLibc, detectNativePackages, muslPrebuildFallbacks } from '../utils/detect-native-packages.js';
 import { buildLauncherShims, buildNativeEnvPreamble } from '../utils/bin-shim.js';
 import { installPackages, makeProgressReporter } from '../utils/install-backend.js';
 import { atomicWriteStrict } from '../utils/install-cache-fs.js';
@@ -1484,6 +1484,23 @@ async function runPostInstallChecks(args: InstallOptions): Promise<void> {
             console.log(`  • ${pkg.name}`);
         }
         console.log('\nUse `gjsify run <bundle>` to launch with LD_LIBRARY_PATH/GI_TYPELIB_PATH set.');
+
+        // A musl host resolves `libc.so.6` to its own loader, so a glibc
+        // prebuild loads and mostly works — the failure lands later, on the
+        // first glibc-only symbol, and reads as a bundler error rather than a
+        // libc one. Say it here, where the fallback is actually chosen.
+        const fellBack = muslPrebuildFallbacks(detectHostLibc(process.platform), native);
+        if (fellBack.length > 0) {
+            console.warn(
+                `\nmusl host: ${fellBack.length} package(s) ship no musl prebuild, so the glibc build was used:`,
+            );
+            for (const name of fellBack) console.warn(`  ! ${name}`);
+            console.warn(
+                '\nThese load only as far as musl\'s libc.so.6 alias carries them. A glibc-only\n' +
+                    'symbol fails at dlopen with "Error relocating … symbol not found", which can\n' +
+                    'surface much later as an unrelated build error.',
+            );
+        }
     }
 
     maybeInstallGitHooks();
--- a/packages/infra/cli/src/commands/install.ts
+++ b/packages/infra/cli/src/commands/install.ts
@@ -1485,20 +1485,27 @@ async function runPostInstallChecks(args: InstallOptions): Promise<void> {
         }
         console.log('\nUse `gjsify run <bundle>` to launch with LD_LIBRARY_PATH/GI_TYPELIB_PATH set.');
 
-        // A musl host resolves `libc.so.6` to its own loader, so a glibc
-        // prebuild loads and mostly works — the failure lands later, on the
-        // first glibc-only symbol, and reads as a bundler error rather than a
-        // libc one. Say it here, where the fallback is actually chosen.
+        // The install is the ONLY moment this is knowable: the libc axis is decided
+        // here, the failure it predicts happens in another process, days later, and
+        // names something else entirely (see `muslPrebuildFallbacks`). The list is
+        // the point — it is what lets a later `Error relocating` line be matched to
+        // a package.
         const fellBack = muslPrebuildFallbacks(detectHostLibc(process.platform), native);
         if (fellBack.length > 0) {
             console.warn(
-                `\nmusl host: ${fellBack.length} package(s) ship no musl prebuild, so the glibc build was used:`,
+                `\nWarning: musl host — ${fellBack.length} of ${native.length} package(s) ship no \`-musl\`` +
+                    ' prebuild, so the\n  default (glibc) build was used:',
             );
-            for (const name of fellBack) console.warn(`  ! ${name}`);
+            for (const name of fellBack) console.warn(`    ${name}`);
             console.warn(
-                "\nThese load only as far as musl's libc.so.6 alias carries them. A glibc-only\n" +
-                    'symbol fails at dlopen with "Error relocating … symbol not found", which can\n' +
-                    'surface much later as an unrelated build error.',
+                '  Most of these run anyway: a bridge that links only GLib records no libc at all.\n' +
+                    '  One that links glibc loads too — musl aliases `libc.so.6` to itself — and then\n' +
+                    '  fails at dlopen on its first glibc-only symbol:\n' +
+                    '    Error relocating <prebuild>.so: <symbol>: symbol not found\n' +
+                    '  A bridge lost that way is silently absent from the next build rather than\n' +
+                    '  reported, so the first symptom can be an unrelated-looking bundler error\n' +
+                    '  (`Could not load <file>` from rolldown, for the CSS bridge). Nothing to\n' +
+                    '  install: no `-musl` package is published for any target yet.',
             );
         }
     }
--- a/packages/infra/cli/src/utils/detect-native-packages.spec.ts
+++ b/packages/infra/cli/src/utils/detect-native-packages.spec.ts
@@ -690,19 +690,19 @@ export default async () => {
                 pkg('@gjsify/lightningcss-native-linux-arm64', 'linux-arm64'),
                 pkg('@gjsify/webgl-linux-arm64-musl', 'linux-arm64-musl'),
             ]);
-            expect(fellBack).toEqual(['@gjsify/lightningcss-native-linux-arm64']);
+            expect(fellBack).toStrictEqual(['@gjsify/lightningcss-native-linux-arm64']);
         });
 
         await it('says nothing when every package has a musl build', async () => {
-            expect(muslPrebuildFallbacks('musl', [pkg('@gjsify/webgl', 'linux-arm64-musl')])).toEqual([]);
+            expect(muslPrebuildFallbacks('musl', [pkg('@gjsify/webgl', 'linux-arm64-musl')])).toStrictEqual([]);
         });
 
         // The warning exists for musl hosts only. A glibc host resolving a
         // `linux-arm64` directory is the correct outcome, not a fallback.
         await it('stays quiet on glibc and where there is no libc axis', async () => {
             const packages = [pkg('@gjsify/webgl', 'linux-arm64')];
-            expect(muslPrebuildFallbacks('glibc', packages)).toEqual([]);
-            expect(muslPrebuildFallbacks(null, packages)).toEqual([]);
+            expect(muslPrebuildFallbacks('glibc', packages)).toStrictEqual([]);
+            expect(muslPrebuildFallbacks(null, packages)).toStrictEqual([]);
         });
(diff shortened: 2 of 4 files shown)


JumpLink:
## Independent review, and what I changed

Reviewed this branch on its own merits and pushed `a77211c99a` on top. Three defects, one
scope note, and one thing left for a maintainer.

### 1. The new function displaced another one's documentation

`muslPrebuildFallbacks` had been inserted **between** `resolveHostLibc`'s doc comment and
`resolveHostLibc` itself. The result was two stacked JSDoc blocks on the new function — the
24-line block explaining the two libc probes, `process.report.getReport()` and the musl loader
probe, now read as documentation of the fallback report — and `resolveHostLibc`, the pure
function the whole libc axis rests on, had none.

Moved below `detectHostLibc`, where everything it reads is already defined, and the displaced
doc is back on its function.

### 2. `endsWith(MUSL_SUFFIX)` is a second reader of the token grammar

This module states in `hostPlatformTokens`' own doc that it is *"the SINGLE definition of the
musl preference order. Three callers share it so it cannot drift."* A raw suffix test on the
directory basename made the report a fourth reader of the libc half of that grammar.

Not just tidiness. `parsePlatformToken` honours `-musl` **on linux only**, deliberately, because
musl targets no other kernel — so `darwin-arm64-musl` is a malformed token, not a musl build. The
suffix test read it as *covered*, which in a one-package list means the report says nothing at
all. Now `parsePlatformToken(basename(dir)).libc !== 'musl'`, with a case pinning that
difference; it goes red against the old implementation.

### 3. The report named a problem and no action

On a musl host the count is *every* prebuild in the tree, and by this PR's own measurement nine
of the ten are fine. The old text hedged that in its second paragraph but led with what reads as
ten failures, and offered nothing to do.

It now leads with the proportion, says most run anyway and why, quotes **both** error strings —
the `dlopen` one *and* the bundler one it surfaces as, which is the reverse mapping the PR body
says took a while to work out — and states that there is nothing to install because no `-musl`
package is published. Rendered:

```
Warning: musl host — 10 of 10 package(s) ship no `-musl` prebuild, so the
  default (glibc) build was used:
    @gjsify/webgl-linux-arm64
    …
  Most of these run anyway: a bridge that links only GLib records no libc at all.
  One that links glibc loads too — musl aliases `libc.so.6` to itself — and then
  fails at dlopen on its first glibc-only symbol:
    Error relocating <prebuild>.so: <symbol>: symbol not found
  A bridge lost that way is silently absent from the next build rather than
  reported, so the first symptom can be an unrelated-looking bundler error
  (`Could not load <file>` from rolldown, for the CSS bridge). Nothing to
  install: no `-musl` package is published for any target yet.
```

That last line is a claim about the world, so it is now a `status/open-todos.md` entry the code
anchors to — `generate-status.mjs` checks the anchor names a real heading *and* that the heading
is not a corpse, so the two halves cannot drift.

### 4. Tests: the seam is testable, but the spec stopped short of the join

`muslPrebuildFallbacks('musl', …)` is the real function with an injected libc, no mock — that
part is sound. But all four cases are handed hand-built `NativePackage` literals, so nothing
proved a real walk spells `prebuildsDir` the way the report reads it. Added that join over the
fixtures already in the file, in **both** shapes:

* the directory fallback (`prebuilds/linux-x64` where no `linux-x64-musl` exists), and
* the **companion-package** fallback — `@gjsify/<name>-linux-x64` with no `-musl` sibling, which
  is the shape that was actually measured on the phone, and which happens one level earlier than
  the directory probe.

### What I deliberately did not do

* **No `-musl` publishing.** Out of scope as declared, and I agree.
* **`gjsify info` still says nothing about the fallback.** It prints `name → prebuildsDir`, so a
  reader who knows what to look for can already see it, and it is the command someone runs while
  *debugging* rather than months earlier at install time — arguably the higher-value site. Doing
  it properly means one shared printer rather than a second copy of the message, so it is a
  separate change, not a silent widening of this one.
* **No `docs/prebuilds.md` change.** A diagnostic message is not an architectural decision; the
  ledger entry is the record.

### For a maintainer, not for me

The PR frames the root cause as *"no `-musl` package exists to fall back from"*. That is half of
it. The other half is that a glibc-only package was installed on a musl host **at all**, which
`applyPlatformFilter` already exists to prevent, and `tests/e2e/install-platform-filter` was
written for that exact outcome (*"on Alpine the installer handed a glibc-only prebuild to a musl
host and failed at `dlopen` instead of at install time"*).

The whole incident reproduces from the committed artifacts, no phone required. Of the ten
`linux-arm64` bridges, exactly one image references a glibc-only symbol:

```
$ readelf -sW packages/infra/lightningcss-native-linux-arm64/prebuilds/linux-arm64/libgjsify_lightningcss.so \
    | grep gnu_get_libc_version
83: 0000000000000000  0 FUNC  GLOBAL DEFAULT  UND gnu_get_libc_version@GLIBC_2.17
```

The other nine reference none — which is exactly why nine of ten loaded. And the same package's
`linux-x64` half **is** refused on musl by npm's `libc`, while the arm64 half is not:

| target | `DT_NEEDED` loader | npm `libc` written | glibc-only symbol |
|---|---|---|---|
| `linux-x64` | `ld-linux-x86-64.so.2` | `["glibc"]` | yes |
| `linux-arm64` | none | none | yes |

`generate-platform-packages.mjs` writes `libc: ["glibc"]` only when a glibc **dynamic loader** is
recorded, and that is a documented, defensible choice — declaring it off a `libc.so.6`
`DT_NEEDED` would refuse the libc-agnostic bridges on the hosts where they work. But the loader
is not recorded consistently for the same source, so one target is filtered and the other
installs and breaks. Had the arm64 companion declared `libc: ["glibc"]`, it would have been
`inert` on that phone and `gjsify build` would have reported a missing CSS bridge instead of
dying inside one.

Widening the predicate to "references a symbol musl does not export" makes more packages `inert`
on musl and needs a musl symbol set to be sound, so it is a policy change to `prebuild-libc`
rather than a repair — a maintainer's call, and written down in the ledger entry rather than
attempted here.

### Verified

Everything below was run and observed, on a glibc x86-64 host with an uninstalled worktree:

* `oxfmt --check .` — clean over 3783 files. This is the step that had `Whole-tree checks` red;
  the merge-base checks clean, so the branch introduced all of it (a `'…musl\'s…'` string oxfmt
  rewrites to double quotes, a `.filter().map()` chain that fits at printWidth 120, two trailing
  blank lines). `a9ceea90da` had already fixed it; confirmed still clean after my change.
* `oxlint` with the repo config — exit 0 on the three files, and the JS-plugin host proven loaded
  with a discriminator (a bare `TODO` in a scratch file makes `gjsify/todo-needs-anchor` fire), so
  the clean run is not the Rust subset passing quietly.
* The 58 spec cases green under a Node harness. **The harness's matchers reproduce
  `@gjsify/unit`'s semantics rather than a convenient approximation** — including `toEqual`
  being `==` — and it fails exactly the four assertions CI failed on `1ede3594c5`, which is the
  discriminator that makes its green worth anything. Each new assertion additionally shown red
  against a deliberately broken implementation (suffix test restored, `basename` dropped, libc
  guard dropped).
* `tsc --noEmit -p packages/infra/cli` — identical error set before and after (207, every one
  cascading from `TS2307` on unbuilt workspace packages).
* `generate-status.mjs --check` — clean, which is what holds the new ledger anchor.
* The warning text rendered from the real code path against the measured ten-package population,
  not retyped.

**Not verified, and not verifiable from here:** that the message is right *on a musl host*. The
fallback branch has never executed outside an injected `libc`, here or in CI.

JumpLink:
## Measured on the device after merge: this warning cannot fire

Ran the merged tree on the OnePlus 6T it was written for — postmarketOS, musl 1.2.6,
aarch64, gjs 1.88.1, no Node — by bundling `main`'s CLI to a single gjs file
(`gjsify build src/index.ts --app gjs`) and installing five native bridges into a
scratch project. The prebuild report printed. **The musl warning did not.**

```
Detected 5 @gjsify/* package(s) with native prebuilds:
  • @gjsify/http2-native-linux-arm64
  • @gjsify/lightningcss-native-linux-arm64
  • @gjsify/sab-native-linux-arm64
  • @gjsify/terminal-native-linux-arm64
  • @gjsify/tls-native-linux-arm64

Use `gjsify run <bundle>` to launch with LD_LIBRARY_PATH/GI_TYPELIB_PATH set.
```

Probing the inputs in the same process, on that host:

```
process.platform          = "linux"
process.report            = undefined
resolveHostLibc({muslLoaderPresent: true}) = "musl"    ← the pure function is right
resolveHostPlatform()     = {"os":"linux","cpu":"arm64","libc":"musl"}   ← right
detectHostLibc("linux")   = "glibc"                    ← WRONG
muslPrebuildFallbacks(detectHostLibc(...), native) = []
muslPrebuildFallbacks('musl', native)              = [all five]
```

So the reported logic is sound and the reporting is unreachable: the call site passes
`detectHostLibc(process.platform)`, which answers `glibc` on a musl host, so
`fellBack` is empty and the block never runs. Ten runs, ten times `glibc`.

### Why `detectHostLibc` is wrong where `resolveHostLibc` is right

`process.report` is undefined under GJS, so the decision falls to
`muslLoaderPresent`, and that is read through the `node:fs` shim:

```ts
existsSync('/lib') && readdirSync('/lib').some((f) => f.startsWith('ld-musl-'))
```

On that host `/lib` is a symlink to `/usr/lib` and holds 2076 entries including
`ld-musl-aarch64.so.1`. Through the shim, in a bundle that does not import Gio:

```
readdirSync('/lib')                     = 77 entries, no ld-musl-*
existsSync('/lib/ld-musl-aarch64.so.1') = false
```

Same bundle with `import Gio from '@girs/gio-2.0'` added, and a `Gio` enumeration of
the same directory beside it for comparison:

```
shim readdirSync : 2076 entries, ld-musl: true
Gio enumerate    : 2076 entries, ld-musl: true
existsSync(...)  = true
```

`readdirSync` returns a **truncated listing and no error**, and `existsSync` answers
`false` for a file that is there. That is a defect in `@gjsify/fs` on GJS, not in this
PR — and it is far wider than libc detection, because a silently short `readdirSync`
is wrong for every caller. Worth its own issue.

This PR's own half is smaller: it rolled a second libc detector, and the one already
in the package — `resolveHostPlatform()` in `platform-check.ts`, whose reader consults
`sharedObjects` rather than listing `/lib` — got the right answer in the very same
process. The fix is to ask that one.

### Correction to this PR's description

The sample output block in the merged description is **mine and it is wrong**. I
rewrote that section before merging, to replace the pre-review message, and
reconstructed the numbers from the format string instead of measuring them — it reads
`1 of 10`, and no `-musl` package is published for any target, so every detected
prebuild is a fallback and the real shape is `N of N`. In the run above it would be
`5 of 5`. Apologies; the commit on `main` carries that error and this comment is the
correction.

PR #1603, opened 2026-09-08
fix(npm-registry): retry past a registry blip, body included

JumpLink:
## What happened

`main` went red on 2026-09-08. The `darwin-x64` leg of **macOS suites** died at 32% of an
`install --immutable`, on one line:

```
gjsify install: downloading 530/1658 (32%)
fetch failed
##[error]Process completed with exit code 1.
```

Nothing else on that commit failed. `darwin-arm64` was green on the same tree, so were the
other eight workflows, and **macOS suites** had passed on the twelve runs before it. Between
the last progress line and the error sit 53 seconds in which nothing downloaded. The registry,
or that runner's route to it, went away for about a minute.

An installer is supposed to absorb that. Three reasons this one did not.

## 1. The budget was 1.75 seconds

`fetchWithRetry` defaulted to `retries: 3` at 250/500/1000 ms.

For comparison, npm's own defaults are `fetch-retries=2` with `fetch-retry-mintimeout=10s`
and `fetch-retry-factor=10`, so npm waits about seventy seconds across three attempts; pnpm
ships npm's numbers. A cold install of this tree fans the whole dependency closure out at
16-way concurrency for minutes, so a blip only has to outlast the budget **once** to fail the
run — and with 1.75 s, most blips do.

Now 1s, 2s, 4s, 8s, 8s: six attempts, ~23 s of patience, still capped at 8 s per delay. It is
a default, so a caller wanting the old impatience passes `retries` / `retryDelayMs`.

## 2. Only the headers were retried

This is the sharper bug, and it survives any retry budget.

`fetchWithRetry` returned a `Response` once the status line was in. Both callers then read the
body **outside** the retried region:

```ts
const res = await fetchWithRetry(url, ...);      // retried
const buf = new Uint8Array(await res.arrayBuffer());  // not retried
```

A connection that dies mid-stream throws there with nothing to catch it. For a tarball that is
the body most likely to be dropped part-way, and for a full packument it is megabytes.

`opts.read` now brings the body inside the loop, so a body that stops arriving is classified and
retried like any other network error.

Validation deliberately stays **outside** it. `assertPackument` throws `TypeError`, and
`isRetryableError` reads every `TypeError` as transient — inside `read`, a malformed packument
would have been re-fetched six times and then reported as a network fault. `read` covers
transport; the caller checks what arrived. A test pins that: a malformed packument is fetched
exactly once.

## 3. The failure named nothing

The raw `TypeError: fetch failed` propagated, which is why the CI log above is one unactionable
line: it says neither which of an install's several thousand requests died, nor how hard it was
tried.

`RegistryUnreachableError` names the URL and the attempts spent and keeps the original error as
`cause`. It is the sibling of the existing `RegistryTimeoutError`, which already did this for
requests that hang rather than die — the same argument, applied to the branch that had been
left raw.

This changes a tested contract: the old test asserted that the underlying `TypeError` came
through verbatim. That test is now the assertion that the URL and the attempt count are named.

## Verified

- `gjsify run test` in `packages/infra/npm-registry` — 48 tests, both **Node 24.19.0** and
  **GJS 1.88.1**. New: a tarball body and a packument body that error mid-stream and are
  retried; a malformed packument fetched exactly once; the six-attempt default.
- `gjsify tsc --noEmit` clean in `npm-registry` and in the CLI that consumes it.
- `oxlint` and `oxfmt --check` clean.

## Scope

This does not explain the outage away, it survives it. Jitter across the concurrent downloads
is a separate question — 16 requests currently retry in lockstep, which is the wrong shape
against a rate-limiting CDN — and is deliberately not in here.

### DIFF
--- a/packages/infra/npm-registry/src/errors.ts
+++ b/packages/infra/npm-registry/src/errors.ts
@@ -42,3 +42,36 @@ export class RegistryTimeoutError extends Error {
         this.name = 'RegistryTimeoutError';
     }
 }
+
+/**
+ * Thrown when every retry attempt against a registry URL fails at the network
+ * layer (`fetch` rejects, the connection resets, the body stops arriving
+ * mid-stream). The sibling of {@link RegistryTimeoutError} for the case where
+ * the request does not hang but dies.
+ *
+ * It exists for the same reason: the raw error the runtime hands us is
+ * `TypeError: fetch failed` (undici) or a bare `FetchError` (GJS/libsoup),
+ * neither of which says WHICH of an install's several thousand requests died,
+ * how many attempts it got, or how long it was given. A CI log that ends in one
+ * unattributed `fetch failed` line is not enough to tell a registry outage from
+ * a bug, which is what happened to gjsify's own macOS run on 2026-09-08.
+ *
+ * The underlying error stays reachable as `cause`.
+ */
+export class RegistryUnreachableError extends Error {
+    constructor(
+        public readonly url: string,
+        public readonly attempts: number,
+        public readonly elapsedMs: number,
+        cause: unknown,
+    ) {
+        const seconds = Math.round(elapsedMs / 100) / 10;
+        const reason = cause instanceof Error ? cause.message : String(cause);
+        super(
+            `@gjsify/npm-registry: GET ${url} failed after ${attempts} attempt(s) over ~${seconds}s: ${reason}. ` +
+                `This usually means the registry or the network dropped out.`,
+        );
+        this.name = 'RegistryUnreachableError';
+        this.cause = cause;
+    }
+}
--- a/packages/infra/npm-registry/src/index.spec.ts
+++ b/packages/infra/npm-registry/src/index.spec.ts
@@ -12,6 +12,7 @@ import {
     fetchTarball,
     PackageNotFoundError,
     IntegrityError,
+    RegistryUnreachableError,
     type NpmrcConfig,
 } from './index.js';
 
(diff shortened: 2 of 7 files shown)


JumpLink:
## Reach: this does not cover CI's bootstrap install until it is published

Worth stating, because a `fetch failed` in CI after this landed reads as the fix not
working, and an independent reviewer reached exactly that conclusion today from the
log alone — inferring that a raw `TypeError` must be bypassing `fetchWithRetry` or
escaping a re-throw inside it.

It is not. Every workflow that installs the tree bootstraps from the PUBLISHED CLI
first, by design — `Fetch the published CLI (a cold tree has no other)`:

```
npm install --no-save --prefix "$RUNNER_TEMP/bootstrap-cli" @gjsify/cli@latest
node "$RUNNER_TEMP/bootstrap-cli/node_modules/@gjsify/cli/lib/index.js" install --immutable
```

So the install that downloads 1658 packages is run by `@gjsify/cli@latest` off npm,
which bundles the published `@gjsify/npm-registry` — 0.48.0, from before this commit.
Its budget is still 1.75s, its body read is still outside the retry, and its
exhaustion still throws the bare `TypeError: fetch failed` that reaches
`commands/install.ts`'s `console.error(err.message)`. Three such failures across two
PRs on 2026-09-08, all on macOS runners mid-download, all from that step.

The fix therefore takes effect for CI's bootstrap only at the next release of
`@gjsify/cli`. It is live now for anything run out of a built tree — which is where
`RegistryUnreachableError` will start naming the URL and the attempt count.

Nothing to change here; the retry code is doing what it says. Recording it so the
next reader of a bootstrap `fetch failed` does not go looking for a bypass that
is not there.

PR #1604, opened 2026-09-08
fix(rolldown-plugin-gjsify): never inline a read of the host

JumpLink:
## What was wrong

The build-time read inliner accepted any statically-resolvable **absolute** path. Its last
gate before reading from disk was:

```ts
if (isAbsoluteFsPath(v)) return v;
```

`/lib` passes that. So a read of the machine's own layout was evaluated on the **builder**
and frozen into the artifact as a literal.

## Measured

OnePlus 6T, postmarketOS, musl 1.2.6, aarch64, gjs 1.88.1, running a `cli.gjs.mjs` built on
Fedora:

```
readdirSync('/lib')                       77 entries, no ld-musl-*
existsSync('/lib/ld-musl-aarch64.so.1')   false
```

The directory really holds 2076 entries and `ld-musl-aarch64.so.1` is one of them. 77 was the
**builder's** `/usr/lib`. Ten runs, ten times the same.

Those two reads are exactly what `detectHostLibc` decides from, so a bundled `gjsify` answered
`glibc` on a musl machine. `detectHostLibc` has four callers, so what was inert for the life of
the bundle is: the `-musl` preference in `prebuildDirCandidates`, the launcher shims' libc,
`install-global`'s target, and the fallback report added in #1602.

Nothing failed loudly. A glibc prebuild loads under musl's loader right up to its first
glibc-only symbol — which is the incident #1602 reported from the other end, and why nine of
ten prebuilds on that phone worked.

## The change

Resolvable is not the same as bundlable. This module exists so a package can carry its **own**
resources — `new URL('../package.json', import.meta.url)` — whose content is a property of the
build. A path outside the reading module's package is either the machine's layout or another
package's business, and its build-time answer is not its run-time answer.

- `resourceRootFor` walks up to the nearest `package.json`, falling back to the module's own
  directory when nothing above declares one — which is what keeps the existing
  `readdirSync(path.join(<tmpdir>))` row inlining.
- `isWithin` gates on `relative()`, not a prefix test: `startsWith` calls `/lib64` a child of
  `/lib`.
- Declining is the safe direction — the call stays put and reads at runtime, which is what a
  probe wants. Silent on purpose: probing `/proc` or `/lib` is ordinary in a runtime shim, and
  a warning per probe would bury the declines that matter.

## Blast radius, instrumented over a full `build:gjs-bundle` of `@gjsify/cli`

The gate declines **four paths / five reads** (`/lib` is `existsSync` + `readdirSync` in one
expression), and every one of them was wrong before:

| path | what it fed | how it was wrong |
|---|---|---|
| `/etc/alpine-release` | `napi-node-addon`'s Alpine test | frozen `false` on a glibc builder; the phone above reports `3.24.1` |
| `/lib` | `detectHostLibc` | the libc probe above |
| `/proc` | `foreach`'s ppid map, the process-tree kill fallback | frozen as the **builder's** PID list |
| `/proc/self` | `install-lock`'s liveness probe | frozen **true**, so on a platform without `/proc` the bundled CLI took the `/proc` branch, found no `/proc/<pid>`, judged every holder dead and **stole the lock** |

**No package-resource read was declined**, so the inlining this module exists for is untouched.

## Verified

- 1715 tests in `@gjsify/cli` on Node 24.19.0. New rows: a declined host read, a declined
  sibling-package file, a still-inlined own resource via `import.meta.url`, plus unit coverage
  for `isWithin` (including the `/lib64` case) and `resourceRootFor`.
- The rebuilt `cli.gjs.mjs` carries no entry name from the builder's `/lib` (`NetworkManager`,
  `Marker.extensions`, `binfmt.d`) and keeps `ld-musl-` as runtime code.

  An earlier revision of this section named `libbd_smartmontools`, `lib++dfb` and
  `libverto.so.1` instead. That was a check that could not fail: on the Fedora builder those
  live in `/lib64`, so they were never candidates for a `/lib` read and were absent from the
  pre-fix bundle too. They appeared in my first measurement because they are in the PHONE's
  single `/usr/lib`, which has no lib64 split. Caught in review.
- On the phone, the fixed bundle reports the musl fallback for all five installed bridges:

```
Warning: musl host — 5 of 5 package(s) ship no `-musl` prebuild, so the
  default (glibc) build was used:
    @gjsify/http2-native-linux-arm64
    @gjsify/lightningcss-native-linux-arm64
    ...
```

The previous bundle printed nothing there. That output also settles the `N of N` shape — the
`1 of 10` in #1602's merged description was my reconstruction and it was wrong; corrected in a
comment on that PR.

- `oxfmt --check` and `oxlint` clean on both touched packages.

## Not in here

`gjsify install --global` returns before `runPostInstallChecks`, so a global install still
reports nothing. That is not a one-line fix: those checks are cwd-scoped by construction
(`detectNativePackages(process.cwd())`, `ensureProjectGjsEngine`), so calling them there would
report an unrelated tree. It needs a root threaded through, and a decision about what a global
install should say.

## Two holes review found in the first version of this gate

**The gate never ran on `new URL(…)`** — the spelling this module was written for.
`evalPathExpr`'s URL branch returned `fileURLToPath(v)` before the containment test, so
`readdirSync(new URL('/proc', import.meta.url))` still froze the builder's PID list while the
same read spelled `fileURLToPath(new URL(…))` was declined. Two spellings of one expression
cannot answer differently. There is now one exit for both shapes.

**And closing that exposed a defect the gate had already shipped.** `resourceRootFor`'s
"nearest ancestor with a `package.json`" stops at a dual-publish `{"type":"commonjs"}` marker
directory, so a module under one had the read of *its own manifest* declined — trading a wrong
constant for the ENOENT-after-move this module exists to prevent. Not hypothetical: in this
tree `engine.io-client/build/{cjs,esm}`, `@socket.io/component-emitter/lib/cjs`,
`lru-cache/dist/{esm,commonjs}` and `entities/lib/esm` are that shape, and some carry the
package's own `name`, so neither presence nor a `name` distinguishes marker from root. The
rule is now structural for installed files — the last `node_modules/<name>` segment on the way
up, scope kept whole, no filesystem probe so a Yarn-PnP zip path resolves too — and
nearest-`package.json` for first-party, which must stay *nearest* or a monorepo package
becomes its workspace.

The two defects cancelled for one spelling, which is why the first measurement looked clean.
Every new row fails against both the pre-PR module and the first version of the gate.

`pnp-zip-static-reads` — the suite most exposed to the structural rule, and the one that could
not be run locally — passes in CI on both its `--app node` and `--app gjs` cases.

### DIFF
--- a/packages/infra/cli/src/inline-static-reads.spec.ts
+++ b/packages/infra/cli/src/inline-static-reads.spec.ts
@@ -130,6 +130,58 @@ export default async () => {
             expect(out.inlined).toBe(0);
         });
 
+        await it("does NOT inline a sibling package's file spelled as a URL either", () => {
+            // THE TWIN OF THE ROW ABOVE, and the reason it has to exist: the gate
+            // sat after the URL branch of `evalPathExpr` had already returned, so
+            // the same read declined as `path.join(<abs>)` was inlined as
+            // `new URL('../../b/…', import.meta.url)` — which is not an exotic
+            // spelling but the one this whole module was written for. Measured
+            // before the branches were merged: `inlined: 1`, with b's bytes in the
+            // output.
+            const dir = mkdtempSync(join(tmpdir(), 'gjsify-sibling-url-'));
+            const reader = join(dir, 'node_modules', 'a', 'lib');
+            const other = join(dir, 'node_modules', 'b');
+            mkdirSync(reader, { recursive: true });
+            mkdirSync(other, { recursive: true });
+            writeFileSync(join(dir, 'node_modules', 'a', 'package.json'), '{"name":"a"}');
+            writeFileSync(join(other, 'package.json'), '{"name":"b"}');
+            writeFileSync(join(other, 'data.json'), '{"secret":1}');
+            const src = `
+                import { readFileSync } from 'fs';
+                const d = readFileSync(new URL('../../b/data.json', import.meta.url), 'utf8');
+            `;
+            const out = inlineStaticReads(src, join(reader, 'index.js'));
+            rmSync(dir, { recursive: true, force: true });
+            expect(out.inlined).toBe(0);
+            expect(out.contents).toContain('../../b/data.json');
+        });
+
+        await it('DOES still inline the manifest a dual-published module reads past its own type marker', () => {
+            // THE COST THE GATE MUST NOT HAVE, and it is not hypothetical: a
+            // package that publishes both formats drops a `{"type":"commonjs"}`
+            // marker beside the emitted output, and that marker IS a
+            // `package.json`. Answering "the nearest one" would put the resource
+            // root at `dist/cjs` and decline the module's read of its OWN manifest
+            // one level up — leaving live in the bundle exactly the read this file
+            // exists to remove. Measured in this repo's installed tree:
+            // `engine.io-client/build/{cjs,esm}` and
+            // `@socket.io/component-emitter/lib/cjs` are that shape.
+            const dir = mkdtempSync(join(tmpdir(), 'gjsify-dual-'));
+            const pkg = join(dir, 'node_modules', 'dual');
+            mkdirSync(join(pkg, 'dist', 'cjs'), { recursive: true });
+            writeFileSync(join(pkg, 'package.json'), '{"name":"dual","version":"7.7.7"}');
+            writeFileSync(join(pkg, 'dist', 'cjs', 'package.json'), '{"type":"commonjs"}');
+            const src = `
+                import { readFileSync } from 'fs';
+                import { fileURLToPath } from 'url';
+                const v = readFileSync(fileURLToPath(new URL('../../package.json', import.meta.url)), 'utf8');
+            `;
+            const out = inlineStaticReads(src, join(pkg, 'dist', 'cjs', 'index.js'));
+            rmSync(dir, { recursive: true, force: true });
+            expect(out.inlined).toBe(1);
+            expect(out.contents).toContain('7.7.7');
+        });
+
         await it("DOES still inline the package's own resource", () => {
             // The happy path the gate must not cost: a package reading its own
             // file through `import.meta.url`, which is the whole reason this
@@ -191,12 +243,52 @@ export default async () => {
         });
     });
 
+    // Two rules, because "the reading module's package" is a different question for
+    // an installed file than for a first-party one, and one answer gets the other
+    // wrong. The pair below is what says so.
     await describe('resourceRootFor', async () => {
-        await it('is the nearest ancestor carrying a package.json', () => {
+        await it('is the nearest ancestor carrying a package.json, for a first-party file', () => {
             const dir = mkdtempSync(join(tmpdir(), 'gjsify-root-'));
             mkdirSync(join(dir, 'pkg', 'src', 'deep'), { recursive: true });
             writeFileSync(join(dir, 'pkg', 'package.json'), '{"name":"p"}');
             expect(resourceRootFor(join(dir, 'pkg', 'src', 'deep', 'x.ts'))).toBe(join(dir, 'pkg'));
+            rmSync(dir, { recursive: true, force: true });
+        });
+
+        await it('is the NEAREST one and not the outermost, so a monorepo package is not the workspace', () => {
+            // The outermost would be the workspace root, and one package could then
+            // inline its siblings' files.
+            const dir = mkdtempSync(join(tmpdir(), 'gjsify-ws-'));
+            mkdirSync(join(dir, 'packages', 'a', 'src'), { recursive: true });
+            writeFileSync(join(dir, 'package.json'), '{"name":"ws","workspaces":["packages/*"]}');
+            writeFileSync(join(dir, 'packages', 'a', 'package.json'), '{"name":"a"}');
+            expect(resourceRootFor(join(dir, 'packages', 'a', 'src', 'x.ts'))).toBe(join(dir, 'packages', 'a'));
+            rmSync(dir, { recursive: true, force: true });
+        });
+
+        await it('is the node_modules package directory for an installed file, marker package.json or not', () => {
+            // Structural, not probed: a `{"type":"commonjs"}` marker deeper in the
+            // tree must not become the root — see the dual-publish row above.
+            const dir = mkdtempSync(join(tmpdir(), 'gjsify-installed-'));
+            const pkg = join(dir, 'node_modules', 'dual');
+            mkdirSync(join(pkg, 'dist', 'cjs'), { recursive: true });
+            writeFileSync(join(pkg, 'package.json'), '{"name":"dual"}');
+            writeFileSync(join(pkg, 'dist', 'cjs', 'package.json'), '{"type":"commonjs"}');
+            expect(resourceRootFor(join(pkg, 'dist', 'cjs', 'index.js'))).toBe(pkg);
+            rmSync(dir, { recursive: true, force: true });
+        });
+
+        await it('keeps a scoped name whole, and takes the LAST node_modules on the way up', () => {
+            // A nested dependency and a scoped one must resolve to THEMSELVES, not
+            // to whatever contains them — the property a Yarn-PnP zip path depends
+            // on as much as a hoisted tree does.
+            const dir = mkdtempSync(join(tmpdir(), 'gjsify-scoped-'));
+            const scoped = join(dir, 'node_modules', '@scope', 'name');
+            const nested = join(dir, 'node_modules', 'a', 'node_modules', 'b');
+            mkdirSync(join(scoped, 'lib'), { recursive: true });
+            mkdirSync(join(nested, 'lib'), { recursive: true });
+            expect(resourceRootFor(join(scoped, 'lib', 'x.js'))).toBe(scoped);
+            expect(resourceRootFor(join(nested, 'lib', 'x.js'))).toBe(nested);
             rmSync(dir, { recursive: true, force: true });
         });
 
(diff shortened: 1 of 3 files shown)


JumpLink:
## Independent review pass — the central claim survives, the gate did not

I re-derived every claim instead of reading it, and pushed one commit. Summary: **the bug
and its four downstream consequences are real and I reproduced all of them**, including on
the musl device — but the gate as merged did not close the hole for the expression shape
this module is built around, and closing it exposed a second, opposite silent wrongness.

---

## 1. The gate never saw `new URL(…)`

`evalPathExpr`'s URL branch `return`ed **before** the containment check ran. So the read the
whole module exists for — `new URL(<lit>, import.meta.url)` — was ungated, while the
identical read spelled `fileURLToPath(new URL(…))` or `path.join(<abs>)` was declined. Two
spellings of one expression answered differently.

Measured by running the pre-PR module, the module as merged, and the fix side by side over
the same sources:

| case | pre-PR | **as merged** | after this commit |
|---|---|---|---|
| `readdirSync(new URL('/proc', import.meta.url))` | inlined | **inlined** | declined |
| `readdirSync(new URL('file:///lib'))` | inlined | **inlined** | declined |
| sibling package via `new URL('../../b/data.json', import.meta.url)` | inlined | **inlined** | declined |
| sibling package via `path.join(<abs>)` | inlined | declined | declined |
| own resource via `new URL('../package.json', import.meta.url)` | inlined | inlined | inlined |

`readdirSync(new URL('/proc', import.meta.url))` came back as the builder's PID list under
the merged gate. Note the existing `does NOT inline a sibling package's file` row uses the
`path.join` spelling — the one that was already covered. Its `new URL` twin is now beside
it, and it fails against both the pre-PR module and the gate as merged.

## 2. Closing it broke a legitimate read, so `resourceRootFor` had to change too

`resourceRootFor`'s "nearest ancestor with a `package.json`" lands on the wrong directory
for any package that dual-publishes: a `{"type":"commonjs"}` marker beside the emitted
output **is** a `package.json`. Measured in this repo's own installed tree —
`engine.io-client/build/cjs`, `engine.io-client/build/esm`,
`@socket.io/component-emitter/lib/cjs` are that shape, and some markers carry the package's
own `name`, so neither presence nor a `name` field distinguishes a marker from a root.

```
resourceRootFor('<t>/node_modules/dual/dist/cjs/index.js')
  as merged        → <t>/node_modules/dual/dist/cjs      ← reading ../../package.json is "outside"
  after this commit → <t>/node_modules/dual
```

Under the merged gate that module's read of **its own manifest** was declined (verified:
`inlined: 0`), which ships the live read this module exists to remove — an ENOENT once the
bundle moves. That is the same silent wrongness in the other direction, which is exactly
what a containment rule risks. It only looked fine because the `new URL` spelling of it was
slipping through the hole in §1; the two defects were cancelling.

The rule is now two rules, because "the package" is two questions: an installed file's root
is **structural** (`node_modules/<name>`, scope kept whole, last segment on the way up — so
a nested dep and a PnP zip resolve to themselves, and no `existsSync` is needed on a
zip-resident path); a first-party file keeps the nearest-`package.json` walk, which has to
stay *nearest* or a monorepo package becomes its whole workspace. Both halves have rows.

## 3. Blast radius, re-measured

Instrumented `evalPathExpr` over a full `build:gjs-bundle` of `@gjsify/cli` (57 resolvable
candidates):

```
27  IN   string        15  IN   URL        15  OUT  string
declines:  /lib ×6   /proc ×3   /proc/self ×3   /etc/alpine-release ×3
```

Four paths, five distinct reads in source (`/lib` is `existsSync` + `readdirSync` in one
expression); each source module is transformed three times, hence the multiples. **No
package-resource read declined, no URL-shaped read declined.** Identical before and after my
commit, so §1 and §2 cost this bundle nothing. Strictly, "declines exactly four reads" is
four *paths* / five *reads* — the substance holds.

`dist/affected.gjs.mjs` rebuilds byte-identical (`cb4038c3…`), so the tracked-artifact
compare passes.

## 4. The four downstream bugs — all verified, read off two real bundles

Two `cli.gjs.mjs` builds, identical except that one has the gate disabled:

| | gate off (pre-fix) | gate on |
|---|---|---|
| `isPidAlive` | `function isPidAlive(n){return existsSync(\`/proc/${n}\`)}` — the `/proc/self` guard folded to `true` and **the signal-0 fallback was dead-code-eliminated** | guard present |
| `buildPpidMap` | `S=\`1.101.10198.102.103.…\`` — the **builder's** PID list, comma-joined by the minifier | `S=readdirSync(\`/proc\`)` |
| `detectHostLibc` | `` …tmpfiles.d,tuned,udev,virtualbox`.split(`,`).some(e=>e.startsWith(`ld-musl-`)) `` | `` existsSync(`/lib`)&&readdirSync(`/lib`).some(…) `` |
| `isMusl` | the string `alpine-release` is **absent from the bundle entirely** (folded to `false`) | `` existsSync(`/etc/alpine-release`) `` |

The `install-lock` one is as strong as claimed and slightly stronger than stated: the
`process.kill(pid, 0)` fallback is not merely unreachable, the minifier removes it, so on a
platform without `/proc` every holder reads dead and `assessLock` returns
`stale: true, reason: "owner pid N is gone"` → the lock is stolen.

Host facts on the builder: `/lib` holds **77** entries and no `ld-musl-*` — the same 77 the
phone reported, which is what makes "the phone was reading the builder's directory"
airtight. `/etc/alpine-release` is absent here and reads `3.24.1` on the phone.

**One correction to the PR's Verified section:** "carries no builder library names
(`libbd_smartmontools`, `lib++dfb`, `libverto.so.1` all absent)" is a check that cannot
fail. Those files live in `/lib64`, not `/lib`, so they were never candidates — they are
absent from the *pre-fix* bundle too. The real discriminator is a `/lib` entry name:
`abrt-java-connector` and `binfmt.d` appear in the pre-fix bundle and are absent from the
fixed one.

## 5. On the musl device, both directions

Same source, same builder, two `--app gjs` bundles of a one-line `detectHostLibc('linux')`
probe differing only in whether the gate runs. OnePlus 6T, postmarketOS, musl 1.2.6,
aarch64, gjs 1.88.1, `/lib` = 2074 entries incl. `ld-musl-aarch64.so.1`:

```
probe-unfixed.gjs.mjs    libc=glibc libc=glibc libc=glibc
probe-fixed.gjs.mjs      libc=musl  libc=musl  libc=musl
```

Three runs each. That is the fix working on the target platform, not just the bug existing.
Scratch files removed from the device.

## 6. What I left, and why

- **Silent decline — agreed, and I strengthened the reason rather than the mechanism.** The
  `isAbsoluteFsPath` precedent is not really "a silent decline is dangerous"; it is "a
  decline no test can see is dangerous". A warning would not have caught that one either,
  because nobody was reading the build log for it. Rows are what catch it, and the declines
  now have rows. There is no debug channel in this plugin to route them to, and inventing
  one for four host probes is not worth it.
- **`isWithin` stays textual — no `realpathSync`.** Both directions of that choice are the
  safe one: a read reaching into the package through a symlink resolves outside the root and
  is *declined* (leaving a working runtime read), and a path textually inside cannot be
  widened by a symlink npm never creates. `resolve()` normalises `..` before comparing, so
  `join(__dirname, '..', …, 'etc', 'alpine-release')` declines — verified. Resolving would
  also cost a syscall per candidate and break the zip-resident PnP case, whose paths are not
  files. Reasoning is now in the doc block.
- **One residual for a maintainer:** `createRequire(<anchor>)` also goes through
  `evalPathExpr`, so an anchor outside the reading module's package is no longer stubbed and
  the yargs-parser PnP crash class would return for it. Unreachable today — every
  `createRequire` in the measured graph is anchored at `import.meta.url` or `__filename` —
  but it is a behaviour the gate now governs and nothing states.
- **A case-insensitive filesystem** (macOS) will decline a case-mismatched own read, because
  `posix.relative` is case-sensitive. Safe direction, no test, worth knowing.
- **`gjsify install --global` still reports nothing** — left exactly as the PR says, for the
  reason it gives.

## 7. Every check I ran

- `gjsify workspace @gjsify/cli test` → **1720 tests passed, 4115 assertions, exit 0**
  (1715 + 5 new rows). The CLI has no GJS test target; Node is the whole suite for it.
- `gjsify workspace @gjsify/rolldown-plugin-gjsify check` and `… @gjsify/cli check`
  (`gjsify tsc --noEmit`) → exit 0
- `npx oxlint packages/infra/rolldown-plugin-gjsify/src packages/infra/cli/src` → exit 0,
  no findings in either touched file
- `npx oxfmt --check` on both touched files → "All matched files use the correct format"
- `node --test tests/e2e/inline-static-reads/run.mjs tests/e2e/published-bundle-reads/run.mjs`
  → **4/4 pass, exit 0**
- full `gjsify run build` → exit 0; `build:gjs-bundle` rebuilt three ways (fixed, gate-off,
  refactored) for the comparisons above
- `dist/affected.gjs.mjs` rebuild → byte-identical
- the on-device runs in §5

**Not run:** `tests/e2e/pnp-zip-static-reads` (needs Yarn 4 + registry). It is the suite most
exposed to the `resourceRootFor` change, and I reasoned it through — a zip-resident path
resolves to `…zip/node_modules/@scope/name` structurally, which *contains* the fixture's
`./data.json` and `./package.json` reads, where the merged rule fell back to the project root
— but reasoning is not a run. Worth watching on this PR's CI leg.

## 8. CI

`Node-pillar suites (darwin-x64 / gjs + node)` is red on the previous head, and it is not
this PR: the failing step is `Install the workspace`, at
`gjsify install: downloading 3/1658 (0%)` → `fetch failed`. A registry blip on the macOS
runner. Everything else was green or skipping. Not re-run by me.

I have not merged this and have not enabled auto-merge.

PR #1605, opened 2026-09-08
fix(ship-oracle): stop blaming the artifact for a fault in the reader

JumpLink:
## What happened

`E2E 1/4` on `main`, 2026-09-08. `tests/e2e/ship-msi`'s "refuses to report on an extraction
that produced nothing" failed because the oracle refused for the **wrong reason**: it said

```
::error title=Ship msi::there is no INSTALLDIR row, so `msiexec INSTALLDIR=…` has
nothing to override and the install location is not addressable
```

for an installer that **has** one. Four other cases in that file read that row out of the same
file in the same run — including the positive oracle 110 ms earlier, which prints
`INSTALLDIR = ProgramFiles64Folder\Ship Demo`, and one that reaches the byte round trip *beyond*
that check. So a `msiinfo export Directory` returned exit 0 without the table it had just
listed, and the accusation landed on the artifact.

That is the direction this file's own header calls the worst one for a diagnostic to point, and
it argues the case at length for `msiextract` — "a reader whose EXIT CODE is not an answer" —
while `idt()` right beside it was ungated:

```bash
idt() {
    msiinfo export "$MSI" "$1" | tr -d '\r'
}
```

The run went green on re-run and the scheduled run of the identical commit was green too, so the
artifact was never at fault. **The mechanism behind the empty export is still unexplained** —
`set -euo pipefail` rules out a crash, and there is no msitools on the machine I could hammer it
from. What is settled is the misattribution, and that is what this fixes.

## The change

**`idt` holds the reader to the same standard as `msiextract`.** An IDT export carries three
header lines — columns, types, table+keys — which is the assumption every `NR > 3` in this file
already makes. Fewer than three means the reader produced nothing usable, and that is a
different fact from a table without the row being looked for. Both outcomes now say so, and say
that nothing below is a claim about the installer.

**`fail` writes to stderr, and this half is load-bearing rather than tidy.** Every reader here
is called as `VAR=$(idt Directory)`, and `fail` inside a command substitution wrote its message
*into VAR*: captured instead of shown, with `exit 1` leaving only the subshell, so `set -e`
aborted the script having printed nothing at all. Without this, the guard above would have made
the failure **less** legible, not more. On stderr the message survives the substitution and
cannot contaminate the value a reader returns on stdout.

GitHub reads `::error` workflow commands from stderr the same as from stdout, and
`oracleExpectingFailure` concatenates both streams, so nothing downstream can tell the
difference except by working.

## Verified

No msitools on this machine, so the four reader outcomes were exercised against a stubbed
`msiinfo` with the real `fail` and `idt` bodies:

| stub | result |
|---|---|
| normal table (3 headers + 1 row) | passed through, 4 lines, `INSTALLDIR` row visible to the caller |
| exit 0, no output | `msiinfo export Directory returned 0 usable line(s)` |
| exit 0, 1 line | `msiinfo export Directory returned 1 usable line(s)` |
| exit 3 | `msiinfo export Directory exited non-zero` |

All three failures are visible on stderr from inside `$(...)`, which is the case that produced
nothing before. `bash -n` clean.

CI's `ship-msi` e2e is the real exercise of the happy path — it is what runs the oracle against
a genuine wixl-built installer.

## Two things review added

**Every deliberate refusal was annotating the job it passed.** `execFileSync` given no
`stdio` of its own re-emits a captured stderr on the parent's, and the runner reads `::error`
workflow commands off stderr as off stdout. So moving `fail` to stderr — necessary for the
guard above to be visible at all — handed all five `oracleExpectingFailure` refusals to the
runner as red annotations inside a *passing* shard. The harness now names its streams.

That fix then missed one call, excluded by reasoning rather than measurement: `the oracle
refuses a file it cannot read at all` goes through plain `oracle()`, and `msiinfo suminfo` on
a non-compound file prints a warning, **exits 0** and names no creating application — so the
script's own `fail` fires there too. Caught by reading the run's annotations. Comparing the
shard across the two heads:

| head | `E2E 1/4` | annotations |
|---|---|---|
| `457a10c0ad` | fail | `exit code 1` + `the summary information names no creating application…` |
| `e97ed3f9fe` | **pass** | **none** |

**The new guard now has the discriminator this file demands of every refusal in it.** A `PATH`
wrapper forwards every `msiinfo` call to the real binary except `export Directory`, which
exits 0 with no output — the smallest reproduction of what was observed. Its second assertion
is the discriminator: `assert.doesNotMatch(/there is no INSTALLDIR row/)` fails if the header
guard is removed. It ran at 80 ms against a real wixl-built installer, having previously died
at 0.7 ms on a missing `mkdirSync` import that neither `node --check` nor `oxlint` can see
(`no-undef` is off for `tests/`, since `process` is not declared a global there).

The four green cases that read `Directory` out of the real installer all passed *through* the
new guard without tripping it, which is the empty-table worry answered by measurement.

### DIFF
--- a/.github/ship-oracle/verify-msi.sh
+++ b/.github/ship-oracle/verify-msi.sh
@@ -96,11 +96,13 @@ require msiinfo msiextract find sort cmp awk tr
 # print `error: libmsi_database_export / msiinfo: internal error (function failed)`
 # on EPIPE, which reads as a corrupt database and is nothing but the pipe closing —
 # the same run that measured the CRLF chased that for ten minutes first.
-# An IDT export, with the reader held to the same standard as `msiextract`
-# below: three header lines (columns, types, table+keys) precede the rows, which
-# is the assumption every `NR > 3` in this file already makes. Fewer than three
-# means the READER produced nothing usable, and that is not the same fact as a
-# table without the row being looked for.
+#
+# THE READER IS HELD TO THE SAME STANDARD AS `msiextract` BELOW. An IDT export
+# carries three header lines — columns, types, table+keys — before any row, which
+# is the assumption every `NR > 3` in this file already makes, and an
+# empty-but-present table still has all three. Fewer than three means the READER
+# produced nothing usable, and that is not the same fact as a table without the
+# row being looked for.
 #
 # Measured 2026-09-08, `E2E 1/4` on `main`: this script reported "there is no
 # INSTALLDIR row, so `msiexec INSTALLDIR=…` has nothing to override" for an
@@ -117,7 +119,7 @@ idt() {
         fail "msiinfo export $1 exited non-zero on $MSI. That is the reader failing, not a finding about the installer."
     fi
     local lines
-    lines=$(printf '%s\n' "$out" | grep -c . || true)
+    lines=$(grep -c . <<<"$out" || true)
     if [ "$lines" -lt 3 ]; then
         fail "msiinfo export $1 returned $lines usable line(s) for $MSI, and an IDT export carries three header lines before any row. The reader produced nothing, so nothing below is a claim about the installer — re-read the file before believing any row is missing."
     fi
--- a/.github/ship-oracle/verify-msi.sh
+++ b/.github/ship-oracle/verify-msi.sh
@@ -58,8 +58,18 @@ MSI=${1:?usage: verify-msi.sh <artifact.msi> <program directory> <producer|!prod
 DIR=${2:?usage: verify-msi.sh <artifact.msi> <program directory> <producer|!producer>}
 PRODUCER=${3:?usage: verify-msi.sh <artifact.msi> <program directory> <producer|!producer>}
 
+# STDERR, not stdout, and it is load-bearing rather than tidy. Several readers
+# here are called as `VAR=$(idt Directory)`, and a `fail` inside a command
+# substitution writes into VAR: the message is captured instead of shown, and
+# `exit 1` leaves only the subshell, so `set -e` aborts the script with nothing
+# printed at all. On stderr the message survives the substitution, and it cannot
+# contaminate the value a reader returns on stdout either.
+#
+# GitHub reads `::error` workflow commands from stderr the same as from stdout,
+# and `tests/e2e/ship-msi`'s `oracleExpectingFailure` concatenates both streams,
+# so nothing downstream can tell the difference except by working.
 fail() {
-    echo "::error title=Ship msi::$*"
+    echo "::error title=Ship msi::$*" >&2
     exit 1
 }
 
@@ -86,8 +96,32 @@ require msiinfo msiextract find sort cmp awk tr
 # print `error: libmsi_database_export / msiinfo: internal error (function failed)`
 # on EPIPE, which reads as a corrupt database and is nothing but the pipe closing —
 # the same run that measured the CRLF chased that for ten minutes first.
+# An IDT export, with the reader held to the same standard as `msiextract`
+# below: three header lines (columns, types, table+keys) precede the rows, which
+# is the assumption every `NR > 3` in this file already makes. Fewer than three
+# means the READER produced nothing usable, and that is not the same fact as a
+# table without the row being looked for.
+#
+# Measured 2026-09-08, `E2E 1/4` on `main`: this script reported "there is no
+# INSTALLDIR row, so `msiexec INSTALLDIR=…` has nothing to override" for an
+# installer that HAS one — four other cases in `tests/e2e/ship-msi` read that row
+# out of the same file in the same run, one of them reaching the byte round trip
+# beyond it. So a `msiinfo export` answered 0 without the table it had just
+# listed, and the accusation landed on the artifact. The mechanism is still
+# unexplained; what is settled is that the artifact was not at fault, and a
+# diagnostic pointing at the file for a fault in the reader is the direction the
+# header of this file calls the worst one.
 idt() {
-    msiinfo export "$MSI" "$1" | tr -d '\r'
+    local out
+    if ! out=$(msiinfo export "$MSI" "$1" | tr -d '\r'); then
+        fail "msiinfo export $1 exited non-zero on $MSI. That is the reader failing, not a finding about the installer."
+    fi
+    local lines
+    lines=$(printf '%s\n' "$out" | grep -c . || true)
+    if [ "$lines" -lt 3 ]; then
+        fail "msiinfo export $1 returned $lines usable line(s) for $MSI, and an IDT export carries three header lines before any row. The reader produced nothing, so nothing below is a claim about the installer — re-read the file before believing any row is missing."
+    fi
+    printf '%s\n' "$out"
 }
 
 [ -f "$MSI" ] || fail "$MSI does not exist"
--- a/tests/e2e/ship-msi/run.mjs
+++ b/tests/e2e/ship-msi/run.mjs
@@ -38,7 +38,7 @@
 import { describe, it, before, after } from 'node:test';
 import assert from 'node:assert/strict';
 import { execFileSync } from 'node:child_process';
-import { cpSync, existsSync, mkdtempSync, readFileSync, rmSync, statSync, writeFileSync } from 'node:fs';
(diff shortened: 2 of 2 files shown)


JumpLink:
## Independent review pass — what I changed, and what I left

I re-derived the claims rather than reading them. The misattribution is real, the
diagnosis is right, and the stderr half genuinely is load-bearing. Two things needed
changing.

### 1. The stderr move handed every DELIBERATE refusal to the runner

This is the one that mattered. `execFileSync` given no `stdio` of its own **re-emits a
captured stderr on the parent's** — measured:

```
$ node -e 'try{execFileSync("bash",["./err.sh"],{encoding:"utf-8"})}catch(e){…}'
::error title=X::message on stderr        ← printed by the PARENT
status= 1
stderr= "::error title=X::message on stderr\n"
```

and with `stdio: ['ignore','pipe','pipe']` the capture is identical while the parent's
stderr stays empty. So while `fail` wrote to stdout, `execFileSync` swallowed it and the
five `oracleExpectingFailure` calls in `tests/e2e/ship-msi` were invisible. On stderr each
of them prints an `::error title=Ship msi::…` line into the job log, and the runner reads
workflow commands off stderr exactly as off stdout — a passing `E2E` shard would carry a
red annotation per refusal the suite asked for. `oracleExpectingFailure` now names its
streams, which is also where its own assertions read the message from.

The three sibling oracles keep `fail` on stdout, and I checked whether they have the same
latent bug: they do not. No `fail` in `verify-deb.sh`, `verify-rpm.sh` or
`verify-app-zip.sh` is reached from inside a command substitution — every one of their
`VAR=$(…)` forms invokes a tool directly, and `verify-rpm.sh:133` already uses this PR's
`if ! VOUT=$(…)` shape. So the divergence is confined to the one file that needs it.

### 2. The new refusal had no discriminator

This file's own header says every refusal in it is driven against a mutated copy, "run it,
do not trust it". Added: a wrapper on `PATH` that forwards every `msiinfo` call to the real
one **except** `export Directory`, which is the smallest reproduction of what `E2E 1/4`
observed. Its second assertion — `assert.doesNotMatch(failure, /there is no INSTALLDIR
row/)` — is what fails if the guard is removed, because the script still exits 1 either
way, just on the sentence about the artifact.

### The empty-table worry, answered by measurement

I exercised the real `fail`/`idt` bodies against a stubbed `msiinfo`:

| stub | rc | outcome |
|---|---|---|
| 3 headers + 1 row | 0 | caller sees 4 lines, INSTALLDIR row present |
| 3 headers, **no rows** | 0 | caller sees 3 lines, no INSTALLDIR row — guard does **not** fire |
| exit 0, no output | 1 | `returned 0 usable line(s)` |
| exit 0, one line | 1 | `returned 1 usable line(s)` |
| exit 3 | 1 | `exited non-zero` |

And the pipeline shape section 5 uses, `SHORTCUTS=$(idt Shortcut \| awk …)`: an
empty-but-present `Shortcut` table passes the guard and reds on `SHORTCUT_COUNT=0` as
before, so the "shortcut row emptied" mutation still fails for its own reason. A
legitimately empty table is not turned into a false red. The comment now says so, where
before it only said "three header lines".

Also `grep -c . <<<"$out"` for `printf | grep`, the here-string idiom the rest of the file
uses.

### What I left, and why

- **The guard for an unexplained mechanism.** I think it earns its place: this file already
  carries two guards of exactly this species (`msiextract` wrote no directory; wrote no
  files), both for the same stated reason, and the *misattribution* is measured even though
  the emptiness is not. What the guard claims is only that the reader failed, which is true
  by construction.
- **Retry instead of refuse.** Worth a maintainer's decision rather than mine. The run was
  green on re-run and the scheduled run of the identical commit was green, so a bounded
  retry with a `::warning` would turn a transient reader fault into a green run *and* leave
  a record — the shape `gjsify publish`'s read-back and #1603 both use. Against it: if the
  empty export ever turns out to be a symptom rather than noise, a retry hides it. I did
  not change the behaviour.
- **`tests/e2e/ship-msi` keeps its own copy of `oracle`/`oracleExpectingFailure`** rather
  than using `tests/e2e/ship/fixture.mjs`'s, which is the drift that file's own comment
  argues against. Pre-existing, and the signatures differ; not this PR's scope.

### Checks run

- `bash -n .github/ship-oracle/verify-msi.sh` — clean
- `node --check tests/e2e/ship-msi/run.mjs` — clean
- `npx oxlint tests/e2e/ship-msi/run.mjs` — exit 0, no findings
- `npx oxfmt --check tests/e2e/ship-msi/run.mjs` — "All matched files use the correct format"
- the five stub cases above, plus the section-5 pipeline shape
- the `execFileSync` stderr behaviour, both with and without an explicit `stdio`

**Not run:** `tests/e2e/ship-msi` itself. No msitools on this machine, so the happy path and
the new discriminator are exercised only by CI's `ship-msi` leg. Do not merge on my word
that the new row passes — it has never run against a real installer.

JumpLink:
## `E2E 1/4` red was mine, and the run settled two things

Run 34227868797 on `457a10c0ad`: **every check green except `E2E 1/4`**, and that
one failed on my own discriminator for a trivial reason —
`ReferenceError: mkdirSync is not defined`, 0.7 ms in, before it built the stub or
reached the oracle. `e57cd34c9b` (yours) fixes the import; I rebased onto it rather
than pushing a duplicate.

Two things worth recording out of that run, because both are the same lesson.

**1. Nothing local could see it, and I said so before it happened but did not act on
it.** `node --check` parses the file — an unresolved reference resolves at run time or
not at all — and there is no msitools here, so the row could not be run. There *is* a
rule that catches the class: `oxlint --deny no-undef` names line 385 before the fix.
It is unusable as a repo default for this tree, because `process` is not declared as a
global here and it reports four false positives on that alone. But it is usable as a
targeted instrument, and I have since verified with it (only the pre-existing
`process` findings remain) plus a standalone replay of the row's body: every
identifier resolves, the stub is written 755, forwards a non-export call and swallows
`export Directory`.

**2. `e97ed3f9fe` — one `::error` was still reaching the runner, and I had reasoned it
away instead of measuring it.** The run's annotations carry, besides the real failure:

```
Ship msi :: the summary information names no creating application, so nothing here
can say which implementation wrote this file
```

That is `verify-msi.sh` line 150, from `the oracle refuses a file it cannot read at
all` — which goes through plain `oracle`, not `oracleExpectingFailure`. I had judged
it could not reach `fail` because `msiinfo` would abort the script first. It does
reach it: `msiinfo suminfo` on a file that is not a compound file prints a WARNING,
**exits 0**, and names no creating application. So a refusal this suite asked for
annotated a job red, exactly the defect the first commit fixed for the other five
call sites, in the one place I had excluded by argument rather than by measurement.
It names its streams now. That is every call in the file that asks for a refusal.

**What the run does confirm:** every other case passed, the new `stdio` included — so
the `idt` guard and the stderr move in `verify-msi.sh` are sound, and the four green
cases that read `Directory` out of the same file went through the guard without
tripping it. The one thing still unproven is the new discriminator itself, which has
never completed. It needs this next run.

JumpLink:
## Both halves now measured green

`E2E 1/4` on `e97ed3f9fe`: **pass**, all 22 `ship-msi` cases, the new row included.

```
✔ refuses to report on an extraction that produced nothing (83.32522ms)
✔ blames the READER, not the installer, when an export comes back empty (80.409492ms)
✔ the oracle refuses a producer the file does not claim (13.562527ms)
...
✔ CLI ship Windows installer E2E (4243.038047ms)
```

80 ms, against 0.7 ms when it died on the import — so it built the stub, drove the
oracle through it and made both assertions against a real wixl-built installer. The
guard's discriminator has now actually run, which it never had before.

**And the annotation half, measured rather than argued.** Annotations on the two runs
of this shard:

| run | shard | annotations |
|---|---|---|
| `457a10c0ad` | fail | `Process completed with exit code 1` + `Ship msi :: the summary information names no creating application…` |
| `e97ed3f9fe` | **pass** | **none** |

The three `::error` strings still in the log are the literal `echo "::error::…"` lines
of an unrelated `run:` block being printed as script text, not emissions. So no
refusal this suite asks for reaches the runner any more, which is what the first
commit claimed and could not yet show.

PR #1607, opened 2026-09-08
feat(prebuilds): declare and ship the first musl targets

JumpLink:
Follow-up to #1602, which made the musl fallback legible. This ends it.

## Why nothing musl was ever published

`build-prebuilds-musl` has been building `@gjsify/lightningcss-native` and
`@gjsify/sab-native` for `linux-{x64,arm64}-musl` and uploading them for a while.
`commit-prebuilds` ignored the artifacts on purpose — its own comment says why: they
are *"`-musl` artifacts no package declares, and committing an undeclared target is
exactly what the `prebuild-artifacts` rule fails on"*.

So the leg proved the build compiles and nothing ever shipped. A musl host silently
resolved the glibc sibling instead.

## What that costs, measured

OnePlus 6, postmarketOS v26.06, musl 1.2.6, aarch64, GJS 1.88.1, against 0.48.0.

Nine of ten prebuilds load anyway — musl's loader aliases `libc.so.6` to itself.
`libgjsify_lightningcss.so` does not:

```
Error relocating …/libgjsify_lightningcss.so: gnu_get_libc_version: symbol not found
```

The CSS bridge is then absent from the next build, which rolldown reports as
`Could not load src/application.css` — nothing in that chain names libc.

## The change, in the prescribed order

Two notes in the tree already spelled this out, and I followed both.
`prebuild-libc.mjs`:

> FOLLOW-UP, and it is a real one: the moment a `-musl` target IS declared,
> `PLATFORM_RE` rejects it and `canonicalPlatform` collapses it. **Both must gain the
> optional suffix in the SAME change that declares the first musl target.**

and `commit-prebuilds`, pointing at the win32 promotion as the shape.

- **`PLATFORM_RE`** gains `-musl`, linux only — npm's `libc` is documented Linux-only
  and musl targets no other kernel, so `darwin-arm64-musl` stays malformed rather than
  becoming a musl build of a mac binary.
- **`canonicalPlatform`** stops folding the suffix away. It kept the first two
  dash-parts, which made a musl target compare equal to the glibc one in every set
  operation the prebuild rules perform.
- **`osCpuForTarget`** splits libc off `cpu`, giving `cpu: ["arm64"]` plus
  `libc: ["musl"]` rather than the `["arm64-musl"]` a last-dash split mints, which npm
  matches against nothing.
- **The generator** lets the token's libc beat the measured one — a `-musl` package is
  musl by construction, and without the field npm installs it on glibc hosts.
- **Both bridges** declare the two targets and their `optionalDependencies`; four
  per-target packages are generated, carrying `platformsUncommitted` until the
  directories land.
- **`commit-prebuilds`** gains the leg and its four downloads.

### One trade worth calling out

Adding the leg to `needs` means a musl failure now takes the commit path down with it,
folding back the 41-hour incident's lesson deliberately. That is the cost of `needs`,
and it is right only because the leg can be green: native on both arches, gated step by
step, no `continue-on-error`. Committing a stale musl directory beside fresh glibc ones
is worse than a red run.

## Still a manual step

The four npm names are queued in `status/pending-npm-bootstrap.json`. OIDC cannot create
a name, so the first publish needs a credential and an OTP. Until that is done a musl
host still resolves the glibc sibling — the gap is recorded rather than silently
shipped, and `verify-published-closure --phase pre-release` reports it as the policy's
queued-action branch.

## Verified, on the musl host itself

- `generate-platform-packages --check` — 70 per-target packages across 13 bridges
  (was 66), 6 exempt
- `verify-published-closure --phase pre-release` — exit 0; 209 publishable names,
  381 pinned edges resolve, 5 declared pending
- `generate-status --check`, `audit-runtimes --check` — exit 0
- `oxfmt --check` and `oxlint` — clean on the changed files
- the workflow re-parsed as YAML: the leg is in `needs` and all four downloads are
  present
- grammar spot-checked: `linux-arm64-musl` valid and canonical-stable,
  `darwin-arm64-musl` rejected, `linux-x86_64-musl` still folds to `linux-x64-musl`

**Not verified, and not verifiable from here:** that the committed musl artifacts load.
Nothing outside `build-prebuilds-musl` has ever produced one, and this host cannot run
that leg. That is what the leg's own `Check the staged prebuilds under musl, both
arches` step is for.

One note on the review of #1602, which found the real asymmetry: `linux-x64` records a
glibc loader and declares `libc: ["glibc"]`, `linux-arm64` records none and declares
nothing — which is why the arm64 companion installed on musl and broke at all. This PR
does not touch that predicate. Widening it needs a musl symbol set to be sound; it stays
a maintainer's call.


JumpLink:
## CI found two real gaps; both were one cause

Pushed `81ce311b3`.

`parseCiPlatforms` dropped every libc-carrying target from the parsed CI coverage, on the premise that *"a libc-carrying target is NOT a `gjsify.platforms` promise"*. True for exactly as long as nothing could declare one. Keeping them makes both failures go away at once:

- **`auditPlatforms`** reported all four musl targets as *"declares … but no CI job produces that target"* — while `build-prebuilds-musl` sits in the same file and builds them.
- **`auditReleaseCoverage`** demanded a release leg. It skips targets `prebuilds.yml` is committing (`// on its way to being committed`, line 647) and reads the *same* map, so the subtraction could not fire. **No release leg is needed** — my first reading of that message was wrong, and the exemption contract covers this case explicitly: *"between 'the leg is green' and 'the artifact is committed' the package legitimately declares a target with no artifact and no release leg."*

### My process error

The failure was reproducible and I did not reproduce it. CI runs `audit-runtimes --check --strict`; I ran `--check`. `--strict` *is* the drift check. The flags were in the workflow — I used the command from memory instead of reading them.

## What the strict audit found on the musl host, which is worth its own look

With the fix in place the only remaining drift is **not from this branch**:

```
PREBUILD-ARTIFACT FAILURES (does the declared platform have a loadable body?) on 1 target(s):
  - @gjsify/lightningcss-native-linux-arm64: `prebuilds/linux-arm64/libgjsifylightningcss.so`
    fails to load on this host with no library-path variable set — OSError: Error relocating
    …/libgjsify_lightningcss.so: gnu_get_libc_version: symbol not found
```

That package is untouched here (unchanged since `68461e4ae chore: release v0.48.0`). **The repo's own `prebuild-artifacts` audit detects the bug this whole line of work started from** — it does a real env-free `dlopen` of host-target prebuilds, and on a musl host it lands on exactly the right file.

It has never fired because it can only fire on a musl host, and there is no musl runner in CI — the same blind spot `detect-native-packages.spec.ts`'s header already notes (*"There is no musl runner in this repo's CI at all"*).

That is the empirical form of the asymmetry the #1602 review identified: `linux-x64` records a glibc loader and declares `libc: ["glibc"]`, `linux-arm64` records none and declares nothing, so the arm64 companion installs on musl and breaks. Still not touched here — widening that predicate needs a musl symbol set to be sound.

### Verified locally after the fix

`generate-platform-packages --check` (70 packages, 6 exempt) · `verify-published-closure --phase pre-release` · `generate-status --check` · `audit-runtimes --check --strict` (only the pre-existing failure above) · oxfmt + oxlint clean · all four exemptions recognised by name.

JumpLink:
## Independent review — three commits pushed, and the one gap this PR admitted is now closed

I reviewed this sceptically rather than signing it off, reproduced every failure locally
before touching it, and took the musl artifacts to a real musl device. Verdict: **the
change is right and should land, but not before the four npm names exist.** Details below,
including one thing I tried, measured and backed out.

---

### The admitted gap is closed: the committed musl artifact loads, the glibc one does not

> *"Not verified, and not verifiable from here: that the committed musl artifacts load."*

It is verifiable, and it now is. `gh run download` from this PR's own `build-prebuilds-musl`
leg (run `34258797167`), copied to a **OnePlus 6T — postmarketOS v26.06, musl 1.2.6,
aarch64, gjs 1.88.1, no Node** — and loaded through GI, glibc build beside musl build:

```
########## linux-arm64 (glibc) ##########
namespace bound
GLib-GIRepository-WARNING: Failed to load shared library 'libgjsifylightningcss.so'
  referenced by the typelib: Error relocating /tmp/.../libgjsify_lightningcss.so:
  gnu_get_libc_version: symbol not found
CALL FAILED: Unsupported type void, deriving from fundamental void      exit=3

########## linux-arm64-musl ##########
namespace bound
CALL OK: a{color:#fff}                                                  exit=0
```

That second line is a real `GjsifyLightningcss.Engine.transform()` over
`a { color: #ffffff; }`. `readelf -d` on the phone confirms the split at the source:
glibc build needs `libc.so.6` + `libm.so.6`, musl build needs `libc.musl-aarch64.so.1`.
`@gjsify/sab-native` loads on **both**, as its ledger entry in `musl-committed-check.sh`
already says it should since its two symbols were removed.

So the chain from #1602 through #1604 to here now has its end-to-end proof. Files removed
from the device afterwards.

---

### The three failures

**`E2E 2/4` — `ci-runner-arch`.** Held up, and the code was right rather than the test.
`parseCiPlatforms` answers "which targets does CI build", read from the workflow; the
declaration comparison is `auditPlatforms`' job and neither answers the other's question.
That the fixture's `@gjsify/tls-native` declares no musl target is not over-crediting — it
is exactly the signal `auditPlatforms` should raise. You had already pushed the same
conclusion in `0d9320b` while I was working; I rebased onto it and kept your wording.

**`E2E 4/4` — Yarn PnP.** Held up, and it is worse than the PR says. See below.

**`Build prebuilds (riscv64) [QEMU]`** was not pending — it was **cancelled** at 58 min,
by `Cancel superseded runs for this PR` firing on the `pull_request: edited` event when the
body was reflowed. Nothing to do with this change; it is running again now.

---

### The Yarn problem, measured outside the harness

Reproduced on yarn 4.14.1 with nothing but a local package declaring one unpublished
`optionalDependencies` entry — no test harness, no `resolutions`:

```
YN0035: @gjsify/sab-native-linux-arm64-musl@npm:0.48.0: Package not found
```

Yarn resolves the whole graph before it links any of it, so a **transitive** optional that
404s fails the install **outright, for every consumer on every platform** — not only musl
hosts. npm and pnpm drop it in silence, which is why no npm-linker suite noticed.

That makes the four ledger entries understatements: they said the cost of the queued publish
is that a musl host keeps resolving the glibc sibling. The real cost, once a release ships
the bridge, is that `@gjsify/sab-native` and `@gjsify/lightningcss-native` do not install
under Yarn at all. Both entries and both `verify-published-closure` warnings now say so, and
`published-closure`'s assertion pins both halves instead of npm's alone.

**What I tried and backed out.** The obvious fix is to pack the four per-target packages so
`resolutions` covers them. It does not work, and the measurement is worth keeping: a `file:`
tarball for a package the current architecture does not support (foreign `cpu`, or a `libc`
the host lacks) resolves and then dies in the **fetch** step with
`ENOENT … .yarn/cache/<name>-file-<hash>.zip` — Yarn skips fetching what it will not link
and stats the cache entry anyway. Isolated against a control: the same tarball for a
*supported* package installs fine, and the same *unsupported* package with its `libc` field
deleted installs fine. So `isForeignPlatformPackage` is load-bearing for PnP and not only a
size optimisation; that is now written where the next person will look for it.

**What I did instead.** A second oracle beside the release-window one, because the two
absences are different facts. `registryVersionState` answers `unverified` on a packument
404 *deliberately* — a rename, a proxy and a typo all 404 identically, and reading any of
them as "not published yet" is how #1533 turned both PnP suites permanently green. A name in
`status/pending-npm-bootstrap.json` is the evidence that probe lacks: a human wrote it down
with the command and the owner, a reviewer saw it in the diff, and the pre-release closure
FAILS on an entry whose name IS published, so the list empties itself at bootstrap. It is
fail-closed twice: the ledger alone disarms nothing, the registry must confirm the name is
absent, and it must be demonstrably answering. Four cases in `release-window-skip` hold it,
and deleting the answering-registry guard reds the fourth (checked by mutation).

**The skip is not the fix, and I want to be blunt about that.** It keeps the tree from being
permanently red on a step only you can take. The fix is publishing the four names — after
which the ledger empties, the packages resolve from npm, and both PnP suites run again with
nothing here to undo. `verify-published-closure`'s own text already puts it correctly:
*"Bootstrap the target BEFORE the bridge — that order is a correctness property, not a style
preference."* **This is the maintainer decision on this PR:** merging before the bootstrap
means a release cut in that window ships two bridges that are uninstallable under Yarn.

---

### The other review findings

**Three claims in the tree had gone false**, each where someone would act on it:
`platforms-ci.mjs`'s `canonicalPrebuildTarget, NOT canonicalPlatform` comment,
`stage-prebuild.mjs`'s parenthetical, and `index.d.ts`'s *"libc-AWARE, unlike
canonicalPlatform"*. Plus two in `prebuilds.yml`: the matrix `libc` key still claims the
parser "drops libc-carrying tokens", and the job banner still claims `commit-prebuilds`
"is not gated on this job" — which this PR's own `needs` line contradicts 300 lines above
where it argues for it.

**`canonicalPrebuildTarget` is gone.** Its own doc says it exists only because
`canonicalPlatform` collapses `-musl`, and asks for both to gain the suffix in the change
that declares the first musl target. Both have, so the second copy goes and its five call
sites take the first. The package is `private: true`, so no published surface moves.

**`PLATFORM_RE` carried the arch list twice** and `KNOWN_ARCH_TOKENS` was a third copy; all
three now read one list, and the alternation still rejects `darwin-arm64-musl`.

**`canonicalPlatform` edge cases**, probed: `linux-x86_64-musl` → `linux-x64-musl`,
`darwin-arm64-musl` preserved *and* rejected by the grammar (so the audit names the token
the author wrote rather than a folded one), `linux-musl` and `musl` untouched, every case
idempotent. It also stopped composing `${os}-undefined` for a dash-less token.

**`osCpuForTarget`** checks out against npm: `cpu: ["arm64"]` + `libc: ["musl"]` is what npm,
yarn and pnpm match; the `arch-musl` a last-dash split would mint matches nothing. Its two
other callers were fine, but `platformReadme`'s JSDoc cast dropped the new third axis, and
the generated README for a musl package did not mention `libc` — the one field that decides
whether it installs. Both fixed; only the four musl READMEs change.

**"The token's libc beats the measured one" is sound**, and now says why: `prebuild-libc`'s
Check A fails a `-musl` directory whose libraries link glibc, so a mislabelled directory is
red at the audit rather than mis-declared by the generator. The comment asserted "by
construction of its own name", which is the weaker claim.

**`generate-platform-packages --check` does hold the `libc` field** — I tried deleting it by
hand and the byte-diff caught it (`line 12: want "libc": [`). The explicit `os`/`cpu` arm
does not compare `libc`, but adding a second check for what the first already catches is the
guard-watching-a-guard shape, so I left it.

**Adding the leg to `needs` is the right trade.** `download-artifact` fails a step when the
artifact is absent, so without `needs` a musl failure leaves the stale musl directories
committed beside fresh glibc ones — which is what the PR says and it is correct.
`build-prebuilds-musl` carries no job-level `if:`, so it can never *skip* and silently
suppress the commit path the way the banner above `needs` warns about. The blast radius does
grow, and that is the cost, stated.

**A dead guard went with the drop.** `if (targets.size === 0) continue;` existed only because
libc-carrying targets were removed mid-loop; every branch above it now adds at least one
target, so it was unreachable and its comment described a mechanism that had gone.

---

### One thing I found and did NOT fix

`HOST_TARGET` is still `${process.platform}-${process.arch}` and libc-blind, while everything
around it is now libc-aware. `prebuild-artifacts` compares a directory against it to choose
what to `dlopen`, so **on a musl host the audit loads the glibc directory (which cannot load
there) and skips the musl one (which can)** — both halves wrong at once. Unreachable until
the first `-musl` directory is committed, which is what this PR arranges. Measured on the
same phone.

Closing it needs a host-libc probe in a module that deliberately reaches for no runtime facts,
plus a CI leg running `audit-runtimes --check` on musl — and I could not validate either from
here, since the musl device has no Node. Recorded in `status/open-todos.md` with the
measurement rather than fixed silently.

Also noticed, pre-existing and not mine: `STATUS.md` is tracked on `main` while `.gitignore`
line 163 and the root `AGENTS.md` both say it is generated and never committed. `gitignore`
does not untrack an already-tracked file, so `generate-status --check` dirties the tree.

---

### What I ran

| | |
|---|---|
| `node --test tests/e2e/ci-runner-arch/run.mjs` | 37/37 (was 35/37) |
| `node --test tests/e2e/release-window-skip/run.mjs` | 26/26, four of them new |
| `node --test tests/e2e/published-closure/run.mjs` | 35/35 |
| `node --test` on `release-window-skip prebuild-declaration-invariant prebuild-change-gate stage-prebuild platform-exemption-clearing staged-prebuild-libc-gate` | 158/158 |
| `node --test` on `package-outputs ci-classifier-output-coverage pack-workspace-protocol install-platform-filter prebuild-loader-path` | 86/87, 1 skip, 0 fail |
| `node --test tests/e2e/{cli-only-pnp,pnp-zip-static-reads}/run.mjs` | both skip, with the reason printed |
| `node scripts/generate-platform-packages.mjs --check` | exit 0, 70 packages / 13 bridges / 6 exempt |
| `node scripts/audit-runtimes.mjs --check --strict` | exit 0 |
| `node scripts/verify-published-closure.mjs --phase pre-release` | exit 0, 209 names, 381 edges, 5 pending |
| `node scripts/generate-status.mjs --check` | exit 0 |
| `oxfmt --check` / `oxlint`, every changed file | clean; oxlint exit 0 |
| yarn 4.14.1 outside the harness | 404-optional fatal; unsupported-arch `file:` tarball ENOENTs at fetch |
| OnePlus 6T, musl 1.2.6 aarch64, gjs 1.88.1 | glibc build fails to relocate, musl build transforms CSS |

Not merged, auto-merge not enabled.

JumpLink:
## Follow-up: the four names went live mid-review, so the skip came back out

Between my first probe (`209/214`, five names 404) and the CI run an hour later
(`214/214`), the whole pending list was bootstrapped — the four musl per-target packages
**and** `@gjsify/effect-platform`. That changes the right answer, so I have force-pushed a
replacement for the third commit.

**What the publish fixes on its own, with no test change.** `yarn install` in
`cli-only-pnp` now completes against the published names — the `YN0035` is gone. The
remaining local failures in that suite are `@gjsify/fs`/`@gjsify/path` having no
`lib/esm` in my worktree (I ran `build:infra`, not the full `build`), which CI does not
have; the install itself, which is what this PR broke, succeeds.

**What came back out.** The ledger-backed skip oracle, its four `release-window-skip`
cases and the `pendingNpmBootstrapNames` reader — written, measured, and removed. The red
is what produced the publish within the hour, and `docs/publishing.md` calls bootstrap
-before-bridge a correctness property rather than a preference; a skip would have deleted
the only pressure toward it. `status/pending-npm-bootstrap.json` is now empty, which its
own `//shape` note calls the correct steady state, and the bidirectional arm that was
reding `Detect runtime-triplet drift` is satisfied: 214 of 214 publishable names exist,
381 pinned edges resolve, exit 0.

**What stays**, because it is measured and would otherwise be relearned:

- Both `verify-published-closure` warnings now name the split rather than npm's half
  alone — npm and pnpm install the bridge with nothing behind it, **Yarn refuses the
  install outright**. `published-closure` asserts both halves. Reproduced on yarn 4.14.1
  outside the harness with nothing but a local package declaring one unpublished
  `optionalDependencies` entry: a *transitive* optional that 404s fails the install for
  every consumer on every platform.
- `unpublishedRegistryDependencies` now says, next to the probe that reports such a name
  as `unverified`, what a `YN0035` on a `@gjsify/*` name means and that the fix is the
  publish. One lookup instead of the investigation it cost me.
- `isForeignPlatformPackage` records the thing I tried first and had to back out: a
  `file:` tarball for a package this architecture does not support (foreign `cpu`, or a
  `libc` the host lacks) resolves and then dies in the **fetch** step with
  `ENOENT … .yarn/cache/<name>-file-<hash>.zip`. Isolated against two controls — the same
  tarball for a *supported* package installs, and the same *unsupported* package with its
  `libc` field deleted installs. So that filter is load-bearing for PnP, not a size
  optimisation, and packing a foreign platform package is never the fix.

Everything in my first comment about the grammar, the `canonicalPrebuildTarget` deletion,
the false comments, the `needs` trade, `HOST_TARGET` and the musl-device proof is
unchanged. Branch is now:

```
7e9b7c6  fix(publishing): empty the ledger, and say what Yarn does
3eac8e5  docs(prebuilds): correct two comments this PR made false
676c10b  refactor: one arch vocabulary, one canonicaliser
0d9320b  test(ci-runner-arch): a musl leg now credits its musl targets   (yours)
```

Re-verified after the rewrite: `verify-published-closure --phase pre-release` exit 0
(214/214, 0 pending), `release-window-skip` + `published-closure` 57/57,
`generate-platform-packages --check` exit 0, `oxfmt --check` and `oxlint` clean.
Still not merged, auto-merge still off.

PR #1612, opened 2026-09-09
feat(gtk-host): list and adjustment at the seam

JumpLink:
Closes the open-todo *"A portable list model reaches every renderer except
GTK"* by building the seam ADR 0042 § 8 built for the menu, for the two
values that were waiting on it.

## What

- `coerce` in `packages/framework/gtk-host/src/props.ts` gains two branches
  beside the `GMenuModel` one, keyed on the ParamSpec and never on a
  property name: a `Gio.ListModel`-typed property accepts `AdwListModelInput`
  and gets a real `Gtk.StringList`; a `Gtk.Adjustment`-typed one accepts
  `AdwAdjustmentInput` and gets a real `Gtk.Adjustment`. The constructions
  live in `list-model.ts` and `adjustment.ts`, each with its inverse.
- The list branch asks TWO facts: `Gtk.ListView:model` is a
  `Gtk.SelectionModel`, which IS a `Gio.ListModel`, so a branch keyed on
  list-ness alone would build a string list GTK refuses with a CRITICAL at
  exit 0. That write is refused by name (`list-model-mismatch`); so is a
  string or an object where a list belongs (`bad-list-model`), and a bare
  number where an adjustment belongs (`bad-adjustment`, ADR 0047 § 1).
- `WithPortableValues<T>` widens the same properties by the same rule on
  all three dialect surfaces; `type-tests/` holds the negatives.
- Gallery: `Adw.ComboRow`, `Gtk.DropDown` and `Adw.SpinRow` get their
  Solid/Vue/React snippets — trees in `adwaita-gallery-trees.mjs`, output
  regenerated, asserted by the three probe showcases (28 widgets, zero
  diagnostics each). Arm 5c of `check-generated-website-data.mjs` holds
  every refusal naming a GType against the types `coerce` converts; it
  printed the three stale refusals before the trees existed.
- The React Native GTK arm hands both values through the seam and imports
  no `gi://` any more (the "one package over" copies the todo named).

## Vectors

`list-model.spec.ts` and `adjustment.spec.ts` drive the shared conformance
tables through the builders, back through the inverses, and through the
real widgets with the diagnostics gate on; the ParamSpec premises sit in
`props.spec.ts` beside the enum block. Measured and stated rather than
implied: an item's `value` does not cross (a `Gtk.StringObject` holds one
string; GTK spells a selection as a position), and a JSON string on this
seam is refused, not parsed (the markup doors are the browser's).

## Decisions recorded in the ADRs

- ADR 0046 § Amendment: `ListController.setRows` is NOT widened — a keyed,
  factory-driven row is a different question from a string list, and an
  `AdwListModel` item has neither a key nor a subtree.
- ADR 0047 § Amendment: at this seam the object is the WHOLE adjustment
  (declarative, like the RN arm; not `SpinState.configure`'s keep-unwritten).
- The METHOD-built widgets stay open in `status/open-todos.md`, on purpose.

## Checks run locally

gtk-host `test:gjs` (588 tests), `check-type-surfaces` (15 measurements),
adwaita-react-native both legs, the three gallery probes plus
`probe:refusals`, `check-generated-website-data`, `audit-runtimes --strict`,
`check-adwaita-rn-platform-split`, `check-adapter-import-direction`,
`check-adwaita-conformance-drivers`, `check-vocabulary-alignment`,
`check-adwaita-collection-reactivity`, lint, format.


JumpLink:
The rebuild-on-write finding is measured and fixed.

`AdwComboRow:selected` is a position into the model, so handing the property a
freshly built `Gtk.StringList` takes the selection away. Measured on GTK 4.22.4
through a real `Adw.ComboRow` and a real `Gtk.DropDown`, selected 2 of
`['a','b','c']`: assigning a new list lands on 0 — with one label changed and
with the SAME three strings alike — while splicing the list the widget already
holds keeps it. A whole-list splice keeps the position; a minimal one keeps the
item under it.

`setProp` now hands the built list to `reconcileStringList`, which splices the
trimmed difference into the held list and emits nothing when the strings are
equal. That silence also removes the need for a consumer memo. Only a list the
seam built is spliced, tracked in a `WeakSet` — a list an application handed in
imperatively is its own object and is replaced, as the imperative line would be.

`list/controller.ts` chose a whole-model splice for a different reason (a
factory re-binds a row when its model object changes); both reasons are now
written where they apply.

Beside it the negatives got an oracle: `toThrow()` alone passes for the wrong
throw, so `testing/refusal.mts` answers the refusal CODE, or `no throw`, or the
wrong error type by name.

PR #1613, opened 2026-09-09
fix: measure musl and glibc as different libcs

JumpLink:
`main`'s `Build Native Prebuilds` has been red since the #1607 merge. All thirteen build
legs are green, including both Alpine ones; the job that fails is `Commit prebuilds to
repo`, at `Sync to origin/main and stage the artifacts` (run 34311463250):

    Error: clear-committed-platform-exemptions: refusing to write
    @gjsify/lightningcss-native-linux-arm64-musl.
    The generator's manifest differs from clearing the exemption, which means it would
    also change a MEASURED declaration (libc / gjsify.glibcRequires).
        line 19: have "        \"platforms\": ["
        line 19: want "        \"glibcRequires\": {"

The guard is right. What it caught is a measurement that is wrong, and there are two of
them. Both read the same DT_NEEDED bytes; both were written when every committed artifact
was glibc-built; each one calls a musl build glibc in its own vocabulary.

## 1. A version need belongs to a FILE, and `readElfGlibcRequires` ignored `vn_file`

`.gnu.version_r` is grouped by the soname that has to SUPPLY each version. The reader
walked every group's Vernaux entries and attributed every `GLIBC_*` name it found to
glibc, no matter who provides it.

The one non-glibc provider of a `GLIBC_*` label is `libgcc_s.so.1`. GCC's unwinder exports
`__register_frame_info` and `__deregister_frame_info` under the version `GLIBC_2.0`,
because those symbols were glibc's before they moved to libgcc; Alpine's libgcc defines
that label exactly as GNU's does, and `ldd` on the OnePlus 6T resolves it to
`/lib/libgcc_s.so.1`. So `lightningcss-native-linux-arm64-musl`'s `libgjsify_lightningcss.so`
— whose entire version-needs section is one group, `File: libgcc_s.so.1`, and whose
DT_NEEDED is `libgcc_s.so.1` + `libc.musl-aarch64.so.1` with no glibc soname at all —
measured a glibc floor of 2.0. Its `linux-x64-musl` sibling, built from the same source in
the same job, measured none: on x86-64 the compiler emits no reference to those two shims.
Same source, two arches, one spurious number. That asymmetry is the tell.

That is where the `glibcRequires` the generator wanted came from, and it is not a floor at
all: a promise about glibc from a binary no glibc host can load.

`isGlibcSoname()` in `binary.mjs` is now the ONE predicate for "does glibc provide this
soname" — glibc's C and maths libraries, the five loader spellings, and the pre-2.34
satellites — and `libcFlavourOfNeeded` uses it too, so the two cannot disagree about the
same bytes. Measured over all 38 committed Linux libraries: exactly one number changes,
the 2.0 above. `libm.so.6` is deliberately in the set, because riscv64's lightningcss
needs `GLIBC_2.35` symbols from it.

## 2. `muslVerdictOfNeeded` had no answer for a musl-linked image

Its question was really "can a GLIBC image load on musl", and its `else` branch read any
libc soname it did not recognise as the glibc loader as `undetermined`. So the first four
`-musl` packages were told:

    declares `libc: ["musl"]`, but its committed targets are glibc-linked with
    musl-loadability undetermined (linux-arm64-musl). The only defensible package-level
    value here is `["glibc"]`

about libraries the same run's own summary counted as "4 musl", and about a
`libc: ["musl"]` that `generate-platform-packages.mjs` writes from the token. One audit
run, two contradictory statements. A third: the `undetermined` note spelled
`declares libc: ["glibc"]` from a hardcoded string while the failure beside it
interpolated the real value — now interpolated.

`musl` is a fourth verdict and its own REQUIRED tier. It is the mirror of `incompatible`,
not a weaker `undetermined`: no `libc.musl-*` file exists on a glibc host under any name,
so the artifact loads on exactly one libc and the ELF settles it in both directions. A
musl target MIXED with a glibc one in one package is a failure of its own — npm's `libc`
is one package-level filter, and since ADR 0017 the suffix is part of the package name.

## 3. The trigger could not have reached either of them

`prebuilds.yml`'s `paths:` and `changed-packages.mjs`' `SHARED_SCRIPTS` name
`scripts/manifest-conformance/**`. ADR 0017 moved `prebuild-artifacts`, `prebuild-libc`
and the one ELF/Mach-O/PE parser under them into
`packages/infra/manifest-conformance/`, which appeared in neither list for as long as it
has existed — so editing what every build's gate ACCEPTS started no prebuilds run, on the
PR or on the merge. That is the silent staleness the entry above it was added to prevent,
re-opened by a move, and it is on this fix's critical path: without it, a PR fixing
`prebuild-libc` could not run the job it fixes. `pr-trigger-parity` cannot see this class
by design — it compares the two lists to each other, and two sides that are equally wrong
look like parity. `selfCheck` now asserts every shared input is in BOTH lists.

## The artifacts land here, in a reviewed commit

The four directories from run 34311463250 are committed and their exemptions cleared, by
`clear-committed-platform-exemptions.mjs` itself, which now writes exactly what the
generator emits. Carried by hand rather than left to the next `commit-prebuilds` for one
reason: the suites build their fixtures from REAL artifacts copied out of the tree, and
the only ELF in the repository whose sole `GLIBC_*` need names a non-glibc file is the one
in this run. Without it the regression for defect 1 cannot be written at all, and the
`-musl` fixture the libc rule needed was `LIBC_AGNOSTIC` — legitimately unconstrained,
which is why nothing here caught defect 2 either.

## Verified

    node scripts/generate-platform-packages.mjs --check   exit 0 (70 packages)
    node scripts/audit-runtimes.mjs --check --strict      exit 0
    node scripts/audit-test-scripts.mjs                   exit 0
    the 5 gate-pushed-tree suites                         185/185

`prebuild-libc` now reports 37 glibc / 4 musl / 7 agnostic with 0 undeclared floors, and
no note for any musl package. Every one of the new cases was run against the unfixed code
first: the reader case fails on the real error text, the five verdict cases fail, and the
end-to-end exemption-clearing case reproduces the production message verbatim.

On the OnePlus 6T (postmarketOS, musl 1.2.6, aarch64, gjs 1.88.1): both typelibs resolve,
`ldd` resolves every DT_NEEDED including `libgcc_s.so.1`, `GjsifySabNative.SharedBuffer`
instantiates, and `GjsifyLightningcss.Engine.transform` — the Rust cdylib carrying the
`libgcc_s.so.1` version need — minifies real CSS.

## For a maintainer to decide

- `libm.so.6` now counts as a glibc soname for `libcFlavourOfNeeded` too, not only for the
  floor. Measured no-op: every directory in the tree recording `libm.so.6` already records
  `libc.so.6`. Stated because it is a widening, not because anything moved.
- The musl legs do not run `check-staged-prebuild-libc.mjs`. Correct as it stands — musl
  ignores symbol versioning and these artifacts have no glibc floor — so nothing is
  deferred, but it is the one PR-side gate the glibc legs have and they do not.
- `changed-packages.mjs` now rebuilds all twelve bridges on a change under
  `packages/infra/manifest-conformance/`, riscv64 included (~97 min). That is what the
  shared-input rule already prescribes for the other registry; this PR pays it.

Closes nothing; unblocks the release.


JumpLink:
## Diagnosis, with the evidence

### The failure

`main` at `c8ae146b84`. Every workflow green except **Build Native Prebuilds**, run
[34311463250](https://github.com/gjsify/gjsify/actions/runs/34311463250). All thirteen
build legs succeeded — both Alpine ones included, and riscv64. The job that failed is
`Commit prebuilds to repo`, step `Sync to origin/main and stage the artifacts`, one line
after `restored 49 of 49 artifact file(s) from the snapshot`.

Reproduced locally byte-for-byte by downloading the four musl artifacts from that run into
the paths the job's own `download-artifact` steps write to (`<pkg>/prebuilds/<target>/`,
which is the layout that path produces — the `stage-prebuild.mjs` rule
`tests/e2e/prebuild-change-gate` enforces is about BUILD legs, not this download):

```
Error: clear-committed-platform-exemptions: refusing to write
@gjsify/lightningcss-native-linux-arm64-musl.
    line 19: have "        \"platforms\": ["
    line 19: want "        \"glibcRequires\": {"
```

### The guard is right; the measurement is wrong — twice

Both defects read the same DT_NEEDED bytes. Both were written when every committed
artifact was glibc-built. Each calls a musl build glibc in its own vocabulary.

**1. `readElfGlibcRequires` ignored `vn_file`.** `readelf -V` on the artifact:

```
Version needs section '.gnu.version_r' contains 1 entry:
  000000: Version: 1  File: libgcc_s.so.1  Cnt: 4
    0x0010:   Name: GCC_3.3     Version: 5
    0x0020:   Name: GLIBC_2.0   Version: 4
    0x0030:   Name: GCC_4.2.0   Version: 3
    0x0040:   Name: GCC_3.0     Version: 2
```

`__register_frame_info@GLIBC_2.0` and `__deregister_frame_info@GLIBC_2.0`, both WEAK, both
against `libgcc_s.so.1`. The label is GCC's — those symbols were glibc's before they moved
to libgcc — and on the OnePlus 6T `ldd` resolves it to `/lib/libgcc_s.so.1`, pmOS's own.
DT_NEEDED for that file is `libgcc_s.so.1` + `libc.musl-aarch64.so.1`, no glibc soname at
all. The x64-musl sibling from the same job has the same Verneed group **without**
`GLIBC_2.0`: on x86-64 the compiler emits no reference to those two shims. Same source,
two arches, one spurious floor — that asymmetry is what makes it a reader bug rather than a
property of the build.

I grouped every `GLIBC_*` need in the tree by its `vn_file` before changing anything. All
38 committed Linux libraries name only `libc.so.6`, `libm.so.6`, `ld-linux-*` or
`ld64.so.*`. The single outlier in the whole repository is that one file. So attribution
subtracts exactly one number, and it is the one that caused the outage.

**2. `muslVerdictOfNeeded` had no musl answer.** Clearing the exemptions by hand to see
past the first failure:

```
declares `libc: ["musl"]`, but its committed targets are glibc-linked with
musl-loadability undetermined (linux-arm64-musl). The only defensible package-level
value here is `["glibc"]`
```

— in the same run whose summary line reads `37 glibc, 4 musl, 7 libc-agnostic`, about a
`libc: ["musl"]` that `generate-platform-packages.mjs` writes from the token, deliberately
and with a comment saying why the token is the stronger source. The rule refused what the
generator emits.

The two contradictory statements in one run come from two readers of the same bytes:
`stats.muslTargets` counts `m.flavour` (correct) while Check C branches on `m.musl`
(`libcFlavourOfNeeded(needed) === null ? 'agnostic' : 'undetermined'` — any non-null
flavour becomes `undetermined`). The *third* contradiction, the note reading
`declares libc: ["glibc"]` about a package declaring `["musl"]`, is a hardcoded string at
what is now line 613 while the failure beside it interpolates. All three are fixed.

### Your loose threads

- `measurePrebuildLibc('packages/infra/lightningcss-native-linux-arm64-musl', [...])`
  returning `{libs: []}` was the call, not the discovery path: the signature is
  `measurePrebuildLibc(dir)` with **one** argument, and `dir` is the directory holding the
  `.so` files — `<pkg>/prebuilds/<target>`, not the package root. The second argument was
  ignored and the package root has no `.so`.
- The two contradictory statements: `stats.muslTargets` vs `m.musl`, plus the hardcoded
  note. Above.
- Your `cp` was fine. `commit-prebuilds` downloads straight into
  `<pkg>/prebuilds/<target>/` and never calls the stager; the ban covers step bodies in
  `prebuilds.yml` / `emulated-build.sh` / `musl-build.sh` / `release.yml`.
- The `prebuild-libc.mjs` caveat you flagged is right and I kept it — but it is about a
  GLIBC-built image running on musl. For a musl-BUILT one the floor is not weak, it does
  not exist, and the file now says so where the caveat is.

### The third defect, which you could not have hit from where you stopped

`prebuilds.yml`'s `paths:` and `changed-packages.mjs`' `SHARED_SCRIPTS` both name
`scripts/manifest-conformance/**` and neither names
`packages/infra/manifest-conformance/**`, where ADR 0017 put `prebuild-artifacts`,
`prebuild-libc` and the ELF/Mach-O/PE parser under them. So a change to the rules that
decide whether a prebuild is acceptable starts **no prebuilds run**, on the PR or on the
merge — the silent staleness the comment above that entry was added to prevent, re-opened
by a move. It is on this fix's critical path: without it, a PR fixing `prebuild-libc`
cannot run the job it fixes. `pr-trigger-parity` documents that it cannot see this class
(it compares the two lists to each other). `selfCheck` now asserts shared-input ⊆ both
lists, and the negative case is in `tests/e2e/prebuild-change-gate`.

### What I ran

```
node scripts/generate-platform-packages.mjs --check    exit 0 — 70 packages, exemptions 6 → 2
node scripts/audit-runtimes.mjs --check --strict       exit 0
node scripts/audit-test-scripts.mjs                    exit 0
node --test <the 5 gate-pushed-tree suites>            185 / 185
node scripts/check-agent-context-size.mjs --check      OK
oxlint / oxfmt --check                                 no new findings; warning set identical to base
```

`prebuild-libc` now: `37 glibc, 4 musl, 7 libc-agnostic`, `37 compared … 0 measured but
undeclared`, highest floor still `GLIBC_2.39 (@gjsify/lightningcss-native-linux-arm64)`,
and no note for any musl package.

Every new case was run against the unfixed code first. The reader case fails on the real
value, the five verdict cases fail, and the end-to-end exemption-clearing case reproduces
the production error text verbatim. `install-platform-filter` and `dlx-native-prebuilds`
fail in my worktree identically on `origin/main` — they need a built
`packages/infra/cli/lib/index.js`.

### On the device

OnePlus 6T, postmarketOS, musl 1.2.6, aarch64, gjs 1.88.1, `/tmp` only and cleaned up.
Both typelibs resolve, `ldd` resolves every DT_NEEDED including `libgcc_s.so.1`,
`GjsifySabNative.SharedBuffer` instantiates, and `GjsifyLightningcss.Engine.transform` —
the Rust cdylib that carries the `libgcc_s.so.1` version need — minifies real CSS.

### Why the binaries are in the PR

With defect 1 fixed there is no measured declaration left to make, so `commit-prebuilds`
would be entitled to land them itself on the next run. I carried them anyway for one
reason: `tests/e2e/prebuild-declaration-invariant` builds its fixtures from **real**
artifacts copied out of the tree, on purpose, and the only ELF in the repository whose sole
`GLIBC_*` need names a non-glibc file is the one from run 34311463250. Without it the
regression for defect 1 cannot be written. The same gap explains defect 2: the suite's only
`-musl` fixture was `LIBC_AGNOSTIC`, which is legitimately unconstrained in a `-musl`
directory, so the case that is not was untestable until a musl artifact existed to copy.

The exemptions are cleared by `clear-committed-platform-exemptions.mjs` itself, not by
hand, so the manifests and the regenerated READMEs are exactly what CI would have written.

### Maintainer calls

1. `libm.so.6` now counts as a glibc soname for `libcFlavourOfNeeded` as well as for the
   floor. Measured no-op — every directory recording `libm.so.6` already records
   `libc.so.6` — but it is a widening and should be seen as one.
2. The musl legs do not run `check-staged-prebuild-libc.mjs`. Defensible as it stands, and
   with this fix a musl artifact measures no floor at all. Left alone rather than deferred.
3. A change under `packages/infra/manifest-conformance/` now rebuilds all twelve bridges,
   riscv64 included (~97 min). That is what the shared-input rule already prescribes for
   the other registry; this PR is the first to pay it, this run included.

Not merging and not enabling auto-merge.

JumpLink:
### CI note: this PR's `pull_request` event produced no Actions runs

Worth flagging separately, because it is not the fix and it is not the branch's content.

Over ~20 minutes and three separate `pull_request` events — `opened`, a `synchronize` from a
force-push, and a `reopened` from a close/reopen — GitHub Actions created **zero** runs and
**zero** check suites for this branch:

```
gh api repos/gjsify/gjsify/commits/<sha>/check-suites
  -> cursor:queued  socket-security:queued        (no github-actions suite, on all three shas)
gh pr checks 1613
  -> no checks reported on the 'fix/musl-prebuilds-land' branch
```

The push webhook clearly fired — both third-party apps created suites on every sha. Actions
itself is healthy: another branch pushed one minute after mine got its full nine-workflow set,
and kept getting them throughout. So it is neither a repo-wide outage nor a queue.

Ruled out from this side:

- **Not a workflow-file parse error.** Both edited `paths:` lists parse and are identical, 43
  entries each, and `changed-packages.mjs`' `selfCheck` agrees (it fails loudly when they
  diverge, which is how I know it is reading them).
- **Not a conflict.** `mergeable=MERGEABLE`; `main` has not moved off `c8ae146b84`, so the merge
  ref is computable. (`mergeStateStatus=BLOCKED` is only the required checks not being green.)
- **Not the ref.** `workflow_dispatch` on the same branch runs fine —
  [GJS 34321164595](https://github.com/gjsify/gjsify/actions/runs/34321164595) and
  [Build Native Prebuilds 34321287477](https://github.com/gjsify/gjsify/actions/runs/34321287477)
  are the verdicts to read for this PR's content.
- **Not workflow state.** All sixteen workflows report `active`.

So the dispatched runs are what I have. They cover the required-check content (`GJS`) and the
workflow this PR is about (`Build Native Prebuilds`, whose `commit-prebuilds` job is correctly
skipped on a non-`main` ref by its own `if:`). What a dispatch does NOT reproduce is the
`pull_request` event itself, so `Commitlint` and `Audit runtimes` — which have no
`workflow_dispatch` trigger — have no run at all. Both were run locally instead: the commit
subject is 46/100 through `check-pr-title-subject.mjs`, this description's longest line is
93/100 through `check-pr-body-lines.mjs`, and `audit-runtimes --check --strict` exits 0.

If the `pull_request` runs are still absent when someone looks at this, a push on top of the
branch is the cheap thing to try before anything else. It should not need a change to this PR.

JumpLink:
### Correction to the CI note above: the `pull_request` runs arrived

They spawned at 07:08Z, roughly 35 minutes after the PR was opened and about 14 minutes
after the last `synchronize` — all eleven workflows at once, including `Commitlint`,
`Audit runtimes` and `Build Native Prebuilds`, which is the set this PR needs read.

So it was a GitHub-side delay in run creation, not anything about this branch, and the
previous comment's ruling-out was right about the cause being external but wrong to imply
it was stuck for good. Nothing needs doing about it; I am leaving that comment in place
because the delay is worth knowing about — for 35 minutes `gh pr checks` reported
`no checks reported on the ... branch`, which reads exactly like green, and AGENTS.md's
"watch the WORKFLOW status, not the check list" is the rule that catches it.

The two `workflow_dispatch` runs are redundant now and were not cancelled by
`Cancel superseded PR runs`, correctly — its selection filters on `event=pull_request`.
Read the `pull_request` runs; the dispatched ones are just further along.

JumpLink:
## The missing CI runs: it was `[skip ci]` in the commit body, not a GitHub delay

Correcting the note above, which concluded that GitHub created the `pull_request` runs
on its own after a ~35 minute delay. It did not. They appeared because the commit
message changed.

This branch's single commit quoted the marker while *explaining* the guard it fixes:

```
declared a floor from a `[skip ci]` job. `isGlibcSoname()` is now the one
```

GitHub honours `[skip ci]` anywhere in a commit message, quoted or not, and suppresses
**every** workflow run for that commit. That is why `opened`, `synchronize` and
`reopened` each produced nothing: all three resolve to the same head commit, and that
commit said not to run.

The run list settles it. Every `pull_request` run sits on the amended SHA, none on the
original:

```
06:53:44  1ece5e07de  workflow_dispatch  GJS
06:55:14  1ece5e07de  workflow_dispatch  Build Native Prebuilds
07:07:56  943dc80124  pull_request       GJS, Audit runtimes, Commitlint,
07:07:56  943dc80124  pull_request       Build Native Prebuilds, + 7 more
```

`943dc80124` is `1ece5e07de` with one sentence reworded to `a skip-ci job` — no
brackets, so no directive. `git diff 1ece5e07de 943dc80124` is empty: the tree is
byte-identical and only the message changed. The eleven runs were created within a
minute of the force-push, ~14 minutes after the "35 minute delay" theory would have
had them arrive anyway.

Worth keeping because the failure mode is nasty in three ways. It is silent — nothing
reports that a commit suppressed its own CI. It survives every retrigger, since the
marker travels with the commit. And for those minutes `gh pr checks` answered
`no checks reported`, which reads exactly like green: the same trap `AGENTS.md`
describes as "a workflow that has not spawned its jobs contributes zero checks".

A commit message is an instruction stream as well as prose. When documenting a marker,
name it (`skip-ci`, `the skip-ci marker`) rather than pasting it.

PR #1615, opened 2026-09-09
feat(adwaita-nativescript)!: methods and signals as GJS spells them

JumpLink:
ADR 0034 converged widget names and is counting down property names.
Methods were the axis no ledger and no gate touched: across the 40
gallery blocks with both a `gjs` and a `nativescript` pane the two
surfaces shared three method names (`add`, `present`, `push`).

And the spelling is not free. Measured on the real GJS:
`Gtk.Button.prototype.add_css_class` is a function, `addCssClass` is
undefined; `Adw.ToolbarView.prototype.add_top_bar` is a function,
`addTopBar` is undefined. GJS installs snake_case and nothing else, so
a method converges to the typelib's spelling or not at all.

## What lands

**`connect` / `disconnect` on every widget** (`widgets/signals.ts`), a
mixin applied where a class meets `@nativescript/core`:
`extends withSignals(GridLayout)` on the 36 classes with a platform
base; the 12 extending a port class inherit it. ADR 0034 § Amendment 13
refused a base class and a mixin for the construct-props bag on
constructor-order grounds, which do not reach a method. Measured before
building it: no view base of `@nativescript/core@9.1.0-alpha.11`
declares either name. Arm 6 of `check-nativescript-xml-doors` prints
`48 widget class(es) reach connect/disconnect — 36 wrap their
@nativescript/core base in withSignals(), 12 inherit it through a port
base`, and goes red on a bare platform base and on a second wrap (A/B
run on both).

**The GIR side is the typelib, committed.** `@girs/<ns>/vocabulary`
carries no method table (measured on 4.6.0: eight exports, none a
method table), so `scripts/generate-widget-methods.mjs` reads the
installed typelib through GIRepository and writes
`gtk-host/src/generated/methods.mts` — the arrangement ADR 0029
§ Amendment 2 chose for enum values, held the same two ways: gtk-host's
`generated.spec.ts` asks the running GJS for every name, chain and host
verb; the no-install vocabulary gate holds the shape. What it proves:
two spellings agree. What it does not: behaviour under the name —
`add_prefix` on the port keeps one slot where libadwaita keeps a box,
declared in place.

**`NS_METHOD_ALIGNMENT`**, the third ledger, with the property ledger's
three verdicts and its rule (same kind of argument, different spelling
= rename). The port side resolves the port's abstract bases. The gate
now prints, derived at run time:

    Methods, on the same surface, against the typelib-read
    generated/methods.mts (Adw-1/1.9.3 Gtk-4.0/4.22.4): 44 widgets
    with a covered counterpart declare 137 public method(s) between
    them — 67 already spell a method of the counterpart's chain, 70 do
    not (20 should converge, 50 declared own, 0 undecided)
    Distance to one vocabulary: 1 widget name(s), 4 property name(s)
    and 20 method name(s), and all three can only go down.

**Forty-eight methods took their GIR names** (breaking, `!`): 49 public
names retired against `main`, `AdwEntryRow`'s `addSuffix`/`setSuffix`
folded into one `add_suffix`, `remove(view)` added on the action row
and the entry row. Carried through the port's specs, the storybook
package, both showcases and the website's NativeScript panes, whose
three `addEventListener` calls became `connect`. Two renames closed
defects: `AdwExpanderRow.addRow(View | ItemSpec)` carried that union
only because the name shadowed `GridLayoutBase.addRow(ItemSpec)`; and
`AdwNavigationView.removePage` was named to avoid a `ViewBase.remove`
that, measured, does not exist.

## Verdicts

- `gir` (20, should converge): the tab view's page methods take the
  page-id string ADR 0048 chose where GTK passes the `Adw.TabPage`;
  `AdwCarousel.scrollToPage` takes an index where `scroll_to` takes the
  widget; `AdwViewStack.removePage` takes a name where `remove` takes
  the child.
- `own` (50): a property of a flattened page type; an action
  (`navigation.push`); a signal (`apply`); private C API the port has
  to expose (`set_show_indicator`); the four collection setters
  (`setViews`, `setItems`, `setSections`, `setToggles`), pointing at
  the open-todo entry on list widgets GTK builds with a method (#1524
  is closed, so they are not gaps); `AdwWrapBox.addChild`, the
  `LayoutBase` override the port cannot not have.
- `gap`: none.

## Left open, with the reason

- Cardinality under a converged name (`add_prefix` one slot vs a box):
  the ledger compares names and cannot see it; ADR 0034 § Amendment 14
  and the widget file declare it.
- The 20 `gir` entries: closing each is an API change (accept the page
  record or the page view), not a rename.
- Signal names: `connect('tap', …)` vs `connect('clicked', …)` — the
  verb converged, the names have no ledger yet.
- Upstream: an `OWN_METHODS` export from ts-for-gir's surface generator
  would let gtk-host emit the table under one provenance and give the
  typelib reading a second source; recorded in `status/open-todos.md`.
- The ambient slice cannot hold a future `Observable.connect`; the
  venue that could is `tests/integration/nativescript`, recorded there.

Gates run green locally: vocabulary alignment (self-test 88 vectors),
xml-doors, widget-coverage, doc-fences, generated-website-data,
storybook-control-parity, theme-classes, disclosure-toggles,
collection-reactivity, enum-values, tag-vs-class, storybook-widget-
coverage, agent-context-size. `gjsify lint` and `gjsify format --check`
clean over every changed file. Port: `check`, `build`, `test:node`;
gtk-host `check`. The GJS test legs are reported in the PR comment.


JumpLink:
Local runs behind the PR body's summary, kept out of the body so its lines stay under 100.

| leg | result |
|---|---|
| `@gjsify/adwaita-nativescript` `check` / `build` / `test:node` / `test:gjs` | green, incl. the two `withSignals` suites |
| `@gjsify/gtk-host` `check` / `test` (GJS) | green — the four new cases: installs every method the artifact names; chains every widget as the running GType system does; reads the host verbs off `GObject.Object.prototype`; excuses only the widgets this host has no class for |
| `gjs -m scripts/generate-widget-methods.mjs --check` | up to date; `gjsify format --check` agrees on the same bytes |
| `check-vocabulary-alignment` | self-test 88 failing vectors; distance 1 / 4 / 20 |
| `check-nativescript-xml-doors` arm 6 | 48 classes reach connect/disconnect — 36 wrap, 12 inherit; A/B red on a bare base and on a double wrap |
| `check-nativescript-widget-coverage`, `check-doc-fences`, `check-generated-website-data`, `check-storybook-control-parity`, `check-nativescript-theme-classes`, `check-nativescript-disclosure-toggles`, `check-adwaita-collection-reactivity`, `check-enum-values`, `check-adwaita-tag-vs-class`, `check-storybook-widget-coverage`, `check-agent-context-size` | green |
| `gjsify lint` / `gjsify format --check` over the 84 changed source files | clean |

Comment budget for `scripts/` moved 0.607 → 0.601 (target 0.592, already over on `main`).

JumpLink:
Two false reds fixed, and they were one blind spot appearing twice.

`AdwSplitViewBase` is the only class in the port whose declaration does not fit
on a line — it is generic AND wrapped in `withSignals(…)` — so the formatter
pushes the base onto its own line. Neither reader over this package could read
that: the optional `withSignals\(` group backtracks to empty and the base
capture then takes `withSignals` itself. So `check-nativescript-xml-doors`
reported "extends withSignals bare" on a class that is wrapped correctly, and
`check-nativescript-widget-coverage` reported a chain that "leaves the package
at 'withSignals'" for the two split views.

Every self-test vector on both sides was single-line, which is why it survived
being written down. The vectors now carry the wrapped-and-broken form — with
and without the trailing comma, and with a type parameter that itself contains
`extends`. Restoring the intolerant pattern turns all four red in both gates,
naming the misread base.

There were TWO readers, which is why one defect appeared twice. They are now
one, in `nativescript-xml-doors.mjs` beside the mixin it reads, with the
vectors, and `classReaderSelfTest()` is called by both gates.

Also: the `!` commit's subject started upper-case, which commitlint refuses.
Reworded, and the branch was rebuilt with identical content (`git diff` against
the pre-rewrite state is empty).

JumpLink:
Rebased onto main after #1614 landed. Three collisions, all of them the
shared-counter kind, and one of them turned out to be a convergence rather than
a compromise.

**The gallery panes.** #1614 rewrote the same `slot="nativescript"` fragments
onto `gi://` plus the construct-props bag while this branch renamed their
methods. Taking either side alone would have thrown away the other's
improvement, so the resolution takes both: main's bag form with this branch's
verbs. `Adw.PreferencesGroup` on `boxed-lists.mdx` is now byte-identical to its
`gjs` sibling.

The rename map was NOT reconstructed by guessing. Ten of the removals had no
mechanical snake_case counterpart (`addRow`, `addGroup`, `setPrefix`,
`setSuffix`, …), so the port was asked instead: `AdwPreferencesGroup` and
`AdwPreferencesPage` both declare `add`, `AdwActionRow` declares `add_prefix`
and `add_suffix`. 36 call sites renamed inside NativeScript panes only, then
audited — every method a pane now calls resolves against a method the port
declares.

**ADR 0034 § Amendment 14.** Both PRs claimed the number. #1614's is on main, so
this one is § Amendment 15, and the seven citations in the port's AGENTS.md and
the five gate scripts moved with it — a citation pointing at the wrong amendment
is exactly the drift this repo does not tolerate. One pre-existing misreference
fixed beside it: the construct bag is § Amendment 13, not 12.

**The class reader.** The renumbering touched the very docblock the reader fix
deletes; my side wins, and the citation survives in the shared module's header.

All six affected gates green locally, format clean. Arm 12's printed distance:
491 → **481** lines.

PR #1617, opened 2026-09-09
feat(react-native): let a tab layout contribute a persistent bottom bar

JumpLink:
`<Tabs bottomBar={…}>` puts a caller's own bar between the content and the view
switcher bar — a now-playing strip, a selection toolbar, an offline notice.

## Why a prop, and why an app could not do this itself

`<Tabs>` reads its children as `<Tabs.Screen>` DECLARATIONS and refuses anything else
by name, so there is no child position for a bar. And wrapping `<Tabs>` from outside
does not work either: the switcher is created with `slot: 'title'`, which resolves
against the PARENT, so a box placed between a tab layout and the header bar takes the
switcher's slot away and the router refuses it. A consumer that wanted a now-playing
strip above its tabs had no way to get one and carried "no mini player" as a recorded
limitation instead.

## Where it lands, measured

`Adw.ToolbarView`'s `bottom` slot is an ADDER, so it takes this bar and the switcher
bar together and the order decides the stacking. MEASURED on libadwaita 1.9.3, two bottom
bars added in sequence in a 480x320 window:

  content       y=0   h=214
  added FIRST   y=217 h=40
  added SECOND  y=257 h=24

The first-added sits closer to the content, so `bottomBar` renders BEFORE the switcher
bar — which puts it above, where a phone puts a now-playing strip relative to its tab
bar.

It is wrapped in a `Gtk.Box` rather than slotted directly, so a caller can hand over
any node — a fragment, or a component whose root element it does not control — without
having to know that a slot name is involved.

## Vectors

`puts a caller's bottom bar between the content and the switcher bar`, windowed at
420x700 so the switcher bar is actually revealed. Two mutants die: dropping
`persistentBar` from the shell, and rendering it after `switcherBar` instead of before.

Two things the vector had to learn, both faults in the vector rather than the router:

  - it settles on `bar.get_reveal() && bar.get_height() > 0`, not on `reveal` alone.
    `reveal` flips before the animation runs, and measured at that moment the bar is
    `height=0` at exactly the y where the bar above it ends — so "above the switcher
    bar" compared two coincident edges.
  - the marker is a `GtkSwitch`. A box or a label is indistinguishable from the wrapper
    the router adds; `AdwActionRow`, tried first, is a `GtkListBoxRow` and outside a
    list box criticals with `gtk_list_box_row_grab_focus: assertion 'box != NULL'
    failed`, which the suite's quiet-GTK gate then failed the test for.

## What is NOT gated, and why the obvious assertion was empty

The implementation renders `null` rather than an empty wrapper when no bar is asked
for. "An empty bottom bar is a strip of dead pixels" was the reason given for that, and
it is FALSE: measured, an empty `Gtk.Box` as a bottom bar takes no height and the view
stack above it is allocated 98 px either way, so a mutant that always rendered the
wrapper passed every geometric assertion and the whole vector. What the guard actually
saves is a widget and an element per tab layout per render — real, invisible, and
unmeasurable without a bar count that Adw exposes no getter for. Said in the code and
in the spec rather than left as a passing mutant nobody mentioned.

`headerShown: false` drops the whole toolbar view, so it drops this bar too. Named in
the prop's docblock rather than worked around: a navigator that asked for no chrome
getting a bar anyway would be the greater surprise.

## Checked

`@gjsify/react-native`: 638 tests, 2760 assertions, green on Gjs 1.88.1;
`gjsify tsc --noEmit` clean.


JumpLink:
## What this pull request actually carried

Recorded here because `git log` shows `(#1617)` against this commit, and this is the
cheapest place a later reader will find it.

The branch was cut from a local topic branch instead of from `main`, so the diff against
`main` carried that branch's single commit as well — `943dc80124 fix: measure musl and
glibc as different libcs`. The merge `b75fa8ed5f` is therefore 37 files, of which 2 are
the router change this title describes:

    packages/framework/react-native/src/router/tabs.ts
    packages/framework/react-native/src/router/router.spec.ts

The other 35 are that musl/libc measurement work — the prebuild toolchain, `prebuilds.yml`,
`docs/prebuilds.md`, ADR 0024, four e2e suites, and the `-musl` platform packages with
their `.gir`/`.typelib`/`.so` artifacts. It is coherent work with its own subject, it was
green, and `main` is green with it; what it lacks is a commit message that names it.

A revert was prepared (#1618) and is being closed instead: reverting would take correct,
already-green work back out of `main` and require it to be re-landed, and the only thing
it would buy is attribution. `STATUS.md` is explicitly a snapshot rather than a log, and
`CHANGELOG.md` is generated from commit subjects, so neither is the place for a note —
which leaves this.

**The consequence to know:** the generated changelog will file those 35 files under a
`feat(react-native)` entry about a tab bar, and `fix: measure musl and glibc as different
libcs` will not appear in it.

PR #1618, opened 2026-09-09
revert: "feat(react-native): let a tab layout contribute a persistent bottom bar"

JumpLink:
This reverts b75fa8ed5f, which landed 37 files under a subject describing 2 of them.

## What went wrong

The branch behind #1617 was cut from a local topic branch instead of from `main`, so
the pull request's diff against `main` carried that branch's work as well. The subject
and the body described only the router change, and the file list was never read before
the merge. What arrived alongside it:

  packages/infra/lightningcss-native-linux-arm64-musl/   .gir, .typelib, .so
  packages/node/sab-native-linux-x64-musl/               .gir, .typelib, .so
  .github/workflows/prebuilds.yml
  .github/prebuild-toolchain/changed-packages.mjs
  docs/adr/0024-ship-installable-artifacts.md
  docs/prebuilds.md
  scripts/generate-platform-packages.mjs
  tests/e2e/prebuild-change-gate/, prebuild-declaration-invariant/, prebuild-loader-path/,
  platform-exemption-clearing/

Binary prebuild artifacts and workflow changes are exactly the content that should not
reach `main` without being reviewed as itself. CI was green on the merge and `main` is
green now, so this is not a revert for breakage — it is a revert for provenance. The
history currently attributes a prebuild toolchain to a commit about a tab bar, which is
the kind of thing nobody finds again when they need to know why it is there.

## What happens next

Two follow-ups, each reviewable on its own:

  - the router change, `<Tabs bottomBar>` with its vector, re-opened from `main`;
  - the prebuild work, from whoever owns that topic branch, with a subject that names
    it and a body that argues for it.

Nothing here judges the prebuild work. It may well be right; it has simply not been
read under a title that would make anyone read it.


JumpLink:
Closed without merging, and the reasoning belongs with it.

The reason for the revert was provenance: `b75fa8ed5f` landed 37 files under a subject
describing 2. That is still true, and it is now recorded on #1617.

What it turned out not to be is a review or a safety problem. The carried work is a
single commit — `943dc80124 fix: measure musl and glibc as different libcs` — by the same
author as the pull request that carried it, continuing a line already merged (#1607,
#1602, #1329, #1219). The argument that binary artifacts and workflow changes reached
`main` unreviewed does not survive contact with who wrote them.

Against that, the revert had a cost understated when it was proposed: it takes correct,
CI-green work back out of `main` and asks for it to be re-landed, buying attribution
alone.

`STATUS.md` is explicitly a snapshot rather than a log and `CHANGELOG.md` is generated
from commit subjects, so neither is a place to write the note — which is why it went on
the pull request that caused it.

PR #1620, opened 2026-09-09
feat(adwaita-nativescript): resolve adwaita icon theme names

JumpLink:
ADR 0034 § Amendment 14 printed a distance between the gallery's `gjs` and
`nativescript` panes, split it four ways, and said one sentence about the
cheapest bucket: *"The `glyph` kind is one decision — the port resolves no theme
name, so every icon pane imports its glyph — and would close a whole bucket at
once."* This makes that decision.

`iconName: 'list-add-symbolic'` now works on the NativeScript port.

## The premise was wrong about which half was missing

Four widget headers, a template generator, a gate's scope paragraph and the ADR
all said the same thing: NativeScript's `Image` decodes no SVG. True — and not
the missing piece. The port already rasterises the document itself
(`icons.android.ts` walks the path data onto a `Bitmap`, `icons.ios.ts` replays
it into a `UIBezierPath`). What was missing was the half in FRONT of that: a
NAME to look a document up by. The port had a rasteriser and no theme.

The tell was in the tree. **Eleven hand-rolled name-to-glyph registries**, one
or two names each, none able to see the others — `avatarIconSvg`, `nsIconSvg`,
`peekIconSvg`, `buttonContentIconSvg` in the port, and seven more across the
NativeScript storybook showcase, each re-implementing the `-symbolic` strip and
a switch with its own fallback. All eleven are gone.

The eleventh had a branch for `edit-delete`, a name its own story control does
not offer, so picking "Trash" in that story drew nothing at all. Nothing could
have found that before: there was nothing to compare a name against.

## The shape is the web pillar's, and the bundle delta is measured

`packages/web/adwaita-web/src/icon-registry.ts` answered this for the browser: a
compiled subset plus a `registerIcon()` door, with the reason recorded as a
measurement (~1.07 MB of data-URI for the whole set, five times the stylesheet).
`widgets/icon-theme.ts` is the same shape one target over, where the argument is
a phone download.

Measured over `packages/web/adwaita-icons/*.ts`, summing the template literals:

- the whole barrel — 644 glyphs, 732 371 bytes of SVG source (715.2 KiB)
- the compiled subset — 31 glyphs, 27 851 bytes (27.2 KiB)
- **what the subset ADDS — 25 glyphs, 24 133 bytes (23.6 KiB), 3.3 % of it**

The third row is the honest one: six of the thirty-one were already imported by
a widget in that directory for its own chrome. An entry is paid for by every app
that touches the module — a bundler cannot tree-shake a property out of an
[...]

PR #1621, opened 2026-09-09
feat(website): gloss fence attributes from the GIR

JumpLink:
The Attributes tab described what a preview PASSES, never what an attribute IS —
and the fence one tab to the left already showed those values. This fills the gap
its deletion leaves, in the markup a reader copies:

```html
<!-- selected: The position of the selected item. -->
<adw-combo-row title="Region" model='["Europe","Americas"]' selected="0"></adw-combo-row>
```

## The mechanism

An HTML comment is inert in the DOM, so the one fence that renders and is shown
carries the gloss for both — no second copy to keep. The text is derived, never
written: tag -> GType -> GIR property -> the first sentence of its `<doc>`, read
straight out of `Adw-1.gir` / `Gtk-4.0.gir`. `@girs/*`'s vocabulary subpath
(ADR 0029) carries names, nicks and `SINCE` and no doc text at all.

The same 110-against-54 shape that killed hand-written attribute prose was
already re-forming: of the 40 fences, seven carried an authored HTML comment and
four of those glossed an attribute — 4 of 193 attribute occurrences, nothing
holding the other 189 and nothing checking the four.

## The line, and what fell on each side

Most GIR property docs restate the property's own name ("The subtitle for this
row.", "The displayed label.", "Whether the row is expanded."). Emitting those
repeats the complaint the pane was deleted for, so a gloss is emitted only where
the doc's first sentence carries a word the NAME does not imply — content words
minus the attribute name, minus the element name, minus a closed list of English
function words, minus a closed list of presentation words (the vocabulary a doc
spends on the fact that a property is displayed, which a documentation fence says
by existing), minus the corpus's own shared vocabulary: every word in 4+ of the
docs this gallery shows, which by construction distinguishes none of them.

The shared vocabulary is DERIVED, and that is what keeps the domain boilerplate
out without hand-naming any of it: "preference" and "represented" sit in 10 and 8
of these docs, because "The title of the preference represented by this row." is
eight elements' answer for `title`.

Of the **107** distinct attributes the 40 fences set:

| | count |
|---|---:|
| glossed | **27** |
| the name suffices | **58** |
| no GIR property (declared) | **20** |
| authored gloss stands | **2** |

**27 comment lines against 332 lines of markup, +8%.** The full lists are in a
comment below; the counts are emitted into the generated module and checked, so
moving the line moves a number rather than an opinion.

Two placement decisions came out of measuring, not asserting:

- **A gloss belongs on the element its block is ABOUT.** Glossing every element
[...]

PR #1624, opened 2026-09-09
feat(adwaita-app): ship the app's own adwaita icon theme

JumpLink:
> "The recommended way is to always copy the icons into the app as assets anyway,
> because we cannot be sure the environment the app runs in even uses the Adwaita
> icon set."

That is right, and gjsify did not do it. Measured repo-wide before this change:
`add_resource_path` and `add_search_path` had **zero** hits in every `.ts`, `.js`,
`.mjs`, `.tmpl` and `.blp` outside `node_modules`. Every gjsify GTK app drew
whatever icon theme the host happened to have, and nothing in the tree noticed.

`@gjsify/adwaita-app` now bundles the Adwaita subset into the app's own GResource
and registers it on `startup`. **On by default.**

## The import form cannot be the shared spelling — verified against the GIR

I checked this rather than taking it on trust, by sweeping `Gtk-4.0.gir` and
`Adw-1.gir` for every class carrying an `icon-name` property and asking which of
them also has a `Gdk.Paintable` or `Gio.Icon` sibling:

**22 classes carry an `icon-name`. 15 have no paintable/GIcon property at all** —
`Gtk.Button`, `Gtk.MenuButton`, `Gtk.Window`, `Gtk.StackPage`, `Adw.ActionRow`,
`Adw.ButtonContent`, `Adw.ButtonRow`, `Adw.ExpanderRow`, `Adw.PreferencesPage`,
`Adw.SplitButton`, `Adw.Toggle`, `Adw.ViewStackPage`, …

The 7 that do are `Gtk.Image`, `Gtk.Entry`, `Gtk.AboutDialog`,
`Gtk.CellRendererPixbuf`, `Adw.Avatar` (`custom-image`, which suppresses the icon
rather than replacing it), `Adw.SidebarItem` and `Adw.StatusPage`. So the claim
was slightly understated — `Adw.StatusPage:paintable` exists too — and the
conclusion is unchanged and stronger: for most icon slots there is no property to
hand an imported SVG to. The guarantee has to be built on the icon THEME.

## What `add_resource_path` buys, and what it does not

This is the part worth reading, because the obvious version of this PR would
claim a guarantee it does not have. `Gtk.IconTheme.add_resource_path()`
*contributes* to the icon theme — its own GIR doc says "make application-specific
icons available as part of the icon theme" — it does not override it. Measured on
gtk4 by asking `lookup_icon(...).get_file().get_uri()`, the file the widget paints:

| host theme | `list-add-symbolic` resolves to |
|---|---|
| Adwaita (ships the name) | `file:///usr/share/icons/Adwaita/…` — **host wins** |
| oxygen, Bluecurve, hicolor | `resource:///…` — **bundle wins** |

So the default (`prefer: 'fallback'`) closes the failure that motivated this — the
name never fails to resolve, on any host — while a user who chose Papirus keeps
Papirus. That is what a desktop app should do and what GNOME apps that ship icons
do.

`prefer: 'bundled'` is the opt-in where the shipped glyph matters more than the
host's taste (screenshot rig, kiosk, a gallery whose pictures must match its
prose). It points the icon theme at a name nothing installs, so the theme chain
contributes nothing and only the app's resource can answer — no second bundling
mechanism, one settings write. Its cost is stated: a name outside the bundle then
has only GTK's builtins behind it.

`prefer: 'host'` / `bundledIcons: false` is the documented way out.

## Where the registration lives, and why

`@gjsify/adwaita-app`, next to the startup CSS bootstrap it is the sibling of.
That package's stated job is "the boilerplate every native Adwaita app repeats",
its README's own quick start writes `icon: 'go-home-symbolic'`, and it is the only
place gjsify constructs an `Adw.Application` for you. `gtk-host` is a renderer for
three frameworks, not an app lifecycle; the CLI is a build tool.

Icons are installed **before** CSS in `_onStartup`: `Gtk.IconTheme` caches a
lookup, so a resource path added after something has already asked for a name
leaves that widget on whatever answered first.

**Not wired into the `create-app` templates**, and that is measured rather than
assumed: all four (`gtk-minimal`, `adw-canvas2d`, `adw-game`, `adw-webgl`) build
`Adw.Application`/`Gtk.Application` by hand rather than through `runAdwaitaApp`,
and **none of them names an icon anywhere** — no `icon-name` in any `.ts` or
`.blp`. Adding the dependency would ship 26.5 KiB and a startup call to apps with
no icons. The day a template draws one it is a one-line change, and
`status/open-todos.md` says so.

## Bundling, and the measured bytes

`scripts/build-icon-gresource.mjs` lays the subset out as a hicolor tree, compiles
it through **the CLI's own `gjsify gresource`** command, and embeds the result as
base64.

Base64 because `gjsify build` emits one bundle file with no asset pipeline, so a
`.gresource` sibling is a file the app cannot find at runtime.
`@gjsify/adwaita-fonts`' `build-embedded.mjs` states the same lesson one package
over: what travels through a JS bundle travels as a value. The bundling is still
`glib-compile-resources`; base64 is transport, not a second mechanism.

41 glyphs:

- 39 674 B of SVG source
- 43 377 B uncompressed GResource
- **20 349 B compiled with `compressed="true"` (19.9 KiB) — 53 % off**
- **27 132 B as the base64 an app actually carries (26.5 KiB)**

The subset is not chosen fresh: it is exactly the set `@gjsify/adwaita-web`
compiles, so one `icon-name` means one glyph on all three renderers.
`scripts/check-bundled-icon-parity.mjs` fails when the two lists drift.
`view-columns` is the one declared exclusion — in no icon theme, hand-drawn for
the web only.

I dropped `preprocess="xml-stripblanks"` after measuring it: it shells out to
`xmllint`, which `.docker/ci-fedora.Dockerfile` does not declare (it names `glib2`
and `gettext` explicitly), and it made the compressed resource **bigger** —
20 810 B with it against 20 349 B without — by re-indenting what zlib then carries.

## Proving the bundled glyph actually wins

`add_resource_path` returns nothing and never fails, so asserting it was called
proves the call happened. Every assertion in `icon-theme.spec.ts` instead goes
through `lookup_icon(...).get_file()` — the file the widget paints — and the
central one compares the BYTES there against the glyph in the GResource, because a
right path to a wrong document reads identically.

**It caught a real defect on its first run.** The generator filed
`emblem-system-symbolic` under `scalable/legacy/`, copying Adwaita's own layout —
but a resource path is looked up as part of hicolor, whose `index.theme` defines
no `legacy` directory. GTK never scanned it, the name fell through to
`resource:///org/gtk/libgtk/icons/…`, and everything was silent: resource present,
path registered, different picture drawn. The generator now refuses any directory
hicolor does not define, and arm 1 is the general form of the failure.

7 tests / 14 assertions, GJS only (it needs `gi://Gtk`) but **no display**:
`Gtk.init_check()` returns false here and every lookup still resolves. Full
package: 22 GJS + 7 GTK + 22 Node, all green.

### Broken on purpose, confirmed red, reverted

- **spec arm 1** — re-filed `emblem-system-symbolic` into `scalable/legacy/`:
  red with `Actual: ["emblem-system-symbolic"]`.
- **spec arms 1+2** — made `addBundledIconsToTheme` a no-op: both red.
- **build guard** — mapped a subpath to a non-hicolor context: the generator
  refuses and names the 13 legal directories.
- **parity gate, 4 arms** — a name only the web map has; a name only the GTK
  bundle has; a glyph taken from the wrong subpath; and the exclusion ratchet
  (pretending adwaita-icons had started exporting the excluded name).

One more failure found while writing the spec and worth recording: `Gtk.IconTheme`
in an uninitialised process **segfaults** — exit 139, no stack, nothing on stderr,
surfacing under the runner only as `gjs exited with signal SIGTERM` part-way
through a suite. The spec calls `Gtk.init_check()` once and says why.

## Scope

This closes the guarantee **for GTK only**. The other two halves already exist:
the browser in `packages/web/adwaita-web/src/icon-registry.ts` (compiled subset +
`registerIcon()`), and NativeScript in
`packages/nativescript-bridge/adwaita/src/widgets/icon-theme.ts` (compiled subset
behind the same property), which is in flight as #1620 — this PR does not depend
on it and touches none of its files.


JumpLink:
Diagnosed and fixed. **It was my change, not a win32 defect** — win32 is just the leg that reaches the build first.

## Evidence

The failing step is `Build the layer under test (+ its workspace deps)`, which runs
`workspace @gjsify/react-native build:gjsify --with-dependencies`. `@gjsify/react-native`
depends on `@gjsify/adwaita-app`, so that closure builds this package — via **`build:gjsify`**,
which I had never wired the icon generator into (I only put it on `build` and `test`).

Reproduced locally in seconds, no CI and no Windows needed:

```
$ rm packages/framework/adwaita-app/src/icons.generated.ts
$ gjsify workspace @gjsify/adwaita-app build:gjsify
UNRESOLVED_IMPORT: Could not resolve "./icons.generated.js" in src/icon-theme.ts
```

Platform-independent — the two darwin legs of the same run were still building when win32
reported. "GTK OS suites" is otherwise green on every recent `main` commit (`37fcbdb37b`,
`c3f31eb276`, `b75fa8ed5f`, `ac42cb74a6` …), so this is not a pre-existing flake and I have
not re-run anything to make it go away.

## Fix

`packages/web/adwaita-web` had already answered this: `src/styles.generated.ts` — a generated
`.ts` that `src/` imports, the exact analogue — is **committed**, *and* its generator is
prepended to every script that reads `src/`. Both halves cover different holes, so this does
both: the artifact is committed, and `build:icons` now runs from `check`, `build:gjsify` and
`build:types`.

## A second defect the first one was hiding

With the generator on `build:gjsify` it now runs wherever this package builds — **including
that same Windows runner, which deliberately has no GTK toolchain** (the job asserts a
gvsbuild GTK is absent, since its whole subject is the shipped closure). `glib-compile-resources`
is not there. So the generator probes for it, keeps the committed artifact when it is missing,
and fails loudly only when there is neither compiler nor artifact.

The cost of committing is staleness, gated rather than hoped for: the generator records
`BUNDLED_ICON_NAMES` and `check-bundled-icon-parity.mjs` fails when that disagrees with
`SUBSET` — no toolchain needed to run the check. Broken on purpose (dropped `avatar-default`
from the recorded list): red, then reverted.

Verified from a cold tree: the exact CI command
`workspace @gjsify/react-native build:gjsify --with-dependencies` exits 0 with the generator
running inside it. Suite unchanged: 22 GJS + 7 GTK + 22 Node.

JumpLink:
Rebased on current `origin/main` (#1619, #1621, #1622 and #1620 all landed) and fixed the red
`Format (oxfmt --check)`.

## The format failure

It was `packages/framework/adwaita-app/src/icons.generated.ts` — which is precisely what
committing a generated artifact buys: it now has to satisfy the formatter like any other
committed source, and nothing about generating it knew that.

One offending construct: `JSON.stringify` of the name list emitted a single long
double-quoted line where oxfmt wants single quotes, one entry per line.

## Fixed in the generator, not by hand

Hand-formatting the artifact would have lasted exactly until the next `build:icons`. The
generator now emits the formatter's shape directly.

**Emitting that shape rather than shelling out to oxfmt**, for the reason
`glib-compile-resources` already taught this file: `build:icons` runs from `build:gjsify`,
which runs on the Windows leg of `gtk-os-suites.yml`, and a second binary on that step is a
second way to fail there. The shape is stable, and `oxfmt --check` in CI is the backstop if
it ever moves.

One knock-on: the staleness reader in `check-bundled-icon-parity.mjs` parsed that array as
JSON, so it **threw** the moment the shape changed. That is the honest failure — a reader
that had caught and returned `[]` would have reported all 41 names as missing and read like a
real finding — but it still had to learn the new shape, so it now reads the quoted names out
of the literal.

## Verified by regenerating, not by formatting

- fresh `build:icons` from scratch, then `oxfmt --check`: clean on the artifact and across all
  3804 files;
- regeneration is byte-identical (deterministic);
- staleness arm broken on purpose against the NEW shape, red, reverted;
- `check-bundled-icon-parity`, `check-adwaita-icon-masks`, `check-agent-context-size --check`
  all exit 0; `gjsify tsc --noEmit` clean; suite unchanged at 22 GJS + 7 GTK + 22 Node.

## Rebase conflict

One, in `.github/workflows/audit-runtimes.yml`: #1620 added its NativeScript icon-name gate
step exactly where this branch adds the GTK parity one. **Both kept**, in both copies of that
job — they are sibling pillars of the same question, not alternatives.

PR #1625, opened 2026-09-10
docs(website): one React, two element vocabularies

JumpLink:
One React, two element vocabularies. The gallery stops pretending they are two
frameworks, and the frameworks docs start saying which is which.

## The gallery window model

Three windows, unchanged in kind. The frameworks window loses one tab:

- **the widget** — the running component, alone, no tab bar
- **GJS** — `TypeScript` | `Blueprint`
- **UI frameworks** — `HTML Web Components` | `Solid` | `Vue` | `React` |
  `NativeScript`, plus the two refusal panes

The `react-native` slot was filled on 3 blocks of 40, all three on
`adwaita/layout.mdx`, against 28 blocks carrying a `react` snippet. Three
blocks on one page is a footnote holding a window pane.

`layout.mdx` now says once in prose what those three snippets showed: the
Adwaita element is the one the React tab already shows, the React Native
children go in its default slot, `configureStyle({ tokens })` is not optional,
and the result is a GTK window rather than a phone screen.

## What moved where

The `AdwWidget.astro` header passage about the tab's "on GTK" qualifier is not
deleted. It is split between the two places a reader meets it:

- **the direction argument** → `frameworks/react-native.mdx` § "What this is
  not", where the claim already lived. It now also records that the qualifier's
  original premise (no Adwaita widget set built out of React Native
  primitives) was written over by #1380 while the conclusion held.
- **the general shape** → `docs/code-anti-patterns.md` § "A rule whose premise
  died and whose conclusion did not". ADR 0034's stage 1 is the second
  instance, resolved the same way.

The component's header keeps a short record and points at both.

## The frameworks pages

`frameworks/react.mdx` leads with the choice. A comparison table, the
`Adw.Clamp` pair written both ways (the Adwaita element is byte-identical, the
children are not), and the decision as a question: bringing React Native or
Expo code, or starting fresh on GTK. It also says the two things a reader
otherwise has to infer: both paths point `jsxImportSource` at
`@gjsify/gtk-host/react`, so both reach the 168 generated tags; and neither
runs on a phone.

`frameworks/react-native.mdx` is reframed, not gutted. Title, description and
opening now say it is the path for code you already have. The primitives, the
lists, the routing, the styling, the prop table and the Adwaita rules all
stand. Its title carries the qualifier the tab label used to.

The packages stay separate, and the reason is measured: a `Gtk.Box` with no
orientation is horizontal and a React Native `View` is a column, so one name
over both would default wrong for half of what it renders, at exit 0.

`frameworks/index.mdx` stops implying a fourth adapter; `overview.mdx` stops
listing four peers where there are three plus a port.

## Two mechanisms, because this shape has now got in three times

**Arm 13 of `check-website-adwaita-gallery.mjs`** holds what a tab covers. A
tab is a fence 40 pages have to write, and arms 5 and 6 are each satisfied by
ONE block writing a slot, so a tab on 3 of 40 and a tab on all 40 are the same
shape to them. Three instances: the `nativescript` XML template as a slot on 4
of 40 until #1502, the hand-written attribute pane (110 of its attributes named
on their page, 54 not), and this one. Each had to be measured by hand first.
The coverage is now printed on every run, and a short tab fails unless
`PARTIAL_TAB_SLOTS` says why. The ledger is empty and a stale entry fails.

**`check-doc-fences.mjs` reads a fence attribute.** Its opener regex matched
```` ```<lang> ```` and nothing else, so a fence carrying an attribute matched
nothing: the dropped opener's closing fence became the next opener and every
fence after it on the page was off by one, with a real one swallowed as the
body of a phantom whose language no arm reads. Measured on
`getting-started.mdx`, the only page in the tree with one: 54 markers, 27
fences, 26 read. The swallowed fence there is `bash`, which no arm reads
either, so nothing was mischecked; a `ts` fence in that position would have
been. `fenceParseProblem` is the floor: markers against fences, per source.
The two comparison fences on the React page are its first consumer.

## Gates

Run before and after, all green: `check-website-adwaita-gallery`,
`check-generated-website-data`, `check-doc-fences`,
`check-website-attr-samples`, `check-nativescript-icon-names`,
`check-agent-context-size --check`, `gjsify format --check`, `oxlint scripts/`.
Site built and the three pages checked in a browser.

Arm 12's line is unmoved: `40 block(s) written twice — 7 are the SAME TEXT
after one normalisation … [vocabulary 6, glyph 0, property 22, composition 5]
— distance 357 line(s) over all pairs.`

`check-doc-fences` moves as expected: 6 → 5 className-bearing gallery fences
and 218 → 217 js/ts fences (three RN fences out of `layout.mdx`, two new ones
on `react.mdx`).

## Mutation tests

Eleven breaks, each confirmed red and reverted: arms 5, 6, 7, 9 and 10 of the
gallery gate, three on the new arm 13, two on the fence reader, and one on the
token arm. The table with what each one said is in a comment below.

## On the four reasons this PR was asked to preserve

All four hold, with one refinement worth knowing. The class-and-token style
system is `@gjsify/gtk-host/style`, framework-agnostic by construction, and
`styling.mdx` records that the only binding consuming it today is
`@gjsify/react-native`. So "the React path has no counterpart to that failure
mode" is true as of today rather than by design: give the host a `className`
binding and the empty-window-at-exit-0 failure reaches `gtk-host/react` too.
Nothing in this PR depends on that staying true.

### DIFF
--- a/docs/code-anti-patterns.md
+++ b/docs/code-anti-patterns.md
@@ -308,3 +308,41 @@ emitted CSS — a selector matching nothing, on a pane that still rendered, so t
 page looked unstyled rather than broken and nothing in the build said a word.
 Astro strips `:global()` only in a SCOPED style block, where it is the escape
 hatch; in a global one there is nothing to escape from.
+
+## A rule whose premise died and whose conclusion did not
+
+**Rule: when a recorded reason says "X, SO Y" and X stops being true, RE-DERIVE Y in
+writing. Deleting the reason keeps a rule nobody can defend; deleting the rule
+because its premise expired throws away a conclusion that may still hold on other
+grounds. Either way the next reader "simplifies" the rule back into the bug.**
+
+Measured on the website's Adwaita gallery. Its React Native pane was labelled
+"React Native on GTK" rather than "React Native", and the label's recorded reason
+was: *there is no Adwaita widget set built out of React Native primitives, SO
+nothing under this window runs on a phone*. The qualifier existed so a reader
+meeting that pane beside the NativeScript one would not take it for the port that
+runs on a device.
+
+#1380 landed `@gjsify/adwaita-react-native`, which is exactly such a widget set.
+The premise was gone, and the `so` says it was load-bearing as written, so nothing
+in the note carried the label any more. The conclusion survived anyway, on a
+different ground: what is at stake is what THAT PANE renders, and it renders
+`@gjsify/react-native`, which takes React Native's components to GTK 4 in-process.
+A package running the other direction, elsewhere in the tree, does not put a phone
+in a pane. Had the reason simply been deleted, the qualifier would have read as
+decoration at the next edit; had the qualifier been dropped with its premise, the
+pane would have claimed a device it cannot reach.
+
+The pane itself is gone now, for an unrelated and smaller reason: it was filled on
+3 of 40 blocks, all on one page, against 28 blocks carrying a `react` snippet, so it
+was a footnote holding a window pane. The claim outlived it, which is the point.
+`website/src/content/docs/adwaita/layout.mdx` says in prose what those three
+snippets showed, and
+`website/src/content/docs/frameworks/react-native.mdx` § "What this is not" carries
+the direction argument for readers who need it.
+
+ADR 0034's stage 1 is the same shape and was resolved the same way: the stage was
+ordered first because adopting the vocabulary clauses on
+`@gjsify/adwaita-react-native` was free before its first publish, the package then
+published at 0.44.0, and the row is left as written because deleting the premise
+deletes the ordering argument. Two instances make it a class, not a one-off.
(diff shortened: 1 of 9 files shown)


JumpLink:
Mutation tests for this PR. Each break was confirmed red, then reverted.

| arm | break | what it said |
|---|---|---|
| 5 | re-add a `react-native` fragment | provides a slot AdwWidget renders nowhere |
| 6 | re-add the tab with no page filling it | port declared in WINDOWS, no block provides it |
| 7 | empty the frameworks window's tabs and groups | window with no pane source at all |
| 9 | drop the `preview` tab only | mounts the fence and renders it on no window |
| 10 | unbold `**UI frameworks**` in `layout.mdx` | draws the window, prose never names it |
| 13 | delete one `blueprint` fragment | only 39 of 40 blocks write that fragment |
| 13 | ledger a slot that is on every block | stale exemption, now on all 40 |
| 13 | ledger `nativescript`, not a tab | arm 13 polices nothing for it |
| fences | restore the old opener regex | 54 markers, 26 fences read, 27 were there |
| fences | append a stray marker to `react.mdx` | 29 markers, 14 fences read |
| tokens | drop `configureStyle` from the new fence | styles with `flex-col gap-m p-m text-title`, nothing above calls it |

PR #1627, opened 2026-09-10
feat(adwaita-core): one authored tree, rendered

JumpLink:
Closes the rung ADR 0051 was written for: the same authored tree, built by more than
one renderer, held to the same `@gjsify/adwaita-core/conformance` rows, with no
per-surface markup branch. That is ADR 0027 § 9's criterion, and until now nothing
held it.

ADR 0051 is now Accepted, with one of its own premises overturned by measurement.

## What was missing

`scripts/adwaita-gallery-shared-trees.mjs` holds seven gallery blocks whose widget
tree is authored once. Arm 11 of `check-generated-website-data.mjs` compares them as
DATA. Nothing built them. Now two renderers do.

* `packages/web/adwaita-core/src/conformance/shared-trees.ts` is the renderer-free
  join: `sharedTreeExpectations(root)` returns the vector ROWS an authored node
  instantiates, by matching the authored value against the row's own input. It writes
  no expected value of its own — ADR 0051 § 3 made structural rather than promised.
* `packages/framework/gtk-host/src/shared-trees.spec.ts` builds each block through
  `createElement`/`setProp`/`insert` into real libadwaita widgets, diagnostics gate on.
* `packages/web/adwaita-web/src/shared-trees.spec.ts` builds the same block out of
  custom elements, in Firefox.
* `packages/web/adwaita-core/src/shared-trees.spec.ts` holds the corpus-and-ledger
  half on Node AND GJS, so neither renderer leg carries a claim about the corpus.

Both drivers assert, on the REAL tree their renderer produced, that the authored nodes
appear in the authored order — the libadwaita revealers and listboxes between them are
the renderer's business, the nesting is not — and then read the reached rows off them.
Each has exactly one seam, a `read(expectation, node)` over a closed observable
vocabulary; everything above it is shared. The per-surface transforms are `hostTagOf`
on both sides plus a camelCase-to-kebab ATTRIBUTE rule on the web, both total over the
corpus, with no per-block branch on either side.

The corpus is read, not transcribed. A hand-written
`scripts/adwaita-gallery-shared-trees.d.mts` lets a TypeScript spec import the `.mjs`
the two gallery generators emit from: `tsc` resolves the declaration and never puts the
`.mjs` in its program, so no package's `rootDir` is crossed, and the corpus stays where
plain-Node generators reach it in a CI job with no `node_modules`.

Distance, derived every run: 7 blocks, 5 reaching a row, 2 declared with a measured
reason, 4 tables, 6 rows.

## The premise that did not hold: stage 2 cannot exist

Stage 2 asked for a NativeScript tree driver "off-device against the port's own
classes, the way the port's specs already run on GJS and Node". Measured:

* every module under `packages/nativescript-bridge/adwaita/src/widgets/` opens with a
  bare `@nativescript/core` import at module scope, because each class EXTENDS an NS
  view;
* `@nativescript/core` is an OPTIONAL peer the workspace install never brings in —
  `node_modules/@nativescript` holds `types`, `types-android`, `types-ios` and nothing
  else;
* the port's own specs say so in their headers and act on it: "this file must NOT
  import `./widgets/adw-banner.js` … the behaviour is exercised through
  `./widgets/chrome.js`, the pure sibling". What runs off-device is the port's pure
  derivations, never its widgets.

So the port has no widget to build a tree out of in any runtime this repo tests. An
Android emulator was offered and deliberately NOT used: a driver that needs a device
cannot be the check that fails a PR, which is the only thing this rung is for. The
second driver is therefore `adwaita-web`, the renderer § 9 named before ADR 0051
re-aimed it. The ADR's "One correction to § 9's own wording" is withdrawn in
§ Amendment 1 — it is still true about the CORPUS, it was wrong about the DRIVER.

## What the drivers found that nothing else had

**A `GParamSpec` default is not a constructed default.** `AdwBanner:use-markup`
declares TRUE (`find_property('use-markup').get_default_value()`); a freshly
constructed `Adw.Banner` answers FALSE. Mechanism measured, not guessed: the banner's
getter DELEGATES to its template `GtkLabel`, `adw-banner.ui` (read out of the installed
GResource) never sets `use-markup` on that label, and a pspec default is not written
through a setter — assigning `label.useMarkup` directly moves what the banner reports.
Both Adwaita ports implement the pspec default, so the same authored banner interprets
`<b>bold</b>` in the browser and paints it literally in GTK. Ledgered in
`status/open-todos.md`, with why it is a rendering change of its own rather than a
drive-by.

**A localized rendering is a fact about the runner.** `SHORTCUT_LABEL_VECTORS` spells
`<Control>C` as `[Ctrl][C]`; `Adw.ShortcutLabel` draws `gtk_accelerator_get_label` —
measured `["Strg","C"]` on this de_DE host, `["Ctrl","C"]` under `LC_ALL=C`, and the
locale cannot be moved from inside the process. Hence `Adw.ShortcutLabel` is a declared
block, not a passing one.

**A notify table's `emitted` half has nowhere to come from.** An authored tree writes a
property; it cannot attach a listener before the write. Only the END STATE is taken.

## The gate learned the tree driver (stage 3)

`check-adwaita-conformance-drivers.mjs` gains a TREE arm, derived and never claimed. It
reads the join's table list out of its CODE, refuses a listed table no import backs,
refuses an imported table left off the list, refuses a driver spec no entry hands to
`run({…})`, and refuses a binding no live suite drives. The one half it cannot decide
statically — a listed table no corpus node REACHES — the drivers assert themselves
against `reachedTables`. `gtk-host` is on the tree-driver list and deliberately NOT
counted as a renderer, so a table only it drives stays core-only.

## Negative controls, runs, and what was left alone

In the first comment below — a markdown table does not survive as a commit body, and
this text becomes one.

## Open item deleted

`status/open-todos.md`'s "the shared source is compared as DATA rather than as a
rendered tree" is gone — not struck through, not marked done. What stands in its place
is what is actually still open: the corpus, the three structural limits, and the new
banner-markup item.

### DIFF
--- a/docs/adr/0051-one-authored-tree-rendered.md
+++ b/docs/adr/0051-one-authored-tree-rendered.md
@@ -1,6 +1,8 @@
 # 51. One authored tree, rendered: ADR 0027 § 9's criterion becomes a suite
 
-- Status: **Proposed**
+- Status: **Accepted** (2026-09-10) — amended, see § Amendment 1: the second driver is
+  `adwaita-web` and not the NativeScript port, because the port has no widget an
+  off-device suite can build. What was actually built is § What landed.
 - Date: 2026-09-09
 - Deciders: Pascal Garber
 - Related: [ADR 0004 (headless Adwaita core)](0004-headless-adwaita-core.md), [ADR 0027 (GTK host layer)](0027-gtk-host-layer.md), [ADR 0028 (widget table provenance)](0028-widget-table-provenance.md), [ADR 0030 (one corpus, GJS as oracle)](0030-one-corpus-gjs-as-oracle.md), [ADR 0034 (widget vocabulary convergence)](0034-widget-vocabulary-convergence.md)
@@ -216,7 +218,7 @@ Each stage is independently useful and breaks nothing that ships.
 | # | stage | what goes red if it is wrong |
 |---|---|---|
 | 1 | A tree driver for `gtk-host`: build `gtkHostTree(widget)` for each corpus block into real widgets, with `installDiagnosticsGate()` on, and assert the `adwaita-core` vectors the tree reaches. | a block that cannot be built; a GTK diagnostic during a build; a vector that disagrees; a block reaching no vector and not declared |
-| 2 | The same for the NativeScript port over `nativeScriptTree(widget)`, off-device against the port's own classes, the way the port's specs already run on GJS and Node. | the same four, plus a block whose two drivers disagree — which is the finding the whole ADR exists to produce |
+| 2 | ~~The same for the NativeScript port over `nativeScriptTree(widget)`~~ — **withdrawn, see § Amendment 1.** The port's widget classes extend `@nativescript/core` bases that no runtime here has, so there is no off-device tree to build. The second driver is `adwaita-web`, over the same corpus. | the same four, plus a block whose two drivers disagree — which is the finding the whole ADR exists to produce |
 | 3 | Teach `check-adwaita-conformance-drivers.mjs` the tree driver, so a table driven only from a tree is not read as undriven, and a tree claiming a table it does not reach fails. | a false coverage claim — the exact class that gate's three incidents are about |
 | 4 | Print the distance: blocks in the corpus, blocks reaching a vector, blocks declared, per renderer. Derived every run; no count in a header or in prose. | any figure that is written down rather than derived |
 | 5 | Put `adwaita-web` on the corpus: emit the gallery `preview` fence from `ADWAITA_GALLERY_SHARED_TREES` instead of authoring it per block, then drive it in `tests/browser`. This is the stage that closes § 9 in its own words. | a preview fence that is not what the shared tree emits; a web tree that fails a vector its two siblings pass |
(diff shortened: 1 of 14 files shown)


JumpLink:
## Negative controls — every one run, every one red

A guard that has never been red is not a guard.

| what was broken | result |
|---|---|
| `ENTRY_TEXT_LENGTH_VECTORS` row `'Ada Lovelace' → 12` changed to `13` | **both** drivers red, and the tree failure is named separately from the per-widget one: gtk-host `1 of 624 tests failed`; browser `3 of 1672`, one of them `[the shared corpus against the adwaita-core vectors it reaches] Adw.EntryRow — … ENTRY_TEXT_LENGTH_VECTORS` |
| `'active'` removed from `<adw-switch-row>`'s `observedAttributes` | 5 browser tests red, **none of them the tree driver** — the declarative path still worked. Kept as a finding: it is why the next control exists. |
| the DECLARATIVE read `this._state.setActive(this.hasAttribute('active'))` in `<adw-switch-row>`'s `connectedCallback` replaced by `false` | **exactly 2 tests red, both of them the tree driver's**, with all 1 670 other browser tests green. The existing per-widget suite drives that property imperatively and cannot see it. This is the class the tree driver adds. |
| `AdwSharedTreesTest` removed from `run({…})` in `test.browser.mts` | gate red: *"drives the shared authored tree, but no test entry of adwaita-web hands its suite to `run({…})` — so it runs NOWHERE"* |
| `'WINDOW_TITLE_VECTORS'` added to `SHARED_TREE_TABLES` | gate red: *"lists WINDOW_TITLE_VECTORS as tree-driven while nothing in the file reads it"* |
| a stale `'Adw.Banner'` entry added to `SHARED_TREE_BLOCKS_WITHOUT_EXPECTATIONS` | adwaita-core spec red on Node: *"declares a block that reaches nothing, and only such a block"* |

The gate also caught one of my own mistakes while it was being written: the ledger export was
originally named `…_WITHOUT_VECTORS`, which `tablesIn`'s `export const *_VECTORS` scan read as a
vector table. Renamed.

## Runs

```
gjsify workspace @gjsify/adwaita-core run check           → 0
gjsify workspace @gjsify/adwaita-core run test:node       → 2114 tests, 13093 assertions, 0 failed
gjsify workspace @gjsify/adwaita-core run test:gjs        → 2114 tests, 13093 assertions, 0 failed
gjsify workspace @gjsify/gtk-host run check               → 0
gjsify workspace @gjsify/gtk-host run test:gjs            → 621 tests, 2872 assertions, 0 failed
gjsify workspace @gjsify/adwaita-web run check            → 0
cd tests/browser && npx playwright test --project=firefox → 5 passed (adwaita-web bundle: 1672 tests)
node scripts/check-adwaita-conformance-drivers.mjs        → 0 (176 tables, 82 live specs,
                                                             2 live tree drivers, 4 joined tables)
node scripts/check-generated-website-data.mjs             → 0 (24 blocks both renderers draw,
                                                             7 from one authored tree)
node scripts/check-node-test-registration.mjs             → 0
node scripts/check-browser-test-registration.mjs          → 0
node scripts/check-vocabulary-alignment.mjs               → 0
node scripts/check-adr-index.mjs                          → 0
node scripts/check-agent-context-size.mjs --check         → 0
gjsify lint                                               → 0 errors
gjsify format --check                                     → clean
```

(The gtk-host count is 621 after the last commit lifted three corpus-level assertions out of
the driver and into `adwaita-core`'s suite, where they run on Node too; it was 624 before.)

**Not run here, and why:** `gjsify workspace @gjsify/gtk-host run test:gjs-on-node` fails to
resolve `@gjsify/node-gi` in this checkout — a `file:` dependency on a package that is not a
workspace member and is neither built nor linked here (`ERR_MODULE_NOT_FOUND` at the bundle's
first import, before any of this branch's code). That leg is `node-gi.yml`'s `gtk-host-node`
job. No e2e suite touches the changed paths.

## Deliberately not touched

- **No AGENTS.md.** The rules this adds live in the ADR and in the two module headers a reader
  of them lands on; touching a context file means `check-agent-context-size --update` and
  committing `status/agent-context-budget.json`, a counter two other branches share right now,
  for content that already has an owner.
- **The corpus did not grow.** ADR 0051's own "Grow the corpus before building the suite"
  rejection stands: the ledgered divergences each need a decision before they need code.
- **`ADWAITA_GALLERY_TREE_DIVERGENCES` and the two gallery generators are unchanged** — this
  PR adds a reader of the corpus, not a second author of it.
- **Stage 5** (emitting the gallery `preview` fence from the shared trees) stays open;
  Amendment 1 moves `adwaita-web` into the driver position, not the website's authored fences.

JumpLink:
## Independent review — every load-bearing claim re-run, seven commits pushed

I did not write this PR. Everything below was re-measured in a clean worktree on
`feat/one-authored-tree-rendered`; where I could not reproduce a number I say so, and where
I disagreed I pushed a change rather than a note.

**Verdict: mergeable.** The claim the PR rests on — that the tree driver sees a class of
defect no per-widget suite sees — reproduces exactly, not approximately. Two of the three
things I was most suspicious of survived measurement and came out *stronger* than the PR
argued them. Two did not, and are fixed on the branch.

---

### 1. The load-bearing negative control — reproduced exactly

The decisive one, run myself end to end:

```
sed -i "71s/setActive(this.hasAttribute('active'))/setActive(false)/" \
    packages/web/adwaita-web/src/elements/adw-switch-row.ts
gjsify workspace @gjsify/adwaita-web run build:test:browser
cd tests/browser && npx playwright test --project=firefox
```
```
Error: 2 of 1669 tests failed:
  [the shared corpus against the adwaita-core vectors it reaches] Adw.PreferencesGroup — AdwPreferencesGroup > AdwSwitchRow[1]: SWITCH_ROW_NOTIFY_VECTORS …
  [the shared corpus against the adwaita-core vectors it reaches] Adw.SwitchRow — AdwPreferencesGroup > AdwSwitchRow[0]: SWITCH_ROW_NOTIFY_VECTORS …
```

Exactly two red, both of them the tree driver's, everything else green. And the control
that motivates it, `'active'` off `observedAttributes`:

```
Error: 5 of 1669 tests failed:
  [<adw-switch-row> still toggles and emits through <gtk-switch>] …
  [<adw-switch-row> notify::active (libadwaita conformance vectors)] programmatic-set-notifies …
  … 3 more, every one of them a per-widget test, not one of them the tree driver's
```

Five red, **none** the tree driver's — as claimed. The two controls together are the actual
argument for this rung, and they hold: the declarative construction path is a real class,
and the tree driver is the only thing in the repo that sees it.

The vector control on the GTK side too, `ENTRY_TEXT_LENGTH_VECTORS` `12 → 13`:
`❌ [Gjs] 1 of 621 tests failed`, named `Adw.EntryRow — AdwPreferencesGroup > AdwEntryRow[0]:
ENTRY_TEXT_LENGTH_VECTORS`. And the gate arm, `'WINDOW_TITLE_VECTORS'` added to
`SHARED_TREE_TABLES`: exit 1, *"lists WINDOW_TITLE_VECTORS as tree-driven while nothing in
the file reads it"*.

**One number is off:** the body says 1 672 browser tests and 1 670 green; the suite reports
**1 669**. Three stale — harmless, but it had also been copied into `docs/release-notes/next.md`,
where it would have gone staler by release. Dropped (`1ff25fa6f5`).

### 2. The premise that took scope AWAY — holds, and is weaker than it should be

This is where a withdrawn task justifies itself too easily, so I checked the three
sub-claims separately rather than the conclusion.

- `node_modules/@nativescript` → `types`, `types-android`, `types-ios`. **True.**
  `peerDependenciesMeta` marks `@nativescript/core` optional, no devDependency. **True.**
- 68 of the modules under `adwaita/src/widgets/` import `@nativescript/core` at module
  scope. The two widget modules that do not (`adw-view-switcher`, `adw-inline-view-switcher`)
  extend `view-switcher-base.ts`, which does — and whose header already says *"a widget
  module cannot be imported off-device"*. **True.**
- Six spec headers carry the *"must NOT import `./widgets/…`"* clause. **True.**

But the *argument* was the wrong one, and I think that matters: "the workspace install does
not bring it in" is a **state of this checkout**. Somebody adds the devDependency and the
argument evaporates while the conclusion stays right — which is how a correct decision gets
re-litigated on a bad premise.

So I measured what actually blocks it:

```
npm pack @nativescript/core          # 9.1.1, into a scratch dir, lockfile untouched
node -e "import('./package/ui/label/index.js')"       → Cannot find module
ls package/ui/label/                 → index.android.js  index.ios.js  (no index.js)
ls package/ui/core/view/             → index.android.js  index.ios.js  (no index.js)
ls package/ui/layouts/grid-layout/   → index.android.js  index.ios.js  (no index.js)
```

`@nativescript/core` **ships no platform-neutral module for a widget class at all**. Picking
between the flavours is NativeScript's own platform-aware resolution — not Node's, not a
bundler's. `class AdwBanner extends GridLayout` has no base class to extend off-device
whether or not the peer is installed. So the answer to "impossible, or merely expensive?" is
**impossible short of building a fake device**, and Amendment 1 now says that
(`c0d490517b`), with `status/open-todos.md` pointing at it.

### 3. The two ledgered findings — one exception holds, one entry was wrong

**Banner `use-markup` — the exception holds.** Re-measured, not taken on trust:

```
pspec default use-markup     : true          libadwaita 1.9.3
constructed get_use_markup() : false
template label found         : Gtk_Label  use-markup=false
after label.use_markup=true  : banner.get_use_markup()=true     ← the getter delegates
```

Every word of it, including the delegation mechanism. And `refs/libadwaita` really is
uninitialised here (`git submodule status` → leading `-`), so the `adw-banner.c` line cites
in `BANNER_DEFAULT_VECTORS`'s `rule` genuinely cannot be checked in this tree. Fixing it
repaints two published renderers and rewrites a conformance row — governance exception (c),
and the PR does land the minimal root part of it: the drivers take **no** pspec-default table
at all rather than picking whichever of the two true facts happens to be green. That is the
right shape.

**`Adw.ShortcutLabel` — the entry was wrong, and I rewrote it (`72003229e0`).** Three
things, in order of how much they matter:

1. The declared reason blamed the locale. The locale is not why the block reaches nothing —
   `SHORTCUT_LABEL_VECTORS` simply is not on `SHARED_TREE_TABLES`. That is a **choice**, and
   the entry should say so; `Adw.WindowTitle` reaches nothing for a genuinely different
   reason (no row matches the authored values) and one wording covered both.
2. **What the silence actually covers was not stated.** `adwaita-web` drives that table from
   its own spec and would pass the row here too. `gtk-host` drives it from **no spec at all** —
   so libadwaita's own keycap rendering is the one thing this declaration leaves unchecked,
   and the entry read as coverage-with-a-reason instead.
3. On the tooling question you'd expect me to answer: **`it.failing(…, { when })` is the
   honester tool, and I can show it rather than assert it.** The locale is immovable from
   *inside* the process — confirmed, `GLib.setenv` does nothing — but not from outside:

   ```
   gjs -m probe.js              → ShortcutLabel keycaps : ["Strg","C"]
   LC_ALL=C gjs -m probe.js     → ShortcutLabel keycaps : ["Ctrl","C"]
   ```

   Per `tests/AGENTS.md` rule 6, `when` scopes the **expectation, not the test**: on a de_DE
   desk a tolerated failure, on an untranslated host — which no workflow here opts out of —
   a real, gating assertion. A declared block asserts nothing anywhere.

   I did **not** implement it. It needs a fifth joined table, a new observable and a GTK-side
   keycap serializer, and only one corpus row would exercise it — a serializer correct for
   exactly one input is a landmine. That is a follow-up, not a review edit. The declaration
   now names it so it does not get lost.

### 4. Craft — what I changed

- **`refactor: lift the tree drivers' shared walk` (`d9867b3f54`).** Both drivers derived
  the same two things privately — the authored classes in authored order, and the index of
  the node an expectation addresses. The second copy is where that gets lifted:
  `authoredTags(root, tagOf)` and `subjectIndexOf(root, path)` now live with the join, so the
  walk a driver addresses into is the same one `shared-trees.spec.ts` proves unique and total
  on Node. The PR claims "each driver has exactly ONE seam"; before this there were three
  shared derivations, two of them copied. The gtk-host driver also stops carrying three
  parallel arrays indexed in step (four fewer non-null assertions), and `widgetOf` reads the
  widget `build` already materialised instead of calling `materialize` a second time.
- **`test(adwaita-web): unmount on the red path too` (`63876f7dde`).** `mount()` handed back
  a host the test removed on its **last line** — so a failing assertion left the block in
  `document.body` and the next test read a document two blocks deep. A suite whose entire
  selling point is an attributable failure cannot have its first red corrupt its second.
  `mounted(node, use)` puts the teardown in a `finally`. Also `querySelectorAll('*')` for the
  hand-rolled recursion, and the double `bannerButton()` call with its `!` collapsed to one.
- **`chore: declare only what a spec imports` (`20e9309084`).** Nothing checks
  `adwaita-gallery-shared-trees.d.mts` against the `.mjs`, so the *only* backstop a declared
  name has is a spec importing it. `nativeScriptTree` and `ADWAITA_GALLERY_TREE_DIVERGENCES`
  had none — their consumers are plain `.mjs` generators that never see the file — so they
  were exactly the unverifiable second truth the file's own header warns about. Removed, and
  the header now states the invariant instead of implying it. **The `.d.mts` trick itself is
  sound**: `skipLibCheck` and `rootDir: src` both hold, and every remaining name is imported
  by a live spec, so a rename in the `.mjs` is now a build error.
- **`docs: derive the denominator, do not write it` (`9537e8f91d`).** `status/open-todos.md`
  said the denominator *"is printed by the drivers rather than written here"* — and then wrote
  it, in the same sentence: seven blocks, four tables, six rows. Stage 4's own failure
  criterion is "any figure that is written down rather than derived".

### 5. `simplicity` applied to the new gate arm — one honest caveat

Governance asks what a new check lets you delete. This one deletes nothing, so I measured
whether it currently *does* anything. Neutering only the credit half:

```
-  byLabel.get(label).add(name); byRenderer.add(name);
node scripts/check-adwaita-conformance-drivers.mjs   → exit 0, byte-identical summary
```

All four joined tables are already driven by live per-widget specs, so **the credit half of
the TREE arm changes no verdict today**. It is not dead — it becomes load-bearing the moment
a table is joined that no per-widget spec drives, which is the blindness it exists for — but
"this arm removes a blindness" is currently a claim about the future, and worth knowing.
The *refusal* half is genuinely live: three of its five refusals were negative-controlled by
the author and I reproduced one of them. I left the arm alone.

### Runs, on the final state of the branch

```
gjsify workspace @gjsify/adwaita-core run test         → 2114 tests / 13093 assertions, Node AND GJS, 0 failed
gjsify workspace @gjsify/gtk-host   run test           → 621 tests / 2872 assertions, 0 failed
tests/browser, playwright --project=firefox            → 5 passed
gjsify workspace {adwaita-core,gtk-host,adwaita-web} run check → 0, 0, 0
check-adwaita-conformance-drivers / -generated-website-data / -adr-index → 0 / 0 / 0
check-node-test-registration / -browser-test-registration / -vocabulary-alignment → 0 / 0 / 0
check-agent-context-size --check  →  0     audit-runtimes.mjs --check → 0
gjsify lint → 0 errors            gjsify format --check → clean
```

Counts unchanged by the refactor in both directions, and I **re-ran the negative control
against the refactored driver** — a refactor that quietly disarms the guard is the failure
mode this repo pays for most: still `2 of 1669`, still both the tree driver's.

No AGENTS.md touched, so `status/agent-context-budget.json` is untouched. Corpus not grown,
stage 5 not pulled in.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

PR #1628, opened 2026-09-10
feat(gtk-host): a toplevel placement beside the portal

JumpLink:
## The class, not the case

`ADR 0045` closed the loudest failure this repository has collected:
`box.append(dialog)` on a box that sits in a window is `g_error()` — SIGABRT, exit
134, a core dump, past every `try`, past `installDiagnosticsGate()`, past the test
runner. It closed it by declaring what an `Adw.Dialog` is: a node presented *against*
its parent, never into it.

The same probe measured the case next door and the result looked harmless. Appending a
`Gtk.Window` to that same rooted box is exit 0 with nothing logged — but afterwards
`win.get_parent()` is the box and `win.get_root()` is the *window itself*: a `GtkRoot`
with a parent, drawn as a toplevel and simultaneously measured and allocated by a
container. Present it and it stays in the child list. One kind shouts, its neighbour
says nothing, and only the shouting one had an answer.

Both have one now (**ADR 0054**). The measurement tables are in
[a comment](https://github.com/gjsify/gjsify/pull/1628#issuecomment-5617763155),
because a squash body is a commit message and tables do not wrap.

## What landed

**A third `NodePlacement` kind, `toplevel`** — `present()` to show it, `destroy()` to
take it down, declared by the eighteen `Gtk.Root` classes in the table that can present
themselves. Not a copy of the portal arm. The arity is what keeps them apart:
`adw_dialog_present` takes a parent and `gtk_window_present` takes nothing, so a portal
has two positions in the tree and a toplevel has one. `destroy` and not `close` because
a `close-request` handler returning true leaves `close()`'s window mapped, and an
unmount is not a user request — the same reasoning that picked `force_close`. There is
no lazy present and no `notify::root` subscription: there is no second position to wait
for. An authored `visible: false` is honoured rather than reversed.

`AdwMessageDialog` is a toplevel and `AdwAlertDialog` is a portal. Same axis, opposite
sides.

**The host asks the CLASS, never a name.** A gtype list in `policies.ts` is the widget
knowledge ADR 0027 rule 1 forbids, and it would be wrong for a consumer's subclass
anyway — so `classPlacementKind` asks `GObject.type_is_a(…, Gtk.Root)` for a toplevel
and, for a portal, whether `present()` takes an argument. Measured over the shipped
table, that second question answers exactly the five `Adw.Dialog` descendants, with the
symmetric difference empty in both directions. `Gtk.Popover.present()` is the
discriminator that puts the ARITY in the question: it exists, it takes none, and a
popover is an ordinary child.

**Two mechanisms cover insertion between them.** `descriptorProblems()` reports a class
that demands a placement and declares none, over the whole table and over a consumer's
descriptors. `refuseUnparentable` refuses such a child at the insert, before any adder,
with a named catchable `GtkHostError` — which for the portal arm is the only report
that can exist, because after the append there is no process left to report from. It
fires only where `placement` is absent, so an explicit `{ kind: 'parented' }` is the
way back. `GtkDragIcon` is the nineteenth `Gtk.Root` and carries no declaration: it has
no `present`, no `close`, no `destroy`, so the refusal is its whole answer and the spec
names it rather than counting it.

**The four sibling walks ask `isUnparented`, not `isPortal`.** ADR 0045 § 5 listed each
as a real defect if it did not skip a portal; each is the same defect for a toplevel.
`@gjsify/react-native`'s `Modal` needed no change — the seam is below every adapter.

## Reversibility: what `remove` means, and what `destroy` means

Found by an independent review of this branch, whose reproducers are vectors here.

`remove` states its own contract — *"Detach only — reversible. Frameworks move nodes;
`remove` must not destroy one"* — and the declared close for the new arm is `destroy()`,
which is terminal. Nothing reconciled the two, so a move destroyed the window and then
presented the corpse: `A window is shown after it has been destroyed`, at exit 0, caught
only because the diagnostics gate was watching. Solid's `removeNode` calls `remove(node)`
directly, so that is the ordinary reorder path.

One cause: the two arms differ in reversibility and one function treated them alike.
`force_close()` followed by `present(parent)` re-hosts a dialog against the same parent
with no diagnostic; `destroy()` followed by `present()` does not. So the axis has two
verbs. `detachOutsideParent` is what `remove` means, `closeOutsideParent` is what a
DISCARD means, and for a portal both are `force_close` — which is why that arm never
showed the defect.

A toplevel detaches with `set_visible(false)`: measured, that leaves it invisible with
one `unmap`, emits no `close-request`, and keeps it in `Gtk.Window.list_toplevels()`,
while `destroy()` drops it from that list. The list is what makes one a detach and the
other a teardown. It is a property write rather than a declared method because a
window's presence on screen IS its `visible` property — which is why `presentToplevel`
already reads it — so a per-row name would be the same string on every declared row.

**What wants the terminal verb is a DISCARD, not a particular caller.** Every caller of
`removeChild` keeps the widget; the question that picks the verb is which code drops
`el.widget`, and that is `materialize`'s rollback, `rebuild` and `destroy`. They share
one `releaseWidget`, which closes first. That matters because a `Gtk.Window` is in
`list_toplevels()` from construction, before any `present()`, and dropping the JS
reference does not remove it — so a half-built `<gtk-dialog>` whose replay was rejected
leaked a window nothing could reach.

**One retraction per operation, on both arms.** `remove` runs the detach once, a discard
runs the close once, and neither runs the other's. That needs a teardown NOT to detach
first: on the portal arm both verbs name the same method, so `destroy` was calling
`force_close` twice. Benign — it is silent on a node that is not presented and drops no
GTK reference — which is exactly why nothing could see it, and why the guarding vector
existing on one arm only was the real defect. Ordering rather than a predicate:
`destroy` releases the widget before it removes, and `rebuild` skips the `removeChild`
that would do nothing else for a non-parented node. A `detachIsClose(placement)` was the
alternative and is a second source for what the two dispatch switches already say.

The identity argument is why `insert` does not rebuild a retracted toplevel instead: a
rebuild hands back a different window after a move, and `solid.spec.ts`'s `<For> keeps
widget identity across a reorder` says it must not — Solid being the framework that
would have paid for it.

For a consumer: unmounting a `<gtk-window>` in Solid takes it off screen at `removeNode`
and destroys it when the root disposer runs `destroyChildren`. The teardown still
happens, one step later, and a reorder no longer costs the window. React and Vue are
unaffected.

## Verification: an abort has to be SEEN to be denied

A test asserting "this no longer aborts" is worthless in the process it asserts in.
SIGABRT takes the runner down, so a suite that merely FINISHES proves it is alive: a
harness that could never observe an abort reports the same green either way.

Three vectors run in a child process and read its exit signal. The negative control
asserts the raw append is killed by signal 6 with `Adwaita-ERROR` on stderr and nothing
on stdout; while it passes, every in-process "does not abort" vector in the file is
falsifiable. They are `it.failing(…, { when: GJS === null })`, never skipped: the suite
is one program across `test:gjs` and `test:gjs-on-node` (ADR 0030) and a node host need
not carry a gjs binary, so they run wherever one exists and go red the day such a host
stops reproducing the abort.

The abort is deliberate; its 2.8 MB core dump is not. The child is spawned under a
`prlimit --core=0` argv prefix, probed with `GLib.find_program_in_path` rather than on
`process.platform`, so a host without it spawns the identical argv and asserts the
identical things.

## What the node leg found

Seven vectors green on gjs, red on node with `TypeError: invalid GObject handle`. Six
were ours: `destroy()` carried its own `widget.destroy()` for "the one node unparenting
cannot reach", written before the placement axis had a name for that node, so a declared
toplevel with a parent was retracted twice. gjs swallows the second call because its
wrapper keeps the object alive; node-gi does not, because `gtk_window_destroy()` drops
GTK's reference. The seventh was the vector itself, reading a property back off a
destroyed window; it is written against the effect now. The divergence is a ledger
entry — whether a JS variable naming a GObject keeps it alive is node-gi's reference
model, not a call site.

## Part B — the measurement, and why nothing was curated

The ledger entry says curation must be "driven by a real window that needs one, with its
vector, not by walking the list alphabetically". So the demand was measured across every
tree that renders through the host, both tag spellings, and the reviewer re-ran it
independently against the packages that actually declare a dependency on
`@gjsify/gtk-host`, reaching the same answer.

No uncurated widget is given a materialised child anywhere. Nineteen are used, every one
as a leaf, and a leaf needs no policy. Seven are attempted as a parent and refused on
purpose in `showcases/gtk/adwaita-gallery-solid/src/refusals.ts`. Everything that wraps
children routes the wrap through an already-curated container.

Two trees that look like consumers are not: `packages/nativescript-bridge` writes its own
NativeScript `View` subclasses, and the `html` preview fences under
`website/src/content/docs/adwaita/` are `@gjsify/adwaita-web` custom elements — confirmed
by that package depending on `@gjsify/adwaita-core` and not on this host.

So nothing was curated, and that is the rule being followed rather than skipped. The
probed refusals are the only measured demand, and the GIR says what each would need;
that is now in `status/open-todos.md` in place of a count, which is what makes the next
curation steerable. Curating any of them is a gallery change with an authored tree,
because `check-generated-website-data.mjs` arm 5b holds the refusal list against the
descriptor directory in both directions.

## Not touched, on purpose

No uncurated widget got a child policy — the measurement says nothing needs one, and the
ones that would are a gallery change. `GtkDragIcon` carries no placement. The
`AdwDialog` subclasses keep `children: { kind: 'uncurated' }`; ADR 0045's reasoning is
unchanged, and `AdwPreferencesDialog.add()` would curate cleanly but is ledgered as
gallery work rather than smuggled in here. node-gi's reference model is measured and
ledgered, not patched. `TOPLEVEL` still lives in `descriptors/gtk.ts` and is imported by
`adw.ts`; a neutral home would be tidier, but the de-duplication is the part that
matters.

A new ADR needs a row in `docs/adr/README.md`, or `check-adr-index.mjs` reds
`Detect runtime-triplet drift`, one of the three required checks. That is what made this
branch red before the first review, and it is worth knowing before the next ADR.


JumpLink:
## Independent review — reproduced the measurements, found two defects in the toplevel retraction

I re-ran every claim in the description rather than reading it, and pushed four small
changes to this branch. Two of the three headline measurements reproduce exactly. One
new defect class does not, and it is why I would not merge this yet.

### Both red checks were the same PR, and one of them was REQUIRED

The description does not mention them. `Manifest checks (Windows)` was red on
`check-adr-index.mjs`, and so was **`Detect runtime-triplet drift` — one of the three
branch-protection checks**, which runs the same script on Linux. The cause is this PR:
`docs/adr/0054-…md` was added with no row in `docs/adr/README.md`.

```
check-adr-index: docs/adr/0054-toplevel-placement-and-the-unparentable-refusal.md
has no row in docs/adr/README.md.
```

Fixed in `docs(adr): index ADR 0054`. `node scripts/check-adr-index.mjs` →
`53 ADR(s) indexed, statuses agree.`

`Lint commit messages` was red on `subject-case`, from a subject opening with
`ADR 0054,`. Reworded to `docs(gtk-host): record ADR 0054 and the sweep` (45 chars).
`npx commitlint --from origin/main --to HEAD` over all six commits → **0 problems**.
One `footer-leading-blank` *warning* remains on
`fix(gtk-host): give a toplevel retraction exactly one owner`; the action runs with
`failOnWarnings: false`, so it does not gate.

Picking 0054 over the free 0053 is right, not a slip: `docs(adr): ADR 0053, a declared
media contract` already exists on `fix/bundle-closure-fonts-and-audio`.

### Verified: the structural oracle is exact

Swept independently over the shipped table on this machine (GTK 4.22.4 / libadwaita
1.9.3), reading `BUILTIN_DESCRIPTORS` out of the built lib:

```
PROVENANCE curated=62 generated=169 total=169
ROOTS 19: GtkWindow(present=0) … GtkPrintUnixDialog(present=0), GtkDragIcon(present=-1)
PRESENT_ARITY_1 5: AdwDialog, AdwAlertDialog, AdwAboutDialog, AdwPreferencesDialog, AdwShortcutsDialog
ADW_DIALOG_DESCENDANTS 5: (identical set)
ARITY1_NOT_DIALOG: (none)      DIALOG_NOT_ARITY1: (none)
GtkPopover.present arity = 0
DECLARED_toplevel 18   DECLARED_portal 5
ZERO_ARITY_PRESENT 21: 18 roots + GtkEmojiChooser, GtkPopover, GtkPopoverMenu
```

Symmetric difference empty in both directions, so "exactly the `Adw.Dialog`
descendants" is a partition and not an observation. Measurements H, I and J in the ADR
all hold. `GtkSvgWidget` is the one row whose `ctor()` answers `undefined` on this
typelib; it is not a root, and both the oracle and the spec skip it correctly.

**The row count in the comment was wrong, though**: `classPlacementKind`'s docblock and
ADR 0054 § 3 both said "all 164 rows". `tableProvenance()` answers **169**, which is
what this PR's own description says — so the live count had drifted inside the commit
that wrote it. That is [the anti-pattern](docs/code-anti-patterns.md) verbatim, and it
is the reason the rule says to write what a number establishes rather than the number.
`refactor(gtk-host): drop counts specs already hold` takes the cardinalities out of
`policies.ts`, `types.ts` and `descriptors/gtk.ts` — `placement.spec.ts` asserts them
and fails when they change; prose does not — and corrects the ADR's own to 169, since a
dated record of a measurement is where a number belongs.

The same docblock also carried a claim its next clause contradicted: *"every `Gtk.Root`
in the table **also** has a 0-argument `present`, and `GtkDragIcon` has no `present` at
all"*. The ADR states it correctly (every root **that has one**); the code now matches.

### Verified: the SIGABRT negative control is real

Ran the raw case standalone, outside the suite:

```
$ gjs -m abort.js
(gjs:1647738): Adwaita-ERROR **: Trying to add AdwDialog 0x… to GtkBox 0x…. Use adw_dialog_present() to show dialogs.
EXIT=134
$ gjs -m toplevel.js
parent=true rootIsSelf=true
EXIT=0
```

134 = 128 + SIGABRT. The control genuinely goes red if the abort stops, and the quiet
half is genuinely quiet. `it.failing(…, { when: GJS === null })` is the right tool here
and hides nothing: `when: false` delegates to a plain `it()` (`unit/src/index.ts`
:1430), so on a host with gjs these three RUN and must PASS — which they do.

**One cost worth a decision:** every run of this suite writes a 2.8 MB `gjs-console`
SIGABRT coredump. `/var/lib/systemd/coredump` on this workstation is at 3.4 GB, and CI
pays it on every run too. `ulimit -c 0` suppresses the *file* and not the *signal* —
verified, `coredumpctl` still lists the abort with `COREFILE: none`, and the control
still passes — so the suite could set it itself without weakening anything. A
`prlimit --core=0` argv prefix in `runInChild` on Linux would do it with no shell.

### Verified: Part B, nothing needed curating

I did my own sweep rather than trusting the list. Restricting to the trees that
actually declare a dependency on `@gjsify/gtk-host` (the ten `showcases/gtk/*`,
`packages/framework/react-native`, `packages/framework/adwaita-react-native`) plus the
gallery tree generators, with comments and string literals blanked and both tag
spellings:

```
files: 612
=== UNCURATED GIVEN A CHILD, in a real gtk-host consumer: 0 ===
=== UNCURATED USED AS A LEAF: 16 distinct ===
```

Zero. The remaining leaves (`AdwViewSwitcherBar`, `GtkListView`, `GtkPicture`,
`GtkTextView`) reach the table through `createElement(tag, props)` with **no third
argument** — I checked all four call sites in `react-native/src/{lists,router}`. So the
19 stands and **not curating is the right answer.**

**The boundary you draw is correct, and I checked it rather than accepting it.**
`packages/web/adwaita-web/package.json` depends on `@gjsify/adwaita-core` and **not** on
`@gjsify/gtk-host`, so the `html` preview fences under `website/src/content/docs/adwaita/`
really are a different vocabulary. My wide sweep found parent usage of
`adw-sidebar`, `adw-carousel`, `adw-tab-view`, `adw-toggle-group`, `adw-bottom-sheet`
and `adw-view-switcher` — all of it in those fences and in `tests/browser/specs/`, none
of it through this host. `packages/nativescript-bridge` likewise does not depend on it.
No real consumer slipped out of the measurement.

Two small notes: `refusals.ts` probes **seven** uncurated parents, not six —
`adw-preferences-dialog` is uncurated too, and is discussed separately under "Not
touched". And the Blueprint trees (`adw-blueprint-layout`, `effect-adw-services`) are
outside the sweep by construction, since GtkBuilder never reaches `insert()`; their only
uncurated tag is a childless `Gtk.MenuButton`, so nothing changes.

---

## Blocking: `remove()` is documented as reversible and is terminal for this arm

`remove` says *"Detach only — reversible. Frameworks move nodes; `remove` must not
destroy one."* For the toplevel arm the declared retraction **is** `destroy()`, and
ADR 0054 § E measures that as terminal. Nothing reconciles the two, and two vectors I
wrote go red on `main`-plus-this-branch. Both are in this comment so they can be lifted
straight in.

**1. A move destroys the window and then re-presents the corpse.**

```ts
const first = rooted();
const second = rooted();
const win = createElement('GtkWindow');
insert(win, first.parent);
remove(win);
insert(win, second.parent);
expect((widgetOf(win) as unknown as Gtk.Window).get_visible()).toBe(true);
```

```
❌ survives the MOVE that remove() documents itself as
   — GTK reported 1 diagnostic(s) that would have passed at exit 0:
     A window is shown after it has been destroyed. This will leave the window
     in an inconsistent state.
```

That is measurement E of your own ADR, reached through the host's own documented
detach. It is exactly the exit-0 class `installDiagnosticsGate()` exists to catch, and
Solid's `removeNode` calls `remove(node)` directly.

**2. `remove()` then `destroy()` runs the declared retraction twice** — the defect the
node leg found, in the other order.

```ts
let retractions = 0;
const realDestroy = widget.destroy.bind(widget);
(widget as unknown as Record<string, unknown>).destroy = () => { retractions += 1; realDestroy(); };
remove(win);
destroy(win);
expect(retractions).toBe(1);   // ❌
```

`remove` unlinks the node, so the `destroy` after it sees `node.parent === null` and
takes the **no-parent arm this PR added**. The `unmap` count stays 1, which is why the
effect-based vector cannot see it — the second `gtk_window_destroy()` is silent on gjs
and `TypeError: invalid GObject handle` on node-gi. So "there is one owner and one call"
holds for `destroy()` alone, not for the pair.

**Why I did not fix it here.** Both come from one place: the two non-parented arms
differ in reversibility — `force_close` → `present` round-trips, `destroy` does not —
and `retractOutsideParent` treats them identically. Closing it is a design call between
"`remove` hides a toplevel and `destroy` destroys it" and "`insert` rebuilds a retracted
toplevel", and the second needs a recorded fact on `HostElement` plus an ADR amendment.
Both change what a `<gtk-window>` unmount means in Solid, where `remove` *is* the
teardown. That is yours to pick, not a reviewer's to impose.

Note this makes the § 4 claim *"two mechanisms cover the class between them"* true for
insertion and not for retraction, which currently has one owner per **order**.

---

### Smaller notes

- `descriptorProblems()`'s new arm and the `is declared by every registered Gtk.Root`
  spec make the same claim about the shipped table. I still think both earn their keep —
  the spec additionally asserts `unpresentable === ['GtkDragIcon']`, which
  `descriptorProblems` deliberately cannot — but it is worth saying so, given
  § `simplicity` asks what a new check lets you delete. `descriptorProblems` + 
  `refuseUnparentable` themselves are not a guard watching a guard: static over the
  table, dynamic over `registerWidget`, different populations.
- **0054 as its own ADR is right.** It reverses a recorded decision of 0045 ("case K is
  routing"), and § `A rule whose premise died` says a reversal gets re-derived in
  writing rather than edited in place. The dated amendment on 0045 pointing here is the
  correct shape.
- `TOPLEVEL` living in `descriptors/gtk.ts` and being imported by `adw.ts` couples two
  sibling descriptor files. The de-duplication is right; a neutral home would be
  tidier. Not worth churn now.
- No new try/catch around a non-throwing call, and no new paranoid probe, in the impl
  diff. `refuseUnparentable`'s `if (!node) return` is unreachable after `materialize`,
  but it mirrors `presentPortal` and narrows the type under `strict: false`.

### Ran

| command | result |
|---|---|
| `gjsify workspace @gjsify/gtk-host run test` | **627 passed**, 2921 assertions |
| `gjsify workspace @gjsify/gtk-host run check` | clean |
| `gjsify workspace @gjsify/react-native run test` | **638 passed**, 2760 assertions |
| `check-generated-website-data` · `check-vocabulary-alignment` · `check-adapter-import-direction` · `check-type-surfaces` · `check-adr-index` · `check-agent-context-size --check` | all OK |
| `gjsify lint` | clean in `gtk-host`; 6 pre-existing warnings elsewhere, unchanged |
| `gjsify format --check` | 3812 files, correct |
| `commitlint --from origin/main --to HEAD` | 0 problems, 1 warning |
| the three child-process vectors, and the raw case standalone | SIGABRT / exit 134 reproduced |

I did **not** run `test:gjs-on-node` — `@gjsify/node-gi` is not built in this checkout
and I chose not to spend the build. That leg is `node-gi.yml`'s `gtk-host-node` job, and
defect 2 above is precisely what it would report.

JumpLink:
## Blocking defect addressed — `remove` detaches, `destroy` closes

Thank you for the two reproducers; both are vectors now, and both were confirmed to
discriminate before I called it fixed.

**The decision.** I took *"`remove` hides a toplevel and `destroy` destroys it"* over
*"`insert` rebuilds a retracted toplevel"*, for two reasons. The second changes widget
IDENTITY across a move, which is the one thing `conformance/vectors.mts` asserts survives a
reorder — a consumer holding a `ref` would be looking at a dead window. And the first is not
a new concept: `remove` and `destroy` already mean different things for a *parented* node
(unparent versus unparent-and-dispose), so the non-parented arms owed the same two answers
and only ever gave one.

**The cause, as you diagnosed it.** The arms differ in reversibility and
`retractOutsideParent` did not:

| case | result |
|---|---|
| `dialog.force_close()`, then `dialog.present(box)` against the SAME parent | re-hosted, `visibleDialog` set, **no diagnostic** |
| `win.destroy()`, then `win.present()` | `Gtk-WARNING **: A window is shown after it has been destroyed` |

So there are two verbs: `detachOutsideParent` (what `remove` means) and `closeOutsideParent`
(what `destroy` means). For a portal both are `force_close` — which is exactly why that arm
never showed the defect, and I added a vector asserting the portal round-trip so that stays
measured rather than asserted.

A toplevel detaches with `set_visible(false)`. Measured in order, one window:

| # | case | result |
|---|---|---|
| O | `set_visible(false)` on a presented window | `visible` false, ONE `unmap`, **no `close-request`**, still in `Gtk.Window.list_toplevels()` |
| P | `present()` after it | mapped again, **no diagnostic** |
| Q | a second `set_visible(false)` | no-op |
| R | `destroy()` on the hidden window | no further `unmap`, and it **leaves** `list_toplevels()` |

O and R are what make it a detach rather than a teardown — GTK's own list still holds a hidden
window, so `destroy` still has work to do. It is a property write rather than a declared
method because a window's presence on screen IS its `visible` property (which is why
`presentToplevel` already reads it), so a per-row name would be eighteen identical strings —
the argument ADR 0045 § 2 already made for `placement` being optional. `set_visible(false)`
and not `hide()`, which is the same call one rename older and deprecated since 4.10.

**A third caller turned up while doing it, and it would have leaked.** `rebuild` discards
`el.widget`; five of `removeChild`'s six call sites re-attach the same widget and only that
one does not. With the detach in place, a construct-only write on a `<gtk-window>` left a
hidden window in `list_toplevels()` forever. `rebuild` now closes it explicitly, and that is a
vector too.

**For a consumer:** unmounting a `<gtk-window>` in Solid takes it off screen at `removeNode`
and destroys it when the root disposer runs `destroyChildren` — which is `destroy`, not
`remove`. The teardown still happens, one step later, and a reorder no longer costs the
window. React and Vue are unaffected.

Recorded as **ADR 0054 § 6** with all six measurements, and ADR 0045 gets one sentence rather
than an edit: its choice of `force_close` is also REVERSIBLE, which is why that arm satisfied
`remove`'s contract by accident. § 4's claim is now scoped to insertion, as you noted it
should be.

### Broke it on purpose

Restoring the terminal call in the detach path and re-running:

```
✖ toplevel placement › survives the MOVE that remove() documents itself as
✖ toplevel placement › runs the declared retraction ONCE across remove() then destroy()
❌ 2 of 631 tests failed
```

Exactly the two, nothing else — so they discriminate rather than merely pass.

### The coredump

Adopted, with the argv prefix rather than `ulimit`: `prlimit --core=0` execs, so `waitpid`
still reports the CHILD's signal instead of a shell's `128 + n`, and there is no command line
to interpolate a path into. Measured side by side in one run — bare spawn
`COREFILE: present 2.8M`, prefixed spawn `COREFILE: none`, both `signalled=true term=6` with
`Adwaita-ERROR` on stderr. Probed with `GLib.find_program_in_path`, because `prlimit` is
util-linux and this suite also runs on darwin and win32; a host without it spawns exactly as
before and asserts exactly the same things.

### Ran

| command | result |
|---|---|
| `gjsify workspace @gjsify/gtk-host run test` | **631 passed**, 2937 assertions |
| `NODE_GI_NATIVE=build … run test:gjs-on-node` | **631 passed**, 2936 assertions — I built `@gjsify/node-gi` for it, and defect 2 is green there too |
| the same with the detach reverted | 2 of 631 failed, exactly the two new vectors |
| `gjsify workspace @gjsify/react-native run test` | **638 passed**, 2760 assertions |
| `run check` · `check-adr-index` · `check-generated-website-data` · `check-type-surfaces` · `check-vocabulary-alignment` · `check-adapter-import-direction` · `check-adwaita-element-properties` · `check-agent-context-size --check` | all OK |
| `commitlint --from origin/main --to HEAD` | 0 problems, 1 pre-existing warning |
| `oxfmt --check` / `oxlint` | clean; 3 pre-existing warnings unchanged |
| `showcase-smoke.mjs` | 11/11 GTK showcases up; refusals probe still 9/9 refused, 0 accepted |

Your other notes are folded into the description: the ADR-index requirement is called out so
the next ADR does not red a required check, the refusal count is 7, and `TOPLEVEL` staying in
`descriptors/gtk.ts` is listed under "not touched" with the reason.

JumpLink:
## Second independent review — the fix holds, and it stopped one caller short

I reviewed the FIX, not the PR again: the three new commits, ADR 0054 § 6, the § 4
restriction and the one added paragraph on 0045. I re-ran the two measurement series
rather than reading them. **Both reproduce exactly.** The decision is the right one. Two
things it asserts about itself are not true, one of them was a live leak, and I pushed a
fix for that one.

### Reproduced: the six-step series, on one window, in order

Written from the ADR's claims rather than lifted from the spec — gjs 1.88.1 / GTK 4.22.4 /
libadwaita 1.9.3, Wayland, one process:

```
BASE visible=true  mapped=true  unmap=0 closeReq=0 inList=true
O    visible=false mapped=false unmap=1 closeReq=0 inList=true
P    visible=true  mapped=true  unmap=1 closeReq=0 inList=true
Q    deltaFromSecondHide=0      closeReq=0 inList=true
R    unmapDelta=0               closeReq=0 inList=false
E    Gtk-WARNING **: A window is shown after it has been destroyed.
```

O and R are exactly as claimed, and they are the load-bearing pair: GTK's own list holds
a hidden window and drops a destroyed one. So `set_visible(false)` really is a detach and
`destroy()` really is the teardown, and the axis has two verbs because it has to.

**`set_visible(false)` as a property write is right here**, and the ADR 0045 § 2 argument
carries: a window's presence on screen IS `visible`, `presentToplevel` already reads it,
and a per-row `detach` name would be the same string on every declared row. Unbracketed is
also right, for the reason given — the window really did leave the screen.

### Reproduced: the negative control discriminates

Restored the terminal call in `detachOutsideParent`'s toplevel arm:

```
✖ toplevel placement › survives the MOVE that remove() documents itself as
✖ toplevel placement › runs the declared retraction ONCE across remove() then destroy()
❌ 2 of 631 tests failed
```

Exactly the two, nothing else.

### Reproduced: the `prlimit --core=0` prefix

Four spawns of the raw abort case in ONE run, through `Gio.Subprocess`, reading
`get_if_signaled()`/`get_term_sig()`, cross-read against `coredumpctl`:

| spawn | result | COREFILE |
|---|---|---|
| bare | `signalled=true term=6`, `Adwaita-ERROR` | **present 3M** |
| `prlimit --core=0` prefix | `signalled=true term=6`, `Adwaita-ERROR` | **none** |

It is a **capability probe, not a platform guard** — `GLib.find_program_in_path('prlimit')`,
no `process.platform` anywhere near it, and the `CORE_FREE ? [...] : run` fallback spawns the
identical argv and asserts the identical things. That is the right shape.

One correction to the ADR's *reason*, not its choice: I ran the two shell forms as controls
and both also report the child's signal — `sh -c 'ulimit -c 0; exec gjs …'` because `exec`
replaces the shell, and even `sh -c 'ulimit -c 0; gjs …'` because a POSIX shell execs the
last command of a `-c` string. So "a shell's `128 + n`" is not what would have happened.
`prlimit` is still the better call for the two reasons that measure true: no command line to
interpolate a path into, and no dependence on a shell's tail-call optimisation (nor on a
shell at all, on win32).

---

## The claim that is not true: `rebuild` was not the only discard, and it leaked

> Five of `removeChild`'s six call sites re-attach the same widget afterwards … The sixth
> is `rebuild` discarding `el.widget`.

`removeChild` has **five** call sites, not six — `materialize`'s rollback, `replaceAt`,
`rebuild`'s child sweep, `rebuild`'s own, `remove` — and the ADR's own enumeration names
five while the sentence says six. That is the live count again, in the commit series that
took the last one out.

More importantly it is the wrong population to have asked. Every caller of `removeChild`
keeps the widget. The question that picks the verb is **which code drops `el.widget`**, and
there are three: `materialize`'s rollback, `rebuild`, `destroy`. Only two of them closed.

Measured, first the premise:

```
constructed:   base=0 after_new=1 inList=true      # new Gtk.Window(), never presented
after destroy: len=0                inList=false
dropped+gc:    len_before=1 len_after=1            # system.gc() twice
```

A `Gtk.Window` is in `list_toplevels()` **from construction**, before any `present()`, and
dropping the JS reference does not remove it. Then the leak, through the host:

```
const dialog = createElement('GtkDialog');   // uncurated children
insert(createElement('GtkButton'), dialog);
materialize(dialog);                          // throws uncurated-placement

PROBE widget=null  toplevels before=31 after=32
```

`materialize` publishes `el.widget` before replaying into it, so a rejected replay has to
unpublish — and for a `Gtk.Root` the reference it drops is not the only one. One refused
child under any declared toplevel whose children stayed uncurated is enough. Same count
before and after this PR, so it is not a regression; it is the class this PR names, at the
site the PR says does not exist.

**Fixed in `fix(gtk-host): close a toplevel at every discard`.** One `releaseWidget` in
`host.ts` owns all three discard sites and closes first, which is also the second copy
lifted rather than a third written — `rebuild`'s two lines and `destroy`'s went into it.
Vector: `a rolled-back materialize does not leak the window it built`. Reverting the new
call reds **exactly that one, 1 of 632**. The ADR paragraph, the README paragraph,
`packages/framework/AGENTS.md`, `removeChild`'s comment and `closeOutsideParent`'s docblock
now say "discard" instead of counting `removeChild`, and measurement U is in the table.

## The claim that is half true: it does NOT happen ONCE on the portal arm

> Here is where a node really is being torn down, and it happens ONCE — `remove` detached,
> this closes.

True for the toplevel arm, and the new vector proves it. For the portal arm both verbs are
the same call, so `destroy` now runs it twice. Counted through a wrapped `force_close`, on
`d9626e9c` and on `b3de1d5c`:

| | before the fix | after |
|---|---|---|
| `destroy(dialog)` | **1** | **2** |
| `remove(dialog)` then `destroy(dialog)` | 2 | **3** |

So the first review's defect 2 is closed for one arm and open — slightly wider — on the
other. It is benign today: `force_close` on a node that is not presented is silent
(measured, ADR 0045), the diagnostics gate is clean, and `force_close` does not drop GTK's
reference, so there is no node-gi handle to invalidate. But the invariant the fix states is
not the one the code has, and the vector that would have said so exists on one arm only.

I did not change it, for the same reason you did not close the reversibility question in
the first round: it is a design call on your axis. The two shapes I can see are (a) an
`detachIsClose(placement)` predicate — the arms differ in reversibility and nowhere else, so
the axis can say so and `destroy` can skip a close its own detach already ran — or (b)
accept 2 as the honest cost of "both verbs are the same call" and write the portal vector
against **that** number so it is measured rather than assumed. Either way the vector should
exist for both arms; it is the asymmetry that hid the defect the first time.

## Smaller

- **The identity argument is right and the citation is not.** `conformance/vectors.mts`
  explicitly excludes it: *"Shape reconciliation — `v-if`, `<For>`, a keyed reorder — is
  each framework's own semantics and each spec already measures it there, with the identity
  assertions that only make sense per framework."* The assertion you want is
  `solid.spec.ts` → `<For> keeps widget identity across a reorder`, which is a **stronger**
  citation: Solid is the framework whose `removeNode` calls `remove` directly, so it is the
  one that would have seen the dead window. `vue.spec.ts` and `react.spec.ts` hold the same
  claim for their own reconcilers. The decision stands; only the sentence in the PR comment
  is wrong, nothing committed says it.
- ADR 0054's sections now run 1, 2, 3, 4, **6**, 5. § 4 ends by pointing forward to § 6, so
  the intent is clear, but the file reads out of order.
- `if (outside && node.widget) closeOutsideParent(...)` in `destroy` re-asked a question
  `closeToplevel`/`retractPortal` already guard (`if (!child.widget) return`). Gone with
  the lift.
- `detachToplevel`'s docblock said a per-row name "would be eighteen identical strings".
  Eighteen is a live count of the declared rows; it now says what the number establishes.

---

## AGENTS.md § Commit conventions is wrong, and the right fix is the doc

> imperative, ≤50-char subject; commitlint enforces on every PR

It does not. `commitlint.config.cjs` extends `@commitlint/config-conventional` and
overrides only `type-enum`; the inherited `header-max-length` is **100**. Three subjects on
this branch are 54, 59 and 60 and run green because of it.

Measured on `main`: **254 of the last 300 subjects are over 50** (85%), longest 91. And 100
is not an oversight — it is the ceiling the SQUASH subject `<title> (#N)` needs, which is
the header that actually lands on `main`.

So of the three options: **correct the document.** Shortening this branch's subjects fixes
one branch and leaves the false claim standing. Entering 50 into `commitlint.config.cjs`
would red #1627 and #1629 the moment it merged, and 85% of the repo's own history with them
— a shared counter, and not this PR's to move even if it were right. What is actually wrong
is that a rule an agent believes a machine is holding is a rule nobody checks; that is the
green-CI-that-checked-nothing class, one line long. `docs: say what commitlint actually
enforces` says the 100 is the enforced number, why it is 100, and that the 50 is house
style. `--update` run, `status/agent-context-budget.json` committed with it.

If you want the shorter number enforced, a separate PR at a quiet moment — 72 rather than
50, on this evidence — with every open branch re-cut in the same round.

## Ran

| command | result |
|---|---|
| `gjsify workspace @gjsify/gtk-host run test` | **632 passed**, 2940 assertions (631/2937 before my vector) |
| the same with the new `releaseWidget` call reverted | **1 of 632 failed** — exactly the new vector |
| the same with the terminal call restored in the detach path | **2 of 631 failed** — exactly your two |
| `gjsify workspace @gjsify/gtk-host run check` | clean |
| `gjsify workspace @gjsify/react-native run test` | **638 passed**, 2760 assertions |
| the O/P/Q/R series, and E, standalone | reproduced, all six rows |
| the toplevel-list premise (`new Gtk.Window()`, drop + gc, destroy) | reproduced |
| `prlimit` A/B + two shell controls, `coredumpctl` cross-read | reproduced; one note above |
| `check-adr-index` · `check-generated-website-data` · `check-vocabulary-alignment` · `check-adapter-import-direction` · `check-type-surfaces` · `check-agent-context-size --check` | all OK |
| `gjsify format --check` | 3812 files, correct |
| `gjsify lint` | clean in `gtk-host`; 12 pre-existing warnings elsewhere, unchanged |
| `commitlint --from origin/main --to HEAD` | **0 problems**, 2 pre-existing `footer-leading-blank` warnings |

**Mergeable once the portal count is decided** — a vector on that arm, whichever number it
asserts. Everything else above is either reproduced or already pushed. I have not merged.

JumpLink:
## The measurement tables

Out of the PR body on purpose: the squash body becomes a commit message on `main`, and a
markdown table does not wrap. `Lint commit messages` measured 10 lines over 100 characters,
all of them tables. The numbers stay; only their home moves.

All on Fedora 44, gjs 1.88.1, GTK 4.22.4, libadwaita 1.9.3, Wayland, one process per case.

### The toplevel arm — why `present`/`destroy` and why the arity separates the arms

| # | case | result |
|---|---|---|
| A | `gtk_window_present` arity | **0**, against `adw_dialog_present`'s 1 |
| B | `close-request` handler returns `true`, then `close()` | window **still mapped and visible** |
| C | then `destroy()` | down, silent |
| D | `present()` after `close()` **or** `destroy()` | `Gtk-WARNING **: A window is shown after it has been destroyed` |
| E | `close()` / `destroy()` on a window never presented | **silent**, both |
| F | `destroy()` twice | silent |
| G | a presented window | `get_parent()` null, `get_root()` itself |

### The structural oracle, over the shipped table

| # | question | answer |
|---|---|---|
| H | `Gtk.Root` implementors | 19, incl. `GtkDragIcon`, which has no `present`/`close`/`destroy` |
| I | classes with a 1-argument `present` | exactly the 5 `Adw.Dialog` descendants |
| J | classes with a 0-argument `present` | 21 — the 18 declared toplevels plus `GtkPopover`, `GtkPopoverMenu`, `GtkEmojiChooser`, none a `Gtk.Root` |

### The quiet half, and the abort it sits beside

| # | case | result |
|---|---|---|
| K1 | `box.append(new Gtk.Window())`, box in an `Adw.Window` | exit 0, silent, `get_parent()` the box, `get_root()` the window itself |
| K2 | then `win.present()` | visible and mapped, **still in the box's child list** |
| L | `box.append(new Adw.Dialog())`, same box, own process | `Adwaita-ERROR`, **killed by SIGABRT**, core dump |

### Reversibility — the pair that separates the two arms

| case | result |
|---|---|
| `dialog.force_close()`, then `dialog.present(box)` against the SAME parent | re-hosted, `visibleDialog` set, **no diagnostic** |
| `win.destroy()`, then `win.present()` | `Gtk-WARNING **: A window is shown after it has been destroyed` |

### The toplevel detach, in order on one window

| # | case | result |
|---|---|---|
| O | `set_visible(false)` on a presented window | `visible` false, ONE `unmap`, **no `close-request`**, still in `list_toplevels()` |
| P | `present()` after it | mapped again, **no diagnostic** |
| Q | a second `set_visible(false)` | no-op |
| R | `destroy()` on the hidden window | no further `unmap`, and it **leaves** `list_toplevels()` |
| U | `new Gtk.Window()`, never presented | already in `list_toplevels()`; dropping the JS reference + `system.gc()` does not remove it |

### One retraction per operation

`force_close` counted through a wrapped method:

| | before § 6 | § 6, first cut | now |
|---|---|---|---|
| `destroy(dialog)` | 1 | **2** | 1 |
| `remove(dialog)` then `destroy(dialog)` | 2 | **3** | 2 |

### The core dump, and the shell controls

| # | spawn | signal | COREFILE |
|---|---|---|---|
| T | bare `gjs -m case.js` | `signalled=true term=6` | **present, 2.8M** |
| T | `prlimit --core=0 gjs -m case.js` | `signalled=true term=6` | **none** |
| W | `sh -c 'ulimit -c 0; exec gjs …'` and the same without `exec` | BOTH report the child's signal | none |

W is a correction to the ADR's earlier *reason*, not to its choice: "a shell would have
reported `128 + n`" is false, because a POSIX shell execs the last command of a `-c` string.
What survives measurement is that an argv array has no command line to interpolate a path
into, does not rest on that behaviour being true of every `sh`, and needs no shell at all —
which matters, because this suite also runs on win32.

### Ran

| command | result |
|---|---|
| `gjsify workspace @gjsify/gtk-host run test` | **646 passed**, 2961 assertions (after merging `origin/main`) |
| `NODE_GI_NATIVE=build … run test:gjs-on-node` | **646 passed**, 2960 assertions |
| the same with the detach reverted to the terminal call | **2 of 631 failed** — exactly the two move/count vectors |
| the same with the `destroy` ordering reverted | **1 of 633 failed** — exactly the portal count vector |
| the same with the `releaseWidget` call reverted | **1 of 632 failed** — exactly the rollback-leak vector |
| `gjsify workspace @gjsify/react-native run test` | **638 passed**, 2760 assertions |
| `run check` · `check-adr-index` · `check-generated-website-data` · `check-type-surfaces` | clean / OK |
| `check-vocabulary-alignment` · `check-adapter-import-direction` · `check-adwaita-element-properties` | OK |
| `check-agent-context-size --check` | OK — `packages/framework/AGENTS.md` 20 421 B, under the 20 KB rule |
| `oxfmt --check` / `oxlint` | clean; 3 pre-existing warnings unchanged |
| `showcase-smoke.mjs` | 11/11 GTK showcases up; refusals probe 9/9 refused, 0 accepted |

JumpLink:
## Portal count decided, plus the two red checks and `origin/main`

### The portal arm: ordering, not a predicate

Taken: **one retraction per operation, on both arms**. `remove` runs the arm's detach once,
a discard runs the arm's close once, and neither runs the other's. `destroy(dialog)` is back
to 1 and `remove` then `destroy` back to 2 — the numbers before § 6, with § 6's fix intact.

`detachIsClose(placement)` was the alternative and I did not take it, for two reasons. It is
a second source for a fact the two dispatch switches already state, so it drifts the day an
arm's verbs stop coinciding. And it cannot answer at `releaseWidget`'s other two sites —
`materialize`'s rollback and `rebuild` — where no detach has run at all, so the predicate
would have been right in one of three places.

So a teardown simply does not detach first: `destroy` releases the widget before it removes,
and `rebuild` skips the `removeChild` that would do nothing else for a non-parented node.
`releaseWidget` nulls the widget, so the detach that follows finds nothing to retract and
still drops the portal's `notify::root` subscription, which has to happen either way. The
invariant is then structural rather than asserted.

Nothing changes on screen for the toplevel arm — `destroy()` alone on a presented window is
the same single `unmap` and the same departure from `list_toplevels()` as hide-then-destroy —
so the cost is one call, not a behaviour.

**Both arms carry the vector now.** Reverting the ordering reds exactly the new portal one,
1 of 633. ADR 0054 § 6 records the decision, the counts before/during/after, and why the
predicate was rejected.

### Your three smaller points

The identity citation is fixed and now lives in the ADR rather than a comment: it cites
`solid.spec.ts`'s `<For> keeps widget identity across a reorder`, notes that
`conformance/vectors.mts` excludes shape reconciliation by construction, and says why Solid
is the right spec to answer to. The same false claim was in `packages/framework/AGENTS.md`
rule (3) — *"conformance vectors … assert IDENTITY across a reorder"* — and is corrected
there too.

The `prlimit` reasoning is corrected in both the ADR and the source comment, with your two
shell controls recorded as measurement W. `prlimit` stays, for the reasons that measure: no
command line to interpolate a path into, no dependence on a shell's tail-call behaviour, and
no shell at all on win32.

ADR 0054 now reads 1, 2, 3, 4, 5, 6.

### `Lint commit messages` — the PR body, not the commits

The gate was right and the fix is the one it names: the description is prose now, every line
under 100, and the measurement tables moved to
[a comment](https://github.com/gjsify/gjsify/pull/1628#issuecomment-5617763155). Nothing was
dropped; the squash body simply stops being a table.

### `GTK suites on the shipped closure (darwin)` — measured, and it was mine

Not a flake. Both darwin arches, win32 green, `1 of 632` — and the failing test is my move
vector:

```
✖ toplevel placement › survives the MOVE that remove() documents itself as
  — GTK reported 1 diagnostic(s) that would have passed at exit 0:
    gdk_surface_thaw_updates: assertion 'surface->update_freeze_count > 0' failed
```

You were right to say a green Linux leg says nothing about that path. What the vector does is
what a MOVE of a `toplevel`-placed node is: `present()`, `set_visible(false)`, `present()` on
a real window — two plain GTK calls. The counter in the assertion is GDK's own, and this
codebase never touches it: `freeze_updates`, `thaw_updates` and `update_freeze` have **zero**
call sites across `packages/{framework,web,dom}`.

That is the premise `ENVIRONMENT_PREFIXES` already rests on for `Vulkan:`, measured the same
way, so `gdk_surface_thaw_updates:` joins it — recorded and counted by `assertQuiet`, never
dropped. The classifier vector takes the new prefix and a control that is NOT it
(`gdk_surface_freeze_updates:` stays a failure), so the gate widens by one function rather
than a namespace. What I cannot establish from here — whether GDK is really the imbalanced
side — is a ledger entry with the reproducer, since this host has no macOS and a VM is out of
scope.

The same log turned up a second one that was plainly mine: with no `gjs` on PATH the child
cases handed `Gio.Subprocess` a null argv[0] and printed
`g_subprocess_newv: assertion 'argv != NULL …' failed` three times before throwing — a GLib
critical from the very cases whose `when` had already declared the host cannot run them. Now
a named error before any spawn.

### `origin/main` merged

`bae4b777` is in, and the four shared files came through with both sides intact — checked
rather than trusted: two `###` sections in `next.md`, ADR 0051 (Accepted) and 0054 both in
`docs/adr/README.md`, both ledger entries in `open-todos.md`, and `test.mts` registering
`placementSuite` and `sharedTreesSuite`. `@gjsify/adwaita-core` needed a rebuild for the new
conformance exports.

### Ran, after the merge

| command | result |
|---|---|
| `gjsify workspace @gjsify/gtk-host run test` | **646 passed**, 2963 assertions |
| `NODE_GI_NATIVE=build … run test:gjs-on-node` | **646 passed**, 2962 assertions |
| the same with the `destroy` ordering reverted | **1 of 633 failed** — exactly the new portal vector |
| `gjsify workspace @gjsify/react-native run test` | **638 passed**, 2760 assertions |
| `run check` · `check-adr-index` · `check-generated-website-data` | clean / OK |
| `check-vocabulary-alignment` · `check-adapter-import-direction` · `check-type-surfaces` | OK |
| `check-agent-context-size --check` | OK — `packages/framework/AGENTS.md` 20 421 B, under the 20 KB rule |
| `commitlint --from origin/main --to HEAD` | 0 problems, 1 pre-existing warning |

On the commitlint doc change: agreed, and thank you for measuring it rather than shortening
three subjects. I have kept to the 50 as house style in the new commits.

JumpLink:
## Third independent review — the darwin assertion is GDK's, measured on a real macOS host

I had the one instrument the first two reviews did not: a macOS Sequoia 15.7.9 guest
(darwin-x64, Homebrew gjs 1.88.1 / GTK 4.22.4 / libadwaita 1.9.2). The gate widening is
**the right call and the argument behind it was not** — so I replaced the argument with a
measurement and pushed it. Everything else reproduces. Three small findings, all pushed.

---

## The headline: environment, and now provable

`gdk_surface_thaw_updates:` joining `ENVIRONMENT_PREFIXES` was justified by a call-site
count — "`freeze_updates`/`thaw_updates`/`update_freeze` have zero call sites across
`packages/{framework,web,dom}`". That is weaker than it reads, and the PR says so itself:
an `update_freeze_count > 0` assertion means somebody thawed **once too often**, and our
own sequence could have been what walked GDK there. So I ran the discriminator.

**Three lines of plain GTK, no gjsify and no gtk-host in the process:**

```js
import Gtk from 'gi://Gtk?version=4.0';
Gtk.init();
const w = new Gtk.Window();
w.present(); w.set_visible(false); w.present();
```

```
(gjs:947): Gdk-CRITICAL **: gdk_surface_thaw_updates: assertion 'surface->update_freeze_count > 0' failed
```

| # | case, bare `gjs -m`, one process each | result |
|---|---|---|
| X1 | `present()`, `set_visible(false)`, `present()` | **FIRES** |
| X2 | `set_visible(true)`, `set_visible(false)`, `set_visible(true)` — no `present()` anywhere | **FIRES** |
| X3 | `present()` alone · `present()` twice · `present()` then `set_visible(false)` | all three **SILENT** |

So it is not our host, and it is **not even `present`** — the trigger is re-showing a
surface that was hidden. X2 is the stronger version of the author's claim than the one
the PR makes.

**The imbalanced side is named.** `G_DEBUG=fatal-criticals` under `lldb` (env set via
`target.env-vars`, because SIP strips `DYLD_*` through the ssh/sh chain):

```
frame #0  _g_log_abort                             libglib-2.0
frame #2  g_log
frame #3  _gdk_macos_toplevel_surface_present      libgtk-4.1   <- the extra thaw
frame #4  gtk_window_present_toplevel
frame #5  gtk_window_show
frame #10 gtk_widget_show
frame #11 _gtk_window_present
frame #12 ffi_call_unix64                          libffi       <- the binding boundary
```

Every frame above the log call is GTK's own. There is no gjsify frame, because there is
no gjsify in the process.

**Loud, not harmful** — I checked rather than assumed. Five hide/show cycles: exactly one
record per re-show, no accumulation, window still `visible` and `mapped`, same
`GdkSurface` object throughout, and with the main loop pumped so the window is really
allocated, the rendered texture is **byte-identical either side** (200x100, 937-byte PNG
before and after).

**And the classifier was checked against the string a real host stores, not a typed one.**
This is the half a synthetic vector cannot cover: `isEnvironmentDiagnostic` does
`startsWith('gdk_surface_thaw_updates: ')`, and if the gate recorded the message with its
domain prefix the whole widening would be inert while `conformance.spec.ts` stayed green.
Reproducing `installDiagnosticsGate`'s own primitive on the guest:

```
LEVEL=8 WARNING=16 recorded<<gdk_surface_thaw_updates: assertion 'surface->update_freeze_count > 0' failed>>
  startsWith("gdk_surface_thaw_updates: ") = true
```

Level 8 is CRITICAL, so it is recorded (<= WARNING), and the prefix matches. The widening
works on the real platform.

**Pushed** as `docs(gtk-host): measure the darwin thaw imbalance`: ADR 0054 gains a
section and rows X1-X6, the `ENVIRONMENT_PREFIXES` docblock states the measurement instead
of the call-site argument, the `conformance.spec.ts` comment says the string was read back
from a real host, the ledger entry is **closed** (its open question is answered), and the
remaining ask moves to `status/upstream-patch-candidates.md` where upstream asks live —
GDK's macOS backend should balance the freeze it thaws.

**What I did not do:** build the shipped closure on the guest and run the real darwin leg.
That would have proved the leg green end to end, but the guest's GTK is Homebrew's, and
this VM's own notes record that a bundle plus Homebrew GTK loads two GObject type systems
in one process — so a leg result there would need its library list audited before it meant
anything. The discriminator above answers the actual question with less to disbelieve.

---

## 1. The ordering invariant is structural, and the negative control lands

I traced it rather than taking it. `releaseWidget` nulls `el.widget`, and **all four arm
verbs are null-guarded**: `detachToplevel` (`if (node)`), `closeToplevel` and
`retractPortal` (`if (!child.widget) return`). So after a discard, any detach that follows
is a genuine no-op — the invariant rests on that plus the ordering, not on a claim.

I enumerated the paths for an `outside` node and found no path where both run:

| operation | detach | close |
|---|---|---|
| `remove`, no parent | `remove`'s own branch, once | — |
| `remove`, with parent | via `removeChild`, once (the two branches are exclusive) | — |
| `destroy` | no-op, widget already null | `releaseWidget`, once |
| `rebuild` | skipped by the new `!outsideParentOf` guard | `releaseWidget`, once |
| `materialize` rollback | never detaches `el` itself | `releaseWidget`, once |
| `replaceAt` | `removeChild`, once — and re-attaches | — |

Two details worth recording because they are what make the skip safe rather than lucky:
`removeChild` for an outside child does **only** `return detachOutsideParent(...)` before
it touches the parent, so `rebuild` skipping it loses the detach and nothing else; and
`rebuild` sets `el.attached = false` unconditionally after `releaseWidget`, so the skip
leaves no stale flag. The portal's `notify::root` subscription still drops, because
`retractPortalWatch` runs before `retractPortal`'s widget guard.

`releaseWidget` also nulls `el.wrapper`, which `remove` would otherwise clear — harmless,
because `attach` returns at the placement dispatch **before** `ensureWrapper`, so an
outside node never has one.

**Negative control, run:**

```
$ # `if (outside) releaseWidget(node)` removed from destroy()
✖ portal placement — the ordinary host operations still hold ›
    runs the declared retraction ONCE per operation, like its neighbour
❌ 1 of 646 tests failed
```

Exactly that vector and nothing else. **The count is 646, not 633** — see finding 3.

---

## 2. The named spawn error does not silence the child cases

Measured rather than reasoned: I ran the built suite with `PATH` scrubbed, through an
absolute `gjs`, so `GLib.find_program_in_path('gjs')` answers null on a host that is
otherwise identical.

```
$ env PATH=/nonexistent /usr/bin/gjs -m packages/framework/gtk-host/test.gjs.mjs
  ✗ kills a child process with SIGABRT on the raw append the seam replaces
      (expected failure — needs a gjs interpreter on PATH to run the case in its own process)
  ✗ exits 0 on the placement the host uses instead                    (same)
  ✗ shows the toplevel half is a silent accept rather than a second abort (same)
✗ 3 expected failures (it.failing — upstream defects)
✔ 646 tests passed · 2952 assertions
```

They are **named on the report line, with the reason**, and 2952 assertions against 2963
with `gjs` present is the visible difference. `when: GJS === null` still delegates to a
plain `it()` on any host that has one, so the day such a host stops reproducing the abort
they go red. Not silent.

One honest gap: **I could not reproduce the "before" symptom on Linux.** With the named
throw removed and `PATH` scrubbed, gjs 1.88.1 produced no `g_subprocess_newv` critical at
all — it rejects the null argv[0] during marshalling. The three criticals the commit
message describes come from the darwin legs' binding, and I did not rebuild that closure.
The change is a strict improvement either way (a named error before a spawn that cannot
work), and it also removes a critical that `gated`'s `afterEach(assertQuiet)` would
otherwise have had to account for.

---

## 3. One live count, in the file the last two rounds took them out of

Both earlier reviews found one of these. Here is the third, and it is in ADR 0054 § 6:

> Reverting the ordering reds exactly the second one, **1 of 633**.

Measured on the tree it is committed on: **1 of 646**. The 633 was taken before
`origin/main` merged and drifted inside the branch that wrote it — the same shape as the
"all 164 rows" the first review corrected. The load-bearing half is "exactly the second
one and nothing else"; the total is the half that cannot stay true. Rewritten to say that.
The same 633 appears in the PR comment above, and 631/632 in the earlier ones — those are
comments and I have left them alone.

---

## 4. `packages/framework/AGENTS.md` — both halves check out

**The identity citation is now correct**, verified against the tree rather than read:
`packages/framework/gtk-host/src/adapters/solid.spec.ts:165` really is
`<For> keeps widget identity across a reorder`, and `conformance/vectors.mts:30` really
does exclude shape reconciliation in so many words. Nothing in the repo is checking
in-repo citations (the ledger says so itself), so this one is hand-verified.

**The trim did not compress away an incident.** Rule (10) lost *"since a window is in
`list_toplevels()` from construction"* — the reason a discard must close rather than
detach — but it survives one hop away in exactly the two places the rule's own opening
sentence points at: `gtk-host/README.md:236` and ADR 0054's measurement U. Both incidents
that justify the rule (the moved window presented as a corpse; the portal arm then running
it twice) are still written out in the file itself. That is the shape AGENTS.md asks for.

20 421 B, and `status/agent-context-budget.json` agrees —
`check-agent-context-size --check` is OK.

---

## 5. The `origin/main` merge — checked by comparing both parents, not by reading

A clean textual merge proves nothing, so I diffed the ledger headings against **both**
sides:

```
$ comm -23 <(git show bae4b777:status/open-todos.md | grep '^### ' | sort) \
           <(grep '^### ' status/open-todos.md | sort)
### 138 of 164 generated widgets have no measured placement rule
### A `Gtk.Window` in a child list is accepted silently, where an `Adw.Dialog` aborts

$ comm -23 <(git show 16b488b79a:status/open-todos.md | grep '^### ' | sort) ...
(empty)
```

Nothing from the branch side is gone, and the only two entries missing from main's side
are the two **this PR closes**. `next.md` carries both `###` sections, `docs/adr/README.md`
carries 0051 and 0054, and `test.mts` imports and registers `placementSuite` and
`sharedTreesSuite`. The merge is sound.

---

## 6. Craft

- **One new lint warning, in a file this PR adds.** `placement.spec.ts` imported
  `GtkHostError` as a value and only ever used it as a type — six `as GtkHostError` casts
  and one return annotation, no `instanceof`. Both earlier reviews reported `gjsify lint`
  as "clean in gtk-host"; it now is. Pushed as `test(gtk-host): import GtkHostError as a
  type`. The two remaining gtk-host warnings (`style/sheet.ts`, `style/theme.ts`) are
  pre-existing and untouched here.
- No new try/catch around a non-throwing call, and no new paranoid probe, in the last four
  commits. `runInChild`'s new `if (!GJS) throw` is a real precondition on a real null.
- The new portal vector counts the CALL rather than an effect, and says why in the comment
  — which is the right exception to the effect-based default, and it is the reason the
  asymmetry was invisible for a round.
- `releaseWidget` is a genuine second-copy lift: `rebuild`'s two lines and `destroy`'s went
  into it and the rollback site joined them. That is the shape the anti-pattern rule asks
  for, one site rather than three.
- Side observation, not this PR's: `STATUS.md` is listed in `.gitignore` **and** tracked,
  and the tracked copy is stale — regenerating it produces a 6 462-line diff. AGENTS.md
  says it is "rendered into a GITIGNORED `STATUS.md` … and never committed", so the tracked
  copy is a second truth nothing checks. Worth a `git rm --cached` in a quiet PR.

---

## Ran

| command | result |
|---|---|
| the macOS discriminator, 6 cases + backtrace + harm + record-shape probe | reproduced; tables above |
| `gjsify workspace @gjsify/gtk-host run test` | **646 passed**, 2963 assertions |
| the same with the `destroy` ordering reverted | **1 of 646 failed** — exactly the portal vector |
| the same with `PATH` scrubbed of `gjs` | 646 passed, 2952 assertions, 3 named expected failures |
| `gjsify workspace @gjsify/gtk-host run check` | clean |
| `gjsify workspace @gjsify/react-native run test` | **638 passed**, 2760 assertions |
| `check-adr-index` | 53 ADR(s) indexed, statuses agree |
| `check-generated-website-data` · `check-vocabulary-alignment` · `check-adapter-import-direction` | OK |
| `check-agent-context-size --check` | OK |
| `gjsify format --check` | 3817 files, correct |
| `gjsify lint` | gtk-host clean apart from 2 pre-existing; 50 warnings repo-wide, unchanged |
| `commitlint --from origin/main --to HEAD` | **0 problems**, 1 pre-existing warning |
| PR description line lengths | longest 89, under the 100 the squash body needs |

Every abort-capable run under `( ulimit -c 0; … )`. The VM was started with the host at
0 qemu processes / 29 GiB available / load 1.29, and stopped after the measurement —
0 guests running now.

---

**Mergeable.** The one change I would not have merged on its argument now stands on a
measurement, and it is the stronger conclusion: the assertion reproduces in three lines of
GTK with nothing of ours in the process, the extra thaw is inside
`_gdk_macos_toplevel_surface_present`, and it is loud rather than harmful. The ordering
invariant is structural and its control discriminates. I have not merged.

Issue #1626, opened 2026-09-10
win32-x64 still ships no MP3, Vorbis or FLAC decoder — the payload half of #1544


JumpLink:
#1544 was closed on the half that was a CHECK. The other half is a payload DECISION and nothing
tracks it, so this is that half: `@gjsify/gtk-runtime-win32-x64` still carries no MP3 decoder, and
MP3 is the format an application reaches for first.

## Measured, on the published tarballs, from Linux

No Windows and no macOS host is needed to read a file list. The three `0.48.0` tarballs were pulled
from the registry and their `gtk/lib/gstreamer-1.0/` directories compared — the same directory
`gtk-runtime.js` points `GST_PLUGIN_SYSTEM_PATH` at:

```
curl -sL https://registry.npmjs.org/@gjsify/gtk-runtime-<target>/-/gtk-runtime-<target>-0.48.0.tgz \
  | tar -tz | grep lib/gstreamer-1.0/
```

| | plugins in the payload | present nowhere else |
|---|---|---|
| `darwin-x64` | 24 | `flac`, `mpg123`, `osxaudio`, `vorbis` |
| `darwin-arm64` | 24 | identical set to `darwin-x64` |
| `win32-x64` | **21** | `directsound` |

So the win32 payload is the darwin payload minus `mpg123`, `vorbis` and `flac`, plus the platform's
own sink. `osxaudio` vs `directsound` is not a gap — a platform's sink is not the other's — and
`wasapi2` is absent on win32 but costs nothing, since `autoaudiosink` resolves to `directsound`.

The three that are missing are decoders, and `GST_AUDIO_DECODERS` in
`packages/node-gi/scripts/gst-plugins.mjs` says what each one costs:

| format | element | plugin | darwin | win32 |
|---|---|---|---|---|
| MP3 | `mpg123audiodec` | `mpg123` | ok | **absent** |
| Ogg / Vorbis | `vorbisdec` | `vorbis` | ok | **absent** |
| FLAC | `flacdec` | `flac` | ok | **absent** |
| WAV / PCM, Opus, A-law, µ-law | | | ok | ok |

`ogg` ships on win32, so an `.ogg` container is demuxed there and then reaches nothing that decodes
the Vorbis stream inside it — the same shape as the mp3 case one format over.

## What the bundle says about itself while this is true

`gtk/manifest.json` on the win32 tarball records

```
"windowingData": { …, "gstPlugins": 21, … }
```

[...]

Issue #1544, opened 2026-09-04
The win32 runtime bundle ships no mp3, vorbis or flac decoder, so its declared audio contract is not met


JumpLink:
Measured on the `win11-gjsify` VM (Windows 11 25H2, no gvsbuild, no MSVC) against
`@gjsify/gtk-runtime-win32-x64@0.47.0` under Node 24 + `@gjsify/node-gi@0.47.0`.

## Symptom

**No audio at all**, in a real application, and neither half of it looks like the same problem:

```
loading assets/audio/sample-episode.mp3 …
[audio] GStreamer error: Your GStreamer installation is missing a plug-in.
[audio] GStreamer error: GStreamer encountered a general stream error.
--- bundled mp3 ---   ticks 10, loaded false, position never reported

loading https://icecast.correctiv.net/salon5low …
[audio] GStreamer error: Internal data stream error.
--- live radio ---    ticks 14, loaded false, position never reported
```

The same application plays both on Linux, and plays the bundled file on darwin-x64.

## What is actually missing

`GST_AUDIO_PLUGINS` in `packages/node-gi/scripts/gst-plugins.mjs` names 21 plugins. Asked of the
running registry inside the shipped win32 bundle:

| plugin | win32 |
|---|---|
| `coreelements`, `app`, `typefindfunctions`, `playback` | ok |
| `audioparsers`, `wavparse`, `isomp4`, `ogg`, `opus` | ok |
| `audioconvert`, `audioresample`, `autodetect`, `directsound` | ok |
| **`mpg123`** | **MISSING** |
| **`vorbis`** | **MISSING** |
| **`flac`** | **MISSING** |
| **`wasapi2`** | **MISSING** |
| `soup` | ok |

Element factories, same process:

```
filesrc ok   souphttpsrc ok   playbin3 ok   decodebin3 ok
mpg123audiodec NULL          wasapi2sink NULL          autoaudiosink ok
```

**One gap explains both failures.** `mpg123` is the mp3 decoder. The bundled asset is an mp3, and
the Icecast stream is mp3 too — so the file fails as *"missing a plug-in"* and the stream fails as
*"Internal data stream error"*, which is decodebin having nothing to autoplug behind a source that
did open. Two symptoms, one cause, and the second one is actively misleading: `Internal data stream
error` out of an https URL is the exact string `gst-plugins.mjs` documents for a MISSING TLS
BACKEND. It is not that here — measured in the same process:

```
[...]

Issue #1536, opened 2026-09-04
A bundled GStreamer plugin g_module_opens Homebrew's libsoup, so two GObject type systems land in one process


JumpLink:
Measured on a macOS 15.7.9 / x86_64 host that also has Homebrew's `glib 2.88.2`, `libsoup 3.6.6`,
`gstreamer 1.28.5` and `glib-networking 2.80.1` installed, against `@gjsify/gtk-runtime-darwin-x64@0.46.0`
under Node 24 + `@gjsify/node-gi@0.46.0`.

## Symptom

An `https://` stream never loads, and it fails SILENTLY: the port reports `loaded: false`,
`position: never reported`, `errors: none`. A bundled file on the same pipeline plays fine
(`0.00s -> 2.96s`, `durationSec 97.47`). So the audio path works and only the network source
does not, which reads exactly like "souphttpsrc is missing" — the thing #1476 fixed.

It is not missing. What arrives instead, on stderr:

```
GLib-GObject-CRITICAL: g_object_new_is_valid_property:
    object class 'GObject-closure-array' has no property named 'user-agent'
GLib-GObject-CRITICAL: signal 'accept-certificate' is invalid for instance '0x7feef2916840'
    of type 'GObject-weak-notifies'
GLib-GObject-CRITICAL: signal 'authenticate' is invalid for instance '0x…' of type 'GObject-weak-notifies'
GLib-GObject-CRITICAL: signal 'restarted' is invalid for instance '0x…' of type 'GObject-weak-notifies'
```

`user-agent` is `SoupSession`'s; `accept-certificate`, `authenticate` and `restarted` are
`SoupMessage`'s. So this is libsoup constructing its own session inside `souphttpsrc`.

**`GObject-closure-array` and `GObject-weak-notifies` are not type names — they are GObject's own
internal qdata QUARK strings** (`strings libgobject-2.0.so.0` matches both exactly). `g_type_name()`
returned them where a class name belongs, which is what a GType id resolved against the wrong type
system looks like.

## Cause, measured with `DYLD_PRINT_LIBRARIES=1`

| library | distinct copies loaded | which |
|---|---:|---|
| `libgobject-2.0.0.dylib` | **2** | the bundle's **and** `/usr/local/Cellar/glib/2.88.2/` |
| `libglib-2.0.0.dylib` | **2** | same pair |
| `libgio-2.0.0.dylib` | **2** | same pair |
| `libgstsoup.dylib` | 1 | **the bundle's** |
[...]

PR #1629, opened 2026-09-10
feat: a runtime bundle declares what it decodes, and initFonts reports family names

### LINKED ISSUE #1542: initFonts() reports files, but only family names are actionable — and the family name differs per platform
`initFonts()` (`@gjsify/gtk-host/fonts`, #1510) reports the FILES it handed to the font map. A
caller can only act on FAMILY NAMES, and the family name is not derivable from the file — it comes
out of the font's naming table, and which name you get depends on which font stack read it.

## Measured

The same face, byte-identical across two machines (SHA-256 verified), staged from
`@expo-google-fonts/merriweather` by the same script:

| host | font stack | family the map ends up with |
|---|---|---|
| Fedora, GTK 4 | fontconfig | `Merriweather` |
| Windows 11, `@gjsify/gtk-runtime-win32-x64@0.47.0` | gvsbuild | **`Merriweather 18pt`** |

Google Fonts ships Merriweather as an optical-size family, and the two readers disagree about
whether the size axis belongs in the family name. `Source Sans 3`, from the same staging run, has
no such axis and reads identically on both.

`initFonts()` reported the same thing on both hosts:

```
registered: 5   declined: 0   failed: 0
```

That report is accurate — five faces did go onto the map. It is also not actionable, and on Windows
the consequence is the exact failure the guide's own premise describes:

```
families on the map: 82 -> 84
  Merriweather     ABSENT
  Source Sans 3    present

couldn't load font "Merriweather 12", falling back to "Sans 12", expect ugly output.
```

So an application that stages a face, calls `initFonts()`, gets `registered: 5, failed: 0`, and then
sets `font-family: Merriweather` renders in Tahoma on Windows and in Merriweather on Linux — from
one payload and one call, both reported successful.

## Why the report is the right place to fix it

`bundled-fonts.md` states the premise itself: *"Pango does not report a missing family.
`set_family('Brand')` against a font map that has never heard of `Brand` resolves to the default
sans, the window draws, the process exits 0 and nothing anywhere says a word."*

`InitFontsResult` is the one moment where something CAN say a word — it is the only place in the
flow that has both the file list and access to the map. Everything downstream has lost the
connection between them: a `font-family` declaration cannot know which file was supposed to back it,
and `PangoFontMap.list_families()` alone cannot say which of 84 names arrived from the staged
directory.

Concretely, what a caller needs and cannot get today:

- **the family names the registration added** — a diff of `list_families()` across the call, which
  `initFonts()` is uniquely placed to take and no caller can reconstruct afterwards;
- ideally the mapping **file -> family**, since a directory can hold several families and a caller
  staging a brand set wants to check its own list against what arrived.

Neither needs a new API surface: `InitFontsResult` already carries `registered`, `declined` and
`failed`, and this is one more field of the same kind.

## What a consumer has to do without it

Read `list_families()` before and after, then match a declared family against what turned up —
accepting an exact name or a single `<family> <n>pt` optical variant, and refusing to guess when
several optical sizes are present, since which size to use at which point size is a design decision.
That is 40 lines of application code whose entire content is a fact about the platform, written
against a `PangoCairo` import the application otherwise has no reason to have. It also has to run
before the first styled element renders, because a class compiler bakes the family into the class it
mints — so getting the ordering wrong is a second silent failure on top of the first.

## Note on the darwin row, because it interacts

On macOS `add_font_file` is declined by the CoreText map and the faces are supposed to arrive
declaratively through a `.app`'s `ATSApplicationFontsPath` — measured, outside a bundle: `declined: 5`,
187 families before and 187 after, all cuts substituted to Helvetica. So `declined` is consistent
BOTH with everything working and with nothing working, and the only thing that separates the two is
what the map has afterwards. A family-name field would make that distinguishable from the result
alone, on the platform where it is least obvious.

Related: #1510 (which added this API), #1512 (the guide), #1539 (a multi-word family — and note that
the Windows alias `Merriweather 18pt` IS multi-word, so the two interact: a family that needs no
quoting on Linux needs it on Windows).

JumpLink:
Two closure defects in the `@gjsify/gtk-runtime-*` bundles, and the mechanism that would have
found the first one without a person on Windows.

## 1. A bundle now declares what it can decode — ADR 0055

The bundles carry GStreamer, so they decide what an application can PLAY, and that answer is
per-platform. Measured on the published `0.48.0` tarballs, read from Linux with `tar -tz` over
the registry tarball — no macOS and no Windows host is needed to read a file list:

| bundle | plugins in `gtk/lib/gstreamer-1.0/` | present nowhere else |
|---|---|---|
| `darwin-x64` | 24 | `flac`, `mpg123`, `osxaudio`, `vorbis` |
| `darwin-arm64` | 24 | identical set to `darwin-x64` |
| `win32-x64` | **21** | `directsound` |

`osxaudio` against `directsound` is a platform's own sink. The other three are decoders: MP3,
Ogg/Vorbis and FLAC play on darwin and not on win32, out of one npm scope, at one version.

Nothing between the builder and the consumer was wrong. The bundle's own `gtk/manifest.json`
records `windowingData.gstPlugins: 21` — **a count cannot be wrong about which.** #1562 closed
the builder's blind spot with `GST_PLUGIN_GAPS`, and that is the right fix aimed at the wrong
audience: a build script is not published.

So each bundle package declares `gjsify.mediaCapabilities` — every format it decodes with the
plugin FILE behind it and the ELEMENT that decodes it, and every format it does not, with a
mandatory reason. The evidence split is ADR 0024 § A3's shape one subject over: `plugin` is a
file any host can read (that is how the table above was measured), `element` is a registry entry
only the target OS can answer for. A present file is necessary and not sufficient, and the rule
prints that on passing runs rather than leaving it in a comment.

`media-capabilities` (portable, in `@gjsify/manifest-conformance`) has four passes, and three
need no payload — which is what makes it a gate on an ordinary Linux PR:

- a package shipping a `gtk/` payload declares `mediaCapabilities` at all;
- shape: three fields per claim, a mandatory `why` per gap, no format both taken and not taken,
  `gstPluginDir` inside `files`;
- **every bundle answers for every format any bundle speaks about** — the pass that would have
  caught this class at declaration time, because the defect is an ASYMMETRY and an asymmetry is
  visible from anywhere;
- every claimed `plugin` is in the payload and every gap's plugin is not — where a payload is
  reachable: the two builders, and `gtk-os-suites.yml`, which passes `--media-payload` at the
  bundle it already stages from npm.

`gst-plugins.mjs` derives `GST_PLUGIN_GAPS`, `gstAudioDecoders(target)` and
`gstFormatGaps(target)` from those manifests, so the builders, the conformance rule and
`gst-elements.test.mjs` read the one declaration the tarball carries. That is a consolidation,
not a fifth guard: three tables leave a build script for the artifacts they describe.
`GST_AUDIO_PLUGINS` stays — it is the SEED of the copy, and #1544 is the distance between a
request and a promise.

The win32 payload question itself is #1626, filed as part A of this work: the gap is declared,
the decision (build libmpg123/libvorbis/libFLAC into the gvsbuild set, take them elsewhere, or
keep the narrowed contract) is not. Two measurements went into it beyond the file listing, both
on the published closure with nothing from a checkout:

- `gtk/bin/` holds 65 DLLs and none of them is `libmpg123`, `libvorbis`, `libFLAC` or `libogg`,
  while the darwin bundles carry all four beside their plugins. The library was never in the
  prefix, so the plugin that links it could not have been built — upstream work, not a builder
  change. `opus-0.dll` being there is why Opus is the one modern codec that works.
- On Windows 11 with `@gjsify/node-gi@0.48.0` + the published bundle: `mpg123audiodec`,
  `vorbisdec` and `flacdec` are all NULL in the running registry, and `playbin3` with a forced
  `fakesink` reaches PLAYING for a WAV and never leaves READY for an MP3 produced by the same
  `ffmpeg` invocation. The WAV is the positive control; a scheme nobody handles is the negative
  one, because a non-existent `file://` URI fails too and would not separate the two.
  `Gst.uri_protocol_is_supported(SRC, 'https')` is now TRUE there, so #1476's widening did land
  and only the decoder behind it is missing — a correction to the older 0.45.0 note.

## 2. `initFonts()` reports family NAMES — closes #1542

It reported the FILES it registered, and a caller can act on none of them: `font-family` takes a
family NAME, the name comes out of the font's naming table, and which name you get depends on
which font stack read it.

Re-measured on `0.48.0` rather than quoted from the report — the same two files on both machines,
SHA-256-verified identical there, one script calling `add_font_file`:

| host | font stack | families | gained |
|---|---|---|---|
| Fedora 44, GJS | fontconfig | 100 → 102 | `Merriweather`, `Source Sans 3` |
| Windows 11, the win32 bundle | gvsbuild | 82 → 84 | **`Merriweather 18pt`**, `Source Sans 3` |

And the substitution as a measurement, since "it resolved" is what a substitution looks like. A
40pt `Wg` on Windows: `Merriweather` 77x66, an invented family 77x66, `Merriweather 18pt` 82x68,
`Source Sans 3` 68x77 — and 68x77 on Linux too, the control having no size axis.

One claim in #1542 is corrected by that run and by this branch: the substitution does NOT write
nothing to stderr. The win32 backend prints `couldn't load font …, falling back to "Sans …"` at
layout time, and the fontconfig host printed nothing for the same invented family. Neither is a
value a caller can branch on, which is why the answer is a returned match — but the sentence was
false on the platform it was about.

`InitFontsResult` now carries `families` (the names the map GAINED across the call, a diff only
the registering call can take) and `matches`. It is deliberately NOT a file → family map: Pango
exposes no link from a file back to the family it produced, and attributing by registering one
file at a time would credit a family to whichever face went first and report the other four as
having contributed nothing.

The diff alone is not actionable, because the name that resolves is not the name that was
written. `matchFontFamily` is that layer — `exact`, `optical`, `ambiguous`, `absent` — and it is
pure, so both sides of a two-OS divergence are exercised on one host. `ambiguous` is a state and
not an error: which optical size to use at which point size is a design decision. Pass
`expectedFamilies` and every name that does not resolve is WARNED about, which is the only place
a missing family is ever loud.

## Negative control — the guard broken on purpose, three ways

A guard that has never been red is not one. Each was run against the real published `0.48.0`
payloads staged into a scratch directory, then reverted:

**1. A claimed format whose plugin never shipped** (win32 declares MP3):

```
MEDIA-CAPABILITY FAILURES on 1 finding(s):
  - @gjsify/gtk-runtime-win32-x64: declares it decodes MP3, and the plugin behind it
    (`mpg123`) is not in gtk/lib/gstreamer-1.0 — 21 plugin file(s) are, none of them it.
```

**2. A declared gap whose plugin arrived** (darwin-x64 declares an `mpg123` gap):

```
  - @gjsify/gtk-runtime-darwin-x64: declares a gap for `mpg123` (MP3), and the payload
    carries it. Delete the entry — a gap that outlives its cause is a promise the bundle
    now keeps and its own declaration still refuses.
```

**3. Silence, with NO payload anywhere** — the pass that runs on an ordinary PR (the FLAC gap
deleted from win32):

```
  - @gjsify/gtk-runtime-win32-x64: says nothing about FLAC, which
    @gjsify/gtk-runtime-darwin-arm64, @gjsify/gtk-runtime-darwin-x64 declare an answer for.
```

And the same three, plus five more shapes, are permanent in
`tests/e2e/media-capability-declaration`, whose last suite asserts the three real bundle
packages by name so the fixtures cannot be the only thing the rule is correct about.

## Vacuity, addressed where it could hide

- The rule PRINTS `payload NOT INSPECTED` on passing runs, per package. An e2e assertion holds
  that line, because a check reporting "clean" over an artifact it never opened is the shape
  this whole area exists to remove.
- `readGstPluginDir` distinguishes an ABSENT directory from an EMPTY one; the two read alike
  from a `length === 0` test.
- `gst-elements.test.mjs` asserts its claim list is non-empty before iterating it.
- The e2e's real-tree suite fails unless it finds exactly the three bundle packages.
- `gstAudioDecoders` REFUSES an unknown target rather than answering `{}`, on
  `expectedGstPlugins`'s precedent: an empty answer relaxes every check downstream.

## What was run

- `node scripts/audit-runtimes.mjs --check` — exit 0. Prints
  `media-capabilities: 3 bundle(s) — 18 declared format(s), 7 declared gap(s),`
  `0 payload(s) inspected here`, plus one `payload NOT INSPECTED` note per bundle.
- the same with three `--media-payload=` pointed at the published 0.48.0 payloads — exit 0,
  `3 payload(s) inspected here`, 24 / 24 / 21 plugin files.
- the same, with each of the three misdeclarations above — exit 1, quoted in full.
- `node --test tests/e2e/media-capability-declaration/run.mjs` — 23/23.
- `node --test` over `prebuild-declaration-invariant`, `runtime-slot-declarations`,
  `ship-declaration`, `storybook-declaration` and this suite — 135/135.
- `node --test packages/node-gi/node-gi/test/gtk-runtime-bundle-gates.test.mjs` — 52/52.
- `gjsify workspace @gjsify/gtk-host run test` (GJS 1.88.1) — 626 tests, 2898 assertions, 0 failed.
- `gjsify run test:gjs-on-node` in `gtk-host` (Node 24.19.0, over node-gi) — 626 tests,
  2897 assertions, 0 failed, 2 pre-existing `it.failing` unrelated to this branch.
- `gjsify workspace @gjsify/gtk-host run check` — exit 0. `gjsify run build` over the tree first,
  since the suite bundle inlines its dependencies' `lib/`.
- `node --test packages/node-gi/node-gi/test/gst-elements.test.mjs` — 6 pass, 2 skipped: no
  bundle resolves on Linux, by design.
- `check-adr-index`, `check-changelog-references`, `check-e2e-suite-coverage`,
  `check-workflow-inline-scripts`, `check-workflow-run-syntax`,
  `check-workflow-script-checkout`, `check-pr-body-lines` — clean.
- `oxfmt --check` and `oxlint` on every touched file — clean.
- On a Windows 11 guest, against `@gjsify/node-gi@0.48.0` + `@gjsify/gtk-runtime-win32-x64@0.48.0`
  installed from npm: a GStreamer registry/effect probe and a font-registration probe, quoted
  above and in full on #1626 and #1542.

## What could not be closed here

- **Whether a shipped plugin LOADS, from a Linux CI leg.** The rule reads file names, by
  construction. That question belongs to `gst-elements.test.mjs` on the target OS, and this
  declaration is what tells it which elements to ask for. Its two bundle-scoped tests SKIP on
  Linux — there is no `@gjsify/gtk-runtime-linux-*` by design, and a win32 DLL cannot be loaded
  here. They run on `gtk-os-suites.yml`'s two OS legs, and the manual Windows run above is what
  says the declaration this branch adds matches the registry there today.
- **The darwin side of the same question, and #1536.** Not touched: the macOS guest wants 16 GiB
  and the workstation did not have them free.
- **The win32 payload decision** (#1626). Declaring the gap does not answer it; it makes the
  answer visible to whoever is choosing the package.
- **`@gjsify/webaudio`'s `canPlayType`**, which answers `audio/aac` from a hardcoded list and is
  wrong about it inside a bundle. Tier 1 cannot import a tier-2 script; the entry stays in
  `status/open-todos.md`, untouched.

No `status/open-todos.md` entry was closed, and no AGENTS.md was touched — so
`status/agent-context-budget.json` is unchanged, which keeps this PR out of the counter two
other branches are sharing.


JumpLink:
## Independent review — and six commits on the branch

I did not write this PR. Everything below was run in a dedicated worktree at
`fix/bundle-closure-fonts-and-audio`; every claim I could not reproduce is marked as such.

**Verdict: mergeable once CI reports.** The mechanism is real and the negative controls are real.
Four defects found, all four fixed here rather than filed. One of them is the vacuity hole this PR
is otherwise built to close, and it was reachable by a one-line edit.

---

### 1. The rule could be made green over nothing, by an ordinary `files` edit — FIXED

The trigger is *`files` names a `gtk/` payload*. `files` is an ordinary field. Narrowing the win32
bundle's to `gtk/bin` + `gtk/lib` + `gtk/share` — a plausible way to trim a tarball — takes the
package out of the trigger, and nothing anywhere says so. Measured on the real manifest, on this
branch as submitted:

```
media-capabilities: 2 bundle(s) — 14 declared format(s), 2 declared gap(s), 0 payload(s) inspected here
bundled-license: 5 bundling package(s) — 5 defer to a shipped notice
EXIT=0
```

Three things at once, all silent:

- the whole MP3/Vorbis/FLAC declaration still travels in the tarball, held by nothing;
- **the coverage pass goes green** — the only bundle that speaks about a gap is gone, so the two
  darwin bundles agree with each other perfectly;
- `bundled-license` loses the same package in the same edit, which is what lets a bundle carrying
  37–45 relocated LGPL/GPL libraries declare `license: MIT`.

`field-coverage` cannot see any of it: it matches key NAMES across the tree, so the bundles that
stayed satisfy coverage for the one that left. And `gst-plugins.mjs` still sees all three, because
it keys on the DIRECTORY name — so the "one declaration, three readers" claim quietly becomes two
readers with different membership.

Fixed in `fix(manifest-conformance): hold the trigger too`: the collector also takes a package that
DECLARES `gjsify.mediaCapabilities` without shipping the trigger, and the audit fails it by name,
then skips it so one untriggered package cannot make every real bundle answer for formats it alone
names. Re-run of the same control:

```
MEDIA-CAPABILITY FAILURES on 1 finding(s):
  - @gjsify/gtk-runtime-win32-x64 (packages/node-gi/gtk-runtime-win32-x64): declares
    `gjsify.mediaCapabilities` and its `files` ship no `gtk/` payload directory …
EXIT=1
```

Two e2e cases hold it, plus one that the failure names only the rogue package.

`bundled-license`'s half of the same blind spot is NOT closed here — it is pre-existing and out of
this PR's scope, but the mechanism to close it now exists one file over.

### 2. `initFonts()` lost its early return — FIXED

`PangoCairo.FontMap.get_default()` moved above the `dir === undefined` branch, so a call with no
staged directory and no `expectedFamilies` now INSTANTIATES the default font map. That is the
ordinary case — `GJSIFY_FONT_DIR` is unset unless `gjsify ship` staged a directory — and the module
doc promises it "does nothing quietly" there.

The comment beside it still said *"There is one early return below"*, describing code that had been
replaced by an `if (dir !== undefined) { … }` block. A comment that describes control flow which no
longer exists is the reason the change reads correct. Restored, with `expectedFamilies` in the
condition so the macOS shape still gets its answer.

### 3. Two disagreeing regexes for one field — FIXED

The rule accepted `^[a-z0-9_]+$` for `element`; `gtk-runtime-bundle-gates.test.mjs` asserted
`^[a-z0-9]+$` over the same field. So `avdec_aac` — the element this ADR names as the one that
would close the AAC gap — was legal in the manifest and illegal in the test.

Lifted to one `isGstElementName`, in a new `lib/gst-payload.mjs` that also takes
`gstPluginBaseName`, `isGstPluginFile` and `readGstPluginDir` out of the RULE module. They were
inside the rule and `gst-plugins.mjs` imported them from there, so both bundle builders and the
on-target probe enrolled a conformance rule into the registry as a side effect of asking what a
plugin file is called (`defineRule` runs at module scope). Same arrangement `binary.mjs` and
`check-prebuild-loader-path.mjs` already have.

### 4. The guide that documents this API was not updated — FIXED

`website/src/content/docs/guides/bundled-fonts.md` is where `initFonts()` is documented. It still
said *"That is the whole API"* over the four fields it had, and its § Check that it worked taught
the hand-rolled `list_families()` scan #1542 asked to be replaced. Two corrections beyond the new
fields, both of them this branch's own measurements:

- **the premise is false on one platform.** "Nothing anywhere says a word" — win32 prints
  `couldn't load font …, falling back to "Sans …"` at layout time, while the fontconfig host printed
  nothing for the same invented family. This is the same correction the PR makes to #1542, and it
  belonged in the doc that carries the sentence.
- **the family NAME differs per platform, and the guide never said so** — one byte-identical face is
  `Merriweather` under fontconfig and `Merriweather 18pt` under the win32 bundle, so a reader
  following this page shipped Tahoma. That is the whole of #1542 and the page it happens on was the
  only one not mentioning it.

The measured transcripts are left exactly as they were measured; nothing new is quoted as a run.

---

### What held up, checked rather than taken

- **The evidence split does hold.** A Linux run asserts three things — declaration shape,
  cross-bundle format coverage, and `element` as a NAME — none of which is a claim about a foreign
  OS. The payload pass runs only where a payload is offered, and each `gtk-os-suites.yml` leg passes
  `--media-payload` for its OWN target only. No `selfReading` back door: the two oracles are a file
  listing and a running registry, and the one place the manifest feeds a builder
  (`missingBundledGstPlugins`' gap classification) SILENCES a build failure with a printed reason —
  the sanctioned `platformsUncommitted` shape, not a self-confirming artifact.
- **`%TARGET%` on the win32 leg expands**: that job sets `defaults.run.shell: cmd`, and line 563
  already used the same spelling.
- **Part C asserts the flip, not the spelling.** Both `it.failing` font tests RAN and PASSED here
  (`when: !REGISTRATION_SUPPORTED` is false on Linux) — not `it.skip` in disguise, confirmed in the
  run output. The pure matcher spec does pin `Merriweather 18pt`, correctly: those are MEASURED
  fixture lists, and the thing under test is the matcher.
- **The `uri_protocol_is_supported` correction is right, and the older note is not wrong.** The
  0.45.0 measurement predates #1476: that bundle carried 20 plugins and no `gstsoup.dll` at all.
  #1544 re-measured on 0.47.0 with soup resolving. The two statements are about different payloads;
  only the version-scoping was missing.
- **#1626 is not a duplicate of #1544.** #1544 closed on the CHECK (#1562's `GST_PLUGIN_GAPS`); its
  own § "What would fix it" names the payload as the second, separable half and explicitly does not
  decide it. #1626 is that half.

### Two limits worth writing down rather than fixing

- **The coverage pass sees asymmetry, never a shared silence.** Delete a format from all three
  manifests and nothing is red: for win32 the seed list catches it (`flac` expected, not shipped,
  not excused → undeclared), but for a format every bundle ships — Opus — the contract narrows
  silently and `gst-elements.test.mjs` simply stops asking. Narrowing is the safe direction, and the
  seed list is the backstop, but the rule cannot be the one that notices.
- **The PR body's longest line is 99/100.** `check-pr-body-lines` passes with one character to
  spare; any edit to it wants a re-run.

### Also on the branch

`main` took 0053 and 0054 (#1630, #1628) while this PR held 0053. Merged and **renumbered to ADR
0055**, with every reference moved and the PR body's own heading with it — the shared-counter
collision two parallel PRs pay on whichever lands second. `check-adr-index`: 55 ADRs, statuses agree.

### Verified, on the merged tree

| | |
|---|---|
| `node --test tests/e2e/media-capability-declaration/run.mjs` | 26/26 (23 → 26) |
| `node --test` bundle-gates + gst-elements | 58 pass, 2 skipped (no bundle on Linux, by design) |
| `node scripts/audit-runtimes.mjs --check` | exit 0, `3 bundle(s) — 18 declared format(s), 7 declared gap(s), 0 payload(s) inspected here` + 3 `payload NOT INSPECTED` notes |
| the fourth negative control, before and after the fix | exit 0 silently → exit 1 naming the package |
| `gjsify workspace @gjsify/gtk-host run test` (GJS 1.88.1) | 664 tests, 3013 assertions, 0 failed |
| `gjsify workspace @gjsify/gtk-host run test:gjs-on-node` (Node 24.19.0) | 664 tests, 3012 assertions, 2 pre-existing `it.failing` |
| `gjsify lint` / `gjsify format --check` | 0 errors / clean |
| `check-agent-context-size --check`, `check-adr-index`, `check-changelog-references`, `check-e2e-suite-coverage`, the four workflow checks | clean |
| `commitlint --from origin/main --to HEAD` | 0 problems |
| `check-pr-body-lines`, `check-pr-title-subject` | 99/100 and 91/100 |

No VM was started. The Windows and macOS measurements in the description are taken on trust and
labelled as such; nothing here re-measured them.

`status/agent-context-budget.json` and every AGENTS.md are still untouched by this branch's own
commits.

PR #1630, opened 2026-09-10
docs(adr): parse blueprint into SharedNode

JumpLink:
ADR 0033 enforces that a widget tree is DECLARED, and on GTK the declarative form is
Blueprint. ADR 0051 authors one tree both renderers build — and does not use Blueprint:
its corpus is `SharedNode { tag, slot?, props?, children? }` in GIR class names, chosen
because authoring in either renderer's markup "would make one of them the reference and
the other a translation".

Read quickly, that rules Blueprint out. Read carefully, it does not, because it is about
a different LEVEL. GIR is the vocabulary. Blueprint and `SharedNode` are both notations
over it. `adw-*` elements and GtkBuilder XML are the runtime formats underneath them.
0051's argument targets that bottom layer, and Blueprint compiles TO GtkBuilder XML
rather than being it. (Full mapping table in a comment below, kept out of commit history.)

## The census, and what it corrected

Measured across every tracked `.blp`. SIX construct classes stand outside `SharedNode`,
not the two an earlier draft of this PR claimed: `template` (11), object ids (41), `_()`
(23), `bind` (6), `Adw.Breakpoint`'s `condition`/`setters` (6+6), and object-valued
properties (19). Zero signal handlers, zero menu blocks, zero inline adjustments.

They are not the same kind of outside. `template` is not a tree construct at all but a
file-level declaration. An object id is addressing, and it is what `bind` resolves
against — a notation without ids cannot express `bind`. `_()` is the expensive one: the
value survives as a string, the translatable MARKING does not, and that marking is the
whole reason ADR 0033 prefers a template.

The gap runs both ways: `SharedNode.props` admits `string | number | boolean`, so none of
the portable values from ADRs 0042, 0046 and 0047 can appear in a shared tree, while
Blueprint writes all three inline. Neither notation contains the other.

## What is decided

1. The parser produces a full AST; `SharedNode` is a DECLARED LOSSY PROJECTION of it. One
   representation cannot serve both halves — clause 4 needs every construct, the
   projection carries none of the six. Losses are named at the seam, not found downstream.
   Blueprint is a second READER, not a second authoring surface: ADR 0051 Decision 1 keeps
   `ADWAITA_GALLERY_SHARED_TREES` the corpus.
2. The equivalence is PROVED, not asserted — a hand-written `SharedNode` expectation per
   corpus file. ADR 0051 rejected a markup translator on a measurement; the projection is
   a translator in one direction and carries that burden rather than an exemption.
3. Subset scope; an unrecognised construct is a hard error naming its line.
4. `blueprint-compiler` proves the emitted XML and nothing built from the AST afterwards.
   It does NOT retire ADR 0028 § 6's other use of the same binary — typelib validation,
   which a parser does not perform: `Gtk.Box { spacinng: 4; }` parses clean.
5. Shadow run until silent, then authoritative.
6. The corpus is WRITTEN, not collected.
7. Done is a deletion list: `resolve-compiler.ts` + spec (505 lines that only find a
   binary), the line-level `oxlint-disable`, the MSYS2 branch of `system-check`. The
   doc-fence skip becomes TWO-STAGE rather than vanishing. `@gjsify/storybook` is a
   separate item — `.oxlintrc.json` scopes the package off the rule, so it needs a scoping
   decision, not a deletion.

## What is deliberately NOT decided

Which notation the corpus is AUTHORED in stays ADR 0051's. The build-from-one-source
horizon stays where 0051 Decision 6 left it. `.blp` as an EMITTED dialect needs no parser
and is separate. Where `SharedNode` has to live before a package can produce it is named
as the first unresolved implementation question rather than assumed away.

## Scope

Docs only: ADR 0053, ADR 0033 to Accepted, the index rows, three `status/open-todos.md`
entries. No AGENTS.md change. ADR 0033 is moved to Accepted because it is already binding
in practice — AGENTS.md, `docs/code-anti-patterns.md` and `.oxlintrc.json` all treat it so
— not because a gate enforces it; 0033 explicitly declines a gate.

Two independent reviews ran over this branch: one fact-checking every number against the
tree, one checking each clause against the ADRs it touches. Both found real defects. The
census above and the AST/projection split in clause 1 are their corrections.

This branch was rebuilt on current `main` after 0034–0052 landed mid-work.

https://claude.ai/code/session_01KAGzf2HfFRVGeAv3u3tr9J

### DIFF
--- a/docs/adr/0033-declarative-templates-preferred.md
+++ b/docs/adr/0033-declarative-templates-preferred.md
@@ -1,9 +1,9 @@
 # 33. A widget tree is declared in a template file; TypeScript holds the behaviour
 
-- Status: **Proposed**
+- Status: **Accepted**
 - Date: 2026-08-28
 - Deciders: Pascal Garber
-- Related: [ADR 0027 (GTK host layer)](0027-gtk-host-layer.md), [ADR 0028 (widget table provenance)](0028-widget-table-provenance.md), [ADR 0032 (React Native on the GTK host)](0032-react-native-on-the-gtk-host.md)
+- Related: [ADR 0027 (GTK host layer)](0027-gtk-host-layer.md), [ADR 0028 (widget table provenance)](0028-widget-table-provenance.md), [ADR 0032 (React Native on the GTK host)](0032-react-native-on-the-gtk-host.md), [ADR 0053 (Blueprint parsed in-repo)](0053-blueprint-parsed-in-repo.md)
 
 ## Context
 
--- a/docs/adr/0053-blueprint-parsed-in-repo.md
+++ b/docs/adr/0053-blueprint-parsed-in-repo.md
@@ -1,0 +1,183 @@
+# 53. Blueprint is parsed in-repo, into the node shape ADR 0051 already renders
+
+- Status: **Accepted**
+- Date: 2026-09-10
+- Deciders: Pascal Garber
+- Related: [ADR 0002 (bootstrap bundle minimization)](0002-bootstrap-bundle-minimization.md), [ADR 0028 (widget table provenance)](0028-widget-table-provenance.md), [ADR 0029 (girs widget vocabulary)](0029-girs-widget-vocabulary.md), [ADR 0030 (one corpus, GJS as oracle)](0030-one-corpus-gjs-as-oracle.md), [ADR 0033 (declarative templates preferred)](0033-declarative-templates-preferred.md), [ADR 0051 (one authored tree, rendered)](0051-one-authored-tree-rendered.md)
+
+## Context
+
+Two things are true at once in this repo and have not been connected. ADR 0033 enforces that a
+widget tree is DECLARED rather than assembled, and on GTK the declarative form is Blueprint.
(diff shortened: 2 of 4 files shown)


JumpLink:
The mapping table, kept out of the PR body so it does not become commit history.

Measured across every tracked `.blp`, 2026-09-10.

| Blueprint | `SharedNode` | GIR-derived? |
|---|---|---|
| `using Adw 1;` | carried by the class name | yes — namespace and version |
| `Adw.HeaderBar { }` | `tag: 'AdwHeaderBar'` | yes — the GIR type |
| `title: "…"` | `props: { title: '…' }` | yes — a ParamSpec |
| `[start]` | `slot: 'start'` | yes — ADR 0029 § 4 derives slot candidates from GIR |
| `styles ["flat"]` | `cssClasses: ['flat']` | yes — ADR 0049 decided style classes are a list |
| `template $Foo: Adw.Bin` | — | **no** — a GtkBuilder composite-template declaration |
| `bind …` | — | **no** — a GObject property binding |

This table is a *reading* of two notations, not a measurement — which is why clause 2 turns it into a suite rather than a claim.

PR #1633, opened 2026-09-10
feat: win32 decodes ogg/vorbis, and the licence rule cannot be switched off

JumpLink:
Stacked on #1629 (`fix/bundle-closure-fonts-and-audio`), which brought the media
DECLARATION and the rule that holds it. This is the payload decision it deliberately left
open (#1626), plus the escape route its review found next door and did not touch.

## The win32 payload: one of three gaps was an omission, two are absences

The Windows GStreamer is not lifted from anywhere. The published
`GTK4_Gvsbuild_<version>_x64.zip` carries none of it, so `node-gi.yml` and `release.yml`
build gstreamer/-base/-good from source with gvsbuild, into that same prefix. **The
argument list is ours; the projects are not** — and that is where the payload is decided,
not in the bundle builder and not in `GST_AUDIO_PLUGINS`.

Read from the pinned `wingtk/gvsbuild@2026.6.0`: `ogg.py` and `libvorbis.py` are in
`gvsbuild/projects/`; there is no `flac.py`, no `mpg123.py`, and no way to load a project
definition from outside the package.

- **Ogg/Vorbis falls.** `libvorbis` joins the first gvsbuild invocation beside `libsoup3`,
  for the reason already written there: the `vorbis` feature is meson `auto` and
  gst-plugins-base does not declare libvorbis as a dependency, so nothing ordered the two
  and -base built green with no `gstvorbis.dll`. The bundle shipped an `ogg` demuxer with
  nothing behind it. The declaration moves with the payload.
- **MP3 and FLAC do not.** gvsbuild has no project for libmpg123 or libFLAC. Not a licence
  question either — libmpg123 is LGPL-2.1, libFLAC BSD-3-Clause. Closing them means a
  project file upstream, and each gap's `why` now says so instead of describing a prefix
  nobody outside this repository can see.

ADR 0056 records the rule that answers the next codec request, and the two alternatives
rejected: MSYS2 side-loading, and building the two libraries beside gvsbuild ourselves.

## The first Windows run refuted this, and that is the most useful thing in the PR

Naming the gvsbuild project was NOT enough. Run 34502880383 built the prefix and came back
without the plugin, because `libvorbis` is in the catalogue and does not build with the
CMake the catalogue ships:

- libogg 1.3.6 declares `cmake_minimum_required(VERSION 3.6)` — configures.
- libvorbis 1.3.7 declares `VERSION 2.8.12`, gvsbuild fetches CMake 4.3.3, and CMake 4
  removed compatibility below 3.5 — `Error: libvorbis build failed`.

That one line is why the `ogg` demuxer was in the payload all along and the Vorbis decoder
never was. Not availability, not licensing, and invisible from a project listing. So step 1
of ADR 0056's rule is necessary and NOT sufficient, and § 2 now says so.

**Three silences the same run exposed, all now named failures.** A failing gvsbuild
invocation did not fail the step — `$ErrorActionPreference = 'Stop'` does not act on a
native command's exit code and pwsh tests `$LASTEXITCODE` only for the last line, so the
first invocation died, the second succeeded and the step went green. `vorbis` in ‑base and
`soup` in ‑good are meson `auto`, which answers a missing library with a `message()` and a
green build; both are pinned to `enabled` through gvsbuild's own `--extra-opts`, so the
library errors at the project that needs it. And the cache key really had moved: the failed
prefix was never saved, so the retry was a clean miss.

## What is measured now, on the artifact itself

Run 34530994036 is green on the bundle build, and the built tarball was downloaded and read:

- `Run-time dependency vorbis found: YES 1.3.7` · `Linking target ext/vorbis/gstvorbis.dll`
- `lib/gstreamer-1.0/gstvorbis.dll`, 775 680 bytes — and no `gstmpg123.dll`, no `gstflac.dll`
- `GStreamer — 22 plugin(s)`, `windowingData.gstPlugins: 22`, up from 21
- `licenses/libvorbis/COPYING` — the Xiph BSD-3-Clause text, in the tarball, named once in
  `THIRD-PARTY-NOTICES.md`
- zero vorbis or ogg DLLs in `bin/`, confirming the static-CMake prediction the licence
  family entry was added against
- the builder printed its three remaining DECLARED GAPs (mpg123, flac, wasapi2) and no
  undeclared absence

And the other half, which a file list cannot give: on `windows-latest` with no gvsbuild on
the host, `gst-elements.test.mjs` passed 8 of 8. It calls `Gst.ElementFactory.make()` for
every format the host bundle's manifest claims and asserts none comes back null, with a
vacuity guard on the loop — so **`vorbisdec` resolving is what its green means**. Its
sibling assertion, `a declared decoder gap is still a gap`, says `mpg123audiodec` and
`flacdec` are still null there, so the two remaining gaps are honest on the target too.

The file and the element are different questions and both are answered, so ADR 0056 is
**Accepted** rather than Proposed.

## `bundled-license` had the same escape route

#1629's review found that `media-capabilities` could go green over nothing: its trigger was
"`files` names a `gtk/` payload", and narrowing `files` to `gtk/bin` + `gtk/lib` +
`gtk/share` drops the package out at exit 0 with the whole declaration still in the tarball.
It fixed that rule and noted the same route stood open next door.

Measured here with that edit on the real win32 manifest: `media-capabilities` fails by name,
and `bundled-license` prints NOTHING — one package fewer audited than the repository
publishes, over a tarball that still carries the notice and the relocated LGPL/MPL/GPL
libraries. `field-coverage` cannot see it either; it matches key names across the tree.

The mechanism is the one #1629 used: a second way in that survives the narrowing. Here it is
the package's own answer — a `license` deferring to a notice INSIDE a payload directory is a
statement that the tarball redistributes one. An ordinary `SEE LICENSE IN LICENSE.md` is
untouched, because the rule is `portable` and firing on packages that redistribute nothing
is how a check gets read as noise and then switched off.

**Negative control:** `tests/e2e/bundled-license-trigger/` carries it, and the suite was
broken on purpose — against the previous rule three of its assertions go red, and they are
the three about the trigger.

## The win32 leg becomes a probe for one release cycle

`media-capabilities` audits this tree's declaration against the PUBLISHED 0.48.0 tarball,
which has no `gstvorbis.dll`. That leg's own comment already called this the honest
direction of disagreement for a widening. What it named as the repair — "land the payload
first" — is not available, and this is the first widening to prove it: the builder refuses
to let declaration and payload travel separately in either direction. A format out of
`gaps` and unclaimed fails the coverage pass; a gap whose plugin arrived fails as RETIRED
and exits 1. Both doors were run rather than argued. They move in one commit by
construction.

An earlier version of this section defended keeping the leg RED as "advisory
(path-filtered, never a required check)". Only the second half is true.
`gtk-os-suites.yml` has no `paths:` anywhere: it runs on every pull request, on every push
to `main`, and nightly. Gating the disagreement therefore reds that job on `main` and on
every unrelated contributor's PR until the release ships, for something they did not cause
— which is this repository's guard rule in the mirror, since the next real win32 media
finding then lands on a job people have learned to skip.

So the win32 step is a PROBE: `continue-on-error` + an `id` + `report-probe-outcome.mjs`,
the shape `rn-probe-win32` already uses eight steps above for the same cause. It reports
rather than silences — the audit still prints the finding, and the reporter puts a
`::warning::` on the PR plus a summary line stating the job is green ONLY because of the
flag. `check-probe-outcomes-read.mjs` refuses a probe nothing reads. Almost nothing stops
gating: `audit-runtimes.yml` runs `--check --strict` on every pull request with no paths
filter of its own, so only the `--media-payload` comparison is advisory. The darwin leg
keeps gating. Retirement is a RELEASE, not an issue number.

One thing this shape does NOT buy, stated because `it.failing` does buy it: nothing fails
the day the probe starts passing. `report-probe-outcome.mjs` reports by design — turning a
probe's outcome into a job failure would red the branch for the defect it documents — and
`check-probe-outcomes-read.mjs` only enforces that somebody reads it. So deleting this
step's flag after the release that carries `gstvorbis.dll` is a person's job, and the
label names the release it waits on.

## Verification

Green on Linux: `audit-runtimes --check`, the media and licence e2e suites, the node-gi
bundle gates, six workflow-reading e2e suites, `check-e2e-suite-coverage`,
`check-adr-index`, `check-doc-fences`, `check-changelog-references`, `gjsify lint`,
`gjsify format --check`, `commitlint`. What no Linux host can run is the win32 bundle leg.


JumpLink:
The measurement tables, kept out of the PR description because that description becomes the
squash body and `body-max-line-length` is 100 (#1628 paid for that one).

## What gvsbuild has, at the tag this repository pins

```
gh api "repos/wingtk/gvsbuild/contents/gvsbuild/projects?ref=2026.6.0" --jq '.[].name' \
  | grep -iE 'vorbis|ogg|flac|mpg|mad|lame|twolame'
libvorbis.py
ogg.py
```

| library | gvsbuild project @ 2026.6.0 | element | win32 before | win32 after |
|---|---|---|---|---|
| libogg | `ogg.py` | `ogg` (demuxer) | shipped | shipped |
| libvorbis | `libvorbis.py` | `vorbisdec` | gap | **claimed** |
| libmpg123 | — | `mpg123audiodec` | gap | gap |
| libFLAC | — | `flac`→`flacdec` | gap | gap |

`gvsbuild/projects/` is the catalogue — gvsbuild loads no project definition from outside its own
package — so "not in that directory" and "not obtainable here" are one sentence.

## The declaration, all three bundles

| format | darwin-arm64 | darwin-x64 | win32-x64 |
|---|---|---|---|
| WAV / PCM | claim | claim | claim |
| Ogg / Vorbis | claim | claim | **claim (was gap)** |
| Opus | claim | claim | claim |
| A-law, µ-law | claim | claim | claim |
| MP3 | claim | claim | gap |
| FLAC | claim | claim | gap |
| AAC (in M4A) | gap | gap | gap |
| `wasapi2` (sink, no format) | — | — | gap |

`audit-runtimes --check` reads `media-capabilities: 3 bundle(s) — 19 declared format(s), 6
declared gap(s)`, up from 18 and down from 7.

## The `bundled-license` escape route, before and after

The edit is `files: ["index.js", "index.d.ts", "gtk"]` → `["index.js", "index.d.ts", "gtk/bin",
"gtk/lib", "gtk/share"]` on the real `@gjsify/gtk-runtime-win32-x64` manifest, licence untouched.

| | `media-capabilities` | `bundled-license` |
|---|---|---|
| before this PR | fails, names the package | **silent** — no line at all, one package fewer audited |
| after | fails, names the package | fails, names the package |

The after-message:

```
@gjsify/gtk-runtime-win32-x64 (packages/node-gi/gtk-runtime-win32-x64): `license` is
"SEE LICENSE IN gtk/THIRD-PARTY-NOTICES.md", deferring the terms of the whole tarball to a
notice inside a `gtk/` payload directory, and `files` ships no `gtk` entry (`files`: index.js,
index.d.ts, gtk/bin, gtk/lib, gtk/share). …
```

And the guard broken on purpose — the new suite against the previous rule:

```
node --test tests/e2e/bundled-license-trigger/run.mjs   # rule reverted to HEAD~
ℹ tests 13   ℹ pass 10   ℹ fail 3
✖ FAILS BY NAME the narrowing that used to switch the rule off
✖ collects that package, which is the half a failure message cannot prove
✖ fails the REAL win32 bundle under the narrowing, licence and all
```

## Local run

| check | result |
|---|---|
| `node scripts/audit-runtimes.mjs --check` | exit 0 |
| `tests/e2e/media-capability-declaration` | 26/26 |
| `tests/e2e/bundled-license-trigger` (new) | 13/13 |
| `packages/node-gi/node-gi/test/gtk-runtime-bundle-gates.test.mjs` | 53/53 |
| six workflow-reading e2e suites | 180/180 |
| `check-e2e-suite-coverage` | 153/156 run, 3 ledgered, 0 unaccounted |
| `check-adr-index` · `check-doc-fences` · `check-changelog-references` | exit 0 |
| `gjsify lint` · `gjsify format --check` · `commitlint` | exit 0 |

`@gjsify/node-gi`'s own `test` needs a built addon (`NODE_GI_NATIVE=build`) and was not run
here; the suite this PR touches is pure JS and runs on Linux, which is why it lives there.

JumpLink:
## Independent review — #1633

Reviewed against the **base branch**, not `main` (`git diff origin/fix/bundle-closure-fonts-and-audio...HEAD`), so #1629's work is not counted here. Six follow-up commits pushed to the branch; each is named below with what it changes.

**Every load-bearing claim in this PR reproduced.** Nothing at the top of this comment is a claim I could not verify. One *justification* did not: the defence of the knowingly-red leg rests on a premise that is false, and that is finding 1.

---

## 1. HIGH — the red leg was defended with a false premise (fixed in `7b22622` + `2ee62b6`)

`d0a6944` accepts `GTK suites on the shipped closure (win32-x64 / node)` being red for one release cycle, on the grounds that the leg is *"advisory (path-filtered, never a required check)"*.

```
$ grep -n "paths" .github/workflows/gtk-os-suites.yml
$ awk '/^on:/,/^jobs:/' .github/workflows/gtk-os-suites.yml
on:
  push: { branches: [main] }
  schedule: [{ cron: '30 6 * * *' }]
  workflow_dispatch:
  pull_request:
```

**There is no `paths:` filter anywhere in that workflow.** It runs on every pull request, on every push to `main`, and nightly. So the window is not this PR's red X — it is that named job red on `main`, on the nightly, and on every unrelated contributor's PR until the release ships. AGENTS.md's own governance line makes the point sharper: *"Only workflows with NO `paths:` filter are eligible"* to be a required check. This is one of the twelve.

"Never a required check" is the true half — nothing is blocked. The cost is attention, and it is paid by everyone, not by this PR.

### The dilemma itself: verified, both doors shut

I ran both rather than believing them.

**Door A — format in neither `audioDecode` nor `gaps`:**

```
$ node scripts/audit-runtimes.mjs --check      # after deleting the Ogg/Vorbis entry
MEDIA-CAPABILITY FAILURES on 1 finding(s):
  - @gjsify/gtk-runtime-win32-x64: says nothing about Ogg / Vorbis, which
    @gjsify/gtk-runtime-darwin-arm64, @gjsify/gtk-runtime-darwin-x64 declare an answer for.
```

**Door B — keep it in `gaps` while the prefix grows the plugin:**

```
$ node -e 'missingBundledGstPlugins([...,"libgstvorbis.dll",...], "win32-x64")'
retired  : ["vorbis"]
```

and `build-gtk-runtime.mjs` has `if (bundledGaps.retired.length > 0) { console.error(…); process.exit(1); }`.

Both shut. **The decision itself stands.** Declaration and payload do move in one commit by construction.

### The third door — it exists, it is adjacent, and I took it

`continue-on-error` + an `id` + `report-probe-outcome.mjs`, with a retirement condition naming a RELEASE. That is not a proposal; `rn-probe-win32` sits *directly above* the conformance step in the same job, for the near-identical cause ("this leg CANNOT see an engine fix that has not been released yet"). It is governed: `check-probe-outcomes-read.mjs` refuses a probe whose outcome nothing reads, and `tests/e2e/ci-probe-outcomes/` holds both halves. A probe is not silence — it emits a `::warning::` and a step-summary row naming the finding.

And it would cost no unique gating on that step:

```
$ grep -n "audit-runtimes" .github/workflows/audit-runtimes.yml
78:  run: node scripts/audit-runtimes.mjs --check --strict     # on: pull_request (no paths:), push main
```

The os-axis half of that step is already held on every PR, more strictly. **Only the `--media-payload` comparison is unique to this leg** — and that is exactly the half which legitimately disagrees during a widening.

**Decision, taken on the branch (`2ee62b6`):** the win32 step is now a **probe** — `continue-on-error` + an `id` + `report-probe-outcome.mjs` — with a retirement condition naming a release. The darwin step keeps gating.

The question that decides it is whether the reporter keeps the finding visible or merely swallows it. It keeps it. Driven exactly as CI drives it:

```
$ PROBE_LABEL='Conformance audit — win32 media claim vs the PUBLISHED payload (red until …)' \
  PROBE_OUTCOME=failure node scripts/report-probe-outcome.mjs
::warning title=Probe failed (not gating)::Conformance audit — win32 media claim … exited non-zero; read the job log for the counts
❌ **probe** `Conformance audit — win32 media claim …` — the probe FAILED — the job is green only because the step is continue-on-error
```

An annotation that lands **on the PR**, a job-summary row, and a sentence that says out loud the job is green only because of the flag — plus the audit's own finding still printed in the log, because the step still runs. `check-probe-outcomes-read.mjs` now counts 19 such steps, every outcome addressable and read.

The honest counter-argument, which I weighed and rejected: your red **self-retires** with the release, a probe needs a human to delete `continue-on-error`, and this repository has measured probes outliving their condition — `rn-probe`'s own comment records that its condition was wrong, and #1556 is two red probes nobody counted. What tips it is the corrected scope. "One advisory leg on the PRs that touch these paths" would have made the red clearly right; "every PR in the repository, plus `main`, plus the nightly, for weeks" makes it a bill other people pay for a change they did not make. I would rather owe a deletion than teach everyone to skip a check — and the deletion is anchored to a release rather than an issue, which is precisely the mistake the probe above it records paying for.

There is also a **structural** answer worth a separate issue, not this PR: the step compares *this tree's* declaration to the *published* payload. The staged tarball carries its own manifest. Auditing published-manifest against published-payload is what "does the artifact a stranger downloads keep this promise" literally means, it can still be red for a true reason (a release that shipped a broken promise), and it can never be red for a widening in the working tree. The crossed pairing produces no unique true finding — a widening that will not be backed is already caught by the named prefix assertion and by `undeclared`, on a fresh payload, which is the right payload for a tree-level claim.

`7b22622` corrects the workflow comment and the ADR consequence and records that both doors were *run*; `2ee62b6` converts the step. I also corrected the same sentence in the **PR description**, since it becomes the squash body and the false claim would otherwise land on `main` permanently.

---

## 2. The catalogue claim — reproduced, and one half was under-proved

```
$ gh api "repos/wingtk/gvsbuild/contents/gvsbuild/projects?ref=2026.6.0" --jq '.[].name' | wc -l
95
```

`ogg.py` and `libvorbis.py` present; no `flac.py`, `mpg123.py`, `mad`, `lame`, `twolame`. ✔

The file listing alone does not prove there is no project — `gstreamer.py` alone defines eleven. So I checked the stronger form over the extracted tag:

```
$ grep -rniE "flac|mpg123|lame|twolame|libmad" gvsbuild/
gvsbuild/patches/gettext/msvc/.../unictype.h:539:  UC_JOINING_GROUP_MANICHAEAN_LAMEDH
gvsbuild/patches/gettext/nmake/Makefile.vc:1:# NMake Makefile for building libflac as a DLL
gvsbuild/patches/gettext/msvc/.../unictype.h:539:  …
```

Three hits, none a project definition. **No project defines those libraries under any file name.** Worth putting in the ADR — it is the claim that carries "MP3 and FLAC cannot be built in that prefix", and a directory listing does not carry it.

The second half — no external loading — holds:

```python
# gvsbuild/projects/__init__.py, all five lines
for module_info in pkgutil.iter_modules(__path__):
    importlib.import_module(f"{__name__}.{module_info.name}")
```

over the package's own `__path__`, with no plugin entry point in `pyproject.toml`. The directory is the catalogue for any supported use.

**One citation you can upgrade for free.** The ADR proves "the published zip carries no GStreamer" by a file list read from a tarball. gvsbuild's own release workflow proves it directly — `.github/workflows/ci.yml` builds `gtk4 libadwaita gtksourceview5 pygobject protobuf-c` and zips the prefix. That is the producer saying it, not an artifact being measured.

---

## 3. The Vorbis finding — reproduced from first principles, and it is the best thing here

Three independent facts, each read from source:

```python
# gvsbuild/projects/gstreamer.py — GstPluginsBase
dependencies=["meson","ninja","gstreamer","opus","ogg","graphene","pango"]     # no libvorbis
```
```meson
# gst-plugins-base-1.28.4/meson.options:66
option('vorbis', type : 'feature', value : 'auto', ...)
# ext/vorbis/meson.build
vorbis_dep = dependency('vorbis', version: '>= 1.3.1', required : get_option('vorbis'))
if vorbis_dep.found()
  gstvorbis = library('gstvorbis', ...)
endif
```

`ogg` **is** a declared dependency and libvorbis is not, and the feature is `auto` — so `-base` built green, shipped `gstogg.dll`, and never built `gstvorbis.dll`. **The bundle shipped a demuxer with nothing behind it, exactly as claimed.** This is the "green because nobody asked" class and it is the most valuable part of the PR.

Two pieces of supporting evidence you did not use, which strengthen the *unmeasured* half:

- libvorbis 1.3.7's CMakeLists installs `vorbis.pc` / `vorbisenc.pc` / `vorbisfile.pc` into `lib/pkgconfig` — which is precisely what meson's `dependency('vorbis')` looks for, so the `auto` feature will actually flip.
- gvsbuild builds `-base` with `-Dc_link_args=<prefix>\lib\ogg.lib`, already forcing static ogg onto the whole `-base` link line. That is the thing a statically-built libvorbis needs in order to link into `gstvorbis.dll`.

## 4. The three safeguards — checked separately

| | verdict |
|---|---|
| prefix assertion names the file | ✔ `foreach` over `gstsoup.dll` + `gstvorbis.dll`, `exit 1`, in **both** workflows. Only these two build GStreamer into the prefix; the other four gvsbuild jobs use the plain zip and are untouched, correctly. |
| absent `vorbis` = UNDECLARED, exit 1 | ✔ measured — moving the format out of `gaps` puts it in `expectedGstPlugins` minus `declaredNames`, and the builder's `undeclared` branch is `process.exit(1)`. Held by the new unit test in both directions. |
| cache key moves | ✔ **and this is the one that decides it**: neither cache step has `restore-keys:`, so `actions/cache@v4` requires an exact key match and cannot fall back to the `-v3` prefix. `grep -rn "gst-soup" .github/workflows/` returns exactly the two new keys — no leftovers. |

## 5. HIGH — the `ogg` licence family could not do the job it is cited for (fixed in `27c5495`)

The new `libvorbis` entry is declared ahead of any match so that a gvsbuild bump flipping it to shared cannot fail the release on a leaf belonging to no family — and its `why` cites the `ogg` entry above as precedent. The precedent could not have prevented one:

```js
{ components: ['ogg'], pattern: /^ogg-\d+\.dll$/i },     // the MESON spelling
```

ogg is a **CMake** project. Measured on the pinned sources: libogg 1.3.6 and libvorbis 1.3.7 both `option(BUILD_SHARED_LIBS ... OFF)` and both set `SOVERSION` on the target — and on Windows CMake spends `VERSION`/`SOVERSION` on the **image-version resource**, leaving the file `ogg.dll`. Only meson puts the soversion in the leaf, which is why the prefix carries `opus-0.dll`. So the bump this entry guards against lands `ogg.dll`, matches nothing, and hits `bundle-licenses.mjs`' *"belongs to no declared license family"* — the exact accident it exists to prevent.

Both spellings now, on both entries, each asserted. Negative control: the new assertions fail against the old pattern (52 pass / 1 fail).

Related: the new test asserted `familyOf('vorbis-0.dll')` — the spelling that cannot occur — and not `vorbis.dll`, which is the one that would. Both are asserted now, plus `vorbisfile.dll`.

## 6. MEDIUM — the "second way in" is conditioned on the licence shape (my own negative control; fixed in `5425e10`)

Your negative control reproduces exactly:

```
$ node --test tests/e2e/bundled-license-trigger/run.mjs           # new rule
tests 13 | pass 13 | fail 0
$ …with the pre-#1633 rule restored
tests 13 | pass 10 | fail 3     ← the three trigger assertions
```

**Mine, which you do not have.** The rule accepts *two* licence shapes and `noticeInsidePayload` only reads one:

```
A  SEE LICENSE IN gtk/… , files narrowed   -> collected: 1 | failures: 1    ← closed
B  compound SPDX,        files narrowed   -> collected: 0 | failures: 0    ← STILL OPEN
C  notice outside gtk/,  files narrowed   -> collected: 0 | failures: 0    ← still open
D  compound SPDX,        files intact     -> collected: 1 | failures: 0    ← control
```

`MIT AND LGPL-2.1-or-later` is a licence this rule **accepts** for a bundling package — your own suite asserts it — and it names no notice path, so the narrowing takes such a package out at exit 0 exactly as before. All six bundles use the notice form today, so the hole is latent, not live. But the PR, the rule doc and the test header all say "the trigger is held against itself" without that qualifier.

I did **not** widen the `files` trigger to match a path segment, although that would close it: the rule is `scope: 'portable'`, which the registry defines as exposed to consumers via `gjsify manifest-check`, and `bin/cli.js` is an ordinary npm `files` entry — firing on packages that redistribute nothing is what your neighbouring test exists to prevent.

What actually closes it is the enumerated `BUNDLING_PACKAGES` list, so I made its precondition machine-checked instead of assumed: every bundling package is now asserted to use the `SEE LICENSE IN <payload>/…` form. Broke it on purpose — switching the real win32 manifest to a compound expression turns it red by name. Suite is 15/15; against the pre-#1633 rule, 12/3.

## 7. MEDIUM — a live fact this PR's own change invalidates (fixed in `0d78bb9`)

`gst-plugins.mjs`, untouched by the PR:

> Seeding a plugin here is a request, not a promise. `mpg123`, `vorbis` and `flac` are seeded **and the win32 archive has none of them** …

One third of that becomes false in the same tree. The point the sentence makes is right and stays; the enumeration goes — it is the live value AGENTS.md says not to restate.

## 8. LOW — craft (fixed in `4fd2f09`)

- `SEE LICENSE IN` was parsed twice per package (`noticeInsidePayload`, then `auditBundledLicense`): two answers to "is this a deferral" that can drift, in a rule whose subject is a package leaving a check unnoticed. One helper now, both readers take the half they want.
- The `bin` note carried two live package counts — one of them to scold an earlier version for measuring the wrong set. The count that mattered had already drifted with the tree. Incident kept, numbers gone, the set asserted in the e2e suite.
- No try/catch around anything non-throwing and no paranoid probes in the diff. ✔
- **Point 5 (a rule ahead of its subject):** justified, not dead — it follows the `ogg` precedent *and*, unlike that precedent, arrives with a test. Finding 5 is what makes the precedent honest.
- Two commit subjects are over the 50-char house style (54, 51), and `f7e03ac`'s body has a stray blank line mid-sentence, which is the `footer-leading-blank` warning `commitlint` reports. Cosmetic; the squash body comes from the description.

## Verification (all run here, Linux)

`node --test` on `media-capability-declaration` + `bundled-license-trigger` (41 pass), `gtk-runtime-bundle-gates` (53 pass), six workflow-reading e2e suites (146 pass), `release-bundle-gate` (44 pass), `audit-runtimes --check`, `check-adr-index`, `check-e2e-suite-coverage`, `check-doc-fences`, `check-changelog-references`, `check-probe-outcomes-read`, `check-bundled-icon-parity`, `check-agent-context-size --check`, `gjsify lint`, `gjsify format --check`, `commitlint` (0 problems), `check-pr-body-lines` (107 lines, longest 92/100).

Plus three negative controls: yours (10/3), the licence-shape guard broken on purpose, and the `ogg` pattern reverted (52/1).

## Verdict

**Mergeable.** The catalogue claim, the Vorbis finding and all three safeguards reproduce; the Vorbis finding is a genuine "green because nobody asked" defect and the strongest thing in the PR. The escape-route closure works, with one shape-conditioned limit now written down and held.

Two notes, neither a code defect:

1. **The probe.** I took the third door rather than only naming it, because the corrected scope changes who pays: not this PR, but `main`, the nightly and every unrelated contributor's PR. If you disagree, reverting `2ee62b6` restores your red exactly and nothing else moves — the factual corrections in `7b22622` stand either way. The counter-argument is written above and it is real.
2. **ADR 0056 stays `Proposed`**, correctly. Its promotion condition is right: the first green win32 windowing-bundle run carrying both.

Issue #1144, opened 2026-08-13
macOS + deno: two GTK copies in one process, so no widget type resolves


JumpLink:
Measured on macOS 15.7.9 (darwin-x64) during the 0.38.0 cross-OS showcase matrix.

## The symptom

Every GUI showcase on `--runtime deno` dies the same way, in this order:

```
objc[…]: Class GtkMacosContentView is implemented in both
         …/homebrew/…/libgtk-4.1.dylib and
         …/@gjsify/gtk-runtime-darwin-x64/…/libgtk-4.1.dylib
GtkWidget is not registered in this process's GObject type registry
TypeError: Adw.Application has no property 'application-id'
```

Two GTK copies are loaded into one process, so the second registration loses and every
widget type lookup afterwards asks a registry that does not have it.

`gjs`, `node` and `bun` on the **same machine, same bundle** are green. Only deno loads
both. `adwaita-storybook` fails identically
(`StorybookApplication has no property 'application-id'`), so it is not a per-showcase
problem — it is the whole deno-on-macOS column.

## What is not yet known

Which of deno's loader behaviours pulls the Homebrew copy in alongside the bundled one.
That is the first measurement: gjs/node/bun resolve to exactly one, and the difference has
not been isolated.

## Why this is a policy question, not just a bug

The bundled-runtime precedence question is already open in `status/open-todos.md`
("The GTK-runtime bundle precedence question is still open") and this touches
[ADR 0023](docs/adr/0023-gtk-source-precedence.md) — the app shipper installs the bundle, a
per-OS default decides, from-source never. What ADR 0023 does not yet answer is what
happens when a *system* GTK is present anyway and a runtime resolves both: whether the
bundle must refuse to load, must win, or must detect and report. A fix that picks one
silently would be deciding the ADR by implementation.

So: decide the precedence rule first, then implement. Filed to keep the measurement, which
otherwise only exists in a session log.

JumpLink:
Reframing this after reading the code rather than the symptoms. **The policy question I
opened this with is not the blocker**, and there is a documented candidate mechanism
already sitting in the tree.

## The repair for this exact signature already exists — and already ran

`activateGiLibraryPath()` (`packages/node-gi/node-gi/gtk-runtime.js`, added in #1132) is
written for exactly the three errors in the report, and names them:

> Node papers over it in `maybeReexecForGtkRuntime` by re-execing with
[...]

Issue #1120, opened 2026-08-11
darwin: the node-gi addon prebuild is never relocated, so a Homebrew host loads GObject twice


JumpLink:
## Symptom

`npx @gjsify/cli@latest showcase excalibur-jelly-jumper --runtime node` on macOS (Intel, Homebrew GTK 4.22.4 / libadwaita 1.9.2, gjsify 0.35.0):

```
** (process:736): WARNING **: node-gi: a Template was set on JellyJumperWindow, which is
   not a Gtk.Widget subclass — composite templates require a Gtk.Widget ancestor
TypeError: Adw.Application has no property 'application-id'
    at new Application (.../@gjsify/node-gi/gi.js:1483:31)
```

Neither statement is true: `JellyJumperWindow` extends `Adw.ApplicationWindow`, and
`application-id` is inherited from `GApplication`. Reduced, **every** property lookup
fails, including on `Gio.Application`, which declares `application-id` directly:

```
Gio.Application -> FAIL: Gio.Application has no property 'application-id'
Gtk.Application -> FAIL: Gtk.Application has no property 'application-id'
Adw.Application -> FAIL: Adw.Application has no property 'application-id'
```

The same script against a locally-built addon passes all three.

## Cause

`DYLD_PRINT_LIBRARIES=1` shows GLib loaded **twice**:

```
.../@gjsify/gtk-runtime-darwin-x64/gtk/lib/libgobject-2.0.0.dylib
/usr/local/Cellar/glib/2.88.2/lib/libgobject-2.0.0.dylib
```

`packages/node-gi/node-gi/scripts/stage-prebuild.mjs` only `copyFileSync`s, so the
published `node_gi.node` keeps the build runner's absolute
`/usr/local/opt/glib/lib/libgobject-2.0.0.dylib` — the #1102 defect, on the one darwin
artifact `scripts/relocate-macho.mjs` does not cover (node-gi has its own stager).

It fails differently from an unrelocated dylib, which is why it went unnoticed for eight
releases: those paths *resolve* on an ordinary Homebrew Mac, so there is no
`ERR_DLOPEN_FAILED`. And `DYLD_FALLBACK_LIBRARY_PATH` — what `maybeReexecForGtkRuntime()`
re-execs with to select the bundle — is consulted only for a path that fails to resolve,
so it cannot redirect one that does. The addon binds Homebrew's libgobject; the bundle's
libgtk/libadwaita bind the bundle's via their own `@loader_path`.

[...]

PR #1634, opened 2026-09-10
fix(gtk-runtime): the bundle's plugin reached past the bundle

### LINKED ISSUE #1536: A bundled GStreamer plugin g_module_opens Homebrew's libsoup, so two GObject type systems land in one process

JumpLink:
Closes #1536.

A bundled GStreamer plugin resolved Homebrew's libsoup instead of the one beside it in
the bundle, so two GObject type systems landed in one process — ADR 0023 § 4's invariant,
broken a third way and at neither layer #1120's and #1144's fixes reach.

## The route, and it was one load command

`libgstsoup.dylib` does not link libsoup; it opens it at runtime by bare leaf name. So
`otool -L` is innocent and every gate that reads it reported a self-contained bundle.
What the image carried instead was a single `LC_RPATH` naming Homebrew's libsoup keg —
the entry its Homebrew build was given so that lookup would work, copied into the bundle
along with the plugin.

`DYLD_PRINT_SEARCHING` on a macOS 15.7.9 x86_64 host with the reporter's Homebrew set
settles what had been argued from the file format. dyld expands a **bare leaf against the
calling image's `LC_RPATH`** — its own phrase is `leaf name using rpath` — and reaches it
**before** the default fallback, which for a leaf `dlopen` is `/usr/lib` alone. Homebrew's
`/usr/local/lib` is never consulted for this lookup. The plugin's own rpath was therefore
not one route among several; it was the only one.

Repointing that entry inside the tree and re-signing makes the same lookup land on the
shipped library, with Homebrew absent from the search. Measured on the same host.

Two assumptions in the issue turn out to be wrong, and both are recorded rather than
quietly dropped: a leaf-name `g_module_open` **is** reachable by a load command, and
`DYLD_FALLBACK_LIBRARY_PATH` had nothing to do with it.

## Why nothing saw it

`relocate()` rewrote `LC_ID_DYLIB` and `LC_LOAD_DYLIB` and never `LC_RPATH`, and
`verifyRelocation()` read only `otool -L`. `stage-prebuild.mjs` has replaced the whole
rpath list for a long time and `docs/prebuilds.md` explains why — two stagers, one
discipline, applied in one of them. Same shape as #1120, where node-gi's stager was the
second one.

## What is here

- **ADR 0057** — inside a bundle an absolute `LC_RPATH` is a redirect, not a fallback,
  and why the prebuild rule that calls the same load command a working fallback stays
  correct where it was made.
- **The builder** replaces the whole rpath list, order included; empty by default, and the
  GStreamer plugins take the payload's own `lib/`. `verifyRelocation()` reads search paths
  and fails a build-host absolute one, with the derived predicate.
- **`bundle-search-paths`**, a portable conformance rule that reads the finished payload's
  Mach-O load commands from any host — a Linux workstation inspects a darwin bundle
  without a Mac — and fails both an escape and, in the other direction, an `@rpath/`
  dependency left with nothing to resolve it.

## Negative control

The rule was written against a real artifact that really carries the defect. Run over the
published `@gjsify/gtk-runtime-darwin-x64@0.48.0` tarball it goes red on three images, one
of them the plugin from #1536. Numbers, dyld traces and the before/after in a comment
below.

Its own GJS leg then caught a defect in the rule: a swallowed read error reported a bundle
it had never opened as NOT INSPECTED. Absence is `existsSync` now and a failed read throws.

## Verification

`test:node` + `test:gjs` legs of the new suite (10 pass), `tests/e2e/prebuild-loader-path`
(61 pass together, 0 fail), `node scripts/audit-runtimes.mjs --check`, `gjsify lint`,
`gjsify format --check`, `check-comment-budget.mjs`, `commitlint --from origin/main`.

## What stays open

An end-to-end run of a bundle built by the fixed builder needs a darwin runner to produce
one, and the leg that would see the defect needs a darwin host **with** Homebrew glib —
the opposite precondition to the batteries-included legs, and one runner setting. Windows
is unmeasured: `soup-3.0-0.dll` is reached the same way, but a PE image has no `LC_RPATH`,
so this rule does not answer for it rather than reporting a false clean.


JumpLink:
## Measurements

### 1. The artifact, read from Linux

`manifest-conformance/lib/binary.mjs`'s `readLibrary()` over the published
`@gjsify/gtk-runtime-darwin-x64@0.48.0` tarball, 2026-09-10:

| question | answer |
|---|---:|
| Mach-O images in the payload | 119 |
| images with an `@rpath/` **dependency** | **0** |
| images carrying any `LC_RPATH` | 10 |
| images carrying a build-host absolute `LC_RPATH` | **3** |

| image | `LC_RPATH` |
|---|---|
| `lib/gstreamer-1.0/libgstsoup.dylib` | `/usr/local/opt/libsoup/lib` |
| `lib/libjpeg.8.dylib` | `/usr/local/Cellar/jpeg-turbo/3.2.0/lib` |
| `lib/libsqlite3.dylib` | `/usr/local/Cellar/sqlite/3.53.4/lib` |

**Not the same images ADR 0024 § A4 recorded.** That probe (darwin-**arm64** 0.41.0, before
#1476 shipped the soup payload) found 2 of 106: `libjpeg.8.dylib` and the addon. libjpeg
persists; the addon is not in this tarball; **libsqlite3 and libgstsoup are both new, and both
arrived with the soup widening** — libsqlite3 because `resolveBrewDep` began following a
keg-only reference, exactly as `gst-plugins.mjs` § soup records.

`libgstsoup.dylib` in full — note the relocated deps, the absent libsoup, and the single rpath:

```
needed:      @loader_path/../libgstreamer-1.0.0.dylib
             @loader_path/../libgstbase-1.0.0.dylib
             @loader_path/../libgsttag-1.0.0.dylib
             @loader_path/../libglib-2.0.0.dylib
             @loader_path/../libintl.8.dylib
             @loader_path/../libgobject-2.0.0.dylib
             /usr/lib/libSystem.B.dylib
             @loader_path/../libgio-2.0.0.dylib
searchPaths: /usr/local/opt/libsoup/lib
id:          @loader_path/libgstsoup.dylib
```

Apple's own `otool -L` / `otool -l` on the guest agree byte for byte — an independent reader
from a different implementation family, per ADR 0024 § A3.

### 2. The loader, asked directly

macOS 15.7.9 / x86_64, Homebrew `glib 2.88.2`, `libsoup 3.6.6`, `gstreamer 1.28.5`,
`glib-networking 2.80.1` — the reporter's configuration.
`DYLD_PRINT_SEARCHING=1 gst-inspect-1.0 souphttpsrc`:

```
find path "libsoup-3.0.0.dylib"
  possible path(original path on disk): "libsoup-3.0.0.dylib"
  possible path(cryptex prefix): "/System/Volumes/Preboot/Cryptexes/OSlibsoup-3.0.0.dylib"
  LC_RPATH '/usr/local/opt/libsoup/lib' from '…/gstreamer-1.0/libgstsoup.dylib'
  possible path(leaf name using rpath): "/usr/local/opt/libsoup/lib/libsoup-3.0.0.dylib"
  possible path(default fallback): "/usr/lib/libsoup-3.0.0.dylib"
  found: dylib-from-disk: "/usr/local/opt/libsoup/lib/libsoup-3.0.0.dylib"
```

Three readings, two of which correct the issue:

- dyld expands a **bare leaf** against the calling image's `LC_RPATH`.
- It reaches the rpath **before** the default fallback.
- The default fallback for a leaf `dlopen` is **`/usr/lib` alone** — not
  `$HOME/lib:/usr/local/lib:/usr/lib`. `/usr/local/lib`, where Homebrew links every keg, is
  never consulted here.

And the chain the issue describes, confirmed — Homebrew's libsoup drags Homebrew's glib family
in behind it:

```
/usr/local/Cellar/libsoup/3.6.6/lib/libsoup-3.0.0.dylib
/usr/local/Cellar/glib/2.88.2/lib/libgio-2.0.0.dylib
/usr/local/Cellar/glib/2.88.2/lib/libgobject-2.0.0.dylib
/usr/local/Cellar/glib/2.88.2/lib/libglib-2.0.0.dylib
```

### 3. The repair, same host, same instrument

`-delete_rpath` the keg, `-add_rpath` a payload-relative entry, ad-hoc re-sign, libsoup placed
where it points:

| | `LC_RPATH` |
|---|---|
| before | `/usr/local/opt/libsoup/lib` |
| after | `@loader_path/../lib` |

```
find path "libsoup-3.0.0.dylib"
  LC_RPATH '@loader_path/../lib' from '…/plugins/libgstsoup.dylib'
  possible path(leaf name using rpath): "…/plugins/../lib/libsoup-3.0.0.dylib"
  found: dylib-from-disk: "…/plugins/../lib/libsoup-3.0.0.dylib"
```

Homebrew's prefix does not appear in the search at all, and `souphttpsrc` still resolves
(`Filename …/fix/plugins/libgstsoup.dylib`). The probe tree keeps plugin and library one level
apart, hence `../lib`; in the payload the plugin is in `lib/gstreamer-1.0/` and the library in
`lib/`, which is the `@loader_path/..` the builder now writes.

### 4. Negative control

The new rule over the published 0.48.0 payload, before any rebuild:

```
stats: {"bundles":1,"inspected":1,"images":119}
FAILURES: 3
  ✗ `gtk/lib/gstreamer-1.0/libgstsoup.dylib` carries a search path that leaves the bundle
      — /usr/local/opt/libsoup/lib
  ✗ `gtk/lib/libjpeg.8.dylib`   — /usr/local/Cellar/jpeg-turbo/3.2.0/lib
  ✗ `gtk/lib/libsqlite3.dylib`  — /usr/local/Cellar/sqlite/3.53.4/lib
EXIT=1
```

Red on a real artifact, naming the defect. The discriminator is in the suite beside it: the
same rule over a payload whose every image resolves inside itself returns zero findings, so a
green run is not a rule that fires on everything.

The GJS leg's own first run was red too, and for a better reason than a wrong constant: it
found `auditPayloadSearchPaths` swallowing a `readdirSync` failure and reporting a payload it
had never opened as NOT INSPECTED — the green-that-checked-nothing shape, in the rule written
against that shape. Fixed in the same commit.

### 5. VM

`macos-sequoia` on `qemu:///session`, started once under the corrected threshold (22 GiB
available, load 3.6, 0 guests), all of the above collected in two passes, stopped immediately
after — `monitor quit`, since the graceful path needs sudo in the guest. No guest is running.

JumpLink:
## Independent review — and the fixes are on the branch

Reviewed in `gjsify-t1-tree`, no VM started. Everything below was measured from Linux against the
**published** artifacts and the repo's own gates; five commits (`427d32c` … `078cebd`) are pushed.

### The load-bearing claim reproduces

`libgstsoup.dylib`'s keg rpath being the ONLY route is the claim the whole PR rests on, and the
half of it that a load command can carry reproduces exactly, on **both** published arches:

| | darwin-x64 0.48.0 | darwin-arm64 0.48.0 |
|---|---:|---:|
| Mach-O images in the payload | 119 | 119 |
| images with an `@rpath/` **dependency** | **0** | **0** |
| images with a build-host absolute `LC_RPATH` | **3** | **3** |
| build-host absolute `LC_LOAD_DYLIB` | 0 | 0 |
| build-host absolute `LC_ID_DYLIB` | 0 | 0 |
| unsigned images | 0 | 0 |

Same three images both times — `libgstsoup`, `libjpeg.8`, `libsqlite3` — with `/usr/local/…` resp.
`/opt/homebrew/…`, which is also the discriminator for the derived predicate. So **premise (b)
holds independently**: nothing in either payload needs a search path to resolve a link, and the
empty default is safe for the artifact as it ships. The dyld half (`leaf name using rpath` before
the `/usr/lib` fallback) I cannot re-run without a guest; the trace in the comment is internally
consistent and the artifact side agrees with it.

One residual the ADR does not name: **7 images per payload carry a self-relative rpath today**
(`libwebp`, `libzstd`, `libgraphite2`, `libsharpyuv`, `libfyaml`, and both svg pixbuf loaders with
`@loader_path/../../..` → `gtk/lib`) and the full-list replace deletes all of them. Nothing *linked*
needs them (0 `@rpath/` deps), so this is safe for the closure — but it is the same blind spot § 3
was written about, one image class over: a leaf-name `dlopen` from one of those seven would lose
its route silently and no gate would see it. Worth a sentence in § 2 rather than a change.

**And the riskiest mechanical failure mode is measured away.** § 2 warns about
*"larger updated load commands do not fit"*; the 24 bundled plugins that now get a new 32-byte
`LC_RPATH` carry **6–17 KiB of Mach-O header pad**, computed from the shipped images. No plugin is
tight.

### Blocker — the fix broke its own build (fixed, `427d32c`)

`verifyRelocation()`'s new search-path half also runs over the node-gi **addon** (§ 6, line 1268),
and `stage-prebuild.mjs` gives that addon `<brew prefix>/lib` LAST *on purpose* — the fallback
**this PR's own ADR § 1 blesses** and `darwin-prebuild-rpaths.test.mjs` asserts by order.

```
$ node -e "import('.../stage-prebuild.mjs').then(m=>console.log(m.darwinAddonRpaths('darwin-x64')))"
[ '@loader_path', '@loader_path/gtk/lib',
  '@loader_path/../../../gtk-runtime-darwin-x64/gtk/lib', '/usr/local/lib' ]
   → verifyRelocation() fails on: /usr/local/lib          (arm64: /opt/homebrew/lib)
```

`node-gi.yml → macos-gtk-runtime` and `release.yml → publish-gtk-runtime-darwin` would both have
died at § 6, on both arches. Fixed by taking the distinction as an argument at the call
(`verifyRelocation([addonDest], { searchPaths: false })`) instead of leaving it to prose — a
distinction stated only in an ADR is one the next builder edit loses.

### Blocker — the new rule never ran (fixed, `26dbdb8` )

You asked me to check the TRIGGER, and this is the `media-capabilities` shape again.
`bundle-search-paths` is registered — so `--rules` lists it and field coverage counts its `files` —
but it was **not in `CHECK_RULES`**, so `node scripts/audit-runtimes.mjs --check` never ran it. Not
on any PR, not on macOS, not on Windows, not in `release-cut`. The PR's own verification line
("`audit-runtimes --check`") is true and says nothing about the rule: grep the output and the rule
is absent entirely, not even a note. Meanwhile ADR § 4 said *"the rule fails the audit and the
publish gate from anywhere"*. Third instance of this exact omission — `audit-runtimes.mjs` narrates
the other two (`platform-packages`, `bundled-license`) in the comments right above the line I added.

Selected, fetched and printed in **both** branches (that is the contract those comments spell out).
The end-to-end control the PR did not have:

```
$ cp -r <published 0.48.0 payload> packages/node-gi/gtk-runtime-darwin-x64/gtk
audit-runtimes --check: DRIFT DETECTED.
BUNDLE-SEARCH-PATH FAILURES on 3 finding(s):
  - … `gtk/lib/gstreamer-1.0/libgstsoup.dylib` … /usr/local/opt/libsoup/lib
  - … `gtk/lib/libjpeg.8.dylib` … /usr/local/Cellar/jpeg-turbo/3.2.0/lib
  - … `gtk/lib/libsqlite3.dylib` … /usr/local/Cellar/sqlite/3.53.4/lib
EXIT=1        (and no other rule fails)
```

Without the payload it prints one NOT INSPECTED note per bundle and passes — honest, and it takes
being selected to say even that. `node-gi.yml`'s macOS bundle job is the one CI job where the
payload exists, so the audit now runs there, right after the build that writes it.

### Blocker — Windows was a false DEFECT, not a non-answer (fixed, `cc4de3a`)

You asked whether (d) is honestly implemented. It is not: `@gjsify/gtk-runtime-win32-x64` declares
the same `files: ["gtk"]`, so it **is** collected, and a payload of PE images satisfied
*"exists and holds no Mach-O image at all"*:

```
$ node ctrl-win32.mjs      # 3 synthetic PE images under gtk/bin
stats: {"bundles":1,"inspected":1,"images":0}
FAILURES: 1
  x … gtk exists and holds no Mach-O image at all. An empty payload satisfies every count-shaped gate…
```

That is worse than the false clean the rule was written against — a false clean lets a good
artifact through, a false defect refuses one. It is latent only because nothing runs the rule where
a win32 payload exists, which stops being true the moment you fix the trigger above. Now a printed
non-answer; asserted in the suite with the real `tests/e2e/pe.mjs` writer.

### My fourth negative control — green over the defect itself (fixed, `cc4de3a`)

The `readdirSync` catch you found had a sibling one field over, and it fails in the direction that
matters. `readLibrary` answers `null` for a file whose magic it does not know (icons, schemas,
locale data — 1097 of the 1216 files) and **throws** for an image it recognised and refused. The
rule's `catch { continue }` conflated the two. So I wrapped the #1536 plugin — same load commands,
same keg rpath — in a fat header:

```
BEFORE                                          AFTER
fat      images=1 findings=0   <-- GREEN        fat      images=1 findings=1
macho32  images=1 findings=0   <-- GREEN        macho32  images=1 findings=1
thin64   images=2 findings=1   (discriminator)  thin64   images=2 findings=1
```

The comment claiming a fat image *"is reported by the readers that own that question"* was not
true: nothing else reads this payload — `checkPrebuildDir()` reads `prebuilds/`, and the builder's
own gate runs on a Mac. An unreadable image is a finding now. The 32-bit case needed one more
line: it reached no refusal at all, falling through `readLibrary`'s dispatch into "not a shared
library", so `binary.mjs` now hands it to the error `readMachO` has always carried. **Inert on the
real artifacts** — 0 unreadable in either published payload, still exactly 3 red.

### Craft

- `otoolRpaths` answered a failed `otool` with `[]`. In `relocate()` that means *delete nothing*
  (so a keg entry survives and the addition lands behind it — precedence inverted, the bug § 2
  exists to prevent); in `verifyRelocation()` it means *no search path outside the bundle*. It
  throws now. `otoolDeps`' identical catch stays: that one runs on closure candidates, where a
  non-Mach-O legitimately fails.
- § 6's `try { -add_rpath } catch { /* already present */ }` also absorbed *"larger updated load
  commands do not fit"*, which is unrepairable after the fact. `otoolRpaths` exists now, so it asks.
- `verifyRelocation()`'s summary said *"0 refs and 0 search paths"* unconditionally; it names which
  question it actually asked.
- **Simplicity — what the new gate lets you delete**: `otool -L "$B"/lib/*.dylib | grep -F
  "$BREW_PREFIX"` in `macos-gtk-runtime`, which printed *"clean — no Homebrew references"* over the
  bundle #1536 was in and could not have done otherwise — `lib/*.dylib` never descends into
  `lib/gstreamer-1.0/`, and `otool -L` lists dependencies so an `LC_RPATH` is invisible to it
  whatever it is aimed at. Diagnostics kept, verdict removed. **The identical grep is still in
  `release.yml`'s "Verify bundle" step** — I left the release path alone deliberately; it is yours
  to decide.
- **No live numbers in the new comments.** I went looking, per the five-for-five record; the diff
  is clean on that.

### Verification run

`bundle-search-paths` + `prebuild-loader-path` 63 pass / 0 fail, **GJS leg included** (`skipped 0`);
the five suites reading the changed binary readers (`ship-macos`, `ship-signing`, `ship-windows`,
`ship-layout`, `prebuild-loader-path`) 177 pass / 0 fail; node-gi's own `gtk-runtime-bundle-gates`
58 pass and `darwin-prebuild-rpaths` 7 pass; `audit-runtimes --check` exit 0; `check-adr-index` (55
ADRs, statuses agree — 0057 is indexed); `check-agent-context-size --check`; `gjsify lint` exit 0;
`gjsify format --check`; `check-workflow-inline-scripts` 16 clean; the CI-workflow suites 104 pass;
`commitlint --from origin/main --to HEAD` exit 0. PR body: no line over 100 characters.
`@gjsify/manifest-conformance` has no `test:node`/`test:gjs` of its own (`echo 'nothing to do'`) —
this e2e suite IS both its legs.

### Still open, and one loose end

- **No end-to-end darwin run**, as the PR says. Everything above is load commands and the repo's
  gates; that a bundle built by the fixed builder plays an `https://` stream on a Homebrew host is
  unmeasured, and it needs the runner setting #1536 names.
- **The addon path is the one I changed without a Mac.** `verifyRelocation([addonDest], {
  searchPaths: false })` is right by the ADR and by `darwin-prebuild-rpaths.test.mjs`; the first
  darwin bundle build after this merges is what confirms § 6 completes.
- **Loose end for the next guest run**: probe files are still sitting at `~/gjsify-probe` on the
  macOS guest. No reason to boot for them — collect or delete them next time the VM is up.

### Verdict

**Not mergeable as it stood** — it would have failed the darwin bundle build on both arches, and
its new rule was wired to nothing. Both are repaired here, together with the win32 false defect and
the silent-skip class. **With these five commits I consider it mergeable**, on the reading the ADR
already takes: the darwin build is the gate that has to go green first, and the published 0.48.0
bundle stays red until it is rebuilt.

I did not merge.

PR #1636, opened 2026-09-10
chore(status): untrack the STATUS.md render and hold it untracked

### LINKED ISSUE #1631: STATUS.md is both gitignored and tracked, so the render nobody regenerates is the one git hands out
ADR 0016 and `docs/status-changelog.md` are explicit: `STATUS.md` is a **generated render** of the authored data under `status/`, it is written by `npm run status:generate`, and it is **not committed** — the doc even says "do not re-commit it, do not add a freshness check", with the reasoning (it derives from every manifest, so a committed copy stales on any merge and serialises unrelated PRs; and its counts come from directory listings on disk, so two correct checkouts legitimately disagree).

The repository does both at once:

```
$ git check-ignore -v STATUS.md
.gitignore:163:/STATUS.md	STATUS.md

$ git ls-tree origin/main STATUS.md
100644 blob 986721767f485cce406e36b362badecf257fde35	STATUS.md
```

A path that is ignored **and** tracked is not a no-op: the ignore rule only suppresses *untracked* files, so git keeps tracking the blob while every tool that consults `.gitignore` reports the file as ignored. The practical effect is the failure mode the ADR set out to avoid, inverted — because the ignore rule hides it, nobody regenerates the render or notices it drift, and `git checkout` hands every clone a copy that is authoritative-looking and unmaintained. The checked-out copy is byte-identical to the tracked blob (1579 lines) precisely because nothing ever rewrites it.

Found as a side observation while reviewing an unrelated PR; it belongs to neither that PR nor any open one.

## What this needs

`git rm --cached STATUS.md` in a PR of its own, so the `.gitignore` entry that already exists starts meaning what it says. Worth deciding in the same change:

- whether anything reads the committed copy today (the website derives its coverage numbers by calling `statusSummary()` directly, not by reading the file — but that should be checked rather than assumed);
- whether the `field-coverage` / `status-data` conformance rules should refuse a tracked render, so the state cannot come back silently. The rule about a declaration needing a check applies to this one too.

## Not established here

How far the tracked copy has drifted from what the generator produces today. Measuring that means running the generator, which rewrites a tracked file in a working tree, so it is left to whoever picks this up.

JumpLink:
`STATUS.md` was gitignored **and** tracked. Git treats that as consistent — an
ignore rule suppresses UNTRACKED files only — so it kept handing the blob out
while every tool reading `.gitignore` called the path ignored.

That inverted the failure ADR 0016's amendment set out to remove. A file nothing
ever shows as dirty is a file nobody regenerates, so `git checkout` gave each
clone an authoritative-looking render of a repository that no longer exists. The
drift is in a comment below; it is not small.

## What this changes

- `git rm --cached STATUS.md`, so the `.gitignore` line starts meaning what it
  says. The file stays on disk; `npm run status:generate` rewrites it.
- The `status-data` conformance rule refuses the render in the git index, on the
  generator's own exported `RENDER_PATH`, read through `git-index.mjs`.
- `audit-runtimes`'s remediation text told readers to regenerate the render "and
  commit both" — the instruction that put it there. It now says the opposite.
- ADR 0016 § Amendment 2 and `docs/status-changelog.md` record the incident.

## Does anything read the committed copy?

No, checked rather than assumed. `website/scripts/generate-coverage.mjs` imports
`statusSummary()` from `scripts/generate-status.mjs`, and its own header says
re-parsing the render is not an option because it is gitignored. No other
non-prose reference resolves to a read: the one remaining mention of the path in
a spec is a `README.md`-shaped fixture string in the affected-classifier test.

## Should a conformance rule refuse a tracked render?

Yes, as an arm of the existing `status-data` rule — no new rule, no new registry
entry, no new declaration kind. "No declaration without a check" applies, and
what makes it more than a slogan here is that the declaration had *only* the
`.gitignore` line behind it, which is a suppression rather than an assertion —
and in this case it suppressed the evidence.

What it lets us delete, which `simplicity` asks: the tracked blob itself, and
the "commit both" sentence that recreates it. The arm is one membership test on
one path, reusing the index reader the registry already owns.

Deliberately **not** generalised to "nothing ignored is tracked". Measured on
the merge commit: `git ls-files -i -c --exclude-standard` reports `refs/gtk` —
tracked, and ignored only by `.git/info/exclude`, which no commit carries. Such
a rule would answer a question about the checkout rather than about the commit
under review. Computing "ignored" from the tracked
`.gitignore`s instead means writing a gitignore matcher — negations, directory
patterns, nested files — to watch a one-line policy.

## Verification

`audit-runtimes --check` exit 0; `check-adr-index`, `check-agent-context-size
--check`, `check-generated-website-data`, `check-adwaita-conformance-drivers`
all exit 0; `gjsify lint` and `gjsify format --check` exit 0; `commitlint --from
origin/main --to HEAD` exit 0. Negative control in a comment below.

Closes #1631


JumpLink:
## How far the tracked copy had drifted

The issue left this open because measuring it means running the generator over a
tracked file. With the file untracked, that is now free:

| | lines | bytes |
|---|---|---|
| the blob `origin/main` handed out | 1579 | 43 KB |
| what `node scripts/generate-status.mjs` writes today | 7015 | 684664 |

`diff` reports 7728 differing lines. The tracked copy describes the repository as
it stood when the last hand-edit landed, not as it is.

## Negative control on the new arm

Track the render again and the rule goes red by name; untrack it and it goes
green. Both directions run, not just the green one.

```
$ git add -f STATUS.md && node <driver calling statusDataRule.run>
failures: 1
  - STATUS.md is staged in the git index. It is the GENERATED render of the
    authored data under status/ (ADR 0016 amendment) and .gitignore already
    names it — but an ignore rule suppresses UNTRACKED files only, so a tracked
    copy never shows up dirty, never gets regenerated, and is handed to every
    clone by `git checkout` anyway. Run `git rm --cached STATUS.md`: the file
    stays on disk and `npm run status:generate` rewrites it whenever you want
    the tables.
summary: status-data: FAILED. status/ validates against 135 published
         package(s); STATUS.md is in the index.

$ git rm --cached STATUS.md && node <same driver>
failures: 0
summary: status-data: OK. status/ validates against 135 published package(s);
         STATUS.md is not in the index.
```

The first version of the arm failed this control in a way worth recording: it
went red correctly but its `summary` still read "STATUS.md is not in the index",
a fixed string. A rule printing a sentence its own finding contradicts is the
shape the header's incidents are about, so the summary now reports what was
measured. That defect is only visible if you run the negative control and read
the *whole* output rather than the exit code.

JumpLink:
## Independent review — mergeable, with three fixes pushed

I reviewed this on `review/open-items` (branched off `chore/open-items`) and pushed two commits
to the PR branch, plus one correction to the description. Everything below was executed, not read.

### Fixed: an unreachable branch in the new summary (`c5a6f3d`)

The summary read `${failures.length === 0 ? 'OK' : 'FAILED'}` and `is ${tracked === null ? 'not ' : ''}in
the index`, with a comment arguing a flat sentence "would have been a sentence this rule prints while
its own finding says otherwise — the shape it exists to catch".

The renderer already forecloses that. `summary` is printed **only on a passing run**, in both
consumers — `report.mjs:35` inside `if (run.ok)`, and `audit-runtimes.mjs:1772` inside its own
`if (run.ok)` block — and `run.ok` is `failures.length === 0` over every rule. A rule that pushes a
failure never prints its summary at all, so neither alternate arm can be taken by any run.

Measured, with the render staged:

```
$ git add -f STATUS.md && node scripts/audit-runtimes.mjs --check
EXIT=1
STATUS-DATA FAILURES on 1 finding(s):
$ grep -c 'status-data: OK' <output>
0
```

Zero lines matching `status-data:` on the red path. So the guard was watching a mechanism, and the
branch it added could not be exercised by a test. Now stated flat with the reason recorded.
`stagedRender` went with it (a function whose only two values are `RENDER_PATH` and `null` is a
boolean wearing a string; both call sites already had the constant), and so did `stats.renderTracked`
— nothing reads it, and `stats` is documented as "rendered by the rule's own summary line".

### Fixed: a claim this PR falsifies (`54d5c2d`)

The header said `git ls-files -i -c --exclude-standard` "reports `refs/gtk` **beside the render**".
It did — until this branch untracked the render. On the merge commit:

```
$ git ls-files -i -c --exclude-standard
refs/gtk
```

Same class as the count `5dadd22` removed from that paragraph, one degree subtler: not a number but
a companion, and the companion is removed by the very change the paragraph documents. The argument
never needed it. The paragraph also carried a 133-character line in a block wrapped at ~87; rewrapped.

### Fixed: the same live number in the PR description

The description still said the command "reports two paths in this repo, and the second is `refs/gtk`".
That string becomes the squash body, so it would have shipped as history. Edited in place; re-checked
with `PR_BODY=… node scripts/check-pr-body-lines.mjs` → `59 line(s), longest 80/100 — fits the squash body`.

## The index reader

Worth stating plainly for the record: **`git-index.mjs` is not touched by this PR** (`git diff
origin/main..HEAD -- scripts/manifest-conformance/git-index.mjs` is empty). It predates this work and
already has two other consumers, `check-nativescript-app-ids.mjs` and `check-build-infra-order.mjs`.
This PR adds a third caller, not a parser. I measured it anyway:

**`.git` as a FILE.** The worktree I reviewed in is exactly that case. `resolveGitDir` follows the
`gitdir:` pointer and lands on the *worktree's own* index, not the main checkout's:

```
gitDir = …/gjsify/.git/worktrees/gjsify-t6-status
entries = 6348 · has STATUS.md = false
```

**Index version 4.** Handled, including the prefix compression. A purpose-built repo with
`index.version 4` and adjacent names that exercise the delta encoding (`alpha.txt`/`alphabet.txt`,
`zzz.txt`/`zzzz.txt`) round-trips byte-identically to `git ls-files`.

**Unknown version** throws, with a message that explicitly forbids downgrading the rule to a skip —
not a false "not tracked". `tests/e2e/ci-workflow-rev-pin/run.mjs` asserts `/index version 5/`,
`/DIRC/` and `/SPLIT index/`. Extensions are walked by their 8-byte headers and `core.splitIndex` is
named rather than silently reported as "absent". A missing `.git` or index throws deliberately —
note the asymmetry with `check-build-infra-order.mjs:455`, which guards with `existsSync` and falls
back to an empty `Set`; different consumers, different calls, but worth knowing they differ.

## Vacuity

Not vacuous, on all three of today's shapes:

- **Selected**: `status-data` is in `CHECK_RULES` and its failures set the exit code (measured above).
- **Triggered**: `audit-runtimes.yml` carries no `paths` filter, deliberately and with the reason in
  the workflow, so staging the render alone still runs it.
- **Printed in both branches**: fetched at `:1739`, summary at `:1796`, failure block at `:2068` — the
  exact trap that file's own comments record twice.

**Author's negative control** reproduced: `git add -f STATUS.md` → exit 1 naming the path and the
`git rm --cached` fix; `git rm --cached STATUS.md` → exit 0.

**Second negative control (mine).** The author's control runs on a directory `.git` with a version-2
index — neither of the two conditions that could make this arm silently green. So I combined them:
`.git` as a **file** and the index at **version 4**, on the real 6349-entry tree.

```
$ git update-index --index-version 4          # the shape feature.manyFiles produces
$ git add -f STATUS.md
  reader: entries 6349 · has STATUS.md true
$ node scripts/audit-runtimes.mjs --check  →  EXIT=1  (finding names STATUS.md)
$ git rm --cached STATUS.md && git update-index --index-version 2
$ node scripts/audit-runtimes.mjs --check  →  EXIT=0
```

This is the control that matters here. A parser blind to v4 prefix compression reads garbage from the
second entry onward and reports the path as **absent** — green, on a repo where the render is
committed. Same false green if `resolveGitDir` had treated a `.git` file as "no checkout". Neither
happens.

## The two decisions

**Arm rather than a new rule** — agreed. The `.gitignore` line was the entire enforcement behind the
amendment, and it suppressed exactly the evidence that would have shown the failure. The deletions
are real: the 1579-line blob, and the "and commit both" sentence in `audit-runtimes`' own remediation
text. I confirmed that sentence is gone by reading the live failure output, not the diff — it now
says "never stage it" and names #1631.

**Not generalising to "nothing ignored is tracked"** — agreed, and I think the argument is stronger
than stated. The real invariant is that a conformance rule must be a function of the **commit**, not
of the checkout. `.git/info/exclude` is per-clone state, so any rule quantifying over
`git ls-files -i -c --exclude-standard` is non-deterministic across machines by construction: green
in CI, red on a developer's tree, or the reverse, with no commit explaining the difference. That is
the "green CI that checked nothing" family, arrived at from the other side. The deterministic general
form does exist — quantify over the tracked `.gitignore`s only — but it needs a gitignore matcher,
which is the mechanism-watching-a-policy the author names. Narrow membership test is right; the cost
is one line per future generated artifact, which is the correct price.

## Also verified

- Nothing reads the committed copy. Broader than the author's check: across every `.mjs/.js/.ts/.mts/.tsx`
  in the tree, the **only** code reference to that path is `RENDER_PATH` in the generator itself.
  Everything else is prose. `website/scripts/generate-coverage.mjs` imports `statusSummary()`.
- The drift numbers hold: the generator writes exactly **7015** lines today; the removed blob was 1579.
- `audit-runtimes --check`, `check-adr-index`, `check-agent-context-size --check`,
  `check-generated-website-data` — all exit 0. `gjsify format --check` exit 0 (3826 files),
  `gjsify lint` exit 0, `commitlint --from origin/main --to HEAD` exit 0.
- I did need a real `gjsify install --immutable` for `lint`/`format`; the `--check` half is genuinely
  zero-install, as claimed.

**Aside, not this PR:** `scripts/check-pr-body-lines.mjs` reads `PR_BODY` from the environment and
ignores `argv`. Handed a file path it prints `the PR body is empty — nothing to check` and exits 0 —
a false green for anyone running it by hand. Worth an `argv.length > 2` refusal one day.

## Verdict

**Mergeable.** The change is correct, the arm is live and non-vacuous on the two shapes that could
have hidden it, and both decisions are the right calls with the right reasons. The three defects I
found were craft, not correctness, and are fixed on the branch. Not merging — that is yours.

PR #1637, opened 2026-09-10
feat(scripts): hold the shared corpus against its preview fence

JumpLink:
**Stage 5 as ADR 0051 wrote it does not carry, and the measurement is the first
thing here.** It asked for the gallery's `preview` fence to be emitted from
`ADWAITA_GALLERY_SHARED_TREES`. Doing that would DELETE documentation — from two
of the seven shared blocks, and for a reason no amount of corpus growth removes.

Measured with a throwaway emitter over the corpus, diffed per block against the
fence with the generated gloss lines set aside: **five** of the seven already
read exactly as the corpus would emit them. Two do not, and neither is an
authoring accident:

- `Adw.PreferencesGroup` teaches with a `<button slot="header-suffix"
  class="adw-button flat">Sign out</button>` and an `<adw-combo-row model='[…]'>`.
  A `SharedNode` names its tag with a GIR class and carries no text, so a
  CSS-classed HTML button is not a node it can be; the combo row's framework
  sibling authors `model` as a JS ARRAY against the fence's JSON string, the same
  one-value-two-doors split the `Adw.SpinRow` ledger entry records.
- `Adw.ShortcutLabel` teaches five accelerators in a `<div style="display:flex…">`.
  The wrapper is the blocker and the count is not — a tree has one ROOT, not one
  node — and a GIR container would emit as `<gtk-box>`, different markup from the
  one the block shows.

Behind the two is one argument this ADR's Risks section already half-stated: the
corpus is **selected for agreement** between two renderers, so what it drops is
exactly what they disagree about — which is what a reader most needs to see. The
divergence ledger had settled the direction already: *"the block's own `preview`
fragment is the authority"*.

## The rung that is actually available

Arm 13 of `check-generated-website-data.mjs`: every node of a shared tree occurs
in the fence its block shows a reader — same element, same attributes, same
values, in the same order — with the fence free to carry more. Containment, not
equality.

It closes what arm 11 structurally cannot see. Arm 11 compares the two authored
TREES to each other, and they can agree while both describe a UI the block
stopped showing. The ledger records that drift having happened once already.

The match is greedy on the element name in document order, and a matched element
must then agree on every authored value. Searching on for a later instance that
fits would turn a drifted value into a silent re-anchor.

Two transforms and no third: `hostTagOf` and `attributeOf`, both exported from
the corpus module and IMPORTED by `adwaita-web`'s driver and by arm 13, so the
check cannot end up agreeing with its own copy of a rule instead of with the
renderer it makes a claim about. Entity decoding is markup's escaping, not a
vocabulary mapping, and an unknown entity FAILS rather than passing through.
Neither gallery generator's refusal of an alias table is weakened.

`markupElements` keeps the attribute values it was already parsing and
discarding, so the one quote-aware fence reader stays one.

## What review changed

- A false boolean is an ABSENT attribute, which the branch's own comment claimed
  and the code did not do: the presence test ran first, so a fence that correctly
  omitted `active` failed against a node authoring `active: false`, and a fence
  that spelled it while the node said false — a real disagreement — passed in
  silence. Both directions now hold, each with a control.
- Arm 13 refused to be vacuous on the node axis and was vacuous on the one that
  carries the claim: neutering the property loop leaves every node matched, the
  note printing the same figure, and the arm green having compared nothing. The
  compared values are counted, printed and asserted non-zero.
- `attributeOf` was spelled twice — once in the driver, once in the arm. It moves
  to the corpus module beside `hostTagOf`; `tsc` holds the declaration.
- Amendment 2's count was three-and-four; re-measured it is five-and-two, and the
  gloss line that made `Adw.SwitchRow`/`Adw.EntryRow` look blocked is a region
  another generator strips and re-inserts on every run.

## Verification

`check-generated-website-data`, `check-adr-index`, `check-adwaita-conformance-drivers`,
`check-agent-context-size --check`, `check-vocabulary-alignment`, `audit-runtimes
--check` — all exit 0, and the worktree is byte-identical before and after.
`oxlint` and `oxfmt --check` clean on every touched file. `commitlint --from
origin/main --to HEAD` exit 0.

Ten negative controls, both directions, in a comment below.

### DIFF
--- a/docs/adr/0051-one-authored-tree-rendered.md
+++ b/docs/adr/0051-one-authored-tree-rendered.md
@@ -348,28 +348,43 @@ at all, and the reason each reaches none is a measurement rather than an omissio
 
 Stage 5 asked for the gallery's `preview` fence to be emitted from
 `ADWAITA_GALLERY_SHARED_TREES` "instead of authoring it per block". Measured against
-the fences on `main` at `5a8895898d`, doing that would DELETE documentation, and the
-reason is structural rather than a matter of how much of the corpus has grown.
+the fences on `main` at `5a8895898d`, doing that would DELETE documentation from two of
+the seven shared blocks, and for a reason no amount of corpus growth removes: what a
+`SharedNode` can NAME is a GIR class, and those two fences teach with markup that has no
+GIR class to be named by.
 
-Three of the seven shared blocks — `Adw.ExpanderRow`, `Adw.Banner`, `Adw.WindowTitle` —
-already read exactly as the corpus would emit them. The other four each carry something
-the corpus cannot express, and none of the four is an authoring accident:
+The measurement is a throwaway emitter over `ADWAITA_GALLERY_SHARED_TREES`, diffed per
+block against the fence with the generated gloss lines set aside. **Five of the seven
+shared blocks already read exactly as the corpus would emit them** — `Adw.SwitchRow`,
+`Adw.EntryRow`, `Adw.ExpanderRow`, `Adw.Banner`, `Adw.WindowTitle`. Two do not, and
+neither is an authoring accident:
 
-- **`Adw.PreferencesGroup`** documents a `<button slot="header-suffix">` and an
-  `<adw-combo-row model='[…]'>`. The corpus uses no `slot` at all — the divergence
-  ledger says why, the two renderers spell slots differently and a block joins only
-  when it needs no alias — and a `Gio.ListModel` is one of the portable values ADRs
-  0042/0046/0047 gave the framework trees and a `SharedNode` deliberately cannot
-  author. Emitting this fence would drop both.
-- **`Adw.ShortcutLabel`** documents five accelerators side by side in a flex `<div>`,
-  including the empty-with-`disabled-text` case. A tree driver builds ONE tree, so the
-  corpus holds one node. Emitting this fence would drop four examples and the wrapper.
-- **`Adw.SwitchRow` and `Adw.EntryRow`** carry a generated gloss line
-  (`<!-- active: … -->`), which `generate-adwaita-attribute-comments.mjs` writes from
-  the GIR and arm 12 holds. A second emitter would be writing into a region a generator
-  already owns, from a source that does not know the GIR.
+- **`Adw.PreferencesGroup`** documents a `<button slot="header-suffix" class="adw-button
+  flat">Sign out</button>` and an `<adw-combo-row model='[…]'>`. A `SharedNode` names its
+  tag with a GIR class name — `hostTagOf` throws on anything else — and carries no text
+  content, so a CSS-classed HTML button is not a node it has any way to be. The combo row
+  is expressible as markup but not as a shared node: its framework sibling authors `model`
+  as a JS ARRAY (one of the portable values ADRs 0042/0046/0047 gave those trees) against
+  the fence's JSON string, which is the same one-value-two-doors split the `Adw.SpinRow`
+  ledger entry records. Emitting this fence would drop both children.
+- **`Adw.ShortcutLabel`** documents five accelerators side by side in a
+  `<div style="display:flex…">`, including the empty-with-`disabled-text` case. The
+  wrapper is the blocker and the count is not: a shared tree has one ROOT, not one node,
+  so five sibling labels are authorable — under a GIR container that would emit as
+  `<gtk-box>`, which is different markup from the one the block shows. Emitting this
+  fence would replace the wrapper and drop four examples.
 
-Behind the four there is one argument, and this ADR's own Risks section already stated
+**What is NOT a blocker, and was claimed as one.** `Adw.SwitchRow` and `Adw.EntryRow`
+differ from the corpus emission by exactly one line each, a generated gloss
+(`<!-- active: … -->`) that `generate-adwaita-attribute-comments.mjs` writes from the GIR
+and arm 12 holds. That is a region another generator owns, but the two compose rather than
+collide: `stripGenerated` removes every generated comment line and `rewriteFence` puts
+them back on every run — "so a rewrite is idempotent", in its own words — over markup it
+does not author. Measured by deleting the gloss line from the fence and re-running
+`applyMeanings`: it comes back. So the count of blocks a corpus emitter would damage is
+two, not four, and the case below rests on those two.
+
+Behind the two there is one argument, and this ADR's own Risks section already stated
 half of it: *"a block joins the shared source only when it needs no alias, so the corpus
 is selected for the property being tested"*. A corpus selected for AGREEMENT is exactly
 the wrong source for documentation, because what it drops is precisely what the two
@@ -386,12 +401,23 @@ today on the whole corpus, and it closes the hole arm 11 structurally cannot see
 while both describe a UI the block stopped showing. The ledger records that drift
 having happened once already, block by block, found by hand.
 
-Two transforms carry it and there is still no third: `hostTagOf`, and the
-camelCase→kebab attribute rule the `adwaita-web` driver already builds with. Entity
-decoding is markup's own escaping rather than a vocabulary mapping, and an entity the
-arm does not know FAILS rather than passing through — an undecoded value reports a
-mismatch between two strings that render identically, which is a worse failure than the
-missing entity it really is.
+**Why not emit the five and author the two**, now that the damage is measured at two
+blocks rather than four. Because the emitter is what stage 5 costs, and it is the same
+whether it serves five blocks or seven: turning a `SharedNode` into markup needs a
+quoting and entity-ENCODING rule, and nothing would hold that rule against the
+quote-aware reader the fences are already read with — one direction generated, the other
+checked, agreeing with each other by construction. A per-block opt-in also puts the
+authority in two places at once, which is what `content` in the divergence ledger was
+decided against. Containment needs no emitter and makes the same drift loud.
+
+Two transforms carry it and there is still no third: `hostTagOf`, and `attributeOf`,
+the camelCase→kebab attribute rule. Both are exported from
+`adwaita-gallery-shared-trees.mjs` and IMPORTED by the `adwaita-web` driver and by arm 13
+rather than restated in either, so the check cannot end up agreeing with its own copy of a
+rule instead of with the renderer it makes a claim about. Entity decoding is markup's own
+escaping rather than a vocabulary mapping, and an entity the arm does not know FAILS
+rather than passing through — an undecoded value reports a mismatch between two strings
+that render identically, which is a worse failure than the missing entity it really is.
 
 **What this does not decide.** Whether the corpus should GROW to carry slots and
 portable values is untouched and stays a corpus question, with the ledger as its
--- a/docs/adr/0051-one-authored-tree-rendered.md
+++ b/docs/adr/0051-one-authored-tree-rendered.md
@@ -1,8 +1,11 @@
 # 51. One authored tree, rendered: ADR 0027 § 9's criterion becomes a suite
 
-- Status: **Accepted** (2026-09-10) — amended, see § Amendment 1: the second driver is
-  `adwaita-web` and not the NativeScript port, because the port has no widget an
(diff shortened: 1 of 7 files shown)


JumpLink:
## Emitted-from-corpus vs. authored fence, block by block

What stage 5 would have produced, against what the page shows today. Run on this
branch's parent (`5a8895898d`).

| block | corpus emits it? | what the fence carries beyond the corpus |
|---|---|---|
| `Adw.ExpanderRow` | identical | — |
| `Adw.Banner` | identical | — |
| `Adw.WindowTitle` | identical | — |
| `Adw.PreferencesGroup` | no | `<button slot="header-suffix">` + `<adw-combo-row model='[…]'>` |
| `Adw.ShortcutLabel` | no | a flex `<div>` and four more accelerators, incl. the `disabled-text` case |
| `Adw.SwitchRow` | no | the generated `<!-- active: … -->` gloss |
| `Adw.EntryRow` | no | the generated `<!-- text: … -->` gloss |

Two of the four also need a transform the corpus does not have: the fence spells
`<Control>C` as `&lt;Control&gt;C`.

Arm 13's own line, derived every run:

```
check-generated-website-data: 14 shared node(s) from 7 authored tree(s) found in
19 preview element(s) — the corpus is that far inside what the gallery documents
```

## Negative controls

Seven, and every one names arm 13's own message. Each perturbation was reverted
before the next.

| # | perturbation | arm 13 says |
|---|---|---|
| 1 | fence value drift: `text="Ada Lovelace"` → `"Grace Hopper"` | `Adw.EntryRow: <adw-entry-row> text is authored "Ada Lovelace" and the preview fence … shows "Grace Hopper"` |
| 2 | fence drops an attribute (`subtitle`) | `authored node 1 <adw-switch-row> sets subtitle, and the preview fence … sets no subtitle on it` |
| 3 | fence drops an element (the switch row) | `authored node 2 <adw-switch-row> has no matching element left in the preview fence …` |
| 4 | fence spells a boolean: `active` → `active="false"` | `active is authored as a boolean, and … spells it "false". The elements read it with hasAttribute()` |
| 5 | fence grows an entity the arm cannot decode (`&nbsp;`) | `carries the entity &nbsp;, which arm 13 cannot decode. Add it to ENTITIES beside the five already there` |
| 6 | drift on the CORPUS side: tree `<Control>C` → `<Control>V` | `accelerator is authored "<Control>V" and the preview fence … shows "<Control>C"` |
| 7 | the block is renamed, so a shared tree has no fence | `Adw.Banner is a shared tree with no preview fence on any gallery page, so the corpus describes a UI the site does not show` |

(4) is the one worth pointing at: `active="false"` would set the attribute, and
every element here reads it with `hasAttribute()`, so the page would render a
switch that is ON while the markup says `false`. The arm names that rather than
comparing two strings.

(6) is the direction that proves the arm is not just re-reading the docs: it
holds the corpus two renderers are TESTED against, so a drift there is red even
though the website was untouched.

Arm 13 is also vacuity-guarded (`no shared node was matched against a preview
fence at all`), because the CI step around this script asserts only that the
tree is unmodified, and an arm that walked nothing would otherwise read green.

## The advisory that moved

`check-comment-budget.mjs` still exits 0 (it runs `--warn`). The `scripts` tree
was already over its ratio; this change adds 67 comment lines against 103 code
lines, a ratio of 0.39 against the 0.59 target, so it pulls the tree's score
down rather than up.

JumpLink:
## Independent review — measured, then fixed on the branch

Verdict: **mergeable.** The refutation's *conclusion* stands; two of its four supporting
claims did not, and arm 13 had one real inversion plus one vacuity hole. All four are
fixed here (5 commits), and the branch is green on every gate.

### 1. The refutation — four blockades, checked one at a time

Method: a throwaway emitter over `ADWAITA_GALLERY_SHARED_TREES` (`hostTagOf` +
camelCase→kebab + boolean-as-presence + entity encoding), diffed per block against the
fence with generated gloss lines stripped. Gallery pages are byte-identical between
`5a8895898d` and `origin/main`, so the pinned measurement and today's are the same
measurement.

| block | claimed | measured |
|---|---|---|
| `Adw.ExpanderRow` | already emits | ✅ identical |
| `Adw.Banner` | already emits | ✅ identical |
| `Adw.WindowTitle` | already emits | ✅ identical |
| `Adw.SwitchRow` | **blocked** — generated gloss | ❌ identical but for one gloss line |
| `Adw.EntryRow` | **blocked** — generated gloss | ❌ identical but for one gloss line |
| `Adw.PreferencesGroup` | blocked | ✅ blocked — `<button slot=… class=…>Sign out</button>` + array-valued `model` |
| `Adw.ShortcutLabel` | blocked | ✅ blocked — but for the `<div style=…>` wrapper, not the count |

**`5/7`, not `3/7`.** The gloss objection does not hold. `generate-adwaita-attribute-comments.mjs`
strips every generated comment line and re-inserts it on every run — "so a rewrite is
idempotent", in `stripGenerated`'s own words — over markup it does not author. Proven by
deleting the gloss line from the `Adw.SwitchRow` fence and re-running `applyMeanings`:

```
generator re-inserts the gloss into the bare fence: true
```

So it is a generator-ORDERING matter, real but composable, not something a `SharedNode`
cannot express. The other two blockades are genuine and stronger than stated: a
`SharedNode.tag` must satisfy `/^(?:Adw|Gtk)[A-Z]\w*$/` (`hostTagOf` throws otherwise) and
the type carries no text content, so a CSS-classed `<button>Sign out</button>` and a
`<div style="display:flex…">` have no node to be. For `Adw.ShortcutLabel` the written
reason ("a tree driver builds ONE tree, so the corpus holds one node") is not the operative
one — a tree has one *root*, and five sibling labels are authorable; the wrapper is what
blocks it. The combo row checks out: its framework sibling authors `model` as a JS array.

ADR Amendment 2 now records the measurement, the corrected reason, and — new — why the
five are not emitted anyway: the emitter is the cost, it is the same emitter for five as
for seven, and nothing would hold its quoting/entity-ENCODING against the quote-aware
reader the fences are already read with.

### 2. Arm 13 — can it be green having proved nothing?

**Trigger.** Sound. A shared tree whose block is renamed, removed, or whose fence stops
parsing falls out of `fenceByTitle` and fails by name rather than silently dropping out of
the loop. Two notes: `fenceByTitle` keeps the *last* fence per title (40/40 titles are
distinct today, so no collision), and `galleryFences` returns the `web` fragment where a
block has one — the arm is correctly slot-agnostic and its messages interpolate
`fence.slot`, but every header said "preview fence" categorically. Wording corrected;
`Adw.Toast` is the one `web` block and it is not shared.

**Vacuity — the eighth control, and it found something.** The arm refused zero matched
nodes and did not refuse zero compared *values*. Replacing the property loop's source with
an empty object:

```
before: 14 shared node(s) from 7 authored tree(s) found in 19 preview element(s)
        OK.                                         ← identical line, exit 0, nothing compared
after:  14 shared node(s) and 0 authored value(s) … 
        arm 13 matched shared nodes but compared no authored value — it proved only that tags line up
```

A ninth, weaker: stopping `depthFirst` from recursing leaves the arm green at 7 nodes
instead of 14. The printed count moves, so it is visible; adding a second independent
walker to hold the first would be paranoia, and it is left as a known residual.

**A real inversion, both directions.** The boolean branch was unreachable for `false`,
because the missing-attribute failure ran first — the opposite of what its own comment
claimed:

```
control  corpus authors active:false, fence correctly OMITS it   → RED   (fence was right)
control  corpus authors active:false, fence SPELLS active        → GREEN (a real disagreement)
```

Fixed; both now behave. Latent today — every boolean in the corpus is `true` — so the
first `false` would have paid for it.

**The seven claimed controls, re-run on the final tree** (plus my three):

| # | control | exit | arm-13 message |
|---|---|---|---|
| 0 | baseline | 0 | — |
| 1 | fence value drift | 1 | ✅ |
| 2 | fence omits an attribute | 1 | ✅ |
| 3 | fence omits the element | 1 | ✅ |
| 4 | fence spells `active="false"` | 1 | ✅ |
| 5 | undecodable entity (`&nbsp;`) | 1 | ✅ (2 lines) |
| 6 | corpus-side value drift | 1 | ✅ |
| 7 | block renamed | 1 | ✅ |
| 8 | **value comparison neutered** | 1 | ✅ *(new)* |
| 9 | **corpus `false` + fence omits** | — | ✅ silent *(new, was RED)* |
| 10 | **corpus `false` + fence spells it** | 1 | ✅ *(new, was silent)* |

Order is genuinely held, not only claimed: swapping two siblings in the
`Adw.ExpanderRow` tree fails on "no matching element left".

### 3. The boundary that must not fall

No alias table appears, and no third transform. But the *second* transform was a
one-liner restated in the checker with nothing holding it against the driver's copy —
unlike `hostTagOf`, which arm 11 runs against `gtk-host`'s generated table precisely
because its first version was wrong (`gtk-glarea`). A checker that re-spells the rule
agrees with its own copy, not with the renderer it is a claim about. `attributeOf` now
lives in the corpus module beside `hostTagOf`; both consumers import it. The `.d.mts`
backstop works — dropping the declaration produces `TS2305` in the driver.

`markupElements` did not become a second parser: it keeps values from the walk it already
did, `null` (bare) stays distinct from `''`, and decoding is the caller's.

### 4. Craft

One live number in a comment written by the commit that added it — "the fence draws five
`<adw-shortcut-label>`s" — removed. No try/catch, no paranoid probe. `simplicity`: arm 13
deletes nothing that arm 11 or arm 12 owns, and the `Adw.ExpanderRow` incident note in the
corpus header stays (it is the reason, not a restatement).

### Gates

```
check-generated-website-data          exit 0   (git status --porcelain identical before/after)
check-adwaita-conformance-drivers     exit 0
check-adr-index                       exit 0
check-agent-context-size --check      exit 0
check-vocabulary-alignment            exit 0
audit-runtimes --check                exit 0
oxlint / oxfmt --check                exit 0   (rig verified against a seeded bad file)
commitlint --from origin/main --to HEAD exit 0
check-pr-body-lines                   82 lines, longest 87/100
```

`gjsify lint` / `gjsify format` could not run in this worktree — `packages/infra/cli/lib`
is not built — so `oxlint`/`oxfmt` were run directly with the repo's own config, which is
what those wrappers call.

JumpLink:
### Addendum — two more, found after the first pass

**A slot arm 13 cannot hold.** `SharedNode` declares an optional `slot`; arm 11 compares it
between the two trees, and arm 13 walks straight past it. A slotted node would therefore be
"contained by" a fence that was never asked to carry it — a rule leaving the check through
its own trigger, the class this PR's own review criteria name. No shared tree has one today,
which is when it is cheap.

Comparing it would be the wrong close: the two renderers spell slots differently
(`start`/`title`/`end` against `startBox`/`titleWidget`/`endBox` in the ledger), so matching
`node.slot` against the fence's `slot=` would bless the gtk-host spelling as the corpus's —
the one thing the corpus exists to refuse, and the reason a slotted block fails the no-alias
admission rule to begin with. So the arm refuses, self-retiring: the day a block is admitted
with a slot it came with a decision about what a shared slot spells, and the failure says to
teach the arm that rather than guess it.

```
control 11  a shared node carrying slot="header-suffix"
→ Adw.EntryRow: authored node 1 <adw-entry-row> carries slot="header-suffix", and arm 13
  does not hold a slot against the fence. […] teach this arm the one the corpus settled on.
```

**A pointer the rename broke, and two counts that survived the sweep.**
`docs(status): stop restating what arm 11 prints` renamed the entry's heading and left the
cross-reference above it naming a heading that no longer exists. Two live counts also
survived the commit that deleted the rest of them: *"a claim about seven blocks"*, and a
`content` bucket introduced by its size (*"two of them"*, *"in all three cases"*) rather than
by the blocks in it. The blocks are named instead, which is what that sentence wanted anyway.
Same commit, same class — the pattern where this keeps being found.

Also observed, not touched: the camelCase→kebab one-liner exists in two further scripts
(`check-vocabulary-alignment.mjs`, `generate-adwaita-framework-snippets.mjs`) for their own
consumers. They are outside this PR's "two transforms carry the corpus" claim, and folding
them in would touch three unrelated generators.

PR #1639, opened 2026-09-11
fix(devtools): a selector matches the type chain, not the exact GType

### LINKED ISSUE #1582: FindWidget matches the exact GType, so a subclass is unreachable
## What

`@gjsify/devtools`' `FindWidget` resolves a selector by comparing the **exact** GType name, so a consumer cannot address a widget by the type it derives from.

Measured in a consumer (`JumpLink/bauplaner`) while verifying a screenshot rig: three of six `BP_APP_DIALOG` values open `Adw.Dialog` **subclasses** — `BauplanerAufbauDialog`, `BauplanerDachDialog`. A rig that waits for `FindWidget AdwDialog` exits 1 with *"opened no AdwDialog"* for all three, while `DumpTree` of the same run shows the dialog mapped and visible.

So the failure is not "the dialog did not open"; it is "the query cannot express what was asked". And it fails in the direction that reads like a defect in the consumer.

## Why it is worth fixing here rather than there

The consumer's workaround is to match a **style class** instead (`floating`, measured unique in a ~5.7k-node tree across all six dialog values). That works and is what shipped, but it is a workaround with two costs:

- it addresses the widget by how it *looks* rather than by what it *is*, so it breaks when the stylesheet changes;
- every consumer that subclasses a GTK or Adwaita type meets the same wall, and each will invent its own workaround.

An application subclassing `Adw.Dialog`, `Adw.ApplicationWindow` or `Gtk.Box` is the ordinary way to write a GTK app, so "the exact GType" excludes the common case rather than an exotic one.

## What a fix would have to decide

- Whether a selector matches the type **chain** by default, or only with an explicit spelling — a default that suddenly matches ancestors would change what existing selectors resolve to, which is a behaviour change on a published surface.
- Where the chain comes from. `GObject.type_parent` walks it at runtime; `@gjsify/gtk-host`'s generated surface data already carries the hierarchy for the types it knows, but a consumer's own subclass is not in it.
- What a match should report. A selector that resolves through three ancestors should probably say which type it matched, or a consumer cannot tell an intended match from an accidental one.

## Not urgent, and not blocking

The consumer shipped the style-class workaround; nothing is stuck. Filed so the next consumer finds this instead of re-deriving it.

JumpLink:
`FindWidget`'s type half compared the **exact** GType name, so a widget of a SUBCLASS of the
named type was unaddressable. Subclassing `Adw.Dialog`, `Adw.ApplicationWindow` or `Gtk.Box` is
the ordinary way to write a GTK app, so the comparison excluded the common case rather than an
exotic one.

Measured in bauplaner: three of six `BP_APP_DIALOG` values open `Adw.Dialog` subclasses, and a
rig waiting for `FindWidget AdwDialog` exited 1 with *"opened no AdwDialog"* while the same
run's `DumpTree` showed the dialog mapped and visible. The failure reads like a defect in the
consumer, and the workaround it invented — matching a style class (`floating`) instead —
addresses the widget by how it LOOKS rather than by what it IS, so it breaks when the stylesheet
changes.

## What changed

`widgetIsA(object, typeName)`, exported from `@gjsify/devtools`:

- the **name comparison first**, which is the whole answer for a type GObject does not know — a
  unit-test shape, or a runtime that never registered it. That is what keeps every exact-name
  spec in the module meaning what it says;
- then `g_type_is_a` where GObject knows both names.

Registration is a **runtime** fact, measured under GJS 1.88.1:
`GObject.type_from_name('AdwDialog')` answers `null` until something in JS touches `Adw.Dialog`,
and registering a SUBCLASS registers its whole ancestor chain. So the lookup happens per call
and is never cached — the base type of a live instance is always registered (the instance's own
class registered it), and a name that resolves to nothing now may resolve a moment later.

### The exact type still wins

An exact match anywhere in the subtree beats a subclass match, **even one earlier in reading
order**. That is the compatibility guarantee and it is what makes the widening measurable:
*every selector that resolved before this change resolves to the same path after it*, and the
subclass leg is consulted only where the answer used to be none. It costs one extra full walk
exactly in the case that already walked the whole tree — the miss.

## The class, not just the site

The same exact-GType comparison was in two more places in `widget-tree.ts`, both failing the
same silent way:

- **`activateWidget`**, on "is the parent a `GtkListBox`?" — every row inside an app's own
  list-box subclass was undrivable. `activate()` returns false for a row, the fallback never
  fired, and the call answered `false`.
- **`sendKeyToWidget`**, on "is this a `GtkEventControllerKey`?" — a key sent to an app's own
  key-controller subclass was delivered nowhere, which reads as "the keyboard does not work".

A repo sweep for exact GType comparisons (`widgetType(…) ===`, `gtypeName(…) ===`, `$typeName
===`, `$gtype.name ===`) finds no fourth site; the one remaining is the deliberate
exact-preference leg inside `findWidgetPath`. All three now call the one helper, so the next
such comparison has something obvious to call.

## Measurement

Red first, with `widgetIsA` stubbed to today's exact-name comparison — **7 of 77 tests failed**,
one per widened behaviour:

```
✖ findWidgetPath › matches a SUBCLASS of the selected type
✖ findWidgetPath › matches a subclass through the toolkit chain too
✖ findWidgetPath › still requires the css class when the type matched through a subclass
✖ widgetIsA › answers for the type itself, its ancestors, and nothing else
✖ widgetIsA › reads the node-gi runtime type off $typeName
✖ activateWidget › drives the row when the parent is a GtkListBox SUBCLASS
✖ sendKeyToWidget › delivers to a SUBCLASS of GtkEventControllerKey
```

That stub IS the negative control: it is `main`'s behaviour, and putting it back reds exactly
those seven and nothing else. With the fix: **77 passed · 189 assertions**.

Four of the new specs are there to pin what must NOT change, and they were green against the
stub too: the exact-match preference, is-a's direction (a `GtkBox` is not a
`DevtoolsSpecDialog`), and the name-only fallback for an unregistered type.

The specs register real GTypes (`DevtoolsSpecDialog extends Gtk.Box`, `DevtoolsSpecListBox
extends Gtk.ListBox`, `DevtoolsSpecKeyController extends Gtk.EventControllerKey`) rather than
mocking the type system — that is exactly the consumer shape, and it is the only way the walk is
proven against GObject rather than against a stand-in for it.

## Review round

An independent review found three things the first pass STATED wrong rather than did wrong.

`g_type_is_a` also answers for an **interface**, so `GtkOrientable` matches every box and
`GtkAccessible` the first visible widget in any tree — a real widening past "the type itself
or any ancestor", and it was neither documented nor tested. Kept, because an interface
describes what a widget can DO and that is the same question a selector asks, but said out
loud in all four docs and pinned by a spec.

The exact-wins-over-subclass guarantee is **per call**, not per tree. `derived` is created
fresh per `findWidgetPath`, so a consumer looping over `ListToplevels` can take a subclass hit
in an early toplevel over an exact hit in a later one. `FindWidget` passes one root and is
unaffected; the exported function's doc now scopes the claim and says what such a caller does
instead.

And one assertion could not fail for its stated reason: `widgetIsA(dialog, 'GtkButton')
=== false` holds both when the types are unrelated and when `GtkButton` is unregistered, so it
now sits beside a positive control on the same chain. The three `void Spec*.$gtype` lines went
with it — `registerClass` registers eagerly, so they were no-ops under a comment claiming they
were load-bearing. **78 passed / 193 assertions.**

Closes #1582


JumpLink:
## Independent review — re-measured, nothing changed

Second pass by a reviewer who did not write this. I re-ran the claims instead of re-reading them.

| claim | how it was checked | result |
|---|---|---|
| three exact-GType comparisons, no fourth | `grep -rnE "(===\|!==)\s*'(Gtk\|Adw\|Gdk\|Gio\|GObject)[A-Za-z]+'"` over `packages/{framework,gjs,nativescript-bridge}/**/src`, plus every `gtypeName(` / `widgetType(` call site | confirmed for `@gjsify/devtools`: `widget-tree.ts:96`, `:138`, `:236`. The other `widgetType()` uses (`:174`, `:375`, `devtools-service.ts:383`) REPORT the type, they do not compare it |
| an existing selector cannot be redirected | read `walkForSelector`: the exact predicate and the DFS order are untouched, and the exact leg returns before the subclass leg is consulted | confirmed — the widening can only turn a former `''` into a path, never move a path |
| negative control | stubbed `widgetIsA` to `return name === typeName`, ran `gjsify workspace @gjsify/devtools run test` | **8 of 78 red** (the PR says 7 of 77) — covering all three sites, the interface arm, and the `$typeName` node-gi path |
| `FindWidget` used as a presence test inside this repo | `grep -rl 'FindWidget\|find_widget'` over `packages showcases examples tests scripts` | only `devtools`, `devtools-mcp`, `devtools-protocol` — no gate in this repo can flip green on the widening |

### The interface arm is wanted, not merely tolerated

It is pinned, not just documented (`widget-tree.spec.ts:264-272` asserts `GtkOrientable` and
`GtkAccessible` true on a `GtkBox` and `GtkEditable` false), and the three mutating sites are
unaffected by it: `GtkListBox` and `GtkEventControllerKey` are classes, so `widgetIsA` there
reaches subclasses and no interface. The breadth is confined to a caller-supplied `FindWidget`
selector, where the compatibility ordering already bounds the damage.

The one semantic that really did change and is not spelled out: a selector naming a BASE type or
an interface can no longer answer "absent". `FindWidget('GtkWindow')` used to be `''` against an
`AdwApplicationWindow` and is now `toplevel:0`. For a concrete type an absence test still works
(nothing is-a `AdwPreferencesDialog` when no dialog is open), so this is narrow — worth a line in
a rig author's head, not a change here.

### Not a defect here, but the same shape one package over

`packages/framework/gtk-host/src/conformance/window-chrome.ts:61,81` compares live GType names
exactly — `isHeaderBar(name)` and `name !== 'GtkWindowControls'`. An app subclassing
`Adw.HeaderBar` under-counts `headerBars`, which that file already declares is "NOT the
invariant"; the invariant counts `GtkWindowControls`, which is not subclassed in practice. Left
alone deliberately — reported so the sweep's boundary ("no fourth") is read as *no fourth in
this package*.

**Verdict: mergeable as it stands. No changes pushed.**

PR #1642, opened 2026-09-11
feat: win32 MP3 is bounded upstream, and the bound is now machine-held

### LINKED ISSUE #1626: win32-x64 still ships no MP3, Vorbis or FLAC decoder — the payload half of #1544

JumpLink:
Closes #1626. The question was whether win32's MP3 and FLAC gaps have to stay, and it is
answered by reading upstream rather than by remembering it. **MP3 is outcome 3 —
upstream-bounded, every route measured shut. FLAC is outcome 2 — obtainable, costed, and
not taken.** ADR 0056 is confirmed, and the two routes its first draft never named are now
in it. The tables are in a comment below, because a wide table in this body becomes a
commit message commitlint rejects.

## Measured, all of it from Linux

**The published 0.48.0 tarballs, re-read with `tar -tz`** (the issue's numbers, confirmed
independently): darwin-x64 24 plugins, darwin-arm64 24, win32-x64 21. `comm` over the
normalised base names — darwin-only `flac mpg123 osxaudio vorbis`, win32-only
`directsound`. #1633 has since landed `vorbis`, leaving MP3 and FLAC.

**gvsbuild's catalogue at the pinned `2026.6.0`, cross-read from two oracles:** the GitHub
contents API for the tag, and the PyPI wheel `pipx install gvsbuild==2026.6.0` actually
unpacks. 95 files each, byte-identical lists, 94 project modules beside an `__init__.py`.
`libvorbis.py`, `ogg.py`, `opus.py` present; **zero matches for `flac` or `mpg123`**, and
no `mad`, `lame` or `twolame`. `main` carried the same list, and so does **2026.8.0**, the
newest release — read with this PR's own `--update` against that tag: nothing added,
nothing removed. A pin bump alone will not close this.

**Two routes out of that same catalogue that ADR 0056 never named, both found and both
read.** `gst-libav` IS there (`gstreamer.py`), and so is `ffmpeg` (`ffmpeg.py`, 8.1.1,
LGPL by default) — and it decodes no MP3, not on principle but on a fact: gvsbuild's own
`gvsbuild/patches/ffmpeg/build/build.sh` configures ffmpeg `--disable-everything` and then
names what comes back — `h264`, `hevc`, `libdav1d`, `mpeg1video`, and on the audio side
exactly `mp2float`, `wmav2`, `wmapro`, under the comment *"audio decoder which aren't
available in native gst plugins"*. No mp3, no flac, no aac — so it would not have closed
the AAC gap either. gst-plugins-ugly 1.28.4's `ext/` is a52dec, cdio, dvdread, mpeg2dec,
sidplay, x264: `mad` is gone.

**FLAC is therefore a price, not a wall.** `claxon` in gst-plugins-rs 0.15.2 is a pure-Rust
FLAC decoder (`claxondec`, MPL-2.0) and gvsbuild already defines that tree as
`gst-plugin-gtk4`. Taking it means `cargo install cargo-c --locked` plus gst-plugins-bad
and gtk4 rebuilt from source (the extracted GTK4 zip carries no gvsbuild build markers),
on the leg whose GStreamer build already runs under a 150-minute timeout — to close the
one of the two gaps no consumer has been measured needing, while MP3 stays absent either
way. Written down in the ADR rather than left for the next reader to rediscover.

## The actual finding: the reason was held by nothing

A win32 gap's `why` says *gvsbuild defines no project for libmpg123*. That is a claim about
an external artifact at a pinned version, and every mechanism in this area is structurally
blind to it — `media-capabilities` compares the declaration to the shipped plugin files,
`missingBundledGstPlugins` to what the builder copied, `gst-elements.test.mjs` to the
registry on the target. **All three stay green on the day the gap could be closed.** That
is #1544's class with the sign flipped: there a decoder was absent and nothing said so;
here a reason expires and nothing says so. Not hypothetical — `GVSBUILD_VERSION` is spelled
in eight `env:` blocks across four workflows, and 2026.8.0 shipped while 2026.6.0 was the
pin.

So the reason leaves prose. A claim or gap may carry `upstream: { catalogue, library }`,
held by who can read it — the split ADR 0055 § 2 draws through `plugin` and `element`, one
level up. `media-capabilities` (portable) holds the SHAPE, so a consumer's
`manifest-check` refuses a malformed one. The new repo-scoped `gvsbuild-catalogue` holds
what the value says, against a committed catalogue snapshot pinned to `GVSBUILD_VERSION` —
no network, so it runs on every PR, which is where the pin bump that expires the reason
lands.

**Both directions on every run, which is the only reason the matcher can be believed.** A
gap's library must match nothing in the snapshot; a claim's must match something. The win32
bundle carries both kinds — Ogg/Vorbis and Opus exist because `libvorbis.py` and `opus.py`
do — so a matcher that had silently stopped matching would fail on the claims in the same
run it passed the gaps. A bundle declaring only one direction is a finding of its own, and
so is a snapshot too short to be a catalogue: a truncated one answers "absent" to
everything, which passes every gap and points the blame at the declarations.

Module basenames rather than project names, deliberately: project names need a Python
parser, and a first attempt at that regex silently missed `opus`, `cairo` and `dav1d` — a
parser that under-reports turns every absence assertion into a pass. A directory listing
cannot be wrong in that direction.

## The guard was broken on purpose, six ways

Six edits, six red runs, each naming the edit — the list is in the comment below. And the
suite itself: neutering `matchLibrary` to return `[]` reds 4 of its assertions.

One thing measured and then *un*-done, because it is the more useful half: the pin reader
looked like it would break on a CRLF workflow file (`core.autocrlf=true` is Git for
Windows' installer default and `.gitattributes` deliberately does not cover `*.yml`), so a
`[ \t\r]*` went in against it — and measured to change nothing. ECMAScript counts CR as a
line terminator, so `$` under `/m` already matches before it. The guard came back out and
the property went into a test instead, since a rewrite splitting on `\n` would lose it
silently.

## What a consumer gets out of this today

No payload — and one thing the bundle did not say before. The gap `why` now records that an
application willing to make the redistribution call can add an MSVC-ABI `gstmpg123.dll` of
the same GStreamer through `GST_PLUGIN_PATH`, which `gtk-runtime.js` deliberately never
sets (it sets `GST_PLUGIN_SYSTEM_PATH`, the non-additive one). Measured: with
`GST_PLUGIN_SYSTEM_PATH` on a bundle-like directory `mpg123audiodec` is absent, adding
`GST_PLUGIN_PATH` on a second directory makes it resolve, and the bundle's own elements
survive it.

## Verification

- `node scripts/audit-runtimes.mjs --check` — green, and the new rule reports `gvsbuild
  2026.6.0 — 94 project module(s), 8 workflow pin(s) agreeing, 2 claimed library(s) present
  and 2 declared gap(s) still absent upstream`
- `node --test tests/e2e/media-capability-declaration/run.mjs` — 41/41, was 26; neutering
  `matchLibrary` reds 4 of them
- the neighbouring suites that read the same files — `ci-workflow-rev-pin`,
  `ci-pr-trigger-parity`, `bundled-license-trigger`, `bundle-search-paths`,
  `ci-probe-outcomes` — 108/108
- `gjsify format --check` clean, `gjsify lint` clean on every touched file (the two
  `audit-runtimes.mjs` warnings are pre-existing, on lines this PR does not touch),
  `check-comment-budget` and `check-agent-context-size --check` both exit 0
- `--update 2026.6.0` round-trips the committed snapshot byte-for-byte

## Left open, in `status/open-todos.md`

Nobody has filed the gvsbuild PR that would close MP3; FLAC's price is not paid; and the
rule re-asks the upstream question only when the pin moves — deliberately, since a project
we are not pinned to cannot be built anyway, so a polling job would add a network
dependency and a leg that can red for something no PR caused.


JumpLink:
The two tables from the description, kept out of the PR body so the squash commit message
stays inside commitlint's 100-character line limit.

### Every route out of gvsbuild's catalogue, measured at the pinned `2026.6.0`

| route | in the catalogue? | decodes MP3? | decodes FLAC? |
|---|---|---|---|
| a `libmpg123` / `libFLAC` project | **no** — 94 modules, zero matches | — | — |
| `gst-libav` + `ffmpeg` | **yes** (`gstreamer.py`, `ffmpeg.py` 8.1.1, LGPL by default) | **no** — `--disable-everything`, audio set is `mp2float`/`wmav2`/`wmapro` | **no** |
| gst-plugins-rs 0.15.2 (`gst-plugin-gtk4`) | **yes** | **no** — no MP3 decoder in the tree | **yes**, `claxon` → `claxondec` (MPL-2.0) |
| gst-plugins-ugly 1.28.4 | yes | **no** — `ext/` is a52dec, cdio, dvdread, mpeg2dec, sidplay, x264; `mad` removed upstream | n/a |

`2026.8.0`, the newest release, carries the same 94 modules — nothing added, nothing
removed.

### Where the two halves of the declaration are held

| held by | what it holds | where it runs |
|---|---|---|
| `media-capabilities` (portable) | the SHAPE of `upstream` — `catalogue` and `library` both filled | any consumer's tree, via `gjsify manifest-check` |
| `gvsbuild-catalogue` (repo, new) | what the value SAYS — against a committed snapshot of gvsbuild's project modules, pinned to `GVSBUILD_VERSION` | every PR, no network |

### The guard broken on purpose, six ways

| break | what it reported |
|---|---|
| MP3 gap's library spelled `libvorbis` (one the catalogue HAS) | *gvsbuild 2026.6.0 defines libvorbis … ADR 0056 § 1 step 1 now answers YES for this format* — plus the reminder that step 1 is necessary and not sufficient |
| Ogg/Vorbis claim's library spelled `libflac` | *CLAIMS the format, and gvsbuild 2026.6.0 defines no project module matching "libflac" (94 module(s) read)* |
| `napi.yml` bumped to `2026.8.0` | *`GVSBUILD_VERSION` is spelled 2 different ways across the workflows*, listing both, plus the snapshot mismatch naming the `--update` command |
| snapshot truncated to 2 modules | *lists 2 project module(s), fewer than the 40 any real gvsbuild carries* — reported before any absence is read off it |
| every `upstream` field deleted | *no bundle declares … `upstream`, so this rule compared nothing and passed* |
| only the gaps keep `upstream` | *every `upstream` declaration points the same way (absent)* |
| the snapshot replaced with `{ "version": 1 }` | the reader's own message under this rule's print block, naming the file and the command that rewrites it |

That last one is seven; it was added with the read-through commit that converted the
snapshot read's throw into a finding.

JumpLink:
## Independent review — verified against the sources, one fix pushed

**The load-bearing fact holds.** I read `gvsbuild/patches/ffmpeg/build/build.sh` at tag
`2026.6.0` myself, from the git tag and from the PyPI wheel. `--disable-everything`, and the
whole audio set is `mp2float`, `wmav2`, `wmapro` under the comment *"Enable audio decoder
which aren't available in native gst plugins, expected to be used via gst-libav"*. No mp3
decoder is registered, so `gst-libav` cannot produce one. "MP3 is upstream-bounded" is a
fact, not an opinion.

**The catalogue numbers reproduce exactly.** GitHub contents API at `2026.6.0`: 95 files,
94 modules beside `__init__.py`. PyPI wheel `gvsbuild-2026.6.0-py3-none-any.whl`: the same
95/94, list-identical, and identical to the committed snapshot. Zero matches for `flac`,
`mpg123`, `lame`, `mad`. `2026.8.0` (still the newest release) carries the same 94 with
nothing added or removed — the ADR's claim, confirmed.

**On "two oracles": it is one and a half, and worth saying so.** The wheel is built by
gvsbuild's own release CI from the same tag, so it is not a second source of truth about
what gvsbuild can build — it is the same tree over a second transport. What it *does*
establish independently is that the PyPI artifact `pipx install` fetches was not built from
a different tree than the tag the snapshot names. That is a real fact and the narrower one.
Note the direction is also inverted from how the header reads: the builder legs install via
`pipx install "gvsbuild==$env:GVSBUILD_VERSION"`, so the wheel is the *operative* artifact
and the contents API is the cheap proxy for it — not the other way round.

**It runs where it claims to.** `audit-runtimes.yml` carries no `paths:` filter, runs on
every `pull_request`, and its job is `Detect runtime-triplet drift` — one of the three
required checks. No `cache-hit` condition anywhere near it, no network. Confirmed by
running `--check` and `--check --strict` locally.

**All eight pins are read.** `readGvsbuildPins` discovers 8 rows across node-gi.yml (3),
release.yml (2), prebuilds.yml (2), napi.yml (1) — every `env:` block, no table to fall
behind.

**Negative controls: I ran six of them against the real `audit-runtimes --check`.**
Snapshot truncated to 2 modules → exit 1, and the vacuity floor fires *before* any absence
is read, so the gaps do not pass. Only one direction left → exit 1. One workflow bumped →
exit 1 naming it. `flac` added to the snapshot → exit 1 naming FLAC and what closing it
takes. Snapshot deleted → exit 1. Snapshot valid JSON but wrong shape → exit 1. None of
them degraded to a pass.

### The one hole, now closed (2295c740)

`readGvsbuildPins` anchored the whole line, so a pin with a trailing YAML comment matched
nothing — and a pin the reader does not see is a pin nothing compares to the snapshot.
Measured on this tree: `napi.yml` moved to `'2026.8.0'  # bumped for GTK 4.22`, the reader
found **7** pins, all seven agreed with the stale 2026.6.0 snapshot, and the audit exited
**0**. That is the rule's own under-report failure mode — the one the module basenames
exist to avoid — reopened one level down, and it bites at the only moment the rule is for:
annotating the line is what a person does at a bump. The pattern now admits a trailing
comment and spells the three quoting forms as alternatives; the new e2e assertion reds
against the old pattern.

### Noted, not changed

`packages/node-gi/node-gi/test/gtk-runtime-bundle-gates.test.mjs:912` is a second reader of
the same pin, with its own regex and a hardcoded two-workflow list — the "table that misses
the ninth block" shape this rule's own comment argues against. It answers a different
question (licence provenance) and `gvsbuild-catalogue` now catches any divergence across
all eight pins before it could matter, so lifting it onto the shared reader is a follow-up
rather than a blocker.

The e2e "real bundles" case uses `collectMediaBundles` while the rule uses its own
`mediaBundles`; the former is a strict superset whose extra members carry no `upstream`, so
the two provably agree today. Left alone.

Suite: 42/42. oxfmt clean, oxlint clean on both files. Comment budget moves `scripts`
-250 → -235 against main (i.e. still an improvement on main) and `tests` -3121 → -3133,
advisory either way.

PR #1643, opened 2026-09-11
fix: the Debian changelog at gzip -9, and a zlib that honours the level

JumpLink:
Closes item 10 of the `gjsify ship` roadmap (ADR 0024) — the last Debian Policy
gap `lintian` still named on a `gjsify ship` artifact.

## The gap, and where it actually was

Policy § 4.4 asks for `gzip -9 -n`. `lintian` reads that claim off the gzip
header's **XFL byte** alone, and raised:

```
W: gjsify: changelog-not-compressed-with-max-compression [usr/share/doc/gjsify/changelog.Debian.gz]
```

The gap was a missing capability in the CORE, not in `ship`. `@gjsify/zlib`'s
`gzipSync`/`deflateSync`/`deflateRawSync` and their async siblings accepted a
`ZlibOptions` **named `_options` and ignored it**, so no caller could ask for a
level on either backend. Nothing noticed, because a dropped level still
round-trips and still shrinks the input — XFL is the only observable, and no
test read it.

So the fix goes home to the core: `Gio.ZlibCompressor` takes a level and now
gets one, `@gjsify/tar`'s `gzip()` grows the same option and routes a levelled
request through `node:zlib` (`CompressionStream` has no level to accept), and
`plan.ts` compresses the changelog at `POLICY_MAX_COMPRESSION`.

## Measured

`lintian` 2.117 on ubuntu-24.04, against gjsify's own `.deb`, built from this
branch with and without the level argument:

| | before | after |
|---|---|---|
| lintian | `W: changelog-not-compressed-with-max-compression` | *(gone)* |
| `file` | `gzip compressed data, from Unix` | `gzip compressed data, max compression, from Unix` |
| error-severity tags | none | none |

`verify-deb.sh` now **gates the tag by name**, so it cannot return quietly: the
level travels four hops (`plan.ts` → `gzipDeterministic` → `@gjsify/tar` →
`node:zlib`) and every one of them has a default-level path to fall back to.

## Two things worth reading the comments for

**Why not just stamp XFL.** mtime and the OS byte are facts about the build
environment, which a reproducible build may normalise; XFL is a statement about
the compression actually performed. The temptation is measurable — the two
changelog members differ in **exactly one byte**, position 9, and are 1152 bytes
either way — and `gzip.ts` records why that is a coincidence of a small input
rather than a rule.

**An out-of-range level is refused, and the guess about why was wrong.**
Measured on gjs 1.86: GObject does **not** clamp a bad construct value to the
paramspec bounds — it discards it with a non-fatal `GLib-GObject-CRITICAL` and
leaves `level` at **0**, zlib's store mode. So `level: 42` would ask for maximum
compression and return an *uncompressed* stream, larger than its input, at exit
0. The guard is range-only and not `Number.isInteger`, because Node accepts a
fractional level — the spec's Node leg failed a stricter version, which is what
that leg is for.

## Negative control

- `packages/node/zlib`: a test asserts **level 0 produces output larger than the
  input**, the one level that must not shrink it — the discriminator for "the
  level reached the compressor at all". Without it every level assertion passes
  against an implementation that ignores the argument, which is the state this
  PR fixes.
- Out-of-range levels must throw; the first version of that list included `2.5`
  and the Node leg went red, and the list now matches Node exactly.
- The before/after `.deb` above is the end-to-end control: the same tree, one
  argument removed, the warning returns.

## Ledgered, not fixed

Two findings the work turned up, both written into `status/open-todos.md`:

1. **`gzipDeterministic` is deterministic per HOST, not per artifact.** Fedora's
   `libz.so.1` is `zlib-ng-compat` 2.3.3; Node bundles `1.3.2.1-motley`. Over
   this repo's `CHANGELOG.md` they agree at **no** level, including the default
   (272 003 vs 272 000 bytes), and zlib-ng's level 9 is ~2.9 % *worse* than its
   own level 8 — which is why the level is asked for only where a reader demands
   it and is not blanket-applied to the payload tarballs. `data.tar.gz` and
   `control.tar.gz` are compressed at pack time and so differ across hosts;
   `changelog.Debian.gz` is immune, being compressed once at assembly time and
   travelling as base64 in the sidecar. `tests/e2e/ship-from-stage` is
   structurally blind to this — both its packs run on one host.
2. **The `.rpm` has the same gap** (`rpmlint` 2.8.0: `no-changelogname-tag`), and
   the blocker is the oracle, not the writer: `rpmlint` appears nowhere in this
   repository, and the RPM changelog is header data built at pack time, so
   `--from-stage` would need a sidecar schema 6 → 7 bump. Both are written down
   with the tag numbers so the next session does not re-derive them.

## Tests

`@gjsify/zlib` and `@gjsify/tar` green on **both** legs — the GJS leg is what
proves the `node:zlib` → `@gjsify/zlib` → `Gio.ZlibCompressor` route, which is
the one `gjsify ship` takes when the CLI runs under `gjs`.

### DIFF
--- a/.github/ship-oracle/verify-deb.sh
+++ b/.github/ship-oracle/verify-deb.sh
@@ -214,6 +214,15 @@ TAG_LIST=$(lintian-explain-tags --list-tags 2>/dev/null || lintian --list-tags 2
 #   no-copyright-file                Policy § 12.5 — the licence overlay
 #   no-changelog                     Policy § 4.4 — the changelog overlay
 #   changelog-file-not-compressed    it ships gzipped, or not at all
+#   changelog-not-compressed-with-max-compression
+#                                    Policy § 4.4 asks for `gzip -9 -n`, and this
+#                                    tag is read off the gzip header's XFL byte
+#                                    alone. GATED rather than merely fixed because
+#                                    the level travels four hops to get here —
+#                                    `plan.ts` → `gzipDeterministic` →
+#                                    `@gjsify/tar` → `node:zlib` — and every one of
+#                                    them has a default-level path it can fall back
+#                                    to without failing.
 #   syntax-error-in-debian-changelog the header/trailer this writer renders
 #   no-md5sums-control-file          the control member's md5sums
 #   md5sum-mismatch                  a digest in it (NOT `md5sums-mismatch`:
@@ -227,6 +236,7 @@ TAG_LIST=$(lintian-explain-tags --list-tags 2>/dev/null || lintian --list-tags 2
 #   malformed-deb-archive            the ar container and its member order
 #   control-file-has-bad-permissions the control member's modes
 for tag in no-copyright-file no-changelog changelog-file-not-compressed \
+    changelog-not-compressed-with-max-compression \
     syntax-error-in-debian-changelog no-md5sums-control-file md5sum-mismatch \
     file-missing-in-md5sums md5sums-lists-nonexistent-file malformed-md5sums-control-file \
     wrong-file-owner-uid-or-gid control-file-has-bad-owner malformed-deb-archive \
--- a/packages/infra/cli/src/utils/ship/changelog.ts
+++ b/packages/infra/cli/src/utils/ship/changelog.ts
@@ -144,20 +144,21 @@ const WIDTH = 80;
  * wrong: exactly one leading space, then `--`, one space, the maintainer, and TWO
  * spaces before the date.
  *
- * TWO lintian tags this file still raises, both measured on lintian 2.117 against
- * gjsify's own `.deb` and both left alone deliberately:
+ * ONE lintian tag this file still raises, measured on lintian 2.117 against
+ * gjsify's own `.deb` and left alone deliberately: `initial-upload-closes-no-bugs`,
+ * because the Debian revision is `-1` and the entry closes no bug. It is a
+ * statement about the Debian ARCHIVE's ITP workflow — the first upload of a new
+ * source package should close its ITP bug — and a package built by `gjsify ship`
+ * is never uploaded there. There is no bug number that would be true to write, so
+ * the tag is inapplicable rather than unmet.
  *
- *  - `initial-upload-closes-no-bugs`, because the Debian revision is `-1` and the
- *    entry closes no bug. It is a statement about the Debian ARCHIVE's ITP
- *    workflow — the first upload of a new source package should close its ITP bug
- *    — and a package built by `gjsify ship` is never uploaded there. There is no
- *    bug number that would be true to write, so the tag is inapplicable rather
- *    than unmet.
- *  - `changelog-not-compressed-with-max-compression`. Policy asks for `gzip -9`
- *    and neither `CompressionStream` nor `@gjsify/zlib` takes a level, so
- *    `gzipDeterministic` cannot ask for one. Ledgered in `status/open-todos.md`:
- *    the fix is a capability in the core, and stamping the header's XFL byte to
- *    claim max compression we did not use is not one.
+ * `changelog-not-compressed-with-max-compression` was a second one, and closing it
+ * was a capability gap in the CORE rather than anything wrong in this file:
+ * neither `CompressionStream` nor `@gjsify/zlib` would take a compression level,
+ * so `gzipDeterministic` had none to pass on. `@gjsify/zlib` honours
+ * `options.level` now and `plan.ts` compresses this file at
+ * `POLICY_MAX_COMPRESSION`. `verify-deb.sh` gates the tag by name, so it cannot
+ * come back quietly.
  */
 export function renderDebianChangelog(settings: PackSettings, source: string | undefined, mtime: number): string {
     const entries = source === undefined ? [] : changelogEntriesFor(settings.version, source);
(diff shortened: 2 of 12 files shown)


JumpLink:
## Independent review — every claim reproduced, one documented gap

Re-measured from scratch in a separate worktree. gjs 1.88.1 / Node 24.19.0 /
lintian 2.117.0ubuntu1.5 (Ubuntu 24.04 container, i.e. what `ubuntu-latest` runs).

### The GObject side-finding

```
asked 9  -> property reports 9
asked 42 -> GLib-GObject-CRITICAL: value "42" … invalid or out of range  ->  property reports 0
asked -7 -> GLib-GObject-CRITICAL: value "-7" … invalid or out of range  ->  property reports 0
asked 0  -> 0      asked -1 -> -1
```

Discard, not clamp, non-fatal, lands on 0 = STORE. Exactly as described.

### Node parity at the edges — `gzipSync(2000 bytes, {level: X})` on Node 24.19.0

| X | Node | `assertLevel` + `Math.trunc` | agree |
|---|---|---|---|
| missing / `undefined` | OK, XFL 0 | OK (`-1`) | yes |
| `-1` | OK, XFL 0 | OK | yes |
| `0` | OK, 2023 bytes (stored) | OK | yes |
| `9` | OK, XFL 2 | OK | yes |
| `10`, `-2` | `ERR_OUT_OF_RANGE` | `RangeError` | yes |
| `2.5` | **OK**, 41 bytes | OK → trunc 2 | yes |
| `-0.5` | **OK**, 2023 bytes (stored) | OK → trunc `-0` → store | yes |
| `9.9` | **throws** | **throws** (`9.9 <= 9` false) | yes |
| `±Infinity` | throws | throws | yes |
| **`NaN`** | **OK**, 35 bytes, XFL 0 (= default) | **throws** | **no** |
| `'9'`, `null`, `true` | `ERR_INVALID_ARG_TYPE` | coerced / `?? -1` | no (TS type errors) |

The 2.5-passes / 9.9-throws asymmetry is the non-obvious one and the expression reproduces it,
because Node compares the raw float against the bound rather than its integer part. The
"`Number.isInteger` reddened the Node leg" story checks out — the Node leg of this spec runs
Node's own `node:zlib` (`gjsify.runtimes.node` is `none`), so it IS the reference.

`NaN` is the one real divergence. The refusal is the right call — `Math.trunc(NaN)` would earn
the same CRITICAL as 42 and hand back a STORE stream — and it reaches no Node consumer for the
same slot reason. But nothing said so, and the comment's measured list read as "we match Node",
which is how a future session deletes the `!` to "restore parity" and reopens the bug. Written
down in a follow-up commit here, with the number.

### The tests discriminate

Reverted `gzipSync`'s level pass-through only (`compressWithGio(data,'gzip')`), rebuilt, re-ran:
**4 of 144 GJS tests fail**, leading with `expected 2, actual 0` on the XFL assertion. Not a
vacuous suite. `@gjsify/tar`'s new block is green on BOTH legs after a clean rebuild — the
`tar → node:zlib → Gio.ZlibCompressor` chain really does stamp XFL 2 under GJS, which is the
path `gjsify ship` takes.

### The lintian gate

Built two minimal `.deb`s differing only in the changelog's gzip level:

```
XFL of default: 0   ->  W: probe: changelog-not-compressed-with-max-compression [...changelog.Debian.gz]
XFL of max:     2   ->  (tag absent)
```

- the tag is raised off XFL alone — confirmed;
- `changelog-not-compressed-with-max-compression` IS in this lintian's `--list-tags`, so the
  gate's "does lintian still know this name" guard passes rather than red-ing the leg;
- the gate's `grep -qE "[ :]$tag( |\$)"` matches the real output line (which carries a `[path]`
  pointer after the tag) — checked against the literal line;
- it is a `W:`, so the non-gating `^E:` warning would NOT have caught it. The named gate is
  load-bearing;
- `ship-pack-linux` is in the `gate:` job's `needs:`, and the job apt-installs and probes for
  `lintian` / `lintian-explain-tags` before use. Fail-closed, blocking, non-vacuous.

`cmp -l` of a default vs `-9` changelog of that shape: `9 0 2` — one byte, position 9, same
size. The "stamping would have produced the identical artifact, by coincidence of a small
input" argument reproduces.

### The `.rpm` deferral is load-bearing, not a half-solution

Verified independently: `rpmlint` appears nowhere in `.github/`, `.docker/` or `scripts/` — the
only hit in the tree is a prose comment inside `rpm.ts` — and `rpm.ts` writes no 1080/1081/1082.
So there is genuinely no oracle to gate a fix against, and the `build-ci-image.yml` ordering
trap is real. Shipping it blind in this PR would be the pattern the repo already paid for with
`msitools`. The ledger entry names both blockers concretely enough to pick up cold. Agreed.

### Not changed, named instead

- non-number `level` (`'9'`, `null`, `true`) diverges from Node's `ERR_INVALID_ARG_TYPE`:
  `null` is swallowed by `?? DEFAULT_LEVEL`, the other two coerce through the relational
  comparison. All three are TS type errors against `ZlibOptions`, so out of contract; closing
  them means a `typeof` guard plus replacing `??` with an `=== undefined` check, which is a
  wider change than this PR.
- `assertLevel` runs twice on the async path (`requestedLevel` then `compressWithGio`).
  Harmless.

PR #1644, opened 2026-09-11
feat(blueprint): resolve identifiers from the GIR, on @girs 4.9.0

JumpLink:
`orientation: vertical` reaches GtkBuilder XML as `<property name="orientation">1</property>`,
and until now the in-repo emitter wrote `vertical`. Eleven of the corpus files diverged on
twenty-three lines under that one cause. **Thirty-seven of the thirty-eight corpus files are
now byte-equal**, and `corpus/divergences.mjs` holds one entry on two lines: an
`accessibility { }` VALUE is typed by GTK's ARIA table, which is built in C and is the one
thing the GIR does not carry. Clause 5's silence is that last entry away.

## Why one PR and not two

The bump is the enabler and the emission is the reason for it: the resolver cannot run on
`@girs` 4.6.0, so a separate emission PR would be stacked and could not be green against
`main`. AGENTS.md § PR size settles that case with a measurement — "land one large feature PR
rather than several small stacked ones". Atomic commits, so the 466 mechanical version lines
and the files of substance read apart:

1. `chore(deps): @girs 4.6.0 -> 4.9.0`
2. `feat(blueprint): resolve identifiers from the GIR`
3. `ci: the corpus counts belong to the script`
4. `fix(blueprint): emit ARIA relations and states` — from the review pass, below

## What the bump brings, measured on the installed tree

The published `@girs/gtk-4.0@4.6.0` vocabulary exports **eight** names, none of them a number.
4.9.0 exports **fourteen**: `ENUM_VALUES` (ts-for-gir #465, released in 4.8.0), `PROP_ENUMS`
(#467, 4.9.0), `FLAG_VALUES`, `ENUM_DEPRECATED` and the two declared-unreadable remainders.
Every pre-existing table is **byte-identical** between the two versions and both carry
`libraryVersion` 4.23.3, so nothing generated from the vocabulary moves and no regeneration
is due.

- `gjsify run check` — 225 workspaces, exit 0. No type fallout from the bump.
- `gjsify upgrade --check` — 130 deps consistently declared.
- The lockfile was **already at version 4**, so this install performed no v2 → v4 migration.
  The diff is versions only.

## Two lookups, and only one was available before

Resolving `vertical` to `1` needs the number behind a nick *and* the join from
`GtkBox.orientation` to `GtkOrientation`. The number was readable from the installed typelib;
the join was not readable at all — searching the nick lists for an enum with a member called
`never` finds several, and guessing between them is the silent-wrong-output ADR 0053 clause 3
exists to prevent. `PROP_ENUMS` is that join. It is keyed by the type that DECLARES the
property, so the walk goes through `DECLS`, the flattened ancestry the vocabulary already
ships — which is how `orientation` is found on `GtkOrientable` and not on `GtkBox`.

`packages/infra/blueprint/src/resolve-ident.mjs` does both lookups **from the pinned npm
dependency, not from a typelib**. That keeps stage C's property that it has no skip path and
runs on every runner, GNOME or not — a typelib-reading resolver would have reintroduced
exactly the hole `--require-oracle` closes one stage over.

## Four things the oracle was asked rather than assumed

Each measured against `blueprint-compiler` 0.20.4, each now a corpus file:

- **A member is spelled with UNDERSCORES.** `halign: baseline_fill` is `4`;
  `halign: baseline-fill` is an error.
- **A flag set is NOT numbered.** `input-hints: word_completion | lowercase` stays
  `word-completion|lowercase`, hyphenated and joined with no spaces.
- **`layout { }` does not resolve through the widget.**
  `Gtk.Grid { Gtk.Label { layout { halign: center; } } }` emits `center` — the entry belongs
  to `GtkGridLayoutChild`. `rules/19-layout.blp` now pins it.
- **`accessibility { }` does not either.**
  `Gtk.Label { accessibility { orientation: vertical; } }` emits `1` although `GtkLabel` is
  not orientable — the ARIA table answers there, not the ParamSpecs.

The last two are why both blocks pass nothing to the value resolver: the widget is the table
nearest to hand and it is the wrong one in each.

## The numbers were read back against the independent oracle

Of the **737** values in gtk-host's typelib-read `enum-values.mts`, **736 agree** with `@girs`
4.9.0 exactly. The single disagreement is the documented version gap —
`GtkEditableProperties.num-properties` is 8 on the installed GTK 4.22.4 and 10 in the GIR of
4.23.3 — and `@girs` additionally carries the two entries that host had to declare
unavailable. Two independent readings of the library agree; where they do not, the reason is
named.

## Negative controls — the guard was broken on purpose, four ways

- Delete `resolve-ident.mjs` — RED, naming the file: "this is a deletion and not an
  environment".
- Resolver always returns `null` — RED on eleven files, first diff line quoted, with no
  ledger entry left to hide behind.
- Off-by-one on every enum number — RED: golden `1`, in-repo `2`.
- Drop the `_` → `-` normalisation — RED: the parser refuses `baseline_fill` and
  `word_completion` by name.

And the positive control came for free: with the resolver in and the ledger untouched, stage C
failed **eleven times** with "byte-equal and still listed — delete the entry". The ledger's
self-retirement is what emptied it, not a hand edit.

## Also here

- ADR 0053 gains **§ Amendment 1**: emission needs introspection, it comes from `@girs` and
  not from a typelib, and clause 4's reservation was too narrow. No new ADR number was taken.
- `status/open-todos.md`: the resolved section is deleted; three entries are new or rewritten
  — the duplicated enum table (`@girs` vs. gtk-host's generator, with the cost of the swap
  measured), the ARIA gap, and the clause-7 demotion, which stays planned.
- The stage C summary line no longer prints three zeros at a reader who has to add them up.

## What the independent review pass changed

The four facts above were measured, and one of them only half. `accessibility { }` had a rule
file holding `label: "an accessible name"` and nothing else — an ARIA property, with a string
value, which is the one shape in that block where nothing shows. Widened to a relation, a
state and an enum, it reports three more things, and two were wrong output rather than a
missing lookup:

- A relation emits `<relation>` and a state emits `<state>`. The emitter wrote `<property>`
  for all three, and GtkBuilder rejects either of the other two spelled that way.
- A name that is none of the three is refused by the oracle by name, and was passed through
  here.

Both are answerable from data this PR already pulls in: the nick lists of
`GtkAccessibleProperty`, `GtkAccessibleRelation` and `GtkAccessibleState`. So
`resolve-ident.mjs` gains a second seam, the emitter asks it per entry, and an unknown ARIA
name now refuses the file with the oracle's own sentence. The VALUE is not answerable — GTK
types the slots in `gtk_accessible_property_init_value` and the GIR carries that function, not
its table — so `checked: true` stays `true` where the compiler writes `1`. That is the single
ledger entry, and it retires itself when ts-for-gir emits the ARIA value types.

Which is why the headline moved: **the shadow run is not silent, so clause 7 stays planned.**
A corpus that does not probe a construct is the second place a tolerated divergence can hide,
and the first place nobody looks — the ledger at least prints itself on every run.

`19-layout.blp` gained `halign: center` for the opposite reason: byte-equal, and it pins the
decision that a layout entry is NOT typed through the widget, which nothing held before.

## Not in this PR, deliberately

- `docs/release-notes/next.md` — v0.49.0 is being cut, so this touches nothing there.
- The `blueprint-compiler` demotion of clause 7 (`resolve-compiler.ts` + spec, the
  `oxlint-disable` in `loading-stack.ts`, the programmatic storybook window, the `not on PATH`
  skip, the MSYS2 branch). Its precondition is met now; it is a deletion PR of its own.
- Retiring `scripts/generate-enum-values.mjs` in favour of the vocabulary. Available and
  measured above, but it has to keep `generated.spec.ts` pointed at the running typelib or the
  artifact becomes a reader agreeing with itself. Written up in `status/open-todos.md`.


JumpLink:
## Independent review — what was re-measured, and what it found

Everything below was run in a separate worktree on `blueprint-compiler` 0.20.4 with
`@girs` 4.9.0 installed. Tables live here and not in the PR body: on a squash merge the
body becomes the commit body, and `body-max-line-length` is 100.

### The claims in the body, checked

| claim | verdict |
|---|---|
| 38/38 byte-equal, 0 divergences | **was true as measured**, `EXIT=0` — but see below |
| baseline on `main`: 26 byte-equal, 11 ledgered | **confirmed**, re-run detached at `fbdfcfc`: "26 byte-equal, 11 ledgered across 1 cause(s) and 23 named line(s)" |
| 466 mechanical version lines | **confirmed** (610 `+` lines minus 144 file headers) |
| `@girs` pins complete and consistent | **confirmed**: 470 declarations, every one exact and at 4.9.0; 25 distinct names, 0 mismatches against the lockfile; the 7 extra lock entries are transitive and also 4.9.0 |
| no package left on an old resolution | **confirmed**: only `@girs/gwebgl-0.1` and `@girs/gjsifywebrtc-0.1` are off the train, both pre-existing documented exemptions in `check-girs-exact-pins.mjs` |
| `layout { }` passes nothing | **confirmed on a case the corpus did not cover**: `Gtk.Grid { Gtk.Label { layout { halign: center; } } }` — oracle `center`, in-repo `center`, byte-equal |
| `accessibility { }` passes nothing | **confirmed, and incomplete** — see the next table |
| the ledger is data, not an `if` | **confirmed**: no per-file or per-widget branch anywhere in `resolve-ident.mjs` or `emit-xml.mjs` |
| the resolver is honest on missing data | **confirmed by injection** — below |

### Negative controls re-run, not taken on trust

| sabotage | result |
|---|---|
| delete `resolve-ident.mjs` | RED, naming the file |
| resolver always returns `null` | RED on 12 files with the diff quoted |
| hand-edit a golden to match a wrong emitter | RED **twice**: stage B says the golden is not what the compiler produces, stage C says the emitter does not reproduce it. The anti-cheat holds |
| move `GtkAlign.baseline-fill` into `ENUM_VALUES_UNREADABLE` | RED: "`baseline_fill` is a member of GtkAlign whose value @girs could not read ((null))". No silent fallback to the name |
| a stale ledger entry on a byte-equal file | RED: "byte-equal and still listed … delete the entry" |
| classifier degraded to always-`property` | RED: "line 13 differs … the entry excuses [14, 15] and nothing else" |

### What the corpus was not asked

`rules/20-accessibility.blp` held `label: "an accessible name"` and nothing else. `label` is
an ARIA *property* with a *string* value — the single shape in that block where three
separate behaviours are invisible. Measured against the oracle:

| written in `accessibility { }` | oracle | in-repo, before | in-repo, now |
|---|---|---|---|
| `label: "an accessible name"` | `<property name="label">` | same | same |
| `row-index: 3` | `<relation name="row-index">3</relation>` | `<property …>` | **fixed** |
| `checked: true` | `<state name="checked">1</state>` | `<property …>true` | `<state …>true` — value **ledgered** |
| `orientation: vertical` | `<property name="orientation">1</property>` | `vertical` | `vertical` — **ledgered** |
| `not-a-real-aria-name: "x"` | refuses the file by name | emitted it | **refuses**, same sentence |

The element half is vocabulary data (three nick lists) and is now resolved. The value half is
not: GTK types the slots in `gtk_accessible_property_init_value`, and the GIR carries that
function, not its table. Hence one ledger entry, two lines, self-retiring.

### The one thing that would have gone red on merge

The PR body carried two Markdown tables, six of whose rows are 102–191 characters. On a
squash merge the body becomes the commit body and `Lint commit messages` is a required check.
Body rewritten as prose lists; the tables are here instead.

### Reported, not fixed

`resolve-ident.mjs` imports Gtk and Adw only, so a `.blp` with a third `using` namespace
resolves nothing and silently keeps the source spelling. Not reachable today — all 38 tracked
`.blp` are Gtk/Adw, and stage C would catch a new one that diverges — but it becomes a
consumer-visible hazard at the clause-7 flip, when the in-repo emitter starts producing the
bytes a build uses. It belongs to the flip PR, with the decision about what a build does when
a real file leaves the subset.

PR #1645, opened 2026-09-11
test: a scaffolded workflow is read by actionlint, not by four regexes

JumpLink:
Closes item 9 of the `gjsify ship` roadmap (ADR 0024): *"a scaffolded workflow
is verified by nothing."*

`gjsify flatpak ci` writes a GitHub Actions workflow into somebody **else's**
repository. Four `assert.match` regexes over raw text were the only thing that
had ever looked at the output — never parsed as YAML, never actionlint'd, never
run. ADR 0024 names this exact class for `ship ci`, which will scaffold too.

## Measured, on the real scaffolded `flatpak.yml`

One mutation each, actionlint 1.7.7. "yaml" is a plain `YAML.parse` — the reader
a hand-rolled structural check would be built on:

| mutation | actionlint | yaml | the 4 original regexes |
|---|---|---|---|
| unclosed `[` in `branches:` | refuses | refuses | **all still match** |
| `runs-on:` → `runs_on:` | refuses | accepts | **all still match** |
| `actions/checkout@v4` → no ref | refuses | accepts | **all still match** |
| `github.sha` → `github.shaX` | refuses | accepts | **all still match** |
| `jobs:` → `jbos:` | refuses | accepts | **all still match** |
| `on:` → `onn:` | refuses | accepts | **all still match** |
| unterminated `if` in a `run:` | **accepts** | accepts | **all still match** |

Each regex asserts a substring the mutation does not touch, which is why all
seven pass — the textbook shape of a green that checked nothing.

## The mechanism, not four more regexes

`scripts/check-scaffolded-workflow.mjs`, with two modes:

- **`--root <dir>`** reads a scaffolder's output with two independent readers,
  neither redundant: **actionlint** for the document (6 of 7 mutations), and the
  existing **`check-workflow-run-syntax.mjs --root`** for the shell inside
  `run:` blocks — the row actionlint is blind to, because it runs with
  `-shellcheck=` empty. Reused rather than reimplemented.
- **`--coverage`** is the static half and needs no build, so it runs in
  `audit-runtimes.yml`: any CLI source that writes a `.github/workflows` path
  and is in no `SCAFFOLDERS` entry fails the check. That is what stops the next
  scaffolder — `ship ci` — from landing with nothing reading it, the same shape
  as `manifest-conformance`'s `field-coverage`.

A YAML parse is deliberately **not** added: it catches one of seven, so
structural assertions of our own would be close to the vacuum they replace, and
`audit-runtimes.yml` already argues actionlint over a hand-rolled checker for
this repo's own workflows. No new dependency enters the tree.

`flatpak ci` emits no `run:` block today, so the second reader is vacuous **on
this scaffolder** and not on the class — the ledger's minimum bar for `ship ci`
was `bash -n` on every extracted `run:` block, and it is wired for the day
`ship ci` exists.

## Negative control

The script **self-tests its discriminators on every invocation** before judging
anything — the same construction as `check-workflow-run-syntax.mjs`'s
`PWSH_MUST_REJECT`. An unmutated control must be accepted (a reader that refuses
everything would "catch" every mutant and prove nothing), then each mutant must
be refused. Driven red deliberately:

- one mutant turned into a no-op → `actionlint accepted a misspelled runs-on —
  it is not discriminating`, exit 1.
- coverage, a new unregistered scaffolder → named, exit 1.
- coverage, a registered source moved away → named in both directions, exit 1.
- `--require-actionlint` with no actionlint on PATH → exit 1.

## Three traps worth not rediscovering

1. **`actionlint` with no file arguments needs a git repository** — pointed at a
   scratch directory it exits 3 with *"no project was found in any parent
   directories"*. Generated output is passed as an explicit file path.
2. **`bash -n` accepts `if [ -z "$x" ; then echo hi; fi`** — a `[` missing its
   `]` is a runtime error from the builtin, not a parse error. The first
   negative control here proved nothing and was replaced with an unterminated
   `if`.
3. **The coverage scan's first version read raw source** and named
   `utils/gjsify-shim.ts`, which only *mentions* `.github/workflows/release-cut.yml`
   in a prose comment. Comments are stripped before the grep now — a rule that
   cries wolf earns an exception list, and an exception list is where the real
   scaffolder eventually hides.

## Not closed

`actionlint` is on no runner image this repo uses and in no `dnf install` in
`.docker/ci-fedora.Dockerfile`, so on CI today only the shell reader runs — and
the script **says so** rather than passing quietly. `--require-actionlint`
exists and is passed nowhere yet: turning it on means adding actionlint to the
CI image in its **own** PR first, because `build-ci-image.yml` publishes only on
a push to `main` and a PR that adds a package *and* a test hard-requiring it can
never go green (the trap `msitools` hit in #1354 M5). Written into the ledger.

> Touches `status/open-todos.md` in the same region as the Debian-changelog PR
> (that one deletes roadmap item 10, this one item 9). Both deletions are
> independent; merging the second will want a two-line conflict resolution that
> takes both.


JumpLink:
## Independent review — the matrix reproduces, and one reader never runs

Re-measured from scratch in a separate worktree, actionlint 1.7.7 (same build), against a
`flatpak.yml` reconstructed byte-for-byte from `renderWorkflow()`'s defaults. `yaml` is
`YAML.parse` from the repo's own `node_modules`.

| mutation | actionlint | yaml | run-syntax | the 4 original regexes |
|---|---|---|---|---|
| CONTROL (unmutated) | accepts | accepts | accepts | all 4 match |
| unclosed `[` in `branches:` | REFUSES | REFUSES | accepts | all 4 match |
| `runs-on:` → `runs_on:` | REFUSES | accepts | accepts | all 4 match |
| `actions/checkout@v4` → no ref | REFUSES | accepts | accepts | all 4 match |
| `github.sha` → `github.shaX` | REFUSES | accepts | accepts | all 4 match |
| `jobs:` → `jbos:` | REFUSES | accepts | accepts | all 4 match |
| `on:` → `onn:` | REFUSES | accepts | accepts | all 4 match |
| unterminated `if` in a `run:` | accepts | accepts | REFUSES | all 4 match |

**6 / 1 / 1 / 0 over seven mutations.** The claim holds exactly, control included. The
argument for two readers and against a hand-rolled structural check stands on the number.

### What did not hold: the pair is vacuous on every CI host

`actionlint` is on no image this repo uses — not `ghcr.io/gjsify/ci-fedora:44` (probed: absent),
and `audit-runtimes.yml`'s download lands in a `mktemp -d`, never on `PATH`, in a step that runs
*after* the new `--coverage` step anyway. Measured on such a host:

```
$ node scripts/check-scaffolded-workflow.mjs --root <dir with runs-on: misspelled runs_on:>
OK — 1 scaffolded workflow(s) parse and lint.
   actionlint is NOT on PATH — the document-level reader did not run on this host.
exit=0
```

The second reader does not cover for it: `flatpak ci` emits no `run:` block, so
`check-workflow-run-syntax.mjs` reads **zero** blocks and exits 0. So on CI the new e2e test
asserted nothing beyond "the file exists" — and reported a **pass** while doing it. The ledger
entry says "only the shell reader runs", which understates it: neither reader reads anything.

Fixed here by the repo's own mechanism rather than a `console.log`: the test now sits behind
`e2eSkipReason('flatpak', [['actionlint on PATH …']])`, so the absence is a visible SKIP,
`GJSIFY_E2E_REQUIRE=flatpak` makes it a named failure for a job that claims the tool (#1550),
and the invocation passes `--require-actionlint` so the tool vanishing between probe and run is
a failure too. The two ways to make the leg actually RUN are ledgered.

### The coverage scan missed one of the two writer spellings

`/writeFileSync/` only. Measured four directions with probe files under the CLI source:

| probe | before | after |
|---|---|---|
| scaffolder using `writeFileSync` | caught | caught |
| scaffolder using `await writeFile` (`node:fs/promises`) | **MISSED** | caught |
| a comment + a string mentioning the path, no writer | no alarm | no alarm |
| reads `.github/workflows`, writes a report elsewhere | false alarm | false alarm (unchanged) |

`actions/build.ts` already writes with `await writeFile` at three call sites, so that is a
shape `ship ci` could plausibly take — on the check whose only job is to catch it. Widened to
`\bwrite(File|FileSync)\s*\(`; measured to add no new hit anywhere in the CLI source today.

The comment-stripping IS load-bearing, confirmed: `utils/gjsify-shim.ts` has `writeFileSync` at
four call sites and mentions `.github/workflows/release-cut.yml` in a prose comment, so the raw
scan really would name it.

Row 4 stays a false-positive shape (the two predicates are file-level, deliberately — the real
scaffolder builds its path lines earlier, so proximity matching would miss `flatpak ci` itself).
Worth knowing, not worth an exception list.

### Two smaller things

- Both new/edited files failed `oxfmt --check` at the pushed commit, which is a gate in
  `main.yml`. Fixed.
- The header table has seven rows; `MUTANTS` runs six (`on:` → `onn:` is not among them), and
  the `--require-actionlint` message says "six of the seven mutations this script measures".
  Consistent if you read "measures" as the table, confusing if you read it as the self-test.
  Left alone.
- `audit-runtimes.yml` pins actionlint **1.7.12**; the matrix was taken on **1.7.7**. Not a
  defect — but if the `e2e` job ever installs the pinned build, re-take the matrix on it.

PR #1646, opened 2026-09-11
fix(workspace): a local dependency is not only a `workspace:` one

### LINKED ISSUE #1587: workspace --with-dependencies builds nothing and exits 0 outside a workspace: tree
## What

`gjsify workspace <pkg> build --with-dependencies` resolves its closure through `collectTransitiveClosure`, which follows **only** dependencies declared with the `workspace:` protocol. In a monorepo whose packages depend on each other by plain semver range — a deliberate choice, and a legal one — it matches nothing, builds nothing, and **exits 0**.

Measured in `JumpLink/Learn6502` while migrating it to 0.48.0: the Flatpak build calls it to assemble `app-gnome` with its local dependencies, and it produced an `app-gnome` **without `@learn6502/core`**. The command reported success. The failure surfaced much later, as a missing module at runtime, in a build nobody suspected.

## Why it is the bad kind

A build tool that builds nothing and says so is a nuisance. One that builds nothing and reports success is the "green that checked nothing" shape: every downstream step then measures an artifact that was never assembled, and each of them passes for the same wrong reason. Here it took a Flatpak build, a container run and a manual `ls` to find out what the exit code had already claimed.

The `workspace:` protocol is the right *hint* — it is unambiguous. But its **absence** cannot mean "this package has no local dependencies", because plain-range monorepos exist and this one is not exotic.

## What a fix has to decide

- **Refuse, or widen?** Refusing (`--with-dependencies` finds no `workspace:` dependency → exit non-zero, naming what it looked for) is the smaller change and turns a silent wrong answer into a loud one. Widening (also resolve a dependency whose name matches a workspace member and whose range the member's version satisfies) makes the flag do what its name promises, at the cost of a resolution rule that has to be exact about version mismatch.
- **What a mismatch means**: a workspace member named `@x/core@2.0.0` against a declared `^1.0.0` is not the same package the consumer asked for. Silently substituting it would be a second silent defect.
- The consumer's own workaround — spelling the chain out (`core → examples → learn`) in meson — is what every caller will invent independently until the flag either works or refuses.

Consumer-side workaround is in place; nothing is stuck.

JumpLink:
`gjsify workspace <pkg> <script> --with-dependencies` resolved its closure through a private
walk that followed **only** `workspace:`-protocol edges. In a monorepo whose packages depend on
each other by plain semver range — legal, deliberate, and not exotic — it matched nothing, built
nothing and **exited 0**.

Measured in `JumpLink/Learn6502`: the Flatpak build calls this to assemble `app-gnome` with its
local dependencies, and it produced an `app-gnome` **without `@learn6502/core`**, reporting
success. The failure surfaced much later, as a missing module at runtime, in a build nobody
suspected. It took a Flatpak build, a container run and a manual `ls` to find out what the exit
code had already claimed.

## Widen, not refuse

The issue left the choice open. Widening, because the `workspace:` protocol is the right *hint*
but its **absence cannot mean "this package has no local dependencies"** — and because there is
an authority for what "a local dependency" means: the rule npm and yarn themselves apply when
they link a workspace instead of fetching from the registry.

`buildDependencyGraph` now makes an edge for the two spellings a package manager links:

1. the **`workspace:` protocol** — explicit, linked whatever the versions say;
2. a **plain registry range** whose named member's version **satisfies** it.

A dist-tag (`latest`), an alias (`npm:other@^1`), a `file:` spec and a URL are not registry
ranges and link nothing. The dist-tag needs its own clause rather than `validRange`'s answer:
`@gjsify/semver` deliberately parses `latest` as `>=0.0.0` so registry packuments resolve, and
that leniency is right there and wrong here.

### A version mismatch is neither linked nor silent

A name that matches a member the range does **not** admit is deliberately not an edge —
`@x/core@2.0.0` against `^1.0.0` is not the package the consumer asked for, and substituting it
would be a second silent defect. It goes into `DependencyGraph.unlinked`, and the command PRINTS
it:

```
gjsify workspace: "@plain/app" depends on "@test/utils": ^9.0.0, which the local
workspace (@test/utils@0.0.1) does not satisfy — NOT built as a dependency;
the installer resolves it from the registry.
```

That is the one case a caller cannot otherwise tell from having no local dependency at all. And
when the expansion genuinely finds nothing, it now says so instead of exiting 0 in silence — a
leaf package legitimately has none, but "none" and "the flag could not see them" produced the
same silent success.

### The duplicate walk is gone

`collectTransitiveClosure` in the CLI duplicated `affectedClosure(buildDependencyGraph(all),
[target])` — the pair `gjsify foreach --with-dependencies` already calls three files over. **The
copy is where the defect lived.** One walk means the rule for "what is a local dependency" has
one place to be wrong in.

## Measurements

**The widening is a no-op in this monorepo.** Against gjsify's own 327 workspaces, comparing the
new edge set against the `workspace:`-only rule:

```
includeDev=false: workspaces=327 newEdges=0 unlinked=0
includeDev=true:  workspaces=327 newEdges=0 unlinked=0
```

So every consumer of this graph resolves exactly as before here — `foreach -t`,
`topologicalSort`, `affected-classify`, selective CI, and the ADR 0006 build cache, whose keys
are computed from it. The change can only add an edge where a plain range already pointed at a
local member, and nothing in this tree has one.

**Negative control.** With `graph.ts` + `workspace.ts` reverted to `main` and the tests kept,
`tests/e2e/workspace-with-dependencies` reports **10 pass / 3 fail** — exactly the three new
cases, with the ten existing ones untouched:

```
✖ --with-dependencies follows PLAIN SEMVER RANGES, not just `workspace:`
✖ does NOT build a name match the local version fails — and reports it
✖ says so when the expansion finds no workspace dependencies
```

With the fix: **13 pass / 0 fail**. The unit suite went 33→37 the same way (4 red first, on both
Node and GJS).

**Also green:**
`tests/e2e/{foreach-command,affected-classifier-bundle,build-cache,workspace-unknown-name}`
(52/52) and `tests/e2e/bootstrap-cold-tree`. The new `@gjsify/semver` dependency of
`@gjsify/workspace` is tier 1 → tier 1 (`audit-runtimes --check` green) and both names are
already in `bootstrap-native-facades.mjs`'s `CLI_RUNTIME_DEPS`, which is what makes the
cold-tree link order hold.

## The class

"Are there other commands that silently return 0 outside their context?" — asked of every
command in `packages/infra/cli/src/commands/`. Most answer well: `foreach --include`, `onboard
--packages`, `storybook`, `ship`, `run`, `check`, `trust`, `dev` and every `flatpak` subcommand
exit non-zero; `prune`, `install`, `info` and `affected` print what they found. **Four do not**,
and they are now a `status/open-todos.md` section with the evidence:

1. **`barrels --paths <dir>`** — a directory it cannot read is `continue`d and contributes
   **0 drift**, so `--check` goes green for a barrel nothing looked at. A guard that checked
   nothing.
2. **`build`** — an entry / `--library` glob matching no files has **no guard at gjsify's
   layer**; `[]` flows into rolldown's `input`, and every post-build guard is offender-based
   and passes trivially on nothing.
3. **`upgrade --workspace <glob>`** — prints one line and exits 0 where `foreach` asserts, so
   a stale name reports success having edited nothing.
4. **`pack`** — an unbuilt package writes a `.tgz` of `package.json` + README and prints its
   name at exit 0; the guard for that lives in a script *outside* the command.

`clear`/`copy` wildcards that match nothing are SILENT-0 too and are left alone — both are
argued for as shell parity and neither produces an artifact a later step measures. The shape of
the fix is the distinction `assertEveryIncludeMatches` already draws: *a pattern the caller
wrote that matched nothing is an error; a filter that emptied a real set is not.*

## Review round

An independent review found three holes in the range reader, **one of them rebuilding the
defect this PR is about**.

A **prerelease** member version dropped every plain-range edge, `*` included. node-semver
excludes a prerelease from a range that does not name one, so — measured against the built
`@gjsify/semver` — `satisfies('0.49.0-rc.1', '^0.49.0')` and even
`satisfies('1.0.0-beta.1', '*')` are false. That rule is a RESOLVER's: it stops one picking an
unstable version by surprise. This graph picks nothing; it asks whether a dependency IS the
local package, which its prerelease tag does not change. **This monorepo cuts `x.y.z-rc.N` on
every release**, so a train mid-RC lost every plain-range edge and built nothing again — with
a warning confidently pointing at the registry for a package that was right there. The
comparison now drops the tag; the range itself is not widened, so `2.0.0-rc.1` against
`^1.0.0` still fails and is still reported.

A **valid range with no digit** was dropped in silence — `x.x.x`, `x` and `''` are legal
any-ranges, and the dist-tag filter looked for a digit, so they linked nothing AND reported
nothing. That is the one shape this whole change exists to remove. The reader now asks npm's
own question — is `validRange` happy? — minus the single word `@gjsify/semver` is lenient
about, `latest`. `next`, `beta` and `latest-3` were already `null` there, and `v2` / `2-legacy`
are ranges to npm too, so they stay ranges here.

And `unlinked` **could name a pair the graph HAS an edge for**: one dependency name in two
blocks (npm documents `optionalDependencies` overriding `dependencies`) gave one linking
spelling and one reported one. Entries are collected per workspace and filtered against the
edge set — the report explains a MISSING edge, and naming a linked pair is worse than saying
nothing.

Negative controls, one per fix: **1, 1 and 2** of the 40 specs fail when each is taken back on
its own. The monorepo measurement is unchanged — 0 new edges, 0 unlinked across 327
workspaces, with and without dev.

Closes #1587


JumpLink:
## Independent review — three commits pushed

Second pass by a reviewer who did not write this. Measurements first.

| claim | how it was checked | result |
|---|---|---|
| 0 new edges against this monorepo | ran both rules side by side over `discoverWorkspaces(root)` and diffed the adjacency sets | **327 workspaces. `includeDev=false`: 414 → 414 edges, 0 added, 0 lost, 0 unlinked. `includeDev=true`: 1068 → 1068, same.** Confirmed |
| node-semver drops a prerelease from a plain range | `satisfies` probes against `@gjsify/semver` | `('1.0.0-beta.1','*')` false, `('0.49.0-rc.1','^0.49.0')` false, `('0.49.0-rc.1','>=0.0.0')` false. Confirmed — `releaseVersionOf` is load-bearing on every release cut |
| `registryRangeOf`'s edges | `validRange` probes | `latest`/`LATEST` → `>=0.0.0` (hence the explicit subtraction), `next`/`beta`/`catalog:`/`catalog:default` → `null`, `v2`/`2-legacy` → ranges, `''` → `>=0.0.0`. All as documented |
| tier discipline for the new `@gjsify/semver` dep | read both manifests | tier 1 → tier 1. Fine |
| committed bundles | `node scripts/verify-committed-bundles.mjs` | all reproduce byte-identically; `build:affected-bundle` leaves no drift |

### 1. There was a second place with the same semantics, and this PR made it diverge

`gjsify foreach -t -p` takes `runTopologicalParallel`, which built its dependency map from its
**own** `workspace:`-only walk (`foreach.ts:690-704`). The sequential leg at `foreach.ts:463-467`
uses the shared graph. Before this PR the two agreed; after it they would not: on a monorepo
declaring local deps by plain range, `-t` orders correctly while `-t -p` sees no edges at all,
releases every workspace at once, and exits 0. That is the defect this PR exists to remove, one
flag over — and the PR's own sentence applies verbatim ("One walk means the rule for 'what is a
local dependency' has one place to be wrong in").

Replaced with `buildDependencyGraph(workspaces, { includeDev })` over the selected set, which
yields exactly the edges the copy computed (it indexes by name from the list it is handed).
Held by a new e2e leg on the `@plain/*` fixture at `--jobs 1`, so ordering is the only variable.
Reverting the walk in the built file reds it: `mid` lands after `app`.

### 2. The `unlinked` warning stated something `gjsify install` does not do

Both `UnlinkedDependency`'s doc and the printed line said the installer resolves such a
dependency **from the registry**. `commands/install.ts` (the `localWorkspace` branch) symlinks a
name-matched member whatever the spec says — only an explicit `link:`/`file:`/`portal:`/`git+`/
`http(s):` opts out, because routing one into the fetch queue would `rmSync` the workspace's own
source tree. So under this repo's own installer the consumer imports the local package while the
graph declines to build it, and the warning explained that away with the one thing it cannot
know.

The decision is untouched — npm/yarn link on name AND range, and substituting a `2.0.0` for a
`^1.0.0` request would be a second silent defect. What changed is that neither the type nor the
message asserts an outcome now: both name the two installers and ask for the range and the
member version to be brought into agreement.

### 3. The barrels finding: the documentation is accurate, and it may not stand

Verified line by line against `actions/barrels-generate.ts:65-76` and `commands/barrels.ts:101,
124-129` — the ledger entry describes the mechanism exactly. But a `--check` that reports 0 drift
for a directory it never read is a guard going green, in the same package this PR already edits,
and the root `AGENTS.md` root-cause rule ("no 'known limitation' notes") bites on a finding this
PR's own sweep produced. Fixed: `unscannableBarrelPaths` refuses every named path that is not a
readable directory, at the command's selection boundary — the shape `assertEveryIncludeMatches`
already has. The generator stays tolerant, so its "does not stop on missing directories" spec
still holds and still asserts 0 drift; the two callers genuinely want opposite things. Four new
specs; stubbing the helper to `[]` reds three. The ledger keeps the incident and carries the
remaining three items.

### Left alone, deliberately

`foreach --with-dependencies` (`foreach.ts:381-390`) uses the shared graph but does **not** print
`unlinked`, while `workspace --with-dependencies` does. Same silent-closure failure mode, and
completing it means lifting `reportUnlinked` somewhere shared — an omission rather than a lie,
and out of this PR's scope. Worth a follow-up.

One CLI trap found while writing the e2e leg, for anyone copying the invocation:
`--include` is a yargs array option, so `gjsify foreach --include '@x/*' run build` swallows the
script name. `assertEveryIncludeMatches` catches it loudly, which is the guard working.

### Verification run locally

`@gjsify/cli` 1724 tests, `@gjsify/workspace` 40 tests, `tests/e2e/workspace-with-dependencies`
14/14, `gjsify tsc --noEmit` clean on both packages, `oxfmt --check` and `oxlint` clean on every
touched file.

**Verdict: mergeable with the three commits above.**

PR #1647, opened 2026-09-11
feat(adwaita-nativescript)!: icon-size is the enum, pixel-size is the number

### LINKED ISSUE #1584: gtk-image's iconSize carries pixel-size semantics under icon-size's name
## What

`@gjsify/adwaita-nativescript`'s `GtkImage.iconSize` takes a **number of DIPs**. `Gtk.Image:icon-size` is a **three-member enum** (`inherit`, `normal`, `large`). The port is carrying `Gtk.Image:pixel-size`'s meaning under `icon-size`'s name.

Measured while censusing value kinds across the NativeScript surface (gjsify#1583): of 109 property names that agree between the port and their GIR counterpart, nine carry a different KIND of value. Seven of the nine are decided portable forms (ADR 0042's menu model, ADR 0046's list model, ADR 0048's page). **The remaining two are this one, twice** — `GtkImage.iconSize` and its sibling — and the same census independently lists `pixelSize` as a property the port does not have.

## The user-visible half

```xml
<gtk:Image iconSize="large" />
```

`large` is what a reader of the GTK documentation writes, and it is one of exactly three legal values there. Here the setter runs it through `xmlNumber`, which cannot parse it, so it falls back — the image renders at **16 DIPs** and nothing is reported. A typo, a GTK-documentation value and a deliberate choice all produce the same silent result.

This is the false-friend shape the widget-vocabulary programme exists to remove: a name that agrees and a value that does not is worse than a name that never matched, because the disagreement is invisible at the call site and to every gate. `check-vocabulary-alignment` counts this pair as agreement — it compares names against a key set and never reads a type.

## Why it was left rather than fixed

Renaming `iconSize` → `pixelSize` renames a **published attribute** on a surface that another change is rewriting right now (`feat/ns-construct-props`, gjsify#1579, adds an optional construct-props bag to all 46 widgets). Landing a rename into that is how two changes become one conflict.

It is declared in the new coverage ledger meanwhile, with a `why` that names the collision.

## What a fix has to decide

- **Rename, or keep the name and take the enum?** `pixelSize` is the honest name for what the port does. Taking `icon-size`'s enum instead would mean mapping three nicks onto DIPs, which is a design decision about what `normal` means on a phone.
- **What an unparseable value should do.** Falling back to 16 in silence is the defect; refusing loudly is a behaviour change on a published attribute. The XML-door gate already holds `number` setters to `xmlNumber` — a third option is a door that refuses a non-numeric string by name.
- Both, if the answer is `pixelSize` **and** an `iconSize` that takes the enum: the port would then carry both GTK properties under their own names, which is the outcome the programme actually wants.

Not urgent — nothing is stuck, and the ledger entry keeps it visible.

JumpLink:
`@gjsify/adwaita-nativescript`'s `GtkImage.iconSize` took a **number of DIPs**.
`Gtk.Image:icon-size` is a **three-member enum** (`inherit`, `normal`, `large`); the number is
`Gtk.Image:pixel-size`. The port carried one property's meaning under the other's name — and
`AdwImageButton` carried the same pair, which is the "twice" in the issue.

## Why this shape is worse than a missing property

A name that agrees and a value that does not is invisible at the call site **and to every
gate**. `check-vocabulary-alignment` and `check-nativescript-widget-coverage` both counted this
pair as agreement, because both compare names against a key set and never read a type. The
coverage ledger went one better and listed `pixel-size` as a property the port did not have —
*beside the property that was it*.

The user-visible half:

```xml
<gtk:Image iconSize="large" />
```

`large` is what a reader of the GTK documentation writes, and one of exactly three legal values
there. Measured against the helper the old setter used:

```
xmlNumber("large", 16)   = 16
xmlNumber("normal", 16)  = 16
xmlNumber("inherit", 16) = 16
xmlNumber("32", 16)      = 32
```

So a typo, a GTK-documentation value and a deliberate choice all produced the same silent 16
DIPs.

## Both properties, under their own names

The issue asked which of three answers to take, and named the one it actually wants: *"the port
would then carry both GTK properties under their own names"*.

| property | takes | means |
|---|---|---|
| `iconSize` | `'inherit' \| 'normal' \| 'large'` | `Gtk.Image:icon-size` |
| `pixelSize` | a number of DIPs | `Gtk.Image:pixel-size`, overriding `iconSize` when set |

The precedence is `Gtk.Image`'s own: `gtk_image_set_pixel_size` documents -1 as "unset, the icon
size decides". Here a non-positive or non-finite value means the same thing, because a 0-DIP
image is not a rendering anyone asked for.

`pixelSize` also READS BACK -1 while unset, as `gtk_image_get_pixel_size` does, rather than the
size the image draws at. The second commit is that correction, and it is worth its own paragraph
because the first version got it wrong in the defect's own shape: every number setter in this
package falls back to its own getter (`xmlNumber(raw, this.<prop>)`), so a getter answering the
drawn size made `image.pixelSize = image.pixelSize` PIN it — set `iconSize = 'large'`, assign an
unparseable `pixelSize`, and 32 became an override a later `iconSize = 'normal'` could no longer
move. The sentinel and the store live in the helper, so the round trip is a spec rather than a
widget nothing off-device can import.

**An unparseable value THROWS**, naming what it was and the three that are accepted. That is the
third option the issue lists, and it is right here specifically: there is no range to be lenient
about, so "not one of them" is a caller mistake and nothing else — the opposite of `xmlNumber`,
which is lenient because a number genuinely has a sensible default.

The collision that deferred this is gone: `feat/ns-construct-props` (#1579) is merged.

## The sizes are measured, not remembered

`refs/gtk` is a submodule nobody initialises for this, so the pixel values were read off a live
toolkit instead — one `Gtk.Image` per member on **GTK 4.22.4 under gjs 1.88.1**, presented and
then `measure()`d on both axes:

| member | min = natural | css class |
|---|---|---|
| `inherit` | 16x16 | (none) |
| `normal` | 16x16 | `normal-icons` |
| `large` | 32x32 | `large-icons` |

`inherit` equalling `normal` is not a coincidence to tidy away: they agree on GTK too, and they
would stop agreeing the day this port grows a context that carries a size.

## The mechanism, so the class is visible

`widgets/gtk-icon-size.ts` is the table, and **arm 7 of `check-nativescript-xml-doors.mjs`**
holds it — the same shape arm 5 gives `Gtk.Align`: the nick list against `GtkIconSizeNick` in
gtk-host's generated props, in that order and that spelling, plus a pixel size for every member
and no entry for a non-member. Order counts because no member of this enum is an alias, so the
positions ARE the GIR constants.

What the arm cannot hold is the three numbers — no in-repo oracle carries GIR enum pixel
semantics — so `icon-size.spec.ts` pins them literally, the same division of labour
`gtk-align.ts` + `construct-props.spec.ts` already use.

## Measurements

**Red first.** The spec was written before the module: `UNRESOLVED_IMPORT: Could not resolve
'./widgets/gtk-icon-size.js'`. After the module and the review round, **1383 passed / 3906
assertions on both Node and GJS** (10 new tests).

**Negative control on the spec** — `large: 32 → 24` and the refusal replaced by a silent
fallback:

```
✖ Gtk.IconSize › resolves each nick to the size GTK renders it at
✖ Gtk.IconSize › takes the three nicks and REFUSES anything else by name
✖ Gtk.IconSize › names the member it could not find, and the three it accepts
✖ Gtk.IconSize › lets an explicit pixel size win over the icon size
✖ Gtk.IconSize › treats a non-positive pixel size as unset rather than as a size
❌ 5 of 1379 tests failed
```

**Negative control on arm 7**, both directions:

```
GTK_ICON_SIZE_NICKS is [inherit, large, normal] where GtkIconSizeNick … is [inherit, normal, large].
Gtk.IconSize 'large' has no entry in GTK_ICON_SIZE_PIXELS. Every member resolves to a size here —
a missing one would fall through to `undefined` and size an image to NaN, which is the silent
shape #1584 was.
```

**The coverage gate caught the ledger on its own**, before I edited it — which is the half worth
showing, because it is the gate answering to SUBTRACTION:

```
GtkImage now sets 'pixelSize' — delete it from KNOWN_GAPS … A declaration whose gap is closed
is a stale reason left standing, and the next reader believes it.
```

**All green:** every `scripts/check-*.mjs` that runs without arguments (three pre-existing
failures — `check-closing-keywords`, `check-vue-program`, `check-staged-prebuild-libc` — fail
identically on a stashed clean tree), `gjsify tsc` for `@gjsify/adwaita-nativescript` and
`@gjsify/storybook-nativescript`, `gjsify format --check`, `gjsify lint`. The vocabulary gate
now counts 111 set properties where it counted 110, and the false-friend row in
`status/open-todos.md`'s census drops from 3 to 2.

## Review round

An independent review of this branch found three things, and the largest was a governance one:
shipping the nick ALONE left **ADR 0034 § 4 asserting a rule this surface did not follow** — an
enum accepts the nick OR the constant, and gtk-host's own
`iconSize?: GtkIconSizeNick | Gtk.IconSize` says the same. A GJS snippet's `Gtk.IconSize.LARGE`
was a type error here and a runtime throw, which is the cross-surface divergence that ADR exists
to prevent. So the constant is IN, arriving the way `Gtk.Align`'s does — through the
construct-props bag, one `ICON_SIZE_PROPS` declaration feeding both the runtime coercion and the
`ConstructProps` widening — and the `open-todos` entry that deferred it is deleted rather than
left standing.

That also bought arm 7 its second oracle. `GTK_ICON_SIZE` derives its constants from the
POSITIONS in the nick list, which is legitimate only for an enum with no alias; `Gtk.Align` is
the counter-example where it is wrong for 2 of 7 members. The arm now holds those positions
against the **typelib-read values in `generated/enum-values.mts`** — committed, readable from a
`checkout` + `setup-node` job with no `@girs` install, and the very thing `gtk-align.ts`'s own
header names as what would retire ITS "no in-repo oracle can check this" caveat. Swapping two
nicks now fails the arm twice, once per oracle; pointing it at a file that does not exist fails
it too.

Two smaller ones: `gtkIconSizeNick`'s comment claimed `fallback` was RETURNED to preserve the
widget's setting — it throws, and the setting survives because nothing is assigned; and
`DEFAULT_GTK_IMAGE_SIZE` / `DEFAULT_ICON_BUTTON_ICON_SIZE` now say they are `pixelSize`'s
default, because after the split their names read as `iconSize`'s, which is a nick.

## What is deliberately NOT here

A plain `image.iconSize = Gtk.IconSize.LARGE` is still a type error, and that is ADR 0034 § 4's
rule rather than a gap: a setter that widened to admit the constant would drag the number into
the XML ATTRIBUTE door, which has no coercer. The constant's door is the construct-props bag,
exactly as for `Gtk.Align`.

`packages/nativescript-bridge`'s comment-budget ratio moves from 0.527 to 0.528 (already over
its 0.495 ceiling before this branch). It is a score, not a gate.

BREAKING CHANGE: `GtkImage.iconSize` and `AdwImageButton.iconSize` no longer take a number of
DIPs. Use `pixelSize` for that; `iconSize` now takes `'inherit' | 'normal' | 'large'` and throws
on anything else. `pixelSize` reads back -1 while unset, as `gtk_image_get_pixel_size` does.

Closes #1584


JumpLink:
## Independent review — breaking surface audited from scratch, one commit pushed

Second pass by a reviewer who did not write this. I grepped the surface myself rather than
taking the enumeration, and re-measured the pixel sizes on a live toolkit.

### The breaking surface really is two properties on two widgets

`grep -rn -E 'iconSize|icon-size|icon_size|pixelSize|pixel-size|pixel_size'` over `*.ts *.tsx
*.mts *.mjs *.js *.xml *.md *.mdx *.astro *.blp *.ui`, everything but `refs/`, `node_modules`,
`lib/esm`, `dist` — 222 hits, and outside `packages/nativescript-bridge/adwaita/src` not one of
them is this API:

- **No XML templates exist in the bridge at all** (`find packages/nativescript-bridge -name '*.xml'` → empty), so there is no template door to migrate.
- **No showcase, website fence, or doc fence** names `iconSize` on this surface. `docs/adr/0034` mentions `gtk-image/pixel-size` in prose; `status/open-todos.md` is this PR's own ledger.
- The other `iconSize` hits are different subjects: `ship/plan.ts`'s `iconSizeDir` (a hicolor path), `adwaita-core`/`adwaita-web`'s `AVATAR_ICON_SIZE_VECTORS` (avatar arithmetic), `framework/react-native` (already maps `size` → `Gtk.Image:pixel-size`), `framework/webgl`'s texel `pixelSize`.
- The four in-repo consumers are all inside the package and all migrated to `pixelSize`: `adw-avatar.ts:115`, `adw-combo-row.ts:74`, `adw-status-page.ts:77`, `adw-tab-view.ts:484`.

So: `@gjsify/adwaita-nativescript` (tier 2, published, 0.48.0), `GtkImage.iconSize` and
`AdwImageButton.iconSize`. Nothing else.

### The throw is loud on every door

| how a number arrives | what happens | message carries the migration |
|---|---|---|
| `img.iconSize = 24` | `gtkIconSizeNick` throws | yes — `…is not a member — it takes one of inherit, normal, large … For a size in DIPs use \`pixelSize\`` |
| `new GtkImage({ iconSize: 24 })` | `iconSizeNickOf` throws | yes — names the three members, their values, and `pixelSize` |
| `<gtk:Image iconSize="24" />` | string `'24'` is no nick → `gtkIconSizeNick` throws | yes |
| `<gtk:Image iconSize="large" />` | `'large'` — the defect, fixed | — |

`applyConstructProps` has no catch around the assignment, so the throw propagates. No silent
fallback and no `undefined` on any path I could reach.

The one quiet corner, and I do not think it needs changing: `iconSize: 0 | 1 | 2` in the
construct-props bag is read as the constant, so old code meaning 0/1/2 DIPs becomes
inherit/normal/large. Only `1` and `2` actually change the rendered size (16→16 is a no-op,
1→16, 2→32), and a 1-DIP icon is not a thing anyone shipped. This is exactly the cost ADR 0034
§ 4 accepts for `Gtk.Align` too.

### The pixel table, measured again on a live toolkit

Not taken from the PR. One `Gtk.Image` per member with `icon_name` set, presented, `measure()`d
on both axes, GTK **4.22.4** / gjs **1.88.1**:

```
inherit(value=0): min 16x16 nat 16x16 classes=[]              pixel_size=-1
normal (value=1): min 16x16 nat 16x16 classes=[normal-icons]  pixel_size=-1
large  (value=2): min 32x32 nat 32x32 classes=[large-icons]   pixel_size=-1
```

Matches the header table, the constants, the css classes, and the `-1` sentinel exactly.
`xmlNumber('large', 16) === 16` reproduces, so the false-friend claim holds too.

### Arm 7, negative-controlled three ways

| what I broke | result |
|---|---|
| swapped `normal`/`large` in `GTK_ICON_SIZE_NICKS` | **3 failures** — the nick-order arm once, the typelib-value arm twice, as the commit message predicts |
| renamed `gtk-icon-size.ts` away | **1 failure**: "is not readable — arm 7 would pass vacuously" |
| changed `large: 32` → `24` | gate exit **0** — as the arm's own comment admits; `icon-size.spec.ts` is what holds it, and it does |

`check-nativescript-widget-coverage`, `check-nativescript-theme-classes`,
`check-vocabulary-alignment` and `check-adwaita-element-properties` all exit 0. Package suite:
1383 tests.

### The ADR question: the code was the right thing to change

ADR § 4 already says, for enums in general, that the nick is the convergent spelling and the
constant is "a second accepted spelling backed by a value table whose numbers are held against
the GIR" — and `gtk-host`'s generated `iconSize?: GtkIconSizeNick | Gtk.IconSize` says the same
on the neighbouring surface. So accepting the constant is **conformance, not over-fulfilment**:
no surface was invented that the ADR did not already require, and amending the ADR to carve out
one enum would have been the bend. Where the constant is coerced (the bag, never the setter) is
forced by the XML attribute door having no coercer, which is the arrangement `gtk-align.ts`
already has. Right call.

### What I pushed

One `docs` commit. The file header still argued for the decision the PR **reversed two commits
later**: that this file "declares NO constants", that the coercion "needs its own gate arm to be
worth having", and that `status/open-todos.md` carries it as an open item. `GTK_ICON_SIZE` is
there, `iconSizeNickOf` is the coercion, arm 7 is the gate arm, and the ledger entry was deleted
in the same PR. Same drift one scale down in `iconSizeNickOf`: "Anything else is returned
unchanged, for the setter to refuse" reads as if `iconSizeNickOf(3)` came back as `3` — it
throws, and the spec asserts that it does. Both rewritten to the decision that shipped, keeping
the reasons the old text existed to carry.

### Recommendation: merge now, do not hold it back

- Landing a break immediately after a release is the *right* window — a full cycle of `main` sits between it and the next tag.
- The surface is two properties on two widgets of one package, with zero in-repo consumers left on the old spelling and no template, showcase or doc door to migrate.
- Every wrong value throws with the migration in the message, so a consumer that has not read the release notes still cannot ship a silent 16.
- Holding it costs the opposite: every day it ships is another consumer writing `iconSize={24}` against a name that means something else.

One thing this review did **not** do, because the brief forbade it: `docs/release-notes/next.md`
still needs the BREAKING entry with the `iconSize` → `pixelSize` migration line.

**Verdict: mergeable with the commit above.**

PR #1649, opened 2026-09-11
fix(react-native): name the element a refusal refused

### LINKED ISSUE #1640: render() drops ParentProvider for every child when any child is a text run, so a sibling's flex-1 throws
`packages/framework/react-native`, at `v0.48.0-48-gb67837c887`.

**Where.** `src/components.ts:441`, the `plan.content === null` arm of `render`:

```ts
const body = children.some(isTextNode) ? children : [wrap(children)];
```

`wrap` (`components.ts:431-432`) is the only thing that emits `ParentProvider`. The test is over the whole child array, so one text child suppresses the provider for every element sibling as well. `isTextNode` is `typeof === 'string' || 'number'` (`child-facts.ts:65`).

**Measured, react 19.2.8 / react-reconciler 0.33.0.** Two behaviours disagree about `''`:

- `Children.toArray(['a','',0,false,null,undefined,<el/>])` → `['a','',0,<el/>]`, so `childNodes` (`child-facts.ts:85-87`) keeps `''` and `0`.
- the reconciler builds a text fiber only for `("string" === typeof newChild && "" !== newChild) || number || bigint` (`react-reconciler.development.js:3893`).

So `''` is a child to `some(isTextNode)` and not a child to the host. No text reaches the widget, nothing is visible, and the provider is gone.

**Why that throws.** The consumer is `primitives/intents.ts:302-305`:

```ts
if (intent.expand !== undefined) {
    if (parent === undefined) remaining.expand = intent.expand;
    else props[parent.orientation === 'horizontal' ? 'hexpand' : 'vexpand'] = true;
}
```

`flex-1` is `{ expand: 'main-axis' }` (`gtk-host/src/style/layout.ts:605`), and an unresolved intent makes `usePlan` throw (`components.ts:217-227`). A child's `flex-1` therefore starts throwing when a SIBLING's value becomes `''`.

**Repro** (not run):

```tsx
const Row = ({ label }: { label: string | null }) => (
  <View className="flex-row">
    {label}
    <View className="flex-1 bg-canvas" />
  </View>
);

<Row label={null} />  // renders: children = [<View>]    -> wrapped
<Row label="" />      // throws:  children = ['', <View>] -> not wrapped
```

With `0` in place of `''` the tree is separately invalid — `View.textSink` is `null` and the host refuses text by tag name (`primitives/table.ts:809-813`, `gtk-host/src/adapters/react.ts:241`) — but the render-phase expand refusal fires before the commit-phase host refusal, so the accurate message is masked by one about a sibling.

**The `plan.content !== null` arm has no hole.** `components.ts:448` wraps unconditionally: a content node is a real host element that children go into, so the optimisation was never applied there. Solid wraps unconditionally too (`solid/index.ts:380-392`), a binding divergence of the kind `components.ts:209-216` argues is worse than a refusal. No parity vector in `solid/solid.spec.ts` mixes a bare string with an element under one parent, and no spec covers this branch.

**The branch has a real reason** (`components.ts:435-440`): a text run must not be wrapped, because a provider is not a host element and nothing under a sink-only primitive has children to publish to, so the wrapper would be overhead on the most common element in any application (233 `Text` uses against 55 `View`s). That concedes correctness and rests on cost. Two fixes keep it:

- **Per-child wrap**: `...children.map(c => isTextNode(c) ? c : wrap([c]))`. Order preserved, text unwrapped, so the 233 `Text` uses still allocate nothing. Cost: one provider per element child rather than one per parent.
- **Branch on the primitive, not the children.** `plan.textSink` (`primitives/resolve.ts:238`) already says whether text is this widget's sink; `textSink !== null && children.every(isTextNode)` is the condition the comment actually describes, at the current allocation.

Always-wrap is the third option, at the overhead the comment measured.

**How it was found.** In a consumer application a `<View className="flex-1 bg-canvas">` at the root of a re-rendered subtree threw `<View> expand — carries layout that cannot be resolved at this position`. The message (`primitives/errors.ts:24-25`) carries the primitive and the subject and no position, so it could not say which of 25 identical elements it was. That half is a separate issue.

### LINKED ISSUE #1641: PrimitiveError cannot name the element it refused, and the unresolved-intent message blames a cause that cannot apply
`packages/framework/react-native`, at `v0.48.0-48-gb67837c887`.

Two defects in one refusal: the message at `src/components.ts:219` identifies no element, and half of its explanation cannot apply at that site.

## 1. The refusal names a primitive, never an element

`PrimitiveError` carries `primitive` and `subject` only (`src/primitives/errors.ts:43-54`), and `primitiveErrorMessage` formats `<View> expand — …` (`errors.ts:24-25`). In an application with twenty-five `<View className="flex-1 bg-canvas">` sites that message identifies none of them. Measured in a consumer app: the stack is bundled host code (`usePlan`, `View$1`, `renderWithHooks`) with no component names, and finding the element required patching the built `lib/esm/components.js` to print `props.className`.

Both identifying fields are already in scope at the throw. `usePlan` binds `props` at `components.ts:185`; `className` is on the shared props (`components.ts:79`, type `ClassNameInput` at `primitives/classes.ts:34`) and `testID` at `components.ts:91`. `nativeID` is refused (`primitives/table.ts:556`), so it can never be present.

Proposal: an optional fourth argument on `primitiveErrorMessage` (`errors.ts:24`) and on the constructor (`errors.ts:50`), carrying those two fields only, appended as one clause; passed at the throw, `components.ts:219`. Optional keeps `prop-table.ts:117,168,169` byte-identical, so ADR 0039 § 1's "same string" holds.

Two named fields, not a props dump: `children` is a React tree, `style` can be a large object, and computed `className`s are common (`classes.ts:42-47`), so flatten an array the way `tokenise` does (`classes.ts:81-89`) and cap the result (~120 chars). The shape already exists in this layer — `className "…"` appears in the subject at `classes.ts:64,71` and `style.ts:170`.

## 2. "or its parent is not a box" cannot be the cause at this site

The message reads: "…and this element is the root of its tree, or its parent is not a box" (`components.ts:224`).

Verified against `primitives/intents.ts`: the only input to resolving `expand` is whether the parent record exists — `if (parent === undefined) remaining.expand = intent.expand; else props[…] = true` (`intents.ts:302-305`). Same for `alignSelf` (`307-310`) and for `overlay` (`312-314`). `widget.box` is tested only by `alignChildren`, `distribute` and `axisSpacing` (`intents.ts:227,239,252`), and `parent.overlay` only by `overlay` (`315`) — all four throw on the spot and never reach `remaining`. `remaining` is the plan's `intent` (`resolve.ts:418,468`), which is the only thing `components.ts:217` reads.

So a non-empty `plan.intent` has exactly one cause: no parent context was published. A developer chasing "is my parent a box?" for a `flex-1` refusal is chasing a clause that cannot apply — several hours, here.

Proposed wording for `components.ts:222-226`:

> carries layout that only a parent can resolve — `flex-1` and `self-*` need the parent's orientation, `absolute` needs the parent to be an overlay — and no parent context was published above this element: either it is the root of its React tree, or it was built outside its parent's provider. Wrap it in a `<View>`, or move the utility to a child.

The "not a box" clause is correct where it already stands and should be kept there: `items-*`, `justify-*` and `gap-x-*/gap-y-*` (`intents.ts:231,242,256`), each of which says it in its own message. "The parent has to be a `Gtk.Overlay`" is likewise correct at `intents.ts:316-320`. No test asserts the `components.ts` string, so the rewording is free.

The second cause named in the proposed wording — "built outside its parent's provider" — is #1640.

## Noted, not part of this issue

The same function memoises the published context on a dependency list containing `plan.childContext` (`components.ts:273-274`), which `resolvePrimitive` rebuilds by spread on every call (`resolve.ts:417,453-467`). The memo therefore always misses and the adjacent `contextKey` string is inert. Flagged so it is not taken as unnoticed.

JumpLink:
Closes #1640. Closes #1641.

Both reported from a consumer application — an experimental GTK4 host for a React Native
app — that spent an evening on a `<View> expand` refusal. One PR rather than two, per
AGENTS.md § PR size, and they are the same evening's findings in the same file.

## The provider, #1640

`render()`'s `plan.content === null` arm dropped `ParentProvider` for every child when
ANY child was a text run:

```ts
const body = children.some(isTextNode) ? children : [wrap(children)];
```

`some` asks whether there is text here, which is a question about the SIBLINGS. One text
child therefore deleted the provider for every element child beside it, and a child's
`flex-1` began refusing because a sibling's value had become a string.

The case is worse than it sounds, because the text need not be visible. `Children.toArray`
keeps `''`, and the reconciler builds a text fiber only for a non-empty string
(`react-reconciler.development.js:3893`) — so `{label}` going from `null` to `''` renders
nothing, changes nothing on screen, and removes the provider.

The condition is the primitive's now:

```ts
const body = plan.textSink !== null && children.every(isTextNode) ? children : [wrap(children)];
```

which is what the paragraph above it already described. A sink-only primitive is the one
with nothing to publish to, and its children are all text — so the 233 `Text` uses ADR 0032
measured still allocate nothing, which was the branch's whole reason. A mixed
`<Text>a<Text>b</Text></Text>` gains one provider, and is correct rather than merely cheap.

Solid's arm wraps unconditionally (`solid/index.ts`), so this also removes a binding
divergence of the kind `components.ts` argues is worse than a refusal.

## Naming the element, #1641

`PrimitiveError` carried the primitive and the subject and no position, and the stack is
this package's own frames in a bundle with no component names. With twenty-five
`<View className="flex-1 bg-canvas">` sites in one application, the message identified
none of them; the element was found by patching the built `lib/` to print `props.className`.

`describeElement` appends `[className="…" testID="…"]`. Two fields and not a props dump:
`children` is a React tree, `style` can be large, and `nativeID` is refused by name
(`primitives/table.ts`) so it can never be present. An array is flattened the way
`splitVariants` flattens one, and the result is capped, so a computed class list cannot
make a refusal scroll.

The fourth argument is OPTIONAL. `prop-table`'s static answers call
`primitiveErrorMessage` with three and produce exactly the string they always did, which
is what keeps ADR 0039 § 1's "same string" true.

## The message's own explanation

"…or its parent is not a box" cannot be the cause at that site. In `primitives/intents.ts`
the only input to resolving `expand`, `alignSelf` and `overlay` is whether the parent
record EXISTS; `widget.box` is tested only by `alignChildren`, `distribute` and
`axisSpacing`, and `parent.overlay` only by `overlay` — all four throw on the spot and
never reach `remaining`, which is what the throw reads. So the clause is kept where it is
true, in those four messages, and the unresolved-intent message says what can actually be
wrong: no parent context was published, either because the element is a React root or
because it was built outside its parent's provider. That second half is #1640.

## Checked

- `gjsify tsc --noEmit` clean for `@gjsify/react-native`.
- `gjsify run build:gjsify` clean.
- `primitives/errors.spec.ts` added, registered in `test.mts`, and all eighteen of its
  expectations exercised against the built `lib/` directly, because `build:test` here
  needs more of the monorepo built than this checkout has. The suite itself has not been
  run — CI is the first full run, and that is worth saying rather than implying.
- Not measured: whether the extra provider on a mixed `Text` run costs anything. It is one
  context per nested run and the 233 all-text uses are unchanged, but if the parity vector
  or a benchmark disagrees, the per-child variant in #1640 is the other option.

### DIFF
--- a/packages/framework/react-native/src/components.ts
+++ b/packages/framework/react-native/src/components.ts
@@ -471,8 +471,7 @@ function render(rendered: Rendered): ReactElement {
         // than `some` keeps a mixed `<Text>a<Text>b</Text></Text>` wrapped, which is
         // one provider on a nested run and correct rather than merely cheap. The 233
         // `Text` uses still allocate nothing, because their children are all text.
-        const body =
-            plan.textSink !== null && children.every(isTextNode) ? children : [wrap(children)];
+        const body = plan.textSink !== null && children.every(isTextNode) ? children : [wrap(children)];
         return createElement(plan.node.tag, nodeProps(plan.node, inherited, extra), ...body);
     }
 
--- a/packages/framework/react-native/src/components.ts
+++ b/packages/framework/react-native/src/components.ts
@@ -54,7 +54,7 @@ import { childFacts, childNodes, isAbsoluteChild, isTextNode } from './child-fac
 import { ParentContext, ParentProvider } from './parent-context.js';
 import { onGesture, onPressStateChange } from './press.js';
 import type { ClassNameInput } from './primitives/classes.js';
-import { PrimitiveError } from './primitives/errors.js';
+import { describeElement, PrimitiveError } from './primitives/errors.js';
 import { createHandle, type TextInputHandle } from './primitives/handles.js';
 import {
     resolvePrimitive,
@@ -214,15 +214,29 @@ export function usePlan(primitive: string, authored: object): Rendered {
     // same authored tree. A feature whose two frameworks differ is worse than a
     // refusal, so this stays a refusal until that difference is understood. The
     // workaround is the container's own expand, which the application owns anyway.
+    //
+    // THE MESSAGE NAMES THE ELEMENT, because naming the primitive is not enough: an
+    // application has many elements per primitive, and a consumer reported spending
+    // hours on a `<View> expand` refusal with twenty-five `<View className="flex-1
+    // bg-canvas">` sites to choose between. `describeElement` says what the author
+    // wrote; `primitives/errors.ts` says why those two fields and no more.
+    //
+    // AND IT NO LONGER OFFERS "or its parent is not a box", WHICH CANNOT BE THE CAUSE
+    // HERE. The only input to resolving `expand`, `alignSelf` and `overlay` is whether
+    // a parent record EXISTS (`primitives/intents.ts`); the four subjects that test
+    // `widget.box` or `parent.overlay` throw on the spot and never reach `remaining`,
+    // so they never arrive here. A reader chasing that clause for a `flex-1` refusal
+    // is chasing something that cannot apply, which is what the same consumer did.
     const unresolved = Object.keys(plan.intent);
     if (unresolved.length > 0) {
         throw new PrimitiveError(
             primitive,
             unresolved.join(', '),
-            'carries layout that cannot be resolved at this position. These need a parent to resolve against — ' +
-                '`flex-1` and `self-*` need the parent orientation, `absolute` needs the parent to be an overlay — ' +
-                'and this element is the root of its tree, or its parent is not a box. Wrap it in a <View>, or move ' +
-                'the utility to a child.',
+            'carries layout that only a parent can resolve — `flex-1` and `self-*` need the parent orientation, ' +
+                '`absolute` needs the parent to be an overlay — and no parent context was published above this ' +
+                'element: either it is the root of its React tree, or it was built outside its parent’s provider. ' +
+                'Wrap it in a <View>, or move the utility to a child.',
+            describeElement(props),
         );
     }
 
@@ -438,7 +452,27 @@ function render(rendered: Rendered): ReactElement {
         // to, so the wrapper would be pure overhead on the most common element in
         // any application (233 `Text` uses against 55 `View`s, ADR 0032's
         // measurement).
-        const body = children.some(isTextNode) ? children : [wrap(children)];
+        //
+        // THE CONDITION IS THE PRIMITIVE'S, NOT THE CHILDREN'S, and it used to be
+        // `children.some(isTextNode)` — which asks "is there any text here" and
+        // answers a question about the SIBLINGS. One text child therefore deleted the
+        // provider for every element child beside it, and a child's `flex-1` started
+        // refusing because a sibling's value had become a string.
+        //
+        // MEASURED, and the case is worse than it sounds because the text need not be
+        // visible: `Children.toArray` keeps `''` (`child-facts.ts`) and the reconciler
+        // builds a text fiber only for a NON-empty string, so `{label}` going from
+        // `null` to `''` renders nothing, changes nothing on screen, and removes the
+        // provider. A consumer hit exactly that and spent hours on a `<View> expand`
+        // refusal about an element they had not touched.
+        //
+        // `plan.textSink !== null` is what the paragraph above actually describes: a
+        // sink-only primitive is the one with nothing to publish to. `every` rather
+        // than `some` keeps a mixed `<Text>a<Text>b</Text></Text>` wrapped, which is
+        // one provider on a nested run and correct rather than merely cheap. The 233
+        // `Text` uses still allocate nothing, because their children are all text.
+        const body =
+            plan.textSink !== null && children.every(isTextNode) ? children : [wrap(children)];
         return createElement(plan.node.tag, nodeProps(plan.node, inherited, extra), ...body);
     }
 
--- a/packages/framework/react-native/src/primitives/errors.spec.ts
+++ b/packages/framework/react-native/src/primitives/errors.spec.ts
@@ -24,9 +24,7 @@ import { describeElement, PrimitiveError, primitiveErrorMessage } from './errors
 export default async () => {
     await describe('describeElement', async () => {
         await it('names a plain className', async () => {
-            expect(describeElement({ className: 'flex-1 bg-canvas' })).toBe(
-                '[className="flex-1 bg-canvas"]',
-            );
+            expect(describeElement({ className: 'flex-1 bg-canvas' })).toBe('[className="flex-1 bg-canvas"]');
         });
 
         await it('flattens an array, which is how half of them are written', async () => {
@@ -36,9 +34,7 @@ export default async () => {
         });
 
         await it('collapses the whitespace an array join leaves behind', async () => {
-            expect(describeElement({ className: '  flex-1\n  bg-canvas ' })).toBe(
-                '[className="flex-1 bg-canvas"]',
-            );
+            expect(describeElement({ className: '  flex-1\n  bg-canvas ' })).toBe('[className="flex-1 bg-canvas"]');
         });
 
         await it('says nothing rather than nothing-in-quotes', async () => {
@@ -51,9 +47,7 @@ export default async () => {
         });
 
         await it('carries testID beside it, and only when it is a non-empty string', async () => {
-            expect(describeElement({ className: 'p-2', testID: 'save' })).toBe(
-                '[className="p-2" testID="save"]',
-            );
+            expect(describeElement({ className: 'p-2', testID: 'save' })).toBe('[className="p-2" testID="save"]');
             expect(describeElement({ testID: 'save' })).toBe('[testID="save"]');
             expect(describeElement({ testID: '' })).toBe('');
             expect(describeElement({ testID: 7 })).toBe('');
@@ -78,9 +72,9 @@ export default async () => {
         });
 
         await it('appends the element clause when there is one', async () => {
-            expect(
-                primitiveErrorMessage('View', 'expand', 'needs a parent', '[className="flex-1"]'),
-            ).toBe('@gjsify/react-native: <View> expand — needs a parent [className="flex-1"]');
+            expect(primitiveErrorMessage('View', 'expand', 'needs a parent', '[className="flex-1"]')).toBe(
+                '@gjsify/react-native: <View> expand — needs a parent [className="flex-1"]',
+            );
         });
     });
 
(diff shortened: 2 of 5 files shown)


JumpLink (commented):
Independent review, from a separate worktree at `fbdfcfc32a`. Everything below is
measured against this branch's commit `f45a1a9895` and against `origin/main`, with an
installed `node_modules` and a real GTK/gjs 1.88.1.

**Verdict: mergeable after two changes** — the format check (A) and a test for #1640
(B). The rest is optional.

## Confirmed, one by one

**The `react-reconciler.development.js:3893` citation is exact.** react-reconciler
0.33.0's dev CJS build, line 3893, is `("string" === typeof newChild && "" !== newChild) ||`
inside `createChild`, and the same guard sits at 3986, 4072 and 4636 (the single-child
path) and in `react-reconciler.production.js`. The claim it supports is true for every
reconciliation path, not just the array one.

**`Children.toArray` keeps `''`.** Measured: `Children.toArray(['', 'a', null, undefined, false, 0])`
→ `["", "a", 0]`. So `isTextNode('')` is `true` and `some(isTextNode)` fires on a child
that renders nothing.

**The defect reproduces, and the fix removes it.** A probe using the real reconciler
(`createRoot` into a `Gtk.Box`, same shape as `widgets.spec.ts`' `mounted`):

| tree | main | this branch |
|---|---|---|
| `<View className="flex-row">{''}<View className="flex-1"/></View>` | REFUSED `<View> expand — carries layout…` | renders |
| `<View className="flex-row"><View className="flex-1"/></View>` (discriminator) | renders | renders |
| `<View className="flex-row">hi<View className="flex-1"/></View>` | REFUSED `<View> expand` | REFUSED `<GtkBox> has no text sink…` |
| `<Text>a b</Text>` | renders | renders |
| `<View>bare</View>` | REFUSED (no text sink) | REFUSED (no text sink) |
| `<Text>a<Text>b</Text></Text>` | REFUSED (`GtkLabel` adopts nothing) | REFUSED (same) |

Row 3 is worth naming as a second improvement the PR does not claim: a visible text
sibling used to blame the SIBLING's `flex-1`; it now names the real defect.

**The "same string" promise holds, byte for byte.** I generated every string
`@gjsify/react-native/prop-table` can produce — `explainProp`, `explainPropValue`,
`propAnswer`, `propRefusedValues`, `acceptsProp*`, plus the unknown-primitive throw —
for every primitive × both `TextInput` variants × every prop × a set of probe values:
5444 rows, from the built `lib/` on each side. `diff` is empty. ADR 0039 § 1 is not
weakened by the optional fourth argument.

**Solid's arm wraps unconditionally** — `childrenUnder(props, published)` at
`solid/index.ts:252`, no condition.

**"…or its parent is not a box" cannot apply there.** `remaining` is filled only when
`parent === undefined` (`intents.ts:303`, `:308`, `:313`); the three `!widget.box` tests
and the one `!parent.overlay` test throw in place. The rewording is right.

**The suite is green.** The PR says "the suite itself has not been run". I ran it:
`gjsify workspace @gjsify/react-native run test` → **648 tests, 2781 assertions, exit 0**
under gjs 1.88.1, 1 expected `it.failing`. All ten new expectations execute.

**No Markdown tables in the PR body** (longest line 96 chars), so the squash is safe.

## A — the format check will be red

`oxfmt --check packages/framework/react-native/src` is clean on `origin/main` and flags
all three touched source files here. The diffs are purely cosmetic — the new code is
wrapped at 100 columns and `.oxfmtrc.json` sets `printWidth: 120`. `gjsify format` fixes
it in one pass (9 insertions, 23 deletions).

## B — #1640 has no test, and it is the defect that cost the evening

`errors.spec.ts` covers `describeElement`, `primitiveErrorMessage` and `PrimitiveError`
— all of #1641. Nothing covers the provider condition. `plan.textSink !== null &&
children.every(isTextNode)` can be reverted to `children.some(isTextNode)` tomorrow and
the whole suite stays green.

The harness already exists: `primitives/widgets.spec.ts` has `mounted` + `createRoot`,
and rows 1 and 2 of the table above are exactly the vector — the empty-string sibling
plus the same tree without it as the discriminator, so a green result cannot come from
the tree simply not refusing anything. I ran both against `origin/main` and this branch
to produce that table, so the test is known to go red on the old condition and green on
the new one.

## C, D, E, F — smaller

**C. The cap is on `className`, not on the message.** `describeElement({ testID: 'x'.repeat(400) })`
returns 411 characters. The PR body says "the result is capped"; the code caps one field.
One line.

**D. No quote escaping.** `describeElement({ className: 'a"b' })` → `[className="a"b"]`.
Unlikely in a class list, ambiguous when it happens.

**E. Two wording points.** "removes a binding divergence" overstates it — the arms still
differ structurally (React skips the provider for a sink-only all-text element, Solid
always wraps); they are only observably equal. And the illustration "a mixed
`<Text>a<Text>b</Text></Text>` gains one provider, and is correct rather than merely
cheap" describes a tree the host refuses on BOTH sides (`<GtkLabel> declares children:
{ kind: 'none' }`), so the added provider has no reachable case behind it. Neither is
wrong; both read as claims with a measurement behind them and have none.

**F. #1641's "Noted, not part of this issue" dies with the issue.** It flagged
`useMemo(() => plan.childContext, [contextKey, plan.childContext])` at `components.ts:273`
— `resolvePrimitive` rebuilds `childContext` by spread on every call, so the dep array
always changes, the memo always misses and `contextKey` is inert. Verified on main.
Nothing in `status/open-todos.md` tracks it, and closing #1641 is the last moment it is
written down anywhere. It is also adjacent to this PR, which ADDS providers: every
consumer below a provider re-renders on every parent render. Worth a `### ` section in
`status/open-todos.md` here.

## Noted, not a change request

`describeElement` is wired at one of 96 `new PrimitiveError(` sites — which is exactly
what #1641 asked for, so it is in scope, not a gap. But the class lives one layer down:
`resolvePrimitive` has `props` in hand for every L2 refusal and serves both bindings, so
the element clause could be supplied once there instead of per throw site. A follow-up,
not this PR.

Also cosmetic: the `errorsSuite` import in `test.mts` sits between `classes` and
`defaults`, breaking the file's alphabetical order. No lint rule enforces it.

---

Method note: I did not push anything to this branch. `correctiv/name-the-refused-element`
is checked out in another worktree on this machine, so committing to it from here would
have been the two-sessions-one-branch collision. Everything above is reproducible from
`origin/main` and `f45a1a9895`.

PR #1652, opened 2026-09-11
fix(vite-plugin-gettext): a catalog lands where gettext looks

JumpLink:
Two independent defects in `@gjsify/vite-plugin-gettext`, both in how a catalog's name
reaches an output path. Triggered by JumpLink/Learn6502#182 (locale names) and
JumpLink/Learn6502#179 (LINGUAS + the JSON names).

The measured lookup tables are in a comment below, so this body stays within the
line length commitlint allows for the squash commit.

## 1. A catalog landed where gettext never looks

Catalogs maintained on Weblate are named in BCP-47, which names simplified Chinese
`zh_Hans`. The compiler used the `.po` basename verbatim as the locale directory
(`src/gettext.ts:51`, language list from `src/utils.ts:43`; `src/msgfmt.ts:167` for the
msgfmt path), so the catalog landed in `locale/zh_Hans/LC_MESSAGES/`.

`zh_Hans` is not a POSIX locale name. `locale -a` has no such entry and glibc never
probes it.

Before, against the real Learn6502 `.mo`, with the glibc `gettext` CLI as the oracle:

```
zh_Hans    -> 黄色       (a selector no real user has)
zh_CN      -> Yellow     <-- MISSING: this is what a real user gets
zh_TW      -> Yellow     <-- MISSING
de         -> Gelb       (control: ordinary names were always fine)
```

A complete Chinese catalog shipped and every Chinese user saw English. The build exited
0, and 14 of the 15 catalogs were fine — which is why nobody looked.

After, same catalogs, same oracle:

```
zh_Hans    -> 黄色       (still works; the mapping is additive)
zh_CN      -> 黄色       FIXED
zh_SG      -> 黄色       FIXED
zh_TW      -> Yellow     correctly NOT claimed (see below)
de         -> Gelb       (control)
```

Three rules came out of the measurements, each counter-intuitive in one direction:

- a bare `<lang>` directory is reached by `<lang>_<TERRITORY>` users, so a script that is
  simply the language's normal one needs no territory;
- `<lang>@<modifier>` is reached by `<lang>_<TERRITORY>@<modifier>` users, while the
  narrower `<lang>_<TERRITORY>@<modifier>` is reached only by itself — here the tempting
  spelling is the wrong one;
- the broad `zh` is reached by `zh_TW` too, so installing a simplified catalog as `zh`
  would serve traditional users simplified text. Chinese territories are therefore
  spelled out and `zh` is never written — which is why `zh_TW -> Yellow` above is
  correct, not a gap.

The original name is kept alongside the mapped ones: additive, so no deployment that
already resolved a catalog stops resolving it, and the compiled tree stays greppable
against `po/`. It is never the only entry, which was the defect. A POSIX target absent
from the building host is still written (`zh_MO` is absent on Fedora 44, present
elsewhere) — the catalog is installed on machines whose locale set the build cannot see.

## 2. The same basename was also an Android resource qualifier

`po2json` wrote the basename as the JSON filename, and `@nativescript/localize` turns
that filename straight into a resource directory (`hooks/converter.js`, via
`` values-${language.replace(/^(.+?)-(.+?)$/, '$1-r$2')} ``). Running that transformation
over the names in play:

```
de       -> values-de        ok
pt-BR    -> values-pt-rBR    ok      <- the name the repo tracks
pt_BR    -> values-pt_BR     BROKEN  <- what po2json wrote
zh       -> values-zh        ok      <- the name the repo tracks
zh_Hans  -> values-zh_Hans   BROKEN  <- what po2json wrote
zh-Hans  -> values-zh-rHans  BROKEN  (a script subtag needs values-b+zh+Hans)
```

An underscore is not legal in a resource qualifier. Both generations sat in the tree at
once: the tracked, correct `pt-BR.json`/`zh.json` and the generated, dead
`pt_BR.json`/`zh_Hans.json`.

So the three targets have three namespaces and must not share a string. The mapping is
data in one table (`src/catalog-names.ts`) read by both plugins, not logic in each — two
copies of this reasoning drift, and the drifted copy wins silently. Evidence that this
restores rather than invents: after the fix the generator produces, against the real 16
catalogs, a filename set byte-identical to the one the repo already tracks.

Not measured: that Android then loads the resulting directory. The qualifier rule is
derived from the hook's source and the transformation above was executed against it, but
no emulator or device ran. The POSIX side is measured end to end against glibc.

## 3. LINGUAS was rewritten on every build

`gjsify workspace @learn6502/translations build` changed `LINGUAS` on an untouched tree
every run — `de`, `es`, `uk` moved to the end and the trailing newline vanished. Causes,
both in `src/utils.ts`: the list came straight from `fs.readdir` (order unspecified; on
ext4 a small directory reads back in creation order, so a language removed and re-added
moves to the end), and `join('\n')` left the last line without a newline.

Sorted at the source in `findAvailableLanguages`, so every consumer is deterministic at
once. No comparator on purpose — the default is UTF-16 code-unit order, identical on
every host, whereas `localeCompare` would order by the builder's locale and make
committed output depend on who built it.

## The mechanism, not just the instances

A silent fix would let the next catalog land wrong the same way. So:

- a catalog whose script subtag maps to no known name FAILS the build, naming every such
  catalog at once rather than aborting on the first. Same asymmetry the existing
  source-pattern guard is built on: a false failure costs one build and one config line,
  a missed one ships a whole language that renders in English. Escape hatch is
  `localeNames`, per catalog, so mapping one language does not disarm the rest.
- two catalogs claiming one output name fail too (`zh_Hans.po` beside a hand-made
  `zh_CN.po`), instead of one overwriting the other at exit 0.
- the determinism test runs the generator twice with the same set in different orders and
  demands byte equality — closing the class rather than the two instances, and doing it
  on every host instead of hoping a CI filesystem hands back a scrambled directory.

## Red to green

The new tests were run against the unmodified implementation (impl files reverted to
`origin/main`, tests kept):

```
10 of 55 tests failed - 88 assertions
  POSIX locale directories > compiles a Weblate script name into the locales glibc probes
  POSIX locale directories > folds a hyphenated catalog name to the POSIX spelling
  POSIX locale directories > refuses a catalog it cannot place instead of compiling it
  glibc as the oracle > a user with a real Chinese locale gets the Chinese catalog
  generateLinguasFile > writes the same bytes whatever order the languages arrive in
  generateLinguasFile > ends with a newline
  generateLinguasFile > sorts by code unit, not by the builder locale
  generateLinguasFile > is stable across two full plugin runs
  po2jsonPlugin > names the JSON so the resource qualifier comes out legal
  po2jsonPlugin > keeps the POSIX and Android names apart for one catalog
```

With the fix: `55 tests passed - 96 assertions`.

The oracle test asks the glibc `gettext` CLI, not our own implementation — asking the
implementation whether its own directory choice was right cannot see it be wrong, which
is how this survived.

The measurement trap, paid for once and now written into the spec header: `LC_ALL` must
name an installed, non-C locale. Under `C`, `C.UTF-8`, `POSIX` or an uninstalled locale,
glibc falls back to C and ignores `LANGUAGE` completely — every probe returns the
untranslated msgid and the suite measures nothing while looking green:

```
LC_ALL=en_US.UTF-8 -> HIT:zh_CN
LC_ALL=C.UTF-8     -> Yellow
LC_ALL=C           -> Yellow
(unset)            -> Yellow
```

The suite therefore discovers a usable locale at runtime, carries a `de` discriminator
that proves the probe can see translations at all, and falls back to
`it.failing(..., { when: !locale })` on a host that has none — so it runs as an ordinary
test wherever it can and self-retires rather than passing vacuously.

## Validation

`gjsify tsc --noEmit`, `gjsify run build`, `gjsify lint` (0 errors, 0 findings in this
package), `gjsify format`, full package suite green.

### DIFF
--- a/packages/infra/vite-plugin-gettext/README.md
+++ b/packages/infra/vite-plugin-gettext/README.md
@@ -34,6 +34,44 @@ Requires `gettext` tools (`msgfmt`, `xgettext`) to be installed on the system.
   fuzzy entries out of the `.mo`. A run that really does replace that many strings raises the
   option.
 
+## One catalog, three namespaces
+
+A catalog is called one thing in the repository and has to be spelled three different ways
+downstream. They disagree, so the `.po` basename is not passed through unchanged:
+
+| Target | Namespace | `zh_Hans.po` | `pt_BR.po` |
+|---|---|---|---|
+| `.mo` directory (`gettextPlugin`, `msgfmtPlugin`) | POSIX locale | `locale/zh_CN/`, `locale/zh_SG/` | `locale/pt_BR/` |
+| JSON file (`po2jsonPlugin`) | Android qualifier | `zh.json` | `pt-BR.json` |
+| the `.po` in the repo | BCP-47-ish (Weblate) | `zh_Hans` | `pt_BR` |
+
+Weblate names simplified Chinese `zh_Hans`, which is BCP-47 and **not** a POSIX locale — glibc
+never probes it, so a catalog compiled under that name ships complete and renders in English for
+every Chinese user while the build exits 0. The compiled catalog is therefore installed under the
+locale directories real users actually resolve, keeping the original name alongside them so
+nothing that already worked stops.
+
+On the JSON side the filename becomes an Android resource qualifier
+(`values-<lang>[-r<REGION>]`), where an underscore is illegal: `pt_BR.json` would yield
+`values-pt_BR`, a directory Android never consults.
+
+**A catalog whose script subtag maps to no known name fails the build**, naming every such
+catalog at once. A build that cannot say where a catalog goes must not quietly put it somewhere —
+"somewhere" is indistinguishable from "correct" until a user of that language complains. Map it
+explicitly with `localeNames`, one catalog at a time, so mapping one language does not disarm the
+check for the rest:
+
+```typescript
+gettextPlugin({
+    poDirectory: 'po',
+    moDirectory: 'dist',
+    localeNames: { az_Arab: ['az_IR'] },
+});
+```
+
+Two catalogs that would land on one output name (a Weblate `zh_Hans.po` beside a hand-made
+`zh_CN.po`) fail the same way, rather than letting one overwrite the other.
+
 ```typescript
 xgettextPlugin({
     sources: ['src/**/*.blp', 'data/**/*.desktop.in', '!src/**/*.generated.blp'],
--- a/packages/infra/vite-plugin-gettext/src/catalog-names.spec.ts
+++ b/packages/infra/vite-plugin-gettext/src/catalog-names.spec.ts
@@ -1,0 +1,159 @@
+// The name mapping on its own — no gettext, no filesystem. `gettext.spec.ts`
+// proves the plugins WIRE it and that glibc agrees; these cases pin down what it
+// answers, including the shapes that occur once and are therefore the ones a
+// later "simplification" quietly changes.
+
+import { describe, expect, it } from '@gjsify/unit';
+import {
+    catalogNames,
+    CollidingCatalogNameError,
+    planCatalogNames,
+    posixLocaleDirectories,
+    UnmappableCatalogNameError,
+} from './catalog-names.js';
+
+const posix = { pluginName: 'vite-plugin-gettext', namespace: 'posix' as const };
+const bcp47 = { pluginName: 'vite-plugin-gettext-po2json', namespace: 'bcp47' as const };
+
+/** What `fn` threw, so a case can assert the TYPE and the message it carries. */
+function thrownBy(fn: () => void): unknown {
+    try {
+        fn();
+    } catch (error) {
+        return error;
+    }
+    return undefined;
+}
+
+export default async () => {
+    await describe('posixLocaleDirectories', async () => {
+        await it('maps a Weblate script subtag to the locales glibc probes', async () => {
+            // The bug: `zh_Hans` alone is a directory nothing looks up. Measured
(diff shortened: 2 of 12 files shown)


JumpLink:
### The measured lookup tables

Moved out of the PR body: a markdown table cannot be wrapped, and the body becomes the
squash commit, where commitlint caps lines at 100.

**How the POSIX column was derived.** Measured on glibc 2.42 / gettext 0.26 by installing a catalog under one target *alone* and asking the `gettext` CLI which selectors reached it:

| installed as | reached by |
|---|---|
| `zh_Hans` | `zh_Hans` only — i.e. nobody |
| `zh_CN` | `zh_CN` only (**not** `zh_SG`) |
| `zh` | `zh_Hans zh_Hant zh_CN zh_TW zh_HK zh_SG zh` |
| `sr_Latn` | **nobody** |
| `sr@latin` | `sr_RS@latin`, `sr@latin` |
| `sr_RS@latin` | itself only |
| `sr` | `sr_RS@latin`, `sr_RS`, `sr` |
| `pt-BR` | **nobody** |
| `pt_BR` | `pt_BR` |

**The three namespaces.**

| target | namespace | `zh_Hans` | `pt_BR` |
|---|---|---|---|
| `.mo` directory | POSIX locale | `zh_CN`, `zh_SG` | `pt_BR` |
| Android resources (JSON) | Android qualifier | `zh` | `pt-BR` |
| Weblate / the `.po` | BCP-47-ish | `zh_Hans` | `pt_BR` |

Scripts with a POSIX home but no safe Android qualifier (`sr_Latn` would need `values-b+sr+Latn`, which the hook cannot emit) carry no `bcp47` entry and are reported rather than given a name that merely looks right.

PR #1653, opened 2026-09-11
fix(cli): a reported failure carries its status

JumpLink:
Reported against the consumer as JumpLink/Learn6502#180.

## What was measured

A gjsify command handler fails one of two ways: it **throws**, or it assigns
**`process.exitCode`** and returns normally. `index.ts` funnelled only the first into a
real exit. Under GJS — which has no atexit hook, so nothing reads the property back at
natural shutdown — the second ended the process at **0**, with the error already printed
on stderr.

```
$ gjs -m dist/cli.gjs.mjs gresource broken.gresource.xml --sourcedir . --target out.gresource
broken.gresource.xml: "does-not-exist.ui" was not found in any source directory
[gjsify gresource] glib-compile-resources failed (exit 1)
$ echo $?
0
```

Same argv under Node exits 1. The same split held for `gjsify gsettings` and
`gjsify barrels --check`. After the fix all three exit 1 on both hosts, and
`gjsify --version` / `gjsify info` still exit 0. The measured before/after table is in a
comment below.

### Why `&&` was no defence

`gjsify run`'s in-process fast path was already immune — it reads `process.exitCode`
back itself (#1568). So the hole survived precisely in the SPAWNED command: every
chained script, and every direct CI invocation. With a script body
`gjsify gresource ... && echo SECOND`, the chained form printed `SECOND` and exited 0,
over a `.gresource` that was never written, while the single-command form exited 1.
Chaining leaves was the documented workaround for this class and was protecting nothing.

## Root cause

`packages/infra/cli/src/index.ts:40-45` — the entry forces a non-zero GJS exit
(`gjsExit(1)`) only inside its `catch`. The success path leaves the status to natural
shutdown, which is Node semantics only.

The hazard itself is already documented in-tree and was simply never applied to the
entry:

- `packages/infra/cli/src/utils/force-exit.ts:3-9` — "Node exits at natural shutdown
  once `process.exitCode` is set. GJS has no atexit hook…"
- `packages/infra/cli/src/utils/oxc-resolve.ts:418-427` — "`process.exitCode` alone is
  NOT honored at natural shutdown."

The call sites that report this way are all correct and deliberate — `gresource.ts:88`,
`gsettings.ts:80`, `gettext.ts:259,277,304`, `barrels.ts:105,127`, `setOxcExitCode`,
`run-node.ts:171`. They propagate the code the underlying tool exited with, which is why
they do not throw.

## The change

`packages/infra/cli/src/utils/cli-exit.ts` is the missing half of the entry's funnel and
a sibling of `force-exit.ts`: that one is for a caller that knows the code it wants NOW,
this one for the status a settled run left behind. `index.ts` calls
`exitOnReportedFailure()` after the try/catch. One exit, every command.

- Node is unchanged: `gjsExit` is a no-op there, and this deliberately does NOT reach
  for `process.exit()`, which would cut short the stdout flush Node's idiom exists to
  allow.
- Daemon commands are exempt via `isDaemonCommand()`. A watch loop resolves as soon as
  it is armed and keeps running; exiting on that resolution is the #1237 incident
  (`gjsify run dev` killing the loop it had just started) one layer up.
  `tests/e2e/dev-command` stays green, 8/8.
- No status is invented: a value that is not a status yields no exit. Guessing would
  make the two runtimes disagree, which is the defect being closed.

## The mechanism

`tests/e2e/cli-exit-status/` is written as a DIFFERENTIAL, not as a list of expected
codes: it runs the same argv on both hosts and asserts they agree and are non-zero.

That is deliberate. Every `process.exitCode = …` call site was correct, so no lint rule
could have caught this — the entry was the thing not reading them. Only running both
hosts can see it. A future command that invents a third way to report a failure shows up
as a disagreement between the hosts, whichever host it is wrong on. It is the
runtime-side complement to `gjsify/deferred-process-exit` (oxlint), which covers the
other GJS exit trap: a bare `process.exit()` that does not halt.

## Red → green

Against the pre-fix build artifacts all four tests fail, including:

```
✖ a chained script stops at the failing leaf instead of running the next step
  AssertionError: the chain continued past a failed `gjsify gresource`
  — `&&` saw a 0 from the child
  actual: 'SECOND\n…'
```

After rebuilding `lib/` + `dist/cli.gjs.mjs` with the fix: `pass 4 · fail 0`.

## Validation

- `gjsify tsc --noEmit` (@gjsify/cli) — 0
- `node dist/test.node.mjs` — 1725 tests passed, 4126 assertions, including the new
  `pendingFailureCode` suite (5 tests)
- `gjsify format --check` — clean, 3837 files
- `gjsify lint` — 0, no warning in any changed file
- e2e regression: `workspace-unknown-name` 4/4, `dev-command` 8/8, `foreach-command`
  22/22, `homepage-run-variations` 32/32
- pre-commit hook: `affected.gjs.mjs` unchanged by the rebuild

No AGENTS.md change: `packages/infra/cli/AGENTS.md` is at 26111 B against a 20 KB
target, and the budget ledger's ceiling is exact so the file may only shrink. The
precedent for exit-discipline modules is in-file documentation (`force-exit.ts`,
`cli-fail.ts`), which is where the incident and the measurements live here too.

## Upstream context

Learn6502#180's original premise — that the in-process dispatch never sets
`process.exitCode` — was fixed in 0.23.0 (#1568) and re-measured as still correct here.
What remained is a residual hole that repo's AGENTS.md records as "a dispatched command
that neither throws nor sets `process.exitCode`". Measured here it is both wider and
differently shaped: the command DOES set `process.exitCode`, and the hole is every GJS
invocation of it EXCEPT the in-process dispatch — which is exactly why chaining, the
documented workaround, did not close it.


JumpLink:
The measurement tables, kept out of the PR body so they never reach the squash commit.

Same argv, same broken input, v0.49.0 bundle vs. this branch:

| command | `node lib/index.js` | `gjs -m dist/cli.gjs.mjs` before | after |
|---|---|---|---|
| `gresource broken.gresource.xml` | 1 | **0** | 1 |
| `gsettings <broken schema dir>` | 1 | **0** | 1 |
| `barrels --check` (no paths) | 1 | **0** | 1 |
| `--version` / `info` (control) | 0 | 0 | 0 |

The script shapes, all run under GJS, with `"build:resources": "gjsify gresource …"`:

| script body | before | after |
|---|---|---|
| single command (in-process dispatch) | 1 | 1 |
| `gjsify gresource … && echo SECOND` (spawn path) | **0, `SECOND` printed** | 1, not printed |

Full red run against the pre-fix artifacts:

```
✖ gresource: a failing glib-compile-resources exits non-zero on both hosts
✖ gsettings: a failing glib-compile-schemas exits non-zero on both hosts
✖ barrels: a usage refusal that sets exitCode exits non-zero on both hosts
✖ a chained script stops at the failing leaf instead of running the next step
```

After the fix: `tests 4 · pass 4 · fail 0`.

PR #1656, opened 2026-09-11
fix(rolldown-native): keep a failing hook's message

JumpLink:
## The measurement

A `.blp` with one deliberate syntax error, one build, two engines.

Node engine (`gjsify build ... --app gjs`) — answers the question:

    Build failed with 1 error:

    [plugin vite-plugin-blueprint]
    BlueprintCompileError: .../src/header-bar.blp: blueprint-compiler failed
    error: Could not determine what kind of syntax is meant here
    at .../src/header-bar.blp line 4 column 30:
       4 |this is not blueprint syntax @@@;
         |                             ^

GJS engine (`gjs -m dist/cli.gjs.mjs build ...`) — whole output, before this PR:

    {"error":"rolldown: Bundler::generate: BatchedBuildDiagnostic([
      BuildDiagnostic { severity: Error, kind: "UNLOADABLE_DEPENDENCY",
        message: "Could not load src/header-bar.blp", .. }])"}

No reason anywhere. In the incident that prompted this — a GitHub Actions run on
`fedora:43` — 22 of 24 `.blp` files failed and the log said this and nothing else.
The cause was a missing `libadwaita`, so no `Adw-1.typelib`, which
`blueprintcompiler/gir.py` resolves `using Adw 1;` through. blueprint-compiler
itself was installed and had printed a precise message. What cracked it was a
count: exactly 22 files contain `using Adw 1;`, exactly 22 failed, and the two
pure-Gtk files loaded.

After this PR, same build:

    {"error":"rolldown: Bundler::generate: BatchedBuildDiagnostic([...])"}
    1 plugin hook failed during this build:
      [plugin vite-plugin-blueprint] load: .../src/header-bar.blp:
      blueprint-compiler failed (/usr/bin/blueprint-compiler)
    error: Could not determine what kind of syntax is meant here
    at .../src/header-bar.blp line 4 column 30:
       4 |this is not blueprint syntax @@@;
         |                             ^

With a second `.blp` broken the headline reads `2 plugin hooks failed during this
build:` and both are listed — the count that made the real incident diagnosable.

## Cause

Not the blueprint plugin. `packages/infra/vite-plugin-blueprint/src/index.ts:67-70`
already reads execa's `.stderr` and wraps it in `BlueprintCompileError`, which is
why the Node engine prints it. The message is lost one layer down, crossing back
from JS into Rust:

- `packages/infra/rolldown-native/src/rust/src/session.rs:379` —
  `.map_err(|e| format!("rolldown: Bundler::generate: {e:?}"))`
- `{e:?}` is Debug, and rolldown's `BuildDiagnostic` has a
  `finish_non_exhaustive()` Debug impl, so everything past the short `message`
  becomes the literal `..` visible in the output above.

Fixing the Rust would mean rebuilding the cdylib, which needs the `refs/rolldown`
submodule and ships as a per-platform prebuild — a release-shaped change, and not
where the information actually lives. The JS side already has every hook error:
the plugin threw it and `respondError` serialised it moments earlier. So it keeps
its own copy instead of hoping the reason survives the round trip.

- `packages/infra/rolldown-native/src/ts/plugins.ts` — `HookErrorLog`, recorded in
  `respondError`, appended in the `error_occurred` handler.

`respondError` is the single point all twelve hooks of every plugin funnel
through, so this is the mechanism and not a blueprint patch: `xgettext`, `msgfmt`,
`msgcat`, `glib-compile-resources` and anything a future plugin spawns are covered
by construction. rolldown's own `msg` stays first and verbatim — it carries
diagnostics no JS hook was involved in.

Collect, not fail fast. By the time a hook rejects, rolldown has already
dispatched the other modules onto its worker pool and batches the diagnostics
before stopping, so aborting early would save no work and would discard the other
failures — and with them the count that named the shared cause. Capped at 8
spelled out in full (each blueprint failure is a five-line excerpt with a caret;
22 of them bury the headline), with the total always stated.

## Sweep — the same class elsewhere

Audited every subprocess spawn in the plugin families; the per-site table is in a
comment below. One background fact decides most of it: execa v9 bakes the child's
stderr into `error.message` itself, so several catches that look like they discard
it do not.

The plugin layer was largely sound, which is what pins the defect on the bridge.
One real finding: `vite-plugin-gettext/src/utils.ts:18`. `checkDependencies`
caught every probe failure as `catch (_error)` and answered all of them with the
install hint — a gettext that is installed but cannot run (missing libintl after a
partial upgrade, a noexec mount, wrong arch under emulation) was told to install
what it already has, with its own stderr dropped. Measured: an absent command
rejects with `code: 'ENOENT'` and empty stderr; a present one that fails has no
`code` and a populated stderr. The second commit branches on that.

Two non-findings worth recording: `cli/src/utils/spawn.ts` never attaches stderr
to an error, but deliberately — it resolves on any exit code and hands `stderr`
back under `'capture'`, leaving the message to each call site. And
`gresource.ts`/`gsettings.ts` hand-roll `execFile` rather than using that helper,
so their correct behaviour is duplicated in two files.

## Red to green

`tests/integration/rolldown-native` — a load hook throws a multi-line message
whose useful line is not the first, mirroring blueprint-compiler's shape. Run
under GJS against the real prebuild.

Without the fix:

    ✖ a failing load hook puts its OWN message in the rejection
      — Expected values to match using ===
    ❌ [Gjs 1.88.1] 1 of 46 tests failed  (53.0ms)

With it:

    ✔ [Gjs 1.88.1] 48 completed  (56.3ms)

The pre-existing `this.error()` test deliberately accepts "either resolution is
acceptable" and asserts only the user-land throw — nobody checked that the message
reached the caller. That gap is this test.

`packages/infra/vite-plugin-gettext` — a real temp executable that exits 1 with
stderr, not a stubbed execa, because the branch reads `error.code`, which the spawn
sets and a mock would have to guess.

Without the fix:

    ✖ checkDependencies › reports an INSTALLED tool that fails with its own stderr
      — Expected values to match using ===
    ❌ [Node.js 24.19.0] 1 of 51 tests failed  (102ms)

With it:

    ✔ [Node.js 24.19.0] 52 completed  (111ms)

The sibling case ("tells someone with no gettext to install gettext") passes in
both runs, so the red is the branch and not the harness.

The cap branch. Reviewing coverage afterwards, 1 and 2 failures were exercised but
`… and N more` was not — an untested branch in the part that carries the count.
Twelve failures now assert the headline total, the eight spelled out and the
remainder; measured before committing it: `12 plugin hooks failed during this
build:`, 8 detail lines, `  … and 4 more`.

## Validation

`gjsify format --check`, `gjsify lint` and `gjsify tsc --noEmit` clean on all three
touched packages. The Rust `{e:?}` remains as-is and is the deeper fix whenever the
cdylib is next rebuilt — it would also cover diagnostics rolldown raises with no JS
hook involved.

### DIFF
--- a/packages/infra/rolldown-native/src/ts/plugins.ts
+++ b/packages/infra/rolldown-native/src/ts/plugins.ts
@@ -255,6 +255,84 @@ export function isResolveMiss(error: string): boolean {
     return error.startsWith('NotFound(');
 }
 
+/** Which plugin's which hook a failure came from. */
+interface HookSite {
+    plugin: string;
+    hook: string;
+}
+
+/**
+ * How many hook failures to spell out in full before summarising the rest.
+ *
+ * Not 1, and not unbounded. ONE is wrong because the COUNT is itself the
+ * diagnosis: 22 of a showcase's 24 `.blp` files failed in a fedora:43 container,
+ * and the two that loaded were the two without `using Adw 1;` — no single
+ * message said "libadwaita is missing", but "22, and exactly the Adw ones" did.
+ * UNBOUNDED is wrong because each blueprint-compiler failure is a five-line
+ * excerpt with a caret, and 22 of those bury the line carrying the count.
+ *
+ * Eight shows that the failures are the same failure repeated, which is the
+ * reader's actual question; the total is stated regardless.
+ */
+const HOOK_ERROR_DETAIL_LIMIT = 8;
+
+/**
+ * The plugin-hook failures of one bundle run, kept so the build can say WHY it
+ * failed.
+ *
+ * Every hook error is already known in JS — the plugin threw it and
+ * `respondError` serialised it — and then the round trip through Rust throws it
+ * away. `Bundler::generate`'s error is Debug-formatted (`session.rs`,
+ * `format!("… {e:?}")`) and rolldown's `BuildDiagnostic` has a
+ * `finish_non_exhaustive()` Debug impl, so everything past the short `message`
+ * collapses into a literal `..`. Measured on this tree, one deliberate syntax
+ * error in a `.blp`, whole output:
+ *
+ *     {"error":"rolldown: Bundler::generate: BatchedBuildDiagnostic([
+ *       BuildDiagnostic { severity: Error, kind: "UNLOADABLE_DEPENDENCY",
+ *         message: "Could not load src/header-bar.blp", .. }])"}
+ *
+ * The same build under the Node engine printed blueprint-compiler's own "Could
+ * not determine what kind of syntax is meant here" with the line and a caret.
+ * One build, two engines, and only one of them answers the question.
+ *
+ * So the JS side keeps its own copy instead of hoping the reason survives the
+ * crossing. Deliberately NOT a blueprint fix: `respondError` is the one place
+ * all twelve hooks of every plugin funnel through, so `xgettext`, `msgfmt`,
+ * `glib-compile-resources` and whatever a future plugin spawns are covered by
+ * construction.
+ */
+class HookErrorLog {
+    private readonly entries: (HookSite & { message: string })[] = [];
+
+    record(site: HookSite, message: string): void {
+        this.entries.push({ ...site, message });
+    }
+
+    /**
+     * The recorded failures as one block, or '' when there were none.
+     *
+     * COLLECTS rather than failing fast, because the JS side has no say in it:
+     * by the time a hook rejects, rolldown has already dispatched the other
+     * modules onto its worker pool, and it batches the diagnostics
+     * (`BatchedBuildDiagnostic`) before it stops. Aborting the session at the
+     * first error would therefore not save work — it would only discard the
+     * other failures, and with them the count that names the shared cause.
+     */
+    format(): string {
+        if (this.entries.length === 0) return '';
+        const shown = this.entries.slice(0, HOOK_ERROR_DETAIL_LIMIT);
+        const lines = shown.map(({ plugin, hook, message }) => `  [plugin ${plugin}] ${hook}: ${message}`);
+        const omitted = this.entries.length - shown.length;
+        if (omitted > 0) lines.push(`  … and ${omitted} more`);
+        const headline =
+            this.entries.length === 1
+                ? '1 plugin hook failed during this build:'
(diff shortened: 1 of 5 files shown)


JumpLink:
Sweep detail for the PR description — kept as a comment because a markdown table
cannot survive commitlint's 100-character body limit when this is squashed.

Every subprocess spawn in the `vite-plugin-*` / `rolldown-plugin-*` families, plus
the two CLI commands that spawn GLib compilers:

| Site | Tool | stderr reaches user? | Note |
|---|---|---|---|
| `vite-plugin-blueprint/src/index.ts:67` | blueprint-compiler | **yes** | reads `.stderr` explicitly, wraps in `BlueprintCompileError` |
| `vite-plugin-gettext/src/gettext.ts:62` | msgfmt | yes | via execa's own message |
| `vite-plugin-gettext/src/msgfmt.ts:199` | msgfmt | yes | via execa's own message |
| `vite-plugin-gettext/src/xgettext.ts:431` | xgettext, msgcat, msgmerge | yes | via execa's own message; `updatePoFiles` throws up into this same catch |
| `cli/src/commands/gresource.ts:83` | glib-compile-resources | yes | writes `e.stderr` through, then a summary line |
| `cli/src/commands/gsettings.ts:75` | glib-compile-schemas | yes | writes `e.stderr` through, then a summary line |
| `vite-plugin-gettext/src/utils.ts:18` | msgfmt/xgettext `--version` | **NO** | fixed in this PR |

Not findings, but worth recording:

- `cli/src/utils/spawn.ts` never attaches stderr to a thrown error — by design. It
  resolves on any exit code ("a non-zero code is data, not an error") and rejects
  only when the child could not be spawned. Under `stdio: 'capture'` it hands
  `stderr` back; under the default `'inherit'` it streams to the terminal but
  captures nothing, so a caller that checks only `result.code` loses it for logs
  and tests.
- `gresource.ts` / `gsettings.ts` hand-roll `promisify(execFile)` rather than using
  that helper, so the correct behaviour above is duplicated in two files.
- `rolldown-plugin-gjsify/src/shims/unicorn-magic.ts` re-exports `execFile` with no
  catch of its own — errors pass through untouched.
- No other `vite-plugin-*` / `rolldown-plugin-*` package spawns a subprocess at all.

PR #1659, opened 2026-09-11
chore(deps): @girs 5.0.0, and signals are typed

JumpLink:
`@girs/*` 4.9.0 -> 5.0.0: 468 exact declarations across 145 manifests, plus `gjsify-lock.json`.

## What 5.0.0 actually changes

Read from the release diff, not the heading. `v4.9.0...v5.0.0` touches 34 files and exactly TWO
of them are generator source: `generators/signal-generator.ts` and the `gobject-2.0.d.ts`
template. ts-for-gir #464 deletes the permissive overloads every generated class carried —
`connect(signal: string, callback: (...args: any[]) => any)`, the same for `connect_after`, and
`emit(signal: string, ...args: any[])` — and drops the `extends [any, ...infer Q] ? Q : never`
that used to strip `emit`'s first argument. The template gains `GObject.SignalMethods<Emitter,
Signals>`, which is how a subclass now declares its own signals.

In `@girs/gtk-4.0` the whole `.d.ts` diff is those three shapes, 271 times each, and nothing
else: filtering the diff for anything but those four line patterns returns empty.

## The three vocabulary tables are unchanged, byte for byte

Not "looks the same". `gtk-4.0-vocabulary.js` and `adw-1-vocabulary.js` are byte-identical
between the published 4.9.0 and 5.0.0 tarballs — sha256
`a0bf129a55cbb50e78e5dc63b8be41236bbc6133149f8e9449216617b53d2bd8` and
`57412014713cfcb801e922a007c0d1509782a7b4fbaac2e45c5ecf7ac59a9870` on both versions. So
`ENUM_VALUES`, `ENUM_VALUES_UNREADABLE` and `PROP_ENUMS` keep their names, their shapes and
their values; the same holds for `ENUM_NICKS`, `FLAG_VALUES`, `FLAG_VALUES_UNREADABLE` and
`DECLS`, which `resolve-ident.mjs` reads beside them. Both tarballs were downloaded and
compared directly — nothing here was read out of `node_modules`.

The numbers, measured twice and both directions, against the typelib this machine runs
(GTK 4.22.4 / libadwaita 1.9.3, vocabulary generated from GTK 4.23.3 / libadwaita 1.10.0):

- vocabulary -> typelib: 739 `ENUM_VALUES` entries, 736 agree, 1 disagrees, 2 nicks this host
  does not register. The one disagreement is the documented version gap,
  `GtkEditableProperties.num-properties` (8 here, 10 in the newer GIR). `FLAG_VALUES` 137, 136
  agree, 1 host-absent. `PROP_ENUMS` 104, all 104 agree. `ENUM_VALUES_UNREADABLE` is empty.
- gtk-host's committed table -> vocabulary, i.e. ADR 0053's own number: 737 values, 736 agree,
  1 disagrees, and it is the same entry. Identical under 4.9.0 and under 5.0.0.

## The corpus did not move

`scripts/check-blueprint-corpus.mjs --require-oracle`, before the bump and after it, on the real
5.0.0 install: stage A 27 rules + 11 reality probes, stage B 38 goldens against
blueprint-compiler 0.20.4, stage C 37 byte-equal and 1 ledgered across 1 cause and 2 named
lines, stage D 38 projections. Same line, character for character.

## What broke, and where it is fixed

The signal typing broke 22 call sites in 9 packages. They are all one shape — a signal name the
object's static type does not declare — and the class and each of its sub-shapes is written up in
`docs/code-anti-patterns.md`. Nothing was fixed by re-widening: every repair either names a
narrower receiver or goes through the lower-level entry point ts-for-gir names for the dynamic
case, and both were measured on a running GJS before being used.

A real defect fell out of it. `@gjsify/http2` connected `accept-certificate` on the
`Soup.Session`, and libsoup 3 installs no signal of that name there: measured,
`GObject.signal_list_ids` on `SoupSession` answers `request-queued, request-unqueued`, and the
call throws `No signal 'accept-certificate' on object 'SoupSession'`. `rejectUnauthorized:
false` had therefore never worked on that path, and the deleted string overload is what let it
compile. It now connects on the MESSAGE, which is both where the signal lives and the only
spelling that does not disable verification for every other stream on a shared session — the
idiom `@gjsify/fetch` already used one package over.

Two of the 22 were a workaround for a gap that was not real: `@gjsify/http2` and `@gjsify/fetch`
each wrote `const signalName: string = 'accept-certificate'` with a comment saying the signal is
not in the typed map. It is, on `Soup.Message`, with its real arguments.

Eleven were webrtcbin's own action signals, invisible because
`Gst.ElementFactory.make('webrtcbin')` is typed `Gst.Element`. `internal/gst-types.ts` already
declared the PROPERTIES that are invisible for the same reason; it now declares the signals
beside them, read out of the running element with `GObject.signal_query` (GStreamer 1.28.6)
rather than off a documentation page, so the names and arities are checked where they used to be
a bare string. An interface could not carry them: redeclaring `emit` over a different key union
is TS2430, measured.

Two were a downcast that is still correct: `widgetOf(node) as Adw.ApplicationWindow` is now
TS2352, while `as Gtk.Button` and `as Adw.Window` still compile. The refusal follows
`Gio.ActionMap` — `Gtk.ApplicationWindow` implements it, `Gtk.Widget` does not, and their
`connect` declarations no longer have a member in common. Casting to `Gtk.Window`, which is what
actually declares `present()`, needs no escape hatch.

## Gates run here, on this machine

`gjsify install --immutable`, `gjsify run build`, `gjsify run check` (225 workspaces, exit 0),
`gjsify run check:examples` (82, exit 0 — `run check` EXCLUDES `@gjsify/example-*`, so the three
showcase repairs are only covered by the second one), `gjsify lint` (exit 0; the warnings it
prints are pre-existing and in other files), `gjsify format --check`, `gjsify upgrade --check
--exclude-workspace '@gjsify/integration-*'`, `audit-runtimes.mjs --check --strict`,
`check-doc-fences.mjs`, `check-blueprint-corpus.mjs --require-oracle`, `check-girs-exact-pins.mjs`
(470 declarations, every one exact, all at 5.0.0), `check-enum-values.mjs`,
`check-type-surfaces.mjs`, `check-gi-import-versions.mjs`, `check-girs-runtime-imports.mjs`,
`check-agent-context-size.mjs --check`, `check-lint-visibility.mjs`, `check-source-visibility.mjs`,
`check-source-control-bytes.mjs`.

One measurement is worth repeating because it is what the webrtc change rests on: `el.emit(…)`
and `GObject.signal_emit_by_name(el, …)` were A/B'd on a live webrtcbin under gjs 1.86 and answer
identically for a boolean (`add-turn-server` -> `false`), for an object, and for a null
(`get-transceiver 0` on a fresh bin -> `null` from both).

## Not fixed here, and why

The `@girs/gtk-4.0`-shaped question upstream may not have measured: a legal downcast from a base
widget to a class implementing a GObject interface now needs `as unknown as` unless a narrower
type is available. That is a consequence of a deliberate contract change, not a generator bug,
so it belongs in a ts-for-gir issue rather than in a consumer patch, and nothing here works
around it.

## Publication window

The premise that the 5.0.0 set was complete does not hold as stated. `gjsify/types` run
34597817768 was still `in_progress` when this started, and of the 30 `@girs` packages this
repository resolves, 29 were up and `@girs/webkit-6.0` was not — its packument still said
`latest: 4.9.0` at 15:00 UTC. It published at 15:04:49 UTC and its tarball became fetchable at
15:07:55 UTC, three minutes after the packument listed it, which is the same accept-then-commit
lag `packages/infra/cli/AGENTS.md` records for `gjsify publish`. Holding webkit at 4.9.0 was
never an option: `@girs` siblings are caret ranges, so a 4.9.0 webkit would have pulled a second
4.9.x closure in beside the 5.0.0 one and split every widget type in two.

## Lockfile

`lockfileVersion` stays 4 — no silent format migration. The diff is 302 insertions against 302
deletions, versions and integrities only. Every `@girs` tarball in it resolves to 5.0.0 except
the `4.0.0-rc.5` closure under `@girs/gwebgl-0.1` and `@girs/gjsifywebrtc-0.1`, which are the two
names `scripts/check-girs-exact-pins.mjs` already lists as `OFF_THE_RELEASE_TRAIN`.


JumpLink:
## The measurements, as tables

Kept out of the description on purpose — the description becomes the squash commit body, where
`body-max-line-length` is 100.

### The published tarballs, 4.9.0 against 5.0.0

| file in `@girs/gtk-4.0` | 4.9.0 vs 5.0.0 |
| --- | --- |
| `gtk-4.0-vocabulary.js` | identical (sha256 `a0bf129a…b53d2bd8`) |
| `gtk-4.0-vocabulary.d.ts` | identical |
| `gtk-4.0-surface.js` / `.d.ts` | identical |
| `gtk-4.0.js`, `index.js`, `index.d.ts`, `*-ambient.*`, `*-import.*`, `tsconfig.json`, `typedoc.json` | identical |
| `gtk-4.0.d.ts` | 1084 lines removed, 271 added |
| `package.json`, `README.md` | version string |

`@girs/adw-1` is the same picture, `adw-1-vocabulary.js` sha256 `57412014…c59a9870`.

The whole `gtk-4.0.d.ts` diff, by shape:

| removed / changed | count |
| --- | --- |
| `connect(signal: string, callback: (...args: any[]) => any): number;` | 271 |
| `connect_after(signal: string, callback: (...args: any[]) => any): number;` | 271 |
| `emit(signal: string, ...args: any[]): void;` | 271 |
| `emit<K …>` rewritten (drops `extends [any, ...infer Q] ? Q : never`) | 271 |
| anything else | 0 |

### Vocabulary against the installed typelib

Host: GTK 4.22.4, libadwaita 1.9.3, Pango 1.57.1. Vocabulary provenance: Gtk 4.23.3, Adw 1.10.0.
Identical numbers for 4.9.0 and for 5.0.0 — the files are the same bytes, and the run was done
against both to say so rather than to assume it.

| table | entries | agree | disagree | nick absent on host |
| --- | --- | --- | --- | --- |
| `ENUM_VALUES` | 739 | 736 | 1 | 2 |
| `FLAG_VALUES` | 137 | 136 | 0 | 1 |
| `PROP_ENUMS` | 104 | 104 | 0 | 0 |
| `ENUM_VALUES_UNREADABLE` | 0 | — | — | — |
| `ENUM_NICKS` | 129 enums | — | — | — |

The single disagreement is `GtkEditableProperties.num-properties`: 8 on this host, 10 in the GIR
of 4.23.3. Same entry in both directions.

### ADR 0053's own number, re-run

| read against | values in the committed table | agree | disagree |
| --- | --- | --- | --- |
| `@girs` 4.9.0 | 737 | 736 | 1 |
| `@girs` 5.0.0 | 737 | 736 | 1 |

### Blueprint corpus, before and after

| stage | before the bump | after |
| --- | --- | --- |
| A completeness | 27 rules + 11 reality probes, 165 nodes, 120 declared losses | same |
| B oracle (blueprint-compiler 0.20.4) | 38 goldens compared | same |
| C shadow | 37 byte-equal, 1 ledgered, 1 cause, 2 named lines | same |
| D projection | 38 SharedNode trees | same |

### What stopped compiling, and what it became

| package | site | shape | repair |
| --- | --- | --- | --- |
| `@gjsify/effect-platform` | `gtk/scope.ts`, `gtk/signal.ts` | name is a parameter | `GObject.signal_connect` |
| `@gjsify/react-native` | `press.ts` | name split out of a joined string | `GObject.signal_connect` |
| `@gjsify/adwaita-react-native` | `rows.gtk.spec.tsx` ×3 | receiver too wide | narrow to `Adw.ButtonRow` / `Adw.EntryRow` |
| `@gjsify/http2` | `client-session.ts` | signal not on the object AT ALL | connect on the `Soup.Message` |
| `@gjsify/fetch` | `request.ts` | `const signalName: string` widening | the literal, now checked |
| `@gjsify/webrtc` | 11 sites | element-class signals | `emitWebRtcBin` + a measured table |
| `example-gtk-effect-adw-services` | `app.ts` | name is a parameter | `GObject.signal_emit_by_name` |
| `example-gtk-adw-host-counter` | `app.ts` | TS2352 downcast | `as Gtk.Window` |
| `example-gtk-solid-host-counter` | `app.tsx` | TS2352 downcast | `as Gtk.Window` |

### webrtcbin's action signals, read with `GObject.signal_query`

GStreamer 1.28.6, on this host. The ten the package uses, out of webrtcbin's 20 signals:

| signal | params | returns |
| --- | --- | --- |
| `add-ice-candidate` | `guint, gchararray` | void |
| `add-transceiver` | `GstWebRTCRTPTransceiverDirection, GstCaps` | `GstWebRTCRTPTransceiver` |
| `add-turn-server` | `gchararray` | `gboolean` |
| `create-answer` | `GstStructure, GstPromise` | void |
| `create-data-channel` | `gchararray, GstStructure` | `GstWebRTCDataChannel` |
| `create-offer` | `GstStructure, GstPromise` | void |
| `get-stats` | `GstPad, GstPromise` | void |
| `get-transceiver` | `gint` | `GstWebRTCRTPTransceiver` |
| `set-local-description` | `GstWebRTCSessionDescription, GstPromise` | void |
| `set-remote-description` | `GstWebRTCSessionDescription, GstPromise` | void |

A/B on a live bin, `el.emit(…)` against `GObject.signal_emit_by_name(el, …)`:

| call | `emit` | `signal_emit_by_name` |
| --- | --- | --- |
| `add-turn-server '…'` | `false` | `false` |
| `get-transceiver 0` (fresh bin) | `null` | `null` |
| `create-data-channel` | `null` | `null` |

### The publication window

| package | 5.0.0 published (UTC) |
| --- | --- |
| `@girs/glib-2.0` | 12:27:23 |
| `@girs/gtk-4.0` | 14:09:06 |
| `@girs/adw-1` | 14:11:54 |
| `@girs/soup-3.0` | 14:37:18 |
| `@girs/webkit-6.0` | 15:04:49 (tarball served 15:07:55) |

`gjsify/types` run 34597817768 was still `in_progress` at 15:21 UTC, 3h08m in, with
`@girs/json-glib-1.0@5.0.0` still 404 — the roster is much larger than the packages this
repository resolves, and all 30 of those are up.

JumpLink:
Upstream observation split out so it does not ride on this PR: gjsify/ts-for-gir#471 — a `Gtk.Widget` is no longer castable to a class implementing a GObject interface (`Gtk.ApplicationWindow`, `Adw.ApplicationWindow`), because the removed string overload was the member those types had in common. Nothing here waits on it; the two call sites cast to `Gtk.Window`, which declares `present()` and needs no escape hatch.

PR #1661, opened 2026-09-11
fix(node-gi): template property setters, and a call with too few arguments

JumpLink:
Two independent node-gi defects, one commit each, both gjs-parity gaps with the
measurement in front of the fix. Learn6502's `--app node` artifacts on macOS and
Windows are blocked on the first.

## 1. A custom property's JS setter never ran (`6302ee5`)

Learn6502's SourceView declares `code`, `copyable` and `line-numbers` as plain
READWRITE GObject properties with JS accessors, and the tutorial sets all three
from a GtkBuilder template. Every code block rendered EMPTY and the copy button
was missing, on both platforms, while the identical source works on gjs.

GtkBuilder applies a NON-construct property AFTER construction via `g_object_set`.
node-gi's set_property vfunc — `src/class.cc:365` `NodeGiSetProperty` — only ever
wrote the engine's per-instance store and never touched the wrapper, so the
class's setter never ran and its getter read an untouched backing field. gjs
routes the same vfunc through the wrapper: `refs/gjs/gi/gobject.cpp:203`
`gjs_object_set_gproperty` → `:56` `jsobj_set_gproperty` → `JS_SetProperty`.

Measured, it was three faces of one gap — a template value, a value passed to
`new`, and a bare `set_property` all reached the store and never the setter.
Only CONSTRUCT-flagged properties were routed, by the construct-property flush
at `gi.js:852`.

The vfunc now delegates, and the gate is the WRAPPER, exactly as gjs bails when
`priv->wrapper()` is null.

**The construction-time trap, kept explicit in the code.** That gate IS the
design. A set landing during construction — before node-gi has built the wrapper
— has nothing to call into and therefore cannot fire a setter early. Firing one
early is what crashed this same SourceView before (`selectable` →
`_signalHandlers.forEach` against state the ctor body had not created yet), which
is why the previous work was scoped to CONSTRUCT only. The two times are not the
same time: GtkBuilder sets non-construct properties AFTER construction, and
delegating there is what gjs does and was never the cause of that crash. A
comment at the gate says so, because the next simplification is what brings it
back.

Construct-time values are replayed from the base ctor once the wrapper exists,
before the user ctor body, in gjs's order. Which ones: those the engine actually
STORED, in first-set order. That replaces the CONSTRUCT-flag filter with the
honest question — a property nobody set is absent from the store — so the
early-setter crash is now excluded structurally rather than by a flag, and a
plain property passed to `new` is no longer missed. Order is recorded because it
is visible output: a GHashTable has none, and conformance caught exactly that.

Setter lookup covers the dash, underscore and camelCase spellings gjs makes
equivalent via `_checkAccessors` (`refs/gjs/modules/core/_common.js:65`), which
is what lets a `line-numbers` property reach a `lineNumbers` setter.

## 2. `get_property()` answered `undefined` on every class (`1b93222`)

Type- and class-independent, and silent:

```
label.get_property('label')
  -> GLib-GObject-CRITICAL: g_object_get_property: can't retrieve property
     'label' of type 'gchararray' as value of type 'gpointer'
  -> undefined
```

Not the store and not the property path: the accessor `label.label` always
answered, and the two-argument `get_property(name, value)` always filled its
caller GValue. The cause is that node-gi checked no argument COUNT at all. A
missing JS argument was read as `undefined` (`src/calls.cc:730/776/799`) and
marshalled, so on a GValue parameter `JsToFreshGValue` (`src/marshal.cc:216`)
applied gjs's guess for null — `G_TYPE_POINTER` — and handed the callee a GValue
of the wrong type. `set_property` mirrors it ("unable to set property ... from
value of type 'gpointer'").

That is not leniency but a wrong call: gjs refuses the same one before any
marshalling (`refs/gjs/gi/function.cpp:891` → `reportMoreArgsNeeded`), so the
one-argument spelling it invited has never worked on either runtime — consumers
got `undefined` and no error to follow. `InvokeFunctionInfo` now throws gjs's
TypeError with gjs's message, down to the singular/plural of "argument" and its
`format_name()` spelling (`refs/gjs/gi/function.cpp:801`). Too MANY arguments
stays permitted: gjs only warns there, through a reporter node-gi has no
equivalent of.

## Before / after

Original reproduction, verbatim. Before:

```
JS-setter ran for : atconstruct=FROM-TEMPLATE
p.plain           : ""
p.atconstruct     : "FROM-TEMPLATE"
(process:...): GLib-GObject-CRITICAL **: g_object_get_property: can't retrieve
   property 'plain' of type 'gchararray' as value of type 'gpointer'
get_property      : undefined
```

After — byte-identical to `gjs -m` on the same program:

```
JS-setter ran for : atconstruct=FROM-TEMPLATE, plain=FROM-TEMPLATE
p.plain           : "FROM-TEMPLATE"
p.atconstruct     : "FROM-TEMPLATE"
get_property THREW: method GObject.Object.get_property: At least 2 arguments
   required, but only 1 passed
```

## Red to green

Both tests written first and watched fail. Red, on the unfixed engine: 10 of 13
failing, the 3 passing ones being the no-regression guards (two-argument
get_property, too-many-arguments, a store-backed property with no accessor) —
so the suite was not vacuous. Green after: 13/13.

## Blast radius

Baseline measured on `origin/main` with the same build, because two tests here
are already flaky: `gc-cross-thread` (red in 2 of 3 baseline headless runs) and
`Gtk.Application.run() runs a real app` (red in 2 of 3 baseline display runs).
Both remain the only failures, neither is touched by this change.

- node:test headless: baseline 634 pass / 0 fail, now 645 / 0, three runs
- node:test display: baseline 656 pass, now 669, same single pre-existing flake
- The counts reconcile exactly: 13 new tests, 2 of them display-gated
- conformance golden-diff: 36 programs x gjs/node/bun/deno, all green
- GC-stress leg: 685 pass / 0 fail; D-Bus leg 8/8
- GIMarshallingTests oracle: 370 pass / 0 fail, unchanged
- Bun 38/38; Deno 36/38 with the known #47 teardown-exit files, 0 test failures

One real hit, and it was a test asserting a call gjs rejects:
`GObject.signal_emitv` with three arguments in `byvalue-elements`. gjs demands
four (measured, arity 4) and throws on three, so that spelling only ever worked
here; the test now passes the `return_value` gjs wants and its stride assertion
is unchanged.

## End-to-end

The real Learn6502 `--app node` bundle, run against the fixed node-gi, its live
widget tree walked from OUTSIDE the app (no file in the easy6502 checkout was
modified; the node-gi swap was reverted afterwards).

SourceViews carrying their template code: 5 of 35 before, 31 of 35 after.
SourceViews with `copyable` true: 0 before, 21 after. The four still empty are
the editor and quick-help views, which legitimately start empty.

## Mechanism, not just a fix

Each defect gets a conformance program whose golden IS gjs's stdout, so both are
held on gjs, node, bun and deno rather than by a Linux-only assertion:
`custom-property-js-setter` and `callable-too-few-args`.

Divergences documented in `docs/node-gi-gjs-surface.md`. No AGENTS.md touched, so
the agent-context ratchet is untouched; `check-agent-context-size --check` green.
`oxfmt --check` and `oxlint` clean on every file changed.

### DIFF
--- a/docs/node-gi-gjs-surface.md
+++ b/docs/node-gi-gjs-surface.md
@@ -97,6 +97,26 @@ accepts BOTH `null` and `undefined` as a NULL string/object (gjs refuses
 `undefined` everywhere and `null` for non-nullable args) — see
 `status/open-todos.md`.
 
+**Too FEW arguments is a REFUSAL, not a pad.** A call supplying fewer than the
+callable's JS arity throws gjs's own `TypeError` before any marshalling —
+`method GObject.Object.get_property: At least 2 arguments required, but only 1
+passed`, down to the singular/plural of "argument" and gjs's `format_name()`
+spelling (`method <ns>.<class>.<name>` / `function <ns>.<name>`). The demanded
+count is `JsInArgCount`, the same pre-scan the invoke loop consumes arguments
+with and the same number `Function.length` reports, so what is demanded is by
+construction what is consumed. It used to pad the missing ones with `undefined`
+and marshal THAT, which is not leniency but a wrong call: on a GValue parameter
+`undefined` becomes gjs's null guess, a `G_TYPE_POINTER` GValue, so
+`label.get_property('label')` printed `g_object_get_property: can't retrieve
+property 'label' of type 'gchararray' as value of type 'gpointer'` and evaluated
+to `undefined` — on stock GTK classes and registerClass'd ones alike, with the
+`set_property` twin mirroring it (`unable to set property … from value of type
+'gpointer'`). Silent, and the one-argument spelling it invited throws on gjs, so
+it never worked anywhere; consumers just got `undefined` and no error to follow.
+Too MANY arguments stays permitted — gjs only warns there, through a JS warning
+reporter node-gi has no equivalent of. Pinned by the `callable-too-few-args`
+conformance program.
+
 ## The raw engine API (`@gjsify/node-gi`)
 
 The low-level entry points the L1 layer is built on. Most code should use L1 below; these
--- a/docs/node-gi-gjs-surface.md
+++ b/docs/node-gi-gjs-surface.md
@@ -405,6 +405,32 @@ defaults to the class name. The parent namespace/type is read from the class's
 `extends` (its `$gtypeName`), so it works for both `GObject.Object` and real GI
 classes (`class extends Gio.SimpleAction { … }`).
 
+**A custom property's JS SETTER runs whenever the property is set.** A class may
+declare a GObject property AND a matching accessor over a backing field; gjs
+routes its set_property vfunc through the wrapper (`gjs_object_set_gproperty` →
+`jsobj_set_gproperty` → `JS_SetProperty`), and node-gi does the same. The lookup
+covers the three spellings gjs makes equivalent — dash, underscore and camelCase
+— so a `line-numbers` property reaches a `lineNumbers` setter, which is what
+`_checkAccessors` (refs/gjs/modules/core/_common.js) buys on gjs by mirroring the
+declared accessor onto all three. A property with NO accessor is untouched and
+keeps the engine's per-instance store as its single backing store.
+
+It happens at **two times, and the split is load-bearing**: a set that lands
+while the instance already has a wrapper (GtkBuilder applying a non-construct
+template property through `g_object_set`, a binding, `set_property`) delegates
+immediately; a set that lands during construction — g_object_new applying
+construct properties, before node-gi has built the wrapper — has nothing to call
+into and is replayed from the base constructor, before the user ctor body, in the
+order the values were actually applied. Only properties that were REALLY SET are
+replayed (the per-instance store's keys, in first-set order): replaying every
+declared property instead runs a setter for one nobody assigned, against state
+the ctor body has not created yet, which is the Learn6502 SourceView
+`selectable` → `_signalHandlers.forEach` crash. Until 0.51 only CONSTRUCT-flagged
+properties reached a setter at all, so a plain READWRITE one set from a
+GtkBuilder template never did — every Learn6502 tutorial code block rendered
+empty on `--app node` while the same source worked on gjs. Pinned by the
+`custom-property-js-setter` conformance program.
+
 Caveats (this is the no-toggle-ref object model): the user class's JS constructor
 body is not run — GObject-idiomatic init belongs in `vfunc_constructed`;
 instances are Proxies over a native handle (but `instanceof` still works — it is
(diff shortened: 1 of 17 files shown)


JumpLink:
### Suite numbers, baseline vs this branch

Baseline = `origin/main`, same build, same host. Two tests are already flaky there
and stay the only failures.

| Leg | Baseline | This branch |
|---|---|---|
| node:test headless | 634 pass / 0 fail (`gc-cross-thread` red 2 of 3 runs) | 645 pass / 0 fail, 3 of 3 |
| node:test display | 656 pass / 1 fail (`Gtk.Application.run()` red 2 of 3) | 669 pass, same single flake |
| conformance golden-diff | 34 programs × 4 runtimes green | 36 × 4 green |
| GC-stress (`--expose-gc`) | — | 685 pass / 0 fail |
| D-Bus | — | 8 / 8 |
| GIMarshallingTests oracle | 370 / 0 | 370 / 0 |
| Bun conformance | — | 38 / 38 |
| Deno conformance | — | 36 / 38, known #47 teardown-exit, 0 test failures |

Counts reconcile exactly: 13 new tests, 2 of them display-gated, so headless
gains 11 (634 → 645) and display gains 13 (656 → 669).

### End-to-end on the real Learn6502 `--app node` bundle

Live widget tree walked from outside the app; no file in the easy6502 checkout
was modified and the node-gi swap was reverted afterwards.

| | SourceViews | with template code | `copyable` true |
|---|---|---|---|
| node-gi 0.50.0 | 35 | 5 | 0 |
| this branch | 35 | 31 | 21 |

The four still empty are the editor and quick-help views, which start empty by
design.

JumpLink:
### Cross-version pairing: the shipped-closure legs caught a real constraint

First push red-flagged `GTK suites on the shipped closure` on darwin-arm64 and
darwin-x64 with `TypeError: native.setPropertySetCallback is not a function`.

That is the job doing its job. It deliberately pairs **this checkout's JS** with
the **published** addon (`NODE_GI_NATIVE=prebuild`, and it logs the pairing:
`node-gi JS: (this checkout)` / `addon: (published)`), so any new native export
called unconditionally from `gi.js` takes those three legs down until the next
release republishes the binary.

Both new calls are now feature-guarded and degrade to 0.50.0 behaviour:

- no `setPropertySetCallback` → the post-construction delegation does not happen
- no `storedPropertyNames` → the replay falls back to the CONSTRUCT-flag filter,
  which is narrower (it misses a plain property passed to `new`) but never wrong

Verified locally against the real published `0.50.0` binary rather than by
reasoning — the published addon exports neither name, and in that pairing every
pre-existing test passes:

```
this JS + published 0.50.0 addon:  691 tests, 659 pass, 11 fail
   the 11 = my 10 new tests (they assert the new engine) + the pre-existing
   Gtk.Application.run() flake
in-tree pairing:                   691 tests, 669 pass, 1 fail (same flake)
```

Also folded into the first commit: the arity count is now read off the pre-scan
the invoke loop already computes instead of calling `JsInArgCount` again, so the
check costs nothing per call on the hottest path in the binding.

PR #1662, opened 2026-09-11
feat(gtk-runtime): hold a typelib's entry points, and bundle the GNOME typeface

JumpLink:
Two defects found by inspection on a Windows 11 VM, both in the GTK runtime
bundle rather than in the app that surfaced them. Neither was a marshalling gap
and neither was visible to any existing gate.

## 1 — `Adw.AboutDialog.new_from_appdata` is not in the win32 bundle

`Adw.AboutDialog.new_from_appdata -> THREW: no static method`, so the About
dialog does not open. Nine other named constructors work in the same run.

Measured symbol by symbol out of each published 0.50.0 bundle's own
`Adw-1.typelib` (see the table in the first comment): the two appdata entry
points are present in both darwin bundles and absent in win32-x64.
**The darwin bundles are NOT affected.**

Cause, verified against the recipe rather than assumed:
`gvsbuild/patches/libadwaita/0001-remove-appstream-dependency.patch` wraps every
`*_from_appdata` entry point in `#ifndef G_OS_WIN32` and makes `appstream_dep`
conditional on `target_system != 'windows'`. libadwaita 1.9.3 — what gvsbuild
*and* Homebrew build — still parses AppStream through the heavy `appstream`
library, for which gvsbuild defines no project; the small `ministream`
replacement landed at 1.10.alpha (`7352d8c8`) and gvsbuild already carries a
`ministream` project. Homebrew's formula `depends_on "appstream"`, which is the
whole of the asymmetry. The patch is still applied on gvsbuild `main`, so a pin
bump does not close it.

**Nothing in this repository can compile that symbol**, so what this does is
make the hole impossible to ship unannounced.
`packages/node-gi/scripts/typelib-symbols.mjs` holds an API floor per namespace,
read off the finished bundle. A missing entry point fails the build unless a
DECLARED gap names its upstream cause; the gap is recorded in
`manifest.typelibApi.gaps`, and `verify-bundle-manifest.mjs` requires that
record before a bundle may publish. The floor is keyed to the namespace the
bundle ships, exactly as `WINDOWING_DATA_SETS` is, so it cannot be dodged by
declaring less.

**The gap expires when the gvsbuild pin is raised, and the catalogue rule makes
that unavoidable.** The committed snapshot now carries the patch list of every
project a gap blames, so a re-read reds the check and names the symbols to
re-measure — and the `gvsbuild-catalogue` rule fails whenever a workflow's
`GVSBUILD_VERSION` disagrees with the snapshot's, which is what compels that
re-read at every bump. It is not automatic: nothing refreshes the snapshot on its
own and no job runs `--update`, so between two bumps a patch upstream has already
deleted is a gap nothing here can see yet. That is correct for the bytes being
built — the pinned gvsbuild still applies it — and it is a weaker claim than an
earlier revision of this body made ("expires by itself"), which is corrected here
and at all three sites in the tree. If this gap survives several bumps, a
scheduled `--update` that opens a PR on a diff is the direct expiry;
`status/open-todos.md` records it as the alternative rather than shipping a cron
nobody reads. A gap whose symbol turns out to be PRESENT fails too — the only
moment anybody would learn it was fixed.

The two upstream routes that actually close it are recorded in
`status/open-todos.md`.

## 2 — no GNOME UI font in any bundle

Two effects, one cause. `gtk/share/` held `glib-2.0`, `gtksourceview-5` and
`icons`; the only `.ttf` in any of the three published tarballs is
GtkSourceView's `BuilderBlocks.ttf`. Fontconfig's *config* shipped
(`etc/fonts/fonts.conf` + `conf.d`) with no faces for it to find, which is why
`windowingData.fontconfig: true` was never an answer to this question. On
Windows 11: 82 families on the map, `Cantarell` / `Adwaita Sans` / `Adwaita
Mono` among none of them, every request answered by Tahoma with `couldn't load
font …, falling back` and exit 0. **All three bundles are affected.**

Both builders now stage Adwaita Sans + Adwaita Mono into `gtk/share/fonts` from
the pinned `refs/adwaita-fonts` submodule — neither Homebrew nor gvsbuild
packages them, and a build-time download would put an unpinned third-party
artifact in a published tarball. The OFL text travels as its own licence
component, so the shared payload writer copies it, the notice names it and
`manifest.licenses.texts` counts it. `fonts` is a declared windowing-data set,
so an empty `share/fonts` is a build failure, not a thinner bundle.

Registration is a second step and not a path: GTK4-on-Windows is pangowin32,
whose font map is filled only from the DirectWrite collection, and a
`FONTCONFIG_FILE` naming a face directory moves it by zero families (ADR 0038
§ W1-W5). So node-gi's loader publishes `GJSIFY_GTK_RUNTIME_FONT_DIR` and the
existing `initFonts()` registers it — two variables, because an app that ships
a brand face must never have to choose between its face and the platform's.

### The size — three states, and the app picks

GTK takes the system UI font from the shell. Measured as `ascent + descent` rather than in
points — points are not comparable across platforms, and macOS reports `gtk-xft-dpi` 72 while
Pango renders at 96, so a points-only reading is wrong there twice over:

| platform | `gtk-font-name` | pango | ascent+descent |
|---|---|---|---|
| Linux (GNOME, the reference) | `Adwaita Sans 11` | 96 dpi | 19.0 px |
| macOS 15.7.9 x86_64 | `.AppleSystemUIFont 12` | 96 dpi | 18.8 px |
| Windows 11 | `Segoe UI 9` | 96 dpi | 16.0 px |

So the gap is **Windows-alone and 16 %**, not the 20 % a 9:11 points ratio suggests — an
earlier revision of this PR carried that derived number and it is corrected throughout.

"Keep the family, raise the size" was the hard-wired default and it was the unclear middle: it
honours the system font only half, so somebody who deliberately set 9 pt got 11 anyway, and
`system` was not expressible at all. Three states now, chosen by the consumer:

| policy | on Windows | what it means |
|---|---|---|
| `system` | `Segoe UI 9`, untouched | the host's font, size included |
| `size` | `Segoe UI 11` | the host's face at the size Adwaita is drawn for. Raise-only |
| `adwaita` | `Adwaita Sans 11` | the GNOME font, identical on every platform |

**Raise-only is measured, not merely defensive:** macOS asks for 12 pt, so `size` leaves it
alone — and at 18.8 px against GNOME's 19.0 that is the correct outcome. A policy that set 11
unconditionally would have made the one platform with no size problem slightly worse.

**Nothing is applied by default.** Registering faces and rewriting a user's font setting are
two different acts, and a runtime that does the second uninvited is a surprise. There is a
second reason: a default would make `system` unreachable, because the host's own value would be
overwritten before a consumer could choose to keep it. `size` is the documented recommendation.

Which is why the baseline exists: `initFonts()` records `gtk-font-name` **as the process first
found it**, and `system` restores exactly that. After a write the host's value is not
recoverable — GTK keeps no previous one, Windows has no GSettings to re-read, and on Linux the
value a session applied may itself override the schema default.

### The API the app builds against

From **`@gjsify/gtk-host/fonts`**:

```ts
type UiFontPolicy = 'system' | 'size' | 'adwaita';
const UI_FONT_POLICIES: readonly UiFontPolicy[];          // enumerate them for a dialog

applyUiFontPolicy(policy | { policy, size?, family? }): UiFontPlan;
uiFontBaseline(): string | undefined;                     // the host's own, captured pre-write
adwaitaUiFontAvailability(family?): UiFontAvailability;   // ask BEFORE offering `adwaita`
initFonts({ uiFont: 'size' });                            // or apply once while registering
planUiFontPolicy({ policy, current, baseline, ... });  // the pure decision, to set it yourself
```

`adwaitaUiFontAvailability()` answers a `FontFamilyMatch`, not a boolean, because `optical` is a
real third state: forcing `Adwaita Sans` where that family never arrived trades one substitution
for another, with a setting that then lies about what it did.

## Red → green

Every guard below was run against the artifacts that actually shipped.

- **Floor, red arm:** the published win32 `Adw-1.typelib` with the gap
  declaration removed → 2 problems naming the file and both symbols. With the
  gap → clean. Both darwin bundles → clean with no gap.
- **Floor, the other red arm:** a gap declared where the symbol is PRESENT (the
  darwin bundles) → 2 problems saying the reason has expired.
- **Expiry, red arm:** a snapshot whose `libadwaita` patch list is empty → the
  gap fails and names the symbols; a snapshot with no patch list at all →
  reported as *uncompared*, never as good news.
- **Publish gate, red arm:** `verify-bundle-manifest.mjs` against the published
  0.50.0 win32 `manifest.json` → FAILED 1 check (no floor record).
- **Fonts, red arm:** `verifyWindowingData` against the published 0.50.0 win32
  and darwin-arm64 bundles → exactly one problem each, the `fonts` set; every
  other set still passes, so the red is specific. Staging one face → clean.
- **Font policy, all three states + the way back:** host-independent in
  `ui-font.spec.ts`, and live against a real `Gtk.Settings` in `fonts.spec.ts`.
  Red arm: with the baseline capture replaced by `undefined`, the round trip
  fails `Expected: restored, Actual: kept`.
- **A vacuous pass, found and closed:** the live round trip first went GREEN with
  the baseline broken — on a current GNOME the host is already `Adwaita Sans 11`,
  so `system` and `adwaita` name the same value and it could not discriminate. It
  now detects that host and says so, and a second row forces an invented family,
  which moves the setting on every host.
- **Symbol reader:** `adw_about_dialog_new` is a prefix of
  `adw_about_dialog_new_from_appdata`, so a substring search would have answered
  PRESENT in all three bundles. Pinned by a test in both directions.

65 tests in `gtk-runtime-bundle-gates.test.mjs` (was 53 on `main` and at
`v0.50.0`; +12), `gjsify tsc --noEmit`
clean on `@gjsify/gtk-host`, new `planUiFont` and `resolveFontSources` suites
green on GJS.

## What only a CI runner can verify

The bundle builders run on Windows and macOS runners; nothing below was executed
locally, and Linux green is not offered as evidence for any of it:

- that the font staging finds `refs/adwaita-fonts` after the new checkout step,
  on all four `--windowing` jobs;
- that `manifest.typelibApi` and `windowingData.fonts` are written as expected by
  a real build;
- that a bundled face actually reaches the pangowin32 map through
  `add_font_file`, and that `gtk-font-name` moves to `Segoe UI 11` — the ADR 0038
  measurement says it does, this wiring has not been re-run on the VM;
- the two darwin legs, where the floor must pass with NO gap declared.

The `node-gi.yml` windowing legs are dispatched with this PR.

## Independent review — what it found and what changed

Reviewed against the branch by a second pass that did not write the original. Six
commits followed; the first is the one that mattered.

- **`tests/e2e/release-bundle-gate` was red on this branch and no check had run
  it.** `verify-bundle-manifest.mjs` now requires `manifest.typelibApi` and that
  suite's `goodManifest()` never gained the field — measured 39 pass / 5 fail,
  including both cases that pin the `--allow-legacy-license-record` retirement
  this PR builds its own allowance on. The `E2E n/4` job appeared in none of the
  41 checks reported green (main.yml's run was still queued), and that suite's
  own header says it IS the only pre-release coverage for that script. Fixed, and
  six cases now pin the new gate in both directions; 44 → 50 tests. Red arm
  measured: defaulting the record instead of requiring it reds 3 of the 6.
- **A fifth "green and still broken", of the same family as the four this PR
  already produced.** The bundle font test was gated on
  `GJSIFY_GTK_RUNTIME_FONT_DIR` and answered an unset variable with a log and a
  `return`. Everything proven about the typeface hangs off one line in
  `gtk-runtime.js`; remove it and the `fonts` data set still counts six files,
  the manifest still names two families, and the strongest guard here goes GREEN
  while nothing registers anything — 0.50.0 again, one layer up. The absence is
  now earned: accepted only where `maybeWireGtkWindowingEnv()`'s own
  preconditions fail.
- **`applyUiFontPolicy({ policy: 'size', family: X })` silently ignored
  `family`.** It type-checks (`ApplyUiFontPolicyOptions extends
  PlanUiFontOptions`), returned a plan and wrote the HOST's family — measured,
  `Segoe UI 11` for a caller that asked for Inter. `planUiFontPolicy` now
  forwards it.
- **`OPTICAL_SIZE_SUFFIX` admitted the tokens its own note names as the
  counterexample.** The comment gives `Noto Sans Display` / `Playfair Display` as
  the reason the set must be closed, and `Display` was in it (with `Poster`, cf.
  `Bodoni Poster`). Measured: `matchFontFamily('Noto Sans', ['Noto Sans
  Display'])` answered `optical` and handed back a DIFFERENT family as the name
  to ask for — the substitution this module exists to surface, caused by the rule
  meant to surface it. Narrowed to what was measured here plus the non-colliding
  `opsz` names; the copy of the rule in `windowing.test.mjs` aligned with it.
- **Two claims corrected** — the expiry above, and the test count above.
- **ADR 0038 amended** rather than left to drift: § 4 and
  `packages/framework/AGENTS.md` rule (9) and `docs/ship-formats.md` all still
  described a single font-directory handover.

Left deliberately unchanged, with reasons, rather than widened in this PR: the
module-level `gi://Gtk` import in `fonts.ts` (a subpath split is its own
decision); `InitFontsResult.dir` answering `undefined` when both variables name
one directory (the collapse is deliberate); and two assertions in `fonts.spec.ts`
that cannot fail.


JumpLink:
## The measurement, published 0.50.0 tarballs

Each bundle's own `Adw-1.typelib`, read as exact NUL-terminated identifiers out of the
file (`npm pack @gjsify/gtk-runtime-<target>@0.50.0`, then the symbol pool):

| symbol | win32-x64 | darwin-arm64 | darwin-x64 | Fedora 44 (libadwaita 1.9.3) |
|---|---|---|---|---|
| `adw_about_dialog_get_type` | present | present | present | present |
| `adw_about_dialog_new` | present | present | present | present |
| `adw_about_dialog_new_from_appdata` | **absent** | present | present | present |
| `adw_about_dialog_get_appdata_resource_path` | **absent** | present | present | present |
| identifiers in the pool | 3369 | 3363 | 3363 | 3363 |

The pool size is the point of the last row: the win32 typelib is the **larger** of the
two, so no count, size or file comparison could ever have found this — only the symbol.

## Fonts, same three tarballs

| | win32-x64 | darwin-arm64 | darwin-x64 |
|---|---|---|---|
| `gtk/share/` subdirectories | glib-2.0, gtksourceview-5, icons | same | same |
| `.ttf` anywhere in the tarball | `gtksourceview-5/fonts/BuilderBlocks.ttf` | same | same |
| `etc/fonts/` (fontconfig config) | present | — | — |
| `verifyWindowingData` with the new `fonts` set | 1 problem | 1 problem | 1 problem |
| every other data set | passes | passes | passes |

On Windows 11 / GTK 4.22.4 / 96 dpi / 1600x1200 scale 1:

| requested | resolved to | ascent+descent |
|---|---|---|
| `Segoe UI 9` (the host's `gtk-font-name`) | Segoe UI 9 | 16.0 px |
| `Cantarell 11` | **Tahoma 11** | 17.7 px |
| `Adwaita Sans 11` | **Tahoma 11** | 17.7 px |

82 families on the map; `Cantarell`, `Adwaita Sans` and `Adwaita Mono` on none of them.

JumpLink:
### One correction while the first CI round was running

The publish gate has TWO roles, and requiring `manifest.typelibApi` outright broke the
second one: `gtk-os-suites.yml` runs the same script against the tarball a consumer
already holds, and no published bundle can ever gain a field that postdates it. All three
"GTK suites on the shipped closure" legs would have gone red over a property 0.50.0 cannot
acquire — which is exactly the mistake `--allow-legacy-license-record` was built to undo
and whose retirement that workflow records in a comment two lines above.

So the same narrow, self-retiring shape is back one field over:
`--allow-legacy-api-record` excuses the record being ABSENT in the published-closure role
and nothing else. A recorded zero still fails, a gap naming no upstream cause still fails,
`release.yml` passes no flag at all, and the run prints that the flag is droppable the
first time a published bundle carries a record. Verified in both directions against the
three published 0.50.0 manifests and against a synthetic one that carries the record.

JumpLink:
### The staged faces, measured rather than counted

A file count is not a capability — the lesson `windowingData.decodeProbe` records one data
set over, where 860 icon files decoded zero times. So the staging was run against the real
pinned checkout and the result handed to a live Pango map.

Staging (`refs/adwaita-fonts` at `bd828c71`, checked out with the exact command the new CI
step runs — `--depth 1` against that SHA, which gitlab.gnome.org serves):

```
faces:   AdwaitaMono-Bold/-BoldItalic/-Italic/-Regular, AdwaitaSans-Italic/-Regular
bytes:   7.29 MiB
licence: adwaita-fonts OFL-1.1, 4554 bytes
fonts set: {"id":"fonts","files":6}
```

The first attempt at the capability half proved nothing: this Fedora host already has both
families installed, so `list_families()` went 100 → 100 and the diff was empty. That is the
`EMPTY IS NOT FAILURE` arm `initFonts` documents, and it is indistinguishable from nothing
having worked. Re-run with a `FONTCONFIG_FILE` that names no font directory, so the host's
own faces are off the map and only the staged bytes can put anything on it:

```
before:    4 families -> Monospace, Sans, Serif, System-ui
registered 6 face(s)
after:     6 families -> Adwaita Mono, Adwaita Sans, Monospace, Sans, Serif, System-ui
GAINED:    Adwaita Mono, Adwaita Sans
```

Exactly the two names `BUNDLED_FONT_FAMILIES` declares, from a map that had neither.

This is the fontconfig backend. The win32 arm — `add_font_file` on a pangowin32 map, and
`gtk-font-name` moving to `Segoe UI 11` — is still only verifiable on the VM or a Windows
runner, and is listed as such in the PR body.

JumpLink:
### Closing the last runner-only gap: the fonts are now proven, not counted

The first round went green, and the win32 `--windowing` build's step list confirmed the
new gates actually ran (`Check out refs/adwaita-fonts` ✓, the builder step ✓ — that one
`process.exit(1)`s on the font data set, the API floor, the gap's upstream reason and the
licence coverage). But it left the honest gap named in the PR body: **the `fonts` data set
proves the faces are IN the bundle, not that they RESOLVE.** That is the same shape as
`iconFiles: 860` with zero of them decoding.

So `test/windowing.test.mjs` — which both windowing-proof jobs already run, on a host with
**no** system GTK — now registers the bundle's own faces and asserts the effect:

- the declared families are on the map after `add_font_file` (on win32 the only mechanism
  that can put them there);
- an invented family is **not**, so `includes()` is a membership test and not a function
  that says yes;
- a request for a declared family **loads a face whose own family is that name** — the
  measured Windows failure stated positively: the name on the map and the text still Tahoma.

Four arms, all run locally:

| arm | result |
|---|---|
| bundle active, host HAS Adwaita | pass — `"Adwaita Sans 11"` loads `Adwaita Sans` |
| bundle active, host has NO Adwaita (the runner's condition) | pass |
| loader names a dir with no face | **fail** — "the bundle promised a typeface it did not ship" |
| a face that is not the declared family | **fail** — names what the map gained instead (`Noto Sans Ethiopic`) |

No bundle → returns with a stated reason, never a silent pass.

**One correction worth recording, because it was a defect in the guard itself.** The first
spelling of the discriminator asserted "resolves to something OTHER than the fallback".
Run under the runner's own condition it FAILED on a healthy bundle: with the host's fonts
off the map, the fallback for a nonexistent family *is* one of the bundled faces. A
discriminator that reds on the healthy state is worse than none, so the comparison is now
against the name asked for, which does not depend on what else the host has. It was found
by running the guard under the condition it will meet — not by reading it.

JumpLink:
### The policy is selectable now — and the cross-platform measurements corrected a number of mine

Three states, all consumer-chosen, from **`@gjsify/gtk-host/fonts`**:

```ts
type UiFontPolicy = 'system' | 'size' | 'adwaita';
const UI_FONT_POLICIES: readonly UiFontPolicy[];

applyUiFontPolicy(policy | { policy, size?, family? }): UiFontPlan;
uiFontBaseline(): string | undefined;
adwaitaUiFontAvailability(family?): UiFontAvailability;
initFonts({ uiFont: 'size' });
planUiFontPolicy({ policy, current, baseline, ... });   // pure, if you'd rather set it yourself
```

`size` is no longer applied by default — **nothing is**. A runtime that rewrites a font setting
uninvited is a surprise, and a default would have made `system` unreachable: the host's own value
would be gone before a consumer could choose to keep it. `size` is the documented recommendation.

### What the measurements changed

The `ascent + descent` numbers are now the basis everywhere, and they corrected me:

- **16 %, not 20 %.** I had derived the figure from the 9:11 points ratio. Measured, the gap is
  16.0 px against GNOME's 19.0. Corrected in `ui-font.ts`, `fonts.ts`, the spec, the win32 README,
  the guide, the PR body **and the commit message that carried it** — a derived percentage sitting
  in a comment that is later read as the justification is exactly the claim nobody re-measures.
- **Raise-only is now measured, not just defensive.** macOS asks for 12 pt and measures 18.8 px
  against 19.0, so `size` correctly leaves it alone. A policy that set 11 unconditionally would
  have made the one platform with no size problem slightly worse. That is now the worked example
  in the policy note and a test row of its own.
- **"All three bundles affected" is no longer an inference on darwin.** 187 families, `Adwaita
  Sans` and `Cantarell` absent from both, both falling back to Helvetica — recorded in the darwin
  builder where the claim is made.
- **`new_from_appdata` on macOS was CALLED, not just byte-scanned** — it ran and failed on the
  deliberately invalid path. Recorded in `typelib-symbols.mjs`: a constructor that gets as far as
  parsing is present in every sense that gate cares about.

### Red → green for the policy

| arm | result |
|---|---|
| all three states + `adwaita → system`, host-independent | pass (`ui-font.spec.ts`) |
| the same against a live `Gtk.Settings` | pass (`fonts.spec.ts`, gated on a real GTK) |
| baseline capture replaced by `undefined` | **fail** — `Expected: restored, Actual: kept` |
| `adwaita` offered where the family is absent | probe answers `absent`, `available: false` |

**And one vacuous pass, found and closed.** The live round trip first went GREEN *with the
baseline broken*: on a current GNOME the host is already `Adwaita Sans 11`, so `system` and
`adwaita` name the same value and the test could not discriminate at all. It now detects that
host and says so rather than passing quietly, and a second row forces an invented family, which
moves the setting on every host and is the row that actually fails when the baseline is broken.

Two of my own assertions were host-dependent and both went green before they were right —
expecting `adwaita` to report `family` (on this host `kept` is correct), and the round trip
above. Both were found by running them, not by reading them.

JumpLink:
### CI found a real defect in the `adwaita` policy — on the one platform it happens on

The Windows windowing proof failed, and not on the test this time:

```
AssertionError: "Adwaita Sans" is not on the font map after registering 6 bundled face(s)
  Map gained: [Adwaita Sans Text, Adwaita Mono]
```

`applyUiFontPolicy('adwaita')` was writing `gtk-font-name = "Adwaita Sans 11"`. On Windows that
family does not exist, so Pango substitutes — and a user who picked "use the Adwaita font" would
have got **Tahoma**, silently. That is precisely the failure the policy exists to avoid, shipped
inside the policy itself.

**Cause, measured in the font file.** `AdwaitaSans-Regular.ttf` declares nameID 1 `Adwaita Sans`
and carries an `opsz` axis (14–32) whose value at 14 is named `Text`. fontconfig reports the
nameID-1 family; gvsbuild's DirectWrite reader composes the STAT name:

| reader | family on the map |
|---|---|
| fontconfig (Linux, macOS) | `Adwaita Sans` |
| gvsbuild DirectWrite (Windows) | `Adwaita Sans Text` |

Byte-identical file, two names — which is the `Merriweather` / `Merriweather 18pt` finding this
repo already wrote down in `font-families.ts`. `OPTICAL_SIZE_SUFFIX` matched only the numeric
spelling, so the documented rule did not cover its own bundled typeface.

**Fixed at the class, not the case:**

- the optical rule now also accepts the standard `opsz` axis-value names, and deliberately as a
  **closed set** — `\w+` would match `Adwaita Sans Condensed`, a different family, and cause the
  substitution the module exists to prevent. `exact` is still tried first; two variants still
  land in `ambiguous`.
- `applyUiFontPolicy('adwaita')` resolves the family against the live map and writes the name
  that map holds. `absent` falls back to the declared name **and warns** — the consumer asked for
  the state explicitly and was told to check `adwaitaUiFontAvailability()` first; what it must
  not do is fail quietly.
- both tests now resolve instead of matching the literal string, which is what the product does.

**Reproduced locally**, by rewriting a fixture's name table to what the Windows reader exposes:

```
fonts: 2 bundled face(s) → map 4 → 6 families;
  declared [Adwaita Sans, Adwaita Mono] resolve to [Adwaita Sans Text, Adwaita Mono]
  and load as [Adwaita Sans Text, Adwaita Mono]
```

Plus two permanent rows in `font-families.spec.ts`: the axis-value name resolves, and the closed
set stays closed.

**For the app side:** nothing changes in the API. `applyUiFontPolicy('adwaita')` does the
resolution for you. If you ever set `gtk-font-name` directly, do not hard-code `Adwaita Sans` —
ask `adwaitaUiFontAvailability().match.family`.

JumpLink:
### And the darwin legs found the other half: macOS cannot register the faces at all

Both macOS windowing proofs failed, on the raw vfunc:

```
Error [GLib.Error]: Adding font files not supported for PangoCairoCoreTextFontMap
```

My test called `add_font_file` directly — but the test was only the messenger. `add_font_file`
is a vfunc the CoreText map **does not implement**, so on macOS every bundled face answers
`G_IO_ERROR_NOT_SUPPORTED`. `initFonts()` has always reported that as `declined` rather than as a
failure, and the reasoning written into that arm is about an application's OWN faces in a shipped
`.app`, where `ATSApplicationFontsPath` activated the directory before any code ran.

**That reasoning does not extend to the runtime bundle's faces** — nothing points that key at
`gtk/share/fonts`. So the darwin bundles now carry ~7.3 MB of faces no process can reach.

| platform | registration route | works |
|---|---|---|
| Linux | fontconfig via `XDG_DATA_DIRS`, plus `add_font_file` | yes |
| Windows | `add_font_file` only — pangowin32 reads no fontconfig path | yes |
| macOS | CoreText map implements no runtime registration | **no** |

The consequence is at least coherent rather than silent: `adwaitaUiFontAvailability()` answers
`absent` on macOS, so a preferences dialog will not offer the `adwaita` policy there. `system`
and `size` are unaffected, and macOS needs no size correction anyway — 18.8 px against GNOME's
19.0.

**What I did NOT do is paper over it.** Asserting the faces arrived would be a lie, and passing
over it quietly would be the third vacuous green in this PR. The decline is now asserted
explicitly — every face declined, none failed for another reason — and the row states what was
and was not proven. The day a darwin map starts accepting registration the count stops matching
and the row reports it.

The faces stay in the darwin bundle deliberately: the payload is not what is broken, and both
routes out need them there —

- **`ATSApplicationFontsPath`**, which is how `gjsify ship` already activates an app's own staged
  faces. It names one directory relative to `Contents/Resources`, so covering both would mean
  staging the bundle's faces into the app's font directory at ship time — a `gjsify ship` change,
  in the layer that owns the `.app` layout.
- **`PANGOCAIRO_BACKEND=fc`**, which would make the existing `XDG_DATA_DIRS` wiring work with no
  further change, and which alters text rendering for the whole application — not a decision a
  runtime bundle may take for its consumer.

Both are in `status/open-todos.md` with that reasoning. Windows, meanwhile, went green on the
family-resolution fix, and all four `--windowing` bundle builds succeeded on both platforms.

PR #1663, opened 2026-09-11
fix(node-gi): set the process locale at load

JumpLink:
## The defect

Under `@gjsify/node-gi`, every application was untranslated, on every platform.

A C program starts in the `"C"` locale, and GNU gettext returns the untranslated
msgid whenever `LC_MESSAGES` is C — whatever `LANG` says, whatever
`bindtextdomain` was told. Node never leaves that locale and neither did
node-gtk, the derivation source. gjs's entry point opens with
`setlocale(LC_ALL, "")` (`refs/gjs/gjs/console.cpp:238`), so the same source
translated under `--app gjs` and not under `--app node`.

Measured, same catalogs, same environment, same directory, only the host differs:

    === GJS ===
    GLib.dgettext : "6502 Stack voll! Überlauf..."
    === node-gi (before) ===
    GLib.dgettext : "6502 Stack filled! Wrapping..."
    === node-gi (after) ===
    GLib.dgettext : "6502 Stack voll! Überlauf..."

Two things hid it, and both are recorded so the rule survives a simplification:

- `GLib.get_language_names()` reads the environment directly and was correct
  throughout. An About dialog resolving its title that way was properly German
  beside English menus, which reads as "some strings are untranslated" rather
  than "the process is in the C locale".
- `gtk_init()` calls `setlocale(LC_ALL, "")` itself (`gtk/gtkmain.c`), so a GTK
  program translates what it looks up AFTER init. Only module-scope strings and
  headless/CLI programs stay wrong — a probe measuring post-init strings would
  conclude the locale was fine.

## The fix

The addon's N-API `Init` performs the `setlocale`. That is the seam every
node-gi program crosses (`gi://`, `requireGi`, the globals shim, a bare import),
whereas `globals.js` is injected only into bundles still referencing the GJS
ambient globals. One process-wide call therefore fixes node, bun and deno from
the same binary.

`gettext.js` was the other half: a passthrough whose binders returned `null`, so
`bindtextdomain(APP_ID, localeDir())` bound nothing and a real locale alone would
still have translated nothing. Lookups now go to GLib (`g_dgettext`,
`g_dngettext`, `g_dpgettext2`) and the binders to libintl through the addon —
the same split gjs makes between `imports.gi.GLib` and `GjsPrivate`.
`LocaleCategory` is read from the host's `<locale.h>`; the hardcoded glibc table
it carried addressed `LC_TIME` on macOS.

## Why LC_ALL whole, LC_NUMERIC included

The hazard is real and was measured here: right after `setlocale(LC_ALL, "")` in
a German locale, the C library's `strtod("3.5")` returns `3.0`. The carve-out is
still wrong, for three reasons.

1. gjs does not make it. Measured under `gjs` with `LANG=de_DE.UTF-8`:
   `LC_NUMERIC = de_DE.UTF-8`. A node-only carve-out would be a silent
   `--app gjs` / `--app node` divergence in number formatting.
2. It would not hold. `gtk_init()` sets the full `LC_ALL` unconditionally, so in
   any GTK program the carve-out is undone moments later — a guard a dependency
   reverts is worse than none.
3. JS is unaffected and GLib has its own answer. `Number`, `JSON` and
   `GLib.Variant` round-trip `3.5` identically before and after; the exposure is
   C-library `strtod`/`printf`, and GNOME's answer to that is `g_ascii_strtod`,
   which GLib and gjs already use throughout for exactly this reason.

## Blast radius

It landed on tests, not the runtime. Assertions matching GLib's own TRANSLATED
error text went red on a German host — `vfunc-chainup` matched `/closed/i` and
got "Datenstrom ist bereits geschlossen". It asserts domain and code now, which
is also stricter, and the node test scripts pin `LC_ALL=C` so a developer's
locale is not a hidden input to the suite — the reason the conformance harness
already did.

## Red to green

`test/locale.test.mjs` builds a real `.mo` from JS (no `msgfmt`, no committed
binary) and measures a msgid whose translation DIFFERS from it. Without the fix
6 of 7 fail on the translation and the 7th on `setlocale(ALL, null)` returning
null; with it, 7/7.

The gate in `test/locale-gate.mjs` reads `locale -a` rather than node-gi's own
`setlocale` on purpose: a gate built on the code under test reports "nothing to
run here" for exactly the defect it exists to catch.

The CI image gains `glibc-langpack-de`, without which no translation can be
measured at all — `C.UTF-8` is the only locale `glibc-minimal-langpack`
generates, and glibc treats it as C for message lookup and ignores `LANGUAGE`
under it (measured, glibc 2.43). The arm64 leg installs it directly so this PR's
own CI runs the full effect test.

## Runtimes checked

node, bun and deno all translate after the fix, from the one addon Init;
`locale` is in the cross-runtime subset so CI proves it rather than assuming it.
GJS applications launched through the gjsify CLI are untouched — the CLI spawns
`gjs -m`, whose `main()` has always done this.

Windows and macOS need libintl named at link time (glibc absorbs it into libc);
`binding.gyp` handles both, and both are optional so a missing import lib yields
the linker's named undefined symbols rather than a blanked library list.


JumpLink:
### Measurements

Suite, `packages/node-gi/node-gi` (Linux, `LC_ALL=C`, `NODE_GI_NATIVE=build`):

| run | tests | pass | fail | skipped |
|---|---|---|---|---|
| `npm test` on `origin/main` (baseline) | 678 | 657 | 0 | 21 |
| `npm test` on this branch | 690 | 669 | 0 | 21 |
| `npm run test:gc` on this branch | 690 | 684 | 0 | 6 |

Cross-runtime:

| leg | result |
|---|---|
| `npm run test:conformance` (gjs · node · bun · deno, 34 programs) | all green, no unexpected results |
| `node scripts/cross-runtime.mjs bun` | 39/39 green (`locale` ✓) |
| `node scripts/cross-runtime.mjs deno` | 38/39 green, 1 known non-gating (`dbus-async`, #47) (`locale` ✓) |

Red→green on the new file, verbatim, same command before and after the fix:

| test | before | after |
|---|---|---|
| a fresh node-gi process adopts the environment locale, not "C" | ✖ `null !== 'de_DE.utf8'` | ✔ |
| setlocale reports the locale rather than the addon guessing it | ✖ `'object' !== 'string'` | ✔ |
| GLib.dgettext translates — the C path, no JS shim involved | ✖ `'Stack filled' !== 'Stapel voll'` | ✔ |
| the Gettext surface translates through the bound domain | ✖ `'Stack filled' !== 'Stapel voll'` | ✔ |
| plural and context lookups read the catalog | ✖ `'%d file' !== '%d Datei'` | ✔ |
| LocaleCategory.MESSAGES is the C library's LC_MESSAGES | ✖ | ✔ |
| the language list and the catalog agree | ✖ | ✔ |

The `LC_NUMERIC` measurement behind the decision, one process, in order:

| step | value |
|---|---|
| locale at start (Node/Python default) | `LC_MESSAGES=C` |
| `g_dgettext` before `setlocale` | `6502 Stack filled! Wrapping...` |
| `g_dgettext` after `setlocale(LC_ALL, "")` | `6502 Stack voll! Überlauf...` |
| `LC_NUMERIC` after | `de_DE.UTF-8` |
| C `strtod("3.5")` after | `3.0` |

And the same question put to gjs, which is why the carve-out was rejected:

| gjs, `LANG=de_DE.UTF-8` | value |
|---|---|
| `Gettext.setlocale(LocaleCategory.NUMERIC, null)` | `de_DE.UTF-8` |
| `Gettext.setlocale(LocaleCategory.ALL, null)` | `de_DE.UTF-8` |

Why `C.UTF-8` cannot stand in for a langpack (glibc 2.43, catalog bound, only
`de/LC_MESSAGES/` present):

| environment | lookup |
|---|---|
| `LC_ALL=C.UTF-8 LANGUAGE=de` | `Stack filled` (untranslated) |
| `LANG=C.UTF-8 LANGUAGE=de` | `Stack filled` (untranslated) |
| `LC_ALL=en_US.utf8 LANGUAGE=de` | `Stapel voll` |
| `LC_ALL=de_DE.utf8` | `Stapel voll` |

That third row is also what lets the test need only ONE fixture language plus any
generated non-C locale, rather than a matching locale/catalog pair per host.

### Known, pre-existing

`scripts/check-comment-budget.mjs` reports `packages/node-gi` OVER its ceiling.
It is over on `origin/main` too — measured in a clean `origin/main` worktree at
ratio 0.495 against a ceiling of 0.434. This branch moves it to 0.498. The check
runs `--warn` only.

JumpLink:
### Four things CI caught that local runs could not

Recorded because each is a distinct class, and three of them are invisible on a
Linux workstation.

**1. A raw NUL byte in a source file.** `test/mo-catalog.mjs` writes gettext's
plural/context separators, and the `'\u0000'` / `'\u0004'` I wrote landed as
literal control bytes rather than escapes. `check-source-control-bytes.mjs`
caught it — git would otherwise have classified the file as binary and stopped
showing its diff. Written as escapes now.

**2. `<windows.h>` defines `SetThreadLocale`.** The MSVC build died with
`addon.cc(150,50): error C2665` because uv.h pulls in `<windows.h>`, whose
object-like macro rewrote the token at the `Napi::Function::New` call site — the
same collision the existing `#undef RegisterClass` exists for. The C++ symbol is
`ApplyThreadLocale` now; the JS export name is unchanged, and the undef list
stays at one entry. (`intl.lib` itself resolved fine:
`C:/gtk-build/gtk/x64/release/lib/intl.lib`.)

**3. The shipped-closure legs pair a PUBLISHED addon with this tree's JS.**
`gtk-os-suites.yml` stages "the tarball a consumer actually receives" on purpose,
so a bare `native.localeCategories()` at module scope took the whole module down
with `TypeError: native.localeCategories is not a function` — and `globals.js`
with it, since it imports gettext.js. The binders are probed now, the same way
`native-prebuilds.js` probes `prependLibraryPath`, and an addon without them
degrades to exactly what it could always do: lookups still work (they go through
GLib, which every addon version has) and the binders are the no-ops they were.
It says so once via `process.emitWarning` rather than silently, because a
silently untranslated UI is the defect this PR closes.

**4. `oxfmt --check`** wanted five of the new files reformatted; done with the
repo's pinned 0.61.0.

Re-verified locally after all four: suite 690/669 pass/0 fail/21 skipped,
`test:gc` 690/684/0/6, conformance all green, bun 39/39, deno 38/39 (the known
`dbus-async` #47 carve-out).

JumpLink:
### Windows: the fix works there, and it corrected a wart in this PR

The Windows leg's node-gi run is where the release blocker was reported from, so
it is worth quoting what it now measures:

| Windows test | before this PR | now |
|---|---|---|
| a fresh node-gi process adopts the environment locale | — | ✔ |
| `GLib.dgettext` translates (the C path) | untranslated | ✔ |
| the Gettext surface translates through the bound domain | untranslated | ✔ |
| plural and context lookups read the catalog | untranslated | ✔ |
| the language list and the catalog agree | — | ✔ |

The three reds it also produced were mine, and one of them was a real defect
rather than a test assumption.

**The defect.** `setlocale(LocaleCategory.MESSAGES, …)` was special-cased on
Windows to cache the name and skip the C library, on the reasoning that the MSVC
CRT has no `LC_MESSAGES`. That made `setlocale(MESSAGES, 'no-such-locale.invalid')`
answer with the PRIOR name — a failure indistinguishable from a success, which is
the exact shape this repo calls green-that-checked-nothing. It goes through the
same call as every other category now and answers `null`, which is the honest
report. Messages still resolve, because GNU gettext reads them from the
environment on Windows rather than from the CRT locale.

**The test assumptions.** Two assertions encoded POSIX behaviour:

- "setlocale answers null for a locale the host does not have" asked on
  `MESSAGES`, where Windows answers null for valid values too — so the assertion
  would have passed there without testing a rejection. It asks on `CTYPE` now, a
  category all three C libraries have.
- "forcing MESSAGES back to C stops message lookup" cannot hold on Windows. That
  case is POSIX-scoped, and a win32 case beside it asserts the divergence
  directly, so neither platform goes unmeasured rather than one being skipped.

`intl.lib` resolved on the first try
(`C:/gtk-build/gtk/x64/release/lib/intl.lib`); the earlier Windows red was the
`SetThreadLocale` macro, not the link.

JumpLink:
### Correction: LC_MESSAGES *is* settable on Windows

The win32 case I added asserted the wrong thing, and the Windows leg said so:
`setlocale(LocaleCategory.MESSAGES, 'C')` answered `'en_US'`, not `null`.

The MSVC CRT has no `LC_MESSAGES`, but GNU gettext's `<libintl.h>` redirects
`setlocale` to its own `libintl_setlocale` on native Windows, and that one does
carry the category. So the divergence from POSIX is **precedence**, not
existence: glibc ignores `LANGUAGE` once the message locale is `"C"` — which is
what makes "force MESSAGES to C, translation stops" a discriminator on Linux —
while gettext's Windows port keeps honouring `LANGUAGE`, so the same program goes
on translating.

Both halves are now asserted, each on the platform where it holds, rather than
one being skipped. The win32 case pins exactly what CI measured: the call reports
a string, and translation survives `LC_MESSAGES=C`.

This also retires the assumption behind the special-case removed in the previous
push. It was wrong twice — the CRT does not refuse the category, and skipping the
call made a failure indistinguishable from a success. Handing every category to
the C library is simply correct on all three platforms.

PR #1664, opened 2026-09-11
feat(cli): ship an AppImage from the Linux payload

JumpLink:
`gjsify ship --target appimage` packs the `/usr` tree `deb` and `rpm` already wrap
into an AppDir and hands it to `appimagetool`. Closes the user request in
JumpLink/Learn6502#93.

## Why this is not the thing ADR 0024 § 9 refused

§ 9 deferred AppImage on one argument: the format promises *one file, no install,
any distro*, the third holds only if the file carries GTK4, libadwaita, GJS and
the typelibs, no relocatable Linux closure exists, and *"that is not an AppImage;
it is a tarball with a launcher"*.

That closure still does not exist. The deferral is lifted anyway, because the
argument counted three promises and weighed one. **No install and no root hold
with no closure at all**, and they are the two nothing else in the table offers:
a `.deb` needs root and the right distribution, a Flatpak needs flatpak installed
and a runtime downloaded. Someone else's machine, someone else's distro, no sudo
is served by none of them.

What would have made § 9 right is silence, so silence is what is refused:

- `appImageHostRequirements` derives what the file does NOT carry from the same
  scan `deriveDepends` derives the `.deb`'s `Depends:` from — `hostProvidedNamespaces`
  is now shared between them, so two artifacts from one payload cannot disagree.
  Distro-neutral on purpose: `the Gtk-4.0 typelib`, not `gir1.2-gtk-4.0`.
- `gjsify ship` PRINTS that list on every pack, not behind `--verbose`.
- `AppRun` carries the interpreter half INSIDE the artifact and refuses at exit
  127 naming it, instead of dying with `gjs: not found` or with nothing at all.

The day `@gjsify/gtk-runtime-linux-<arch>` lands, the list shrinks and nothing
else changes — which is § 9's own prediction, kept. ADR 0024 § A24-§ A27 records
all of it.

## The four decisions

**Where `appimagetool` comes from: the host's PATH.** The repo's existing line,
not a new one — `flatpak-builder` for a Flatpak, WiX for an `.msi` on Windows.
Writing the format ourselves stays refused by § A6's rule (a squashfs writer is a
project, not a target). Vendoring the binary is refused on licence: GPL-3.0 into
185 MIT manifests, ADR 0023's argument. Downloading at pack time is refused
because every other packer here runs offline. The cost is honest and paid in the
`installHint`: this is the first row that cannot name a distro package, so it
names the GitHub release and the `--stage` route.

**FUSE, measured in a `fedora:44` container with no `/dev/fuse`.** Without
`APPIMAGE_EXTRACT_AND_RUN=1` appimagetool exits 127 — it is itself an AppImage
and mounts itself through libfuse to start. The packer sets that variable
unconditionally (a native build ignores it), so the FUSE-less path is the default
rather than something a CI author has to know. Same container: it also needs
`file(1)` even with `ARCH` set, and the produced artifact runs there only through
`--appimage-extract-and-run`. All three are now in the CI image or the e2e suite.

**Host-boundness: `finishOn: ['linux']`, flatpak's kind and not the `.dmg`'s** —
the container is an ELF runtime for Linux, so the format is bound the way the
application is. Not `'any'`, so `defaultFormatIds('linux')` stays `deb` + `rpm`
and the format is opt-in through `--target`, per ADR 0024's rule for a row that
needs extra tooling.

**Architecture: `uname -m`'s vocabulary**, in its own table rather than shared
with rpm's — `x86_64`, `aarch64`, `i686`, `armhf`, and an arch with no runtime is
refused rather than labelled. No `all`/`noarch` row: the file begins with an ELF.
The artifact is `<binary>-<version>-<release>.<arch>.AppImage`, the convention
the `.dmg`, the `.msi` and both zips already use for a download.

## The oracle is not the artifact

`--appimage-offset` and `--appimage-extract` are the embedded runtime reading the
file it is embedded in — `selfReading` with extra steps.
`.github/ship-oracle/verify-appimage.py` derives the filesystem offset from the
ELF section-header table and lists the tree with `unsquashfs`. The derived value
was checked against `--appimage-offset` on a real artifact: 944632 both ways.
That chain catches the failure this format has and the `.deb` does not — an
executable ELF of plausible size with nothing behind it, which mounts an empty
directory at exit 0. The e2e suite produces exactly that file on purpose and
asserts the oracle refuses it.

## One defect found by building it

appimagetool CREATES `.DirIcon` when the AppDir has none: a symlink, with the
wall clock, written after the packer has stamped every path, which also moves the
AppDir root's own mtime. mksquashfs stores both, so two packs of ONE build
differed in sha256 while every listing, mode, size and content byte was identical
— 965 differing bytes, the first inside `.digest_md5`. Four packs over the SAME
AppDir were byte-identical, which is what separated "nondeterministic tool" from
"the tool edits the tree". The packer writes `.DirIcon` itself now, and the
determinism assertion is what would notice if it stopped.

## Red then green

`appimage.spec.ts` (33 assertions) plus `tests/e2e/ship-appimage` (8 tests, three
tiers with four negative controls). Eight targeted mutations were applied to the
implementation one at a time and each produced a named red; the full list is in a
comment below.

The ninth attempt is worth naming here because it stayed GREEN: deleting the
`/symbolic/` icon guard failed nothing, because `<appId>-symbolic.svg` is already
excluded by the basename match. The assertion was passing for the wrong reason.
It now tests the case where the guard actually decides — a plain-named icon in
the symbolic context, alone — and that mutation reds.

## Verified end to end, on this workstation

Built from `tests/e2e/ship/fixture.mjs`'s project on Fedora 44 / x86_64 with
appimagetool 1.9.1 and RAN: `[object GIRepositoryNamespace] …`, exit 0, both by
FUSE mount and by `--appimage-extract-and-run`. Read back by the oracle: 8 staged
paths at exactly the planned modes, plus AppRun, entry and icon. Two independent
`gjsify ship` runs over one build: identical sha256. With `gjs` off PATH the
artifact printed its requirement list and exited 127.

## What this PR's own CI cannot prove, and why

The CI image gains `appimagetool` (pinned by SHA-256), `squashfs-tools` and
`file`, so tier 3 stops being a printed skip — the same argument that bakes
`msitools` rather than probing for it. That new layer was build-tested with
podman (digest verified, `appimagetool --version` runs in the image).

But `build-ci-image.yml` BUILDS on a PR and pushes only on main, by design. So
`main.yml`'s e2e shard in this PR consumes the last-good image, which has no
`appimagetool`, and tier 3 will print its skip here and become real on the next
run after merge. Every tier-3 assertion did run on this workstation — see above.

Also unverified by anything here: `aarch64`, `i686` and `armhf`. The arch table
names them, the image is `linux/amd64` only, and no leg exercises the other
three.


JumpLink:
### Red → green, measured

Each row is one mutation applied to the implementation alone, the CLI test bundle
rebuilt, and the suite re-run. Baseline: **1793 passed, 0 failed**.

| mutation | result | assertion that caught it |
|---|---|---|
| drop `--no-appstream` from the arg vector | 1 failed | `appImageToolArgs › turns OFF the AppStream validation` |
| drop `ARCH` from the child env | 1 failed | `appImageToolEnv › sets ARCH` |
| drop `APPIMAGE_EXTRACT_AND_RUN` | 1 failed | `appImageToolEnv › takes the FUSE-less path by default` |
| write the payload at the AppDir root, not under `usr/` | 5 failed | `appDirPayload › puts the whole prefix under usr/`, + 4 |
| stamp nothing | 1 failed | `appDirPayload › stamps every path with the run's mtime` |
| delete the `/symbolic/` icon guard (first attempt) | **0 failed** | — see below |
| delete the `/symbolic/` icon guard (after the fix) | 1 failed | `selectAppDirIcon › refuses the symbolic CONTEXT too` |
| `AppRun` without the interpreter refusal | 2 failed | `renderAppRun › refuses with a SENTENCE…`, + 1 |
| format row claims `finishOn: 'any'` | 5 failed | `format descriptors › keeps the host-bound format out of the default target set`, + 4 |

The sixth row is the one worth reading. The assertion *"never picks a symbolic
icon"* passed with the guard deleted, because `<appId>-symbolic.svg` is already
excluded by the basename match — it was testing a different rule than the one it
named. The case where the guard actually decides is a **plain-named** icon in the
symbolic context, alone, and that is what the test asserts now.

### The failure modes the artifact has, produced on purpose

| what | where | what it would look like without the check |
|---|---|---|
| hollow AppImage (ELF, no filesystem) | e2e negative control | executable file of plausible size, mounts an empty directory, exit 0 |
| `AppRun` not executable | `verify-appimage.py` | image mounts, runtime has nothing to exec |
| launcher packed 0644 | `verify-appimage.py` mode compare | installs and will not start |
| root icon is the `-symbolic` glyph | `verify-appimage.py` | a monochrome 16px glyph as the app's face in every file manager |
| AppDir without `usr/` | e2e negative control | launcher resolves the mountpoint as prefix, dies at the first `Gio.Settings.new()` |
| non-reproducible pack | e2e determinism assertion | two identical builds, two different sha256 |

### Measured in a `fedora:44` container with no `/dev/fuse`

| step | result |
|---|---|
| `appimagetool` without `APPIMAGE_EXTRACT_AND_RUN=1` | exit 127, FUSE message |
| with it, no `file(1)` | exit 1, `file command is missing but required` (even with `ARCH` set) |
| with it, `file` + `squashfs-tools` installed | built OK, 948 728 bytes |
| oracle: derived offset | 944 632, equal to `--appimage-offset` |
| `unsquashfs -ll` | full tree with modes |
| run the artifact by mount | fails (no `/dev/fuse`) |
| run with `--appimage-extract-and-run` | `hello from /tmp/appimage_extracted_…/usr/bin/hello` |

That is why the CI image gains `file` and `squashfs-tools` beside the pinned
`appimagetool`, and why the e2e suite asserts the extract path always and the
mount path only where `/dev/fuse` exists.

### Nothing was adapted from `gtkx`

`refs/gtkx` was read for the approach only — how it obtains `appimagetool`, what
it puts in an AppDir — and **no code was taken**. gjsify's answers differ where
it matters: gtkx downloads a pinned `appimagetool` **and** a pinned type2 runtime
at deploy time; this requires the tool on the host and passes no runtime at all,
because 1.9.1 embeds its own (measured). The AppRun, the icon selection, the
`.DirIcon` handling, the refusals, the reproducibility stamp and the oracle have
no counterpart there.

JumpLink:
### Independent review — one measurement in the PR body is wrong, and it is the load-bearing one

Reviewed in a separate worktree, against the pinned tool (1.9.1, build 296, git `8c8c91f`).

**`gjsify ship --target appimage` is not offline.** The PR body and ADR § A26.1 say
*"it embeds its own runtime … so this packer passes no runtime and `gjsify ship --target
appimage` stays offline like every other format."* Measured, `ARCH=x86_64` on an `x86_64`
host — so not only the cross-arch case:

```
Downloading runtime file from https://github.com/AppImage/type2-runtime/releases/download/continuous/runtime-x86_64
Downloaded runtime binary of size 944632
Embedding ELF...
```

No cache (four consecutive packs, four downloads). With the network blocked: exit 1, no file,
*"Failed to download runtime file … pass it to appimagetool with `--runtime-file`"*. So § A25's
refusal of a fetching pack step is not achieved, only relocated into the tool; the Dockerfile's
SHA-256 pin on `appimagetool` is defeated one level down, because the ~940 KB of ELF a user
executes comes from a **rolling** tag with no digest in this tree; and the determinism assertion
can only ever see two packs close enough together that `continuous` did not move. `--runtime-file`
is the fix and needs a decision about where a pinned runtime comes from — `status/open-todos.md`
→ *"The AppImage pack is not offline"*.

**Three claims verified and standing.** `APPIMAGE_EXTRACT_AND_RUN=1` is harmless on a FUSE-capable
host (full pack 0.27 s with, 0.26 s without; no `appimage_extracted_*` survives either path).
The `.DirIcon` repair is complete, and now for a known reason: appimagetool replaces a **symlink**
`.DirIcon` ("Deleting pre-existing .DirIcon") and leaves a **regular file** alone — with one
present and every path stamped it touches nothing, two packs byte-identical. And the arch table is
better than the PR claims: all four labels produce a correct-arch container from one x86_64 host
(`file`: x86-64, AArch64, Intel i386, ARM EABI5), because the runtime is fetched per `$ARCH`.

**Where the ninth mutation had company.** The `/symbolic/` guard was not the only assertion passing
for the wrong reason: the oracle accepted an empty `staged[]` at exit 0 (`verify-modes.py` has had
the guard since the `.deb`); nothing checked `.DirIcon` at all, so the repair was held up by a
sha256 comparison between two packs a second apart; the determinism leg ended in
`assert.ok(DIR_ICON_NAME === '.DirIcon')`, a constant compared to its own literal; the MOUNT leg
inherited `APPIMAGE_EXTRACT_AND_RUN` from the environment; the flat-AppDir control asserted only
that AppRun throws, which it does at the interpreter probe on a host with no `gjs`; and deleting
`assertAppImageIsPackable`'s **call** from `appDirPayload` left all three of its tests green.
Fixed, each with the mutation that reds it.

**Two open items, described rather than rebuilt** — both are yours to decide:

1. `AppRun` prints `gjs (>= 1.86)` and checks `command -v gjs`. A too-old GJS passes the refusal
   and dies later, which is the failure § A24 says the format exists to prevent; the `.deb` beside
   it enforces the floor through `Depends:`.
2. `assertAppImageIsPackable` fires inside `packOne`, i.e. **after** the project's build — while
   the tool gate deliberately fires before it, on the argument that a knowable refusal must not
   cost a build. `kind: "cli"` is knowable from settings alone.

Verified after the changes: CLI unit suite 1796/1796; `ship`, `ship-appimage`, `ship-from-stage`,
`ship-layout`, `ship-declaration` 130/130 with `appimagetool` on PATH — tier 3 built, read back
and **ran** a real AppImage; `tsc` clean, format clean, committed bundles reproduce.

JumpLink:
### The runtime is pinned, and the two MEDIUMs are built

**`cf4dcce160`** — plus `2608240f43` for the oracle's setuid gap.

**The pin.** `.docker/ci-fedora.Dockerfile` installs `runtime-x86_64` from type2-runtime's
**dated `20251108`** release under `sha256sum -c`, the same shape the tool above it uses.
`findPinnedRuntime` looks for `runtime-<arch>` in `/usr/local/share/gjsify/appimage-runtime`
(moved by `GJSIFY_APPIMAGE_RUNTIME_DIR`), and `appImageToolArgs` passes `--runtime-file` when
there is one. The rolling tag really does move: `continuous`'s `runtime-x86_64` and `20251108`'s
are the same 944 632 bytes and **different content**, measured today — so the old determinism
assertion was green only because two packs happen seconds apart.

**Two packs with a pinned runtime are byte-identical, and the pack succeeds with the network
blocked.** Both measured, and both are now asserted: the e2e derives a runtime from the artifact
it just built (the ELF prefix the oracle already locates), points the env var at it, and packs
**with a dead proxy in the environment** — which only a pinned runtime survives.

**An arch with no pin is announced, not refused.** `appImageRuntimeNotice` prints one of two
sentences on every pack, never behind `--verbose`: the pinned file, or that this pack fetches from
a rolling tag and embeds bytes nothing here can name. The three unpinned arches keep working.

**`AppRun` checks the floor, fail-open.** Below it: exit **126** (found, cannot run) rather than
127 (not found), naming the version it got. An answer it cannot parse, or an interpreter that
fails `--version`, **runs** — and silently.

**`kind: "cli"` is refused before the build**, from `FormatDescriptor.requiresDesktopEntry` and a
`resolveShipKind` shared with `resolveShipSettings`. The e2e proves the order: the fixture's build
script writes a marker, and the marker's absence is the assertion. The icon half stays in the
packer, because an icon is discovered.

### Three of my own assertions could not fail, and the mutations found them

Worth naming, because they are the same class as the `/symbolic/` guard:

| mutation | first result | why it was green | after |
|---|---|---|---|
| drop `--runtime-file` from the arg vector | **0 red** | the e2e asserted the printed "pinned" line, not that the flag arrived | packs with the network cut → red |
| invert `-lt` on the **major** | **0 red** | every case used `1.70` vs a `1.86` floor, settled by `-eq` + minor either way | added `0.99` and `2.0` → red |
| delete the fail-open gate | **0 red** | `execFileSync` drops stderr on success, so `stderr === ''` was vacuous exactly there | `spawnSync` → red |

A `case "$major$minor" in ""\|*[!0-9]*)` guard was **deleted** rather than kept: the `sed` above it
can only emit two runs of digits or nothing, so it guarded a state nothing produces — a guard that
cannot fail, which is what I came here to find.

Six mutations now red: `findPinnedRuntime` → undefined (unit 2, e2e 1), the dropped flag (unit 1,
e2e 1), `requiresDesktopEntry` removed (unit 1, e2e 1), the major comparison, the fail-open gate,
and `exit 126` → `127`. Plus `resolveShipKind` → always `'app'` (e2e 1).

### Verified

Unit **1805/1805** (4292 assertions). E2E `ship-appimage`, `ship`, `ship-from-stage`,
`ship-layout`, `ship-declaration`, `ship-flatpak`: **141/141** with `appimagetool` 1.9.1 on PATH —
tier 3 built, oracle-read and **ran** a real AppImage, pinned and unpinned. `tsc` clean, format
clean, committed bundles reproduce byte-identically.

ADR § A26.1 now records the measurement **and** the pin; § A25's refusal of a fetching pack step
holds again for the architecture this image builds. `status/open-todos.md` carries what is left:
three architectures without a pinned runtime, blocked on the same `platforms:` line as the arch
entry beside it.

JumpLink:
### My own test would have red this PR's e2e shard — `732af6e79e`

Worth writing down, because it is the rig-lies shape rather than a code defect.

The new pre-build test asserts the `gjsify.ship.kind` refusal. Every local run had
`appimagetool` on PATH, so it passed. **Without** the tool — which is today's CI image, where
tier 3 prints its skip — `assertCanPack` answered first with *"appimagetool is not on PATH"* and
the test red. The e2e shard was still queued, so nothing had reported it yet.

The fix is not a skip guard. The two refusals are not peers:

- `assertCanPack` says **not on this machine** — answered by installing something.
- `assertKindCanPack` says **not for this project** — installing appimagetool would not help.

So the project-level refusal now runs **first**. A CLI author on a host without the tool gets the
sentence that is actually about them, instead of being sent to a GitHub release for a tool that
was never going to fix it. The test then runs on every host, with the tool and without.

Verified both ways after the change: **10/10** with `appimagetool` on PATH and **10/10** without;
unit **1805/1805**; the six-suite ship family **141/141**.

And one more configuration, because it is the one the CI image will have once
`build-ci-image.yml` pushes the pinned runtime: the whole suite with
`GJSIFY_APPIMAGE_RUNTIME_DIR` pointed at the **pristine `20251108` release runtime** and
**no network at all** (dead proxy) — 10/10, real AppImage built, oracle-read and **run**. The
digest in the Dockerfile was re-verified against a fresh download using the exact `ARG`s the file
declares, and the layer's command chain was executed locally to a temp root: `OK`, exit 0.

PR #1665, opened 2026-09-12
fix(cli): compile the AppImage's own schemas

JumpLink:
Learn6502 0.8.0's first AppImage built in real CI mounts and dies on its first line:

```
JS ERROR: Error: GSettings schema eu.jumplink.Learn6502 not found
  _init@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:833:23
```

It carries `usr/share/glib-2.0/schemas/eu.jumplink.Learn6502.gschema.xml` and no
`gschemas.compiled` beside it. GSettings does not fall back to reading sources.

## The premise, which was true until #1664

`commands/ship.ts` compiles the cache for every layout except Linux, and that was correct when it
was written: on Linux the `.deb`/`.rpm` postinst compiles the SYSTEM schema directory at install,
where our schemas merge with every other package's, so a cache staged there is this app's ~700
bytes installed over every other package's — and one `rpm -e` takes the system's own away with it.

The condition is spelled `layout.os === 'linux'`. What it MEANT is "this artifact has an install
step". Those were the same sentence for a year. An AppImage is the first artifact for which they
differ: it takes the Linux layout because it wraps the same `/usr` payload (§ A27), and it has no
install step because not having one is the whole point of the format. It inherited a skip written
for a postinst it does not have.

## The fix

The compile goes where the ARTIFACT is, not where the layout is. `compileSchemasForPayload` runs on
the AppImage's own pack path and adds `share/glib-2.0/schemas/gschemas.compiled` to the AppDir. It
reads the PAYLOAD rather than the project's files, which is what keeps `--from-stage` working on a
host that never saw the project. The Linux stage is untouched, so § A27's claim that the AppImage
stages nothing of its own still passes unchanged.

`FORMATS.appimage.host.requiredTools` gains `glib-compile-schemas` — the only row in the table
where that tool sits on a pack path. The comment that had it absent argued a `--from-stage` pack
would arrive with the cache already in it; no Linux stage has ever carried one.

`appDirPayload` now REFUSES a payload that stages schema sources with no cache. That guard is on the
pure function rather than in `packOne`, because every reader downstream is blind to this failure:
appimagetool packs the tree, `verify-appimage.py` reads it back and agrees with the plan, and two
packs are byte-identical. Every listing in the suite was green over an artifact that cannot start.

## AppRun was the obvious suspect and was innocent

Exporting `GSETTINGS_SCHEMA_DIR` from `AppRun` would have changed nothing, and that is measured
rather than reasoned. On the real 0.8.0 artifact: compile `gschemas.compiled` into the extracted
AppDir, touch NO environment variable anywhere, and the application starts. The staged launcher's
`XDG_DATA_DIRS` was pointing GSettings at the right directory the whole time; the directory had
nothing in it. The recorded measurement is in `renderAppRun`'s docstring so the next reader does not
re-suspect it.

## The gate asserts the effect, and it was broken on purpose

`gsettings` — glib's own CLI, resolving through `g_settings_schema_source_get_default()`, which is
the path `Gio.Settings` uses and is not ours — is handed the `XDG_DATA_DIRS` the staged launcher
exports, RECORDED from inside the chain rather than composed by the test, and must read a key out of
the app's schema. Its negative control deletes the cache and requires the read to FAIL, which is how
the suite learns the AppDir answered and not the host's own `/usr/share`. Tier 3's fixture now
constructs a `Gio.Settings` and prints a key off it, so "and it starts" is a claim about the
payload.

Two things had made the suite blind: the fixture's schema was `<schemalist/>`, which compiles to a
valid cache containing no schema, and its bundle was `print(Gtk, Adw)`, which proves the typelibs
resolved and says nothing about `share/`. Both replaced.

Three mutations, each red on the assertions that should see it — details in a comment below.

## Only one of the four install-time refreshes aborts

`cacheRefreshCommands` names four and an AppImage runs none. `glib-compile-schemas` aborts;
`gtk-update-icon-cache` does not — measured against the rebuilt image on GTK 4 / glib 2.88.3, with
no `icon-theme.cache` anywhere, `Gtk.IconTheme.has_icon('eu.jumplink.Learn6502')` is `true` and
`false` for a name that is not there. `update-desktop-database` and `update-mime-database` write
into a system database an install owns, and an AppImage is not installed. `share/locale` needs no
step on any layout.

## Verified

A real Learn6502 0.8.0 AppImage packed with this branch on Fedora 44 (gjs 1.88.1, Wayland) carries
`gschemas.compiled` (728 B), starts, takes `eu.jumplink.Learn6502` on the session bus and exports
`/eu/jumplink/Learn6502/window/1` — a window stands. `tests/e2e/ship-appimage` is 12/12 with tier 3
exercised against pinned appimagetool 1.9.1; the CLI unit suite is 1814/1814; `ship`,
`ship-from-stage`, `ship-layout`, `ship-flatpak`, `ship-macos`, `ship-windows`, `ship-signing`,
`ship-declaration` and `gsettings` all green. `ship-msi` needs `msitools`, which this workstation
does not have.

Closes the 0.8.0 release blocker.


JumpLink:
### Red → green, and each mutation broken on purpose

The tables live here rather than in the PR body, which is checked as a squash commit body.

**Before the fix** — `packages/infra/cli` unit suite, the two assertions that encoded the false premise:

```
✖ the appimage format descriptor › is Linux-bound the way flatpak is, and execs appimagetool alone
    Expected: ["appimagetool"]
    Actual:   ["appimagetool","glib-compile-schemas"]
❌ [Node.js 24.19.0] 2 of 1805 tests failed · 4290 assertions
```

**After** — `✔ [Node.js 24.19.0] 1814 tests passed · 4305 assertions`, and
`tests/e2e/ship-appimage` 12/12 with tier 3 exercised (pinned appimagetool 1.9.1 + runtime 20251108).

### Three mutations

| # | what was broken | result |
|---|---|---|
| A | `compileSchemasForPayload` returns `undefined`; the `appDirPayload` refusal disabled | 3 red |
| D | the cache staged at the AppDir ROOT instead of under `usr/` — **file present in the AppDir** | 2 red |
| E | the effect test's own negative control neutered (`rmSync` removed) | 1 red |

Mutation A reproduces the incident inside the fixture, same GJS override frame as the Learn6502 crash:

```
✖ resolves its own GSettings schema — read by `gsettings`, not by a file listing
✖ REFUSES to assemble an AppDir whose schemas were never compiled
✖ builds, reads back and RUNS a real AppImage
    JS ERROR: Error: GSettings schema org.example.ShipDemo not found
    _init@resource:///org/gnome/gjs/modules/core/overrides/Gio.js:833:23
```

Mutation D is the one that matters for the shape of the assertion: `gschemas.compiled` **is** in the
AppDir, so a presence check passes, and both effect assertions still go red. Mutation E proves the
negative control is live rather than vacuous.

### The repaired artifact

Packed from this branch against the real Learn6502 checkout, out-of-tree:

```
$ node .../lib/index.js ship linux --target appimage --skip-build --out <tmp>
[gjsify ship] appimage: eu.jumplink.Learn6502-0.8.0-1.x86_64.AppImage (1874424 bytes)

$ ls usr/share/glib-2.0/schemas/
eu.jumplink.Learn6502.gschema.xml   2657
gschemas.compiled                    728      ← was absent in the shipped 0.8.0

$ ./eu.jumplink.Learn6502-0.8.0-1.x86_64.AppImage --appimage-extract-and-run
onStateChanged {"mainButtonState":"hidden", ... }
[Editor] First initialization with controller
GNOME GameConsole: Initialization complete
  → still running, holds eu.jumplink.Learn6502 on the session bus,
    exports /eu/jumplink/Learn6502/window/1
```

Host: Fedora 44, gjs 1.88.1, glib 2.88.3, Wayland.

### Suites run locally

| suite | result |
|---|---|
| `packages/infra/cli` units | 1814 pass |
| `ship-appimage` | 12 pass (tier 3 exercised) |
| `ship`, `ship-from-stage`, `ship-layout` | 115 pass |
| `ship-flatpak`, `ship-macos`, `ship-windows`, `ship-signing`, `ship-declaration`, `gsettings` | 117 pass |
| `ship-msi` | needs `msitools`, absent on this workstation |
| `gjsify format --check`, `gjsify lint`, `verify-committed-bundles.mjs` | clean |

PR #1666, opened 2026-09-12
fix(gtk-host): attribute a face to its font source

JumpLink:
The two darwin legs of `gtk-os-suites.yml` went red on `main` at b84c214 with no source change
behind them. Those legs pair the JS in the checkout with the PUBLISHED GTK bundle, so what moved
was the release: 0.51.0 is the first bundle to carry the six Adwaita faces (#1662), and the
prebuild-pointer commit aimed `main` at it.

`initFonts()` resolves TWO font sources — the application's own directory (`GJSIFY_FONT_DIR`) and
the runtime bundle's (`GJSIFY_GTK_RUNTIME_FONT_DIR`) — but reported them in ONE flat pair of
lists. `InitFontsResult.dir` was deliberately kept singular when the second source landed, so a
caller asking "was my staged face found" could not start answering yes because the platform's
arrived. `registered`/`declined`/`failed` were not given the same treatment, and seven assertions
in `fonts.spec.ts` had been reading them as "what MY staged directory contributed" since before
any published bundle carried a face. 0 became 6; one staged path became seven paths.

Finding those faces is correct and is the whole point of #1662, so this is a test defect. But
pinning 6 would be worse than the bug: the seventh face breaks it again, and it would still be
asserting nothing about the directory the test staged.

So the attribution moves into the result. `InitFontsResult.sources` now carries a
`FontSourceOutcome` per directory — what THAT source registered, what it declined, what failed —
computed in the loop that hands each file to the font map, which is the only place that knows
where a face came from. A caller cannot re-derive it afterwards without comparing path prefixes,
which is a guess about filesystem layout rather than a measurement. The spec now asks the source
it staged what it contributed, so no count in it knows how many faces a bundle carries.

win32 carries the same assumption and was ALREADY failing it — seven tests, three more than
darwin, because its font map registers faces where CoreText declines them. That leg reports green
only because `@gjsify/gtk-host` runs there as a `continue-on-error` probe, gating blocked on the
Unix-only table rows (#1446), so the failures never reached the job status. Linux has no
batteries-included runtime bundle at all, so the variable is never set there and the assumption
is latent rather than triggered.

Which is why the fix is not only the seven reads. A new suite stages BOTH directories from the
options instead of waiting for an environment to supply one, so the two-source shape is exercised
on every host — linux included, where nothing would otherwise reach it. That suite is also the
discriminator for the per-source read itself: it fails if the attribution stops excluding the
bundle, and it fails if it starts excluding everything.

Measured on Fedora 44 / Node 24 / `PangoCairoFcFontMap`, with the host's own six Adwaita faces
staged as a stand-in for the bundle's font directory:

- before, `GJSIFY_GTK_RUNTIME_FONT_DIR` set: 7 of 697 failed — the same seven the win32 leg lists
- after, same variable set: 700 of 700 passed
- after, variable unset: 700 of 700 passed, identical 3119 assertions

Each touched assertion was then broken on purpose to check it can still fail. Dropping the
`families` diff guard turns one red. Making the per-source read answer with the flat lists turns
nine red — the original seven plus both new ones. Emptying the attribution inside `initFonts()`
turns six red. One assertion could NOT be made to fail and is gone rather than kept: with nothing
staged and no family named, `initFonts()` takes its early return and reports an empty `families`
whatever the diff below it does, so the guard is held by the sibling test that reaches it.

ADR 0038's amendment records the widening, since it is the half that was missing when the second
source landed.

### DIFF
--- a/docs/adr/0038-shipped-application-fonts.md
+++ b/docs/adr/0038-shipped-application-fonts.md
@@ -563,6 +563,20 @@ it once. `InitFontsResult.dir` still means the APPLICATION's directory alone, so
 caller asserting "my staged face was found" cannot start answering yes because the platform's
 arrived.
 
+**AND `InitFontsResult.sources` NOW CARRIES WHAT EACH SOURCE CONTRIBUTED, because keeping `dir`
+singular was not enough.** The flat `registered` / `declined` / `failed` lists SPAN both
+directories, and that widening stayed invisible for as long as the published bundles shipped no
+face. The day `@gjsify/gtk-runtime-*` started staging the six Adwaita faces, every count taken
+from those lists moved by six — including seven assertions in `fonts.spec.ts`, which turned the
+darwin legs of `gtk-os-suites.yml` red on the first scheduled run after the 0.51.0 prebuild
+pointer landed, with no source change behind it and five months after those assertions were
+written. Those legs pair the JS in the checkout with the PUBLISHED bundle, so the trigger was a
+release, not a commit. A caller cannot re-derive the attribution afterwards without comparing
+path prefixes, which is a guess about filesystem layout rather than a measurement; the loop that
+hands each file to the font map is the only place that knows which directory it came from. So
+each entry is a `FontSourceOutcome` — the same argument `families` makes about the family diff,
+one field over. Ask a source what it contributed; the flat lists are the SUM.
+
 **The per-OS honesty rows of § 3 carry over unchanged, and one of them now bites harder.** Linux
 finds `share/fonts` through the `XDG_DATA_DIRS` the loader already sets; Windows can only be
 reached by `add_font_file`, which is what makes the handover load-bearing there exactly as § 4
--- a/packages/framework/gtk-host/src/fonts.spec.ts
+++ b/packages/framework/gtk-host/src/fonts.spec.ts
@@ -21,7 +21,9 @@ import PangoCairo from 'gi://PangoCairo?version=1.0';
 import {
     adwaitaUiFontAvailability,
     applyUiFontPolicy,
+    type FontFaceFailure,
     initFonts,
+    type InitFontsResult,
     isUnsupportedByFontMap,
     matchFontFamily,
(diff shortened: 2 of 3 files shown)


JumpLink:
### The seven assertions, and what each one was reading

Kept out of the PR body because a squash merge writes that body as the commit message, and a
markdown table there fails commitlint's `body-max-line-length`.

| `fonts.spec.ts` test | read that moved | darwin | win32 |
|---|---|---|---|
| does nothing … when no directory is named | `registered.length` 0 → 6 | red | red |
| reports a named directory that cannot be read | `registered.length` 0 → 6 | green | red |
| ignores the strays a font directory carries | `registered.length` 0 → 6 | green | red |
| never silently drops a face that will not open | `[failed…, declined]` 1 → 7 paths | red | red |
| leaves a face OUTSIDE the directory invisible | `registered.length` 0 → 6 | green | red |
| finds the staged face and accounts for it | `[registered, declined]` 1 → 7 paths | red | red |
| routes it to `declined` … never to `failed` | `declined`/`registered` 1 → 7 | red | red |

Three of them are green on darwin purely because a CoreText map registers nothing: every bundle
face lands in `declined`, so `registered.length` stays 0 and the assertion passes for the wrong
reason. On win32 and on Linux the same map registers them, which is why the win32 probe shows
the full seven.

### The deliberate breaks

| break | where | result |
|---|---|---|
| drop the `families` diff guard | `initFonts()` | 1 of 700 red |
| per-source read answers with the flat lists | `fonts.spec.ts` helper | 9 of 700 red |
| attribution emptied per source | `initFonts()` | 6 of 700 red |

### Runtime pairing

| leg | bundle fonts | gtk-host step | font tests |
|---|---|---|---|
| darwin-x64 / darwin-arm64 | 6 | gate | red before, green after |
| win32-x64 | 6 | `continue-on-error` probe (#1446) | red before, green after |
| linux (`node-gi.yml` gtk-host-node) | no bundle — system GTK | gate | latent; now covered by the new suite |

PR #1669, opened 2026-09-13
fix(cli): carry the offline sources to Flathub

JumpLink:
`gjsify flatpak sync-flathub` repointed `tag` and `commit` and nothing else.

For any app that uses `gjsify flatpak sources`, that produces a Flathub manifest
which parses, passes both linters, and cannot build. Flathub builds with the
network unshared, and the install reads a generated tarball list that has to
live in the Flathub repo *and* be named in the module's sources. Repointing the
git source is half the bump.

**The missing half is invisible until a build runs**, and then it surfaces four
steps away from its cause. Measured on Learn6502 0.8.0, the first release after
that app dropped its vendored dependency cache:

    ../meson.build:23:0: ERROR: Command `gjs-console -m build-aux/gjsify.gjs.mjs
    install --immutable` failed with status 1

Nothing in that line points at the manifest. The same manifest built fine for
every earlier release, because until then the git checkout carried a committed
dependency cache and was self-contained.

## What changed

`sync-flathub` reads the list from the tag it is about to pin, using
`git show <tag>:<file>` rather than the working tree, so the file that travels
to Flathub is the one the pinned commit contains. It copies it into the clone
and names it in the manifest, once however often the command runs. Absent at the
tag is a skip, not an error: apps that install online or vendor their
dependencies must not grow a reference to a file their repo does not have.

The "nothing to do" test now reads the copied file as well as the manifest. The
tarball list changes whenever a dependency does, so a release can need a PR with
the pin unmoved, and that case used to return "already at `<version>`" over a
Flathub repo carrying a stale list.

`sources` is typed as objects **and** bare strings, which is what a flatpak
source list holds.

## Tests, and one honest negative

`editManifest` had no coverage at all. It has six cases now, including the run
that used to be impossible: a second sync, over an array that by then holds an
object and a string.

    break: ignore sourcesFile              ->  3 of 1821 failed
    break: append on every run             ->  1 of 1821 failed
    break: assume index 0 is the git pin   ->  1 of 1821 failed
    fixed:                                 ->  1821 passed, 4317 assertions

The `flatpak-sync` E2E gained two cases, and it earned them: the first version
of this change **failed** that suite, correctly. Its git stub answered every
unknown subcommand with exit 0 and no output, and the code read only the exit
status of `git show` — so an empty answer became "the file exists and is empty",
and an empty list was written into the Flathub repo and named in the manifest.
Strictly worse than skipping: the build still cannot install, and now the
manifest claims it can. Real git exits non-zero there, which is exactly why the
bug survives reading. The stub now answers `show` the way git does.

    break: read an empty answer as a file  ->  1 of 6 failed
    break: compare tags only               ->  1 of 6 failed
    break: never copy the list             ->  1 of 6 failed

Two of those did not discriminate until a follow-up commit, both found by
mutation and not by reading: the changed-list fixture carried no
`x-checker-data`, so `editManifest` injected one and the PR opened for the wrong
reason.

A fourth mutation stayed green and is worth naming rather than hiding: removing
the string guard from the search changes nothing in JS, because `"file.json".type`
is `undefined` and the search skipped it either way. The type fix is for the next
reader, who would write `source.url.startsWith(...)` and get a compile error
rather than a crash. The spec says so in as many words.

## Not in this PR

`gjsify flatpak check` wraps `appstreamcli` and `flatpak-builder-lint`, and
neither knows this offline-cache convention. There is still no way to ask "will
this manifest install with the network off" short of building it, which matters
for hand-written manifests and for repos whose release automation does not go
through `sync-flathub`. That check belongs in `check`, and it is a separate
change.

---


JumpLink:
## Refactor pass — two real defects, four cleanups

Reviewed adversarially against this branch. Every count in the original body reproduced exactly; the mutation table below was re-run rather than trusted.

### 1. A silent skip that ships the very bug this PR exists to prevent

`readSourceListAtTag` caught **every** `execFileAsync` error and returned `null`, i.e. "the tag carries no list". Measured in a scratch repo: git cannot support that reading — path-absent, unknown-tag and not-a-git-repository are **all exit 128**, and the message is localized (`Schwerwiegend: Pfad … existiert nicht` on this host), so neither the status nor the text separates them. `maxBuffer` overrun and a missing `git` never reach an exit status at all (`code` is `ERR_CHILD_PROCESS_STDIO_MAXBUFFER` / `ENOENT`) and were swallowed too.

Failure scenario, confirmed end to end: `sync-flathub --version v2.0.0 --commit <sha>` against a checkout that never fetched the tag. `--commit` is the one path that skips `resolveCommitForTag`, so nothing else verifies the tag exists. The list is skipped, the pin moves, the Flathub repo keeps its stale list, and the PR opens — the original incident, reintroduced through the fix for it, on the path release automation actually uses.

The tag is now asked about separately (`git rev-parse --verify --quiet <tag>^{commit}`) and the answer believed as a **SHA, not an exit code** — the same lesson this PR already learned for `git show`. A non-numeric error code throws instead of skipping.

### 2. `sources.includes()` doubled the entry

Run against the built `editManifest`: a manifest already holding `./gjsify-sources.json` came back as `["./gjsify-sources.json", "gjsify-sources.json"]`. flatpak-builder resolves a bare string against the manifest's directory, so those name one file — and the same tarballs get spliced in twice, producing exactly the duplicate `dest`/`dest-filename` pairs that `flatpak sources` dedupes its own output to avoid. Now compared with a leading `./` stripped, with a spec case.

### 3–6. Cleanups

- **The list's name was a cross-command contract spelled four times.** `flatpak sources` writes it, `sync-flathub` carries it, each with its own literal — changing one would leave the other reading a file nobody writes. One `FLATPAK_SOURCES_FILE` in the group's `utils.ts` now.
- **`flatpak diff` inherited the hazard this PR creates.** This PR is what first puts a bare string into `modules[0].sources`, and the sibling command still typed that array as objects only — so the stated benefit ("the next reader gets a compile error, not a crash") did not hold one file over. Same union and narrowing; behaviour unchanged, diff e2e 4/4.
- **`mainModule.sources = sources` was dead** past the `sources.length === 0` throw. Removing it fails 0 unit and 0 e2e tests. Gone.
- **The incident was written out four times.** It now stands in full beside the code that would be "simplified" back into the bug — the push in `editManifest` — and the other three say only what is locally true. Nothing measured was dropped; the spec keeps its honest negative about the type change. Tree comment ratio for `packages/infra/cli` unchanged at 0.547.

**Kept deliberately:** `args.sourcesFile ?? '…'` is unreachable from the CLI (yargs `default:` always applies; deleting it fails 0 tests), but the sibling `sources.ts` uses the identical belt-and-braces pattern twice, so removing it here would make this file the odd one out. With the shared constant it is no longer a magic string, which was the real cost.

### Mutation results

Original claims, all reproduced:

| break | red |
|---|---|
| never name the list | 3 of 1821 |
| append on every run | 1 of 1821 |
| assume index 0 is the git pin | 1 of 1821 |
| read an empty answer as a file | 1 of 6 |
| compare manifests only | 1 of 6 |
| never copy the list | 1 of 6 |

Added:

| break | red |
|---|---|
| swallow every git failure again | 1 of 7 |
| remove the stub's `rev-parse` arm (exit 0, no output) | **3 of 7** — the SHA check is load-bearing |
| compare `./` spellings as written | 1 of 1822 |
| drop `mainModule.sources = sources` | 0 of 1821 + 0 of 6 (dead) |
| drop the `??` default | 0 of 6 (unreachable) |

One methodology note worth carrying: the first mutation harness restored `src` but left `lib` built from the mutated source, so one run measured the *previous* mutation's binary. Caught, harness hardened to rebuild after restore, table re-run.

### Verified independently of the authoring pass

    node --test tests/e2e/flatpak-sync/run.mjs   →  7 pass, 0 fail, EXIT=0
    node packages/infra/cli/dist/test.node.mjs   →  1822 tests · 4319 assertions, EXIT=0

Also green during the pass: `flatpak-diff` 4, `flatpak-sources` 6, `flatpak-release` 4, `flatpak` 18, `ship-flatpak` 9. `gjsify tsc --noEmit`, `gjsify lint`, `gjsify format --check` all clean.

### Left for later, on purpose

`gjsify flatpak diff` still reports `✅ in sync` on tags alone, so it can bless a Flathub repo carrying a stale list — the same asymmetry this PR fixed in sync-flathub's "nothing to do". Closing it means fetching and comparing the list over the network, which belongs with the deferred `flatpak check` work named in the original body, not here.

https://claude.ai/code/session_01EiXBPydVkcbfcCBb2S95wV

PR #1671, opened 2026-09-13
feat(cli): convert the app icon for every OS

JumpLink:
A shipped app has no icon. Measured on the released Learn6502 0.8.0, on both
platforms it ships to:

    macOS 15.7.9, "Learn 6502 Assembly.app"
      Info.plist | grep -i icon     → no icon key at all
      find <app> -name "*.icns"     → nothing
      ls Contents/Resources         → lib  share

    Windows 11, installed from the .msi
      Start-menu entry              → the generic blank-document tile

The taskbar button and the GNOME window DO show the icon, because GTK sets the
window icon at runtime from the theme inside the GResource. Everything that asks
the *artifact* — Explorer, the Start menu, Finder, the Dock — got nothing,
because `plist.ts` wrote neither an `.icns` nor `CFBundleIconFile`,
`pe-launcher.ts` emitted no resource directory, and `msi.ts` had no Icon table.

## One seam, three writers

`utils/ship/icons.ts` renders the app's declared icon once; `ico.ts`, `icns.ts`
and the layout rows encode it per target. Not a Windows fix with a macOS patch
beside it — the same gap had already been recorded twice under two different
headings, which is what a missing seam looks like.

**The pixels come from `Rsvg.Handle.render_document` onto a cairo
`ImageSurface`, in a child `gjs -c`.** Each half of that was measured, not
assumed:

* **Not gdk-pixbuf.** `/usr/lib64/gdk-pixbuf-2.0/2.10.0/loaders/` on this host
  holds only `libpixbufloader-xpm.so`, and the CI image has no loaders directory
  at all. `new_from_file_at_size` on an SVG happens to work here through a
  different loader path — a property of this host's packaging, not of the API.
* **Not `rsvg-convert` or ImageMagick.** Neither is in the CI image;
  `rsvg-convert` is not on the workstation either.
* **A child process, not `gi://` in-process**, because the CI leg that assembles
  the Windows artifact runs the CLI under **Node** (`lib/index.js`). The CLI
  source has zero GI imports and Node reaches GI only through a node-gi prebuild
  that leg does not have. GJS is on every assembling host by definition (ADR 0024
  § A2). The script is a string constant, so the single-file `dist/cli.gjs.mjs`
  needs no asset.
* **Reproducible across hosts**: the same SVG at 16 and 256 px gave one sha256
  (`18cf46a3…`) on the workstation and inside `ghcr.io/gjsify/ci-fedora:44`,
  librsvg 2.62.3 on both.

Precedence: a project's own sized PNG wins at its size — decided by the **IHDR**,
not the path, so a `32x32/` file that is really 48 px is refused by name — the
SVG fills the rest, and a `-symbolic` icon is never a source.

## The mechanism, which matters more than any one format

An app with no icon source, or PNGs that leave a declared size uncovered with no
SVG to fill it, **fails the stage** and names the file to add. Two guards, not
one: `commands/ship.ts` renders for every row declaring `Layout.icon`, and the
row itself (`iconFor`) refuses an `app` that arrives without rasters. A
`kind: 'cli'` project gets nothing and no key.

Linux declares no sizes on purpose: the theme carries the SVG and GTK renders
it. AppImage's `.DirIcon` was checked and still accepts it.

## What each target gets

**win32** — `pe-launcher.ts` appends a `.rsrc` section to the image it already
hand-emits: the three-level directory, ids ascending because the loader binary
searches, data entries as **RVAs** (the classic silent mistake), plus
`NumberOfSections` 2→3, `SizeOfInitializedData`, `SizeOfImage`, one more section
header and data-directory slot 2. `RT_GROUP_ICON` is written in the **resource**
form — 14-byte entries naming resources by id, not the `.ico` file form's
16-byte offset entries. Sizes 16/24/32/48/64/256.

The MSI names the launcher itself as the Icon-table binary, on the advertised
shortcut's `Icon_` and as `ARPPRODUCTICON`. Embedding alone covers Explorer,
taskbar, Alt-Tab and a pinned entry; the MSI row covers the two consumers the
`.exe` cannot reach — an *advertised* shortcut resolves through a descriptor, and
Add/Remove Programs reads only `ARPPRODUCTICON`. That last reasoning is from the
Windows Installer docs, not from a run.

**darwin** — `icns.ts` writes the ten PNG-backed elements `iconutil` itself
produces (`icp4 ic11 icp5 ic12 ic07 ic13 ic08 ic14 ic09 ic10`, `TOC ` first) to
`Contents/Resources/<binaryName>.icns`, and `plist.ts` emits `CFBundleIconFile`.
`CFBundleIconName` is deliberately NOT emitted — it needs an `Assets.car`.

## Verified on a real macOS, by Apple's own tool

Run in the macOS 15.7.9 guest against the generated file:

    iconutil -c iconset <file>   → exit 0, exactly ten iconset files,
                                   16/32/32/64/128/256/256/512/512/1024 px,
                                   no warnings
    iconutil -c icns back        → exit 0

Honest limit: the roundtrip is 100 830 bytes against 68 793, so iconutil
re-encoded. This is "reads and accepts", **not** byte identity.

And the rendering, since `qlmanage` is unusable from an SSH session (no Aqua
session — it hangs with no output; `launchctl asuser` needs root): the file plus
the key were injected into a 0.8.0 bundle, launched through LaunchServices, and
the Dock photographed through the QEMU monitor. **The Dock draws the artwork.**
Three other copies of the identical bundle in the same guest still read
`CFBundleIconFile=NONE` — that is the negative control.

## Red arms

Nine unit mutations, each restored and re-run green afterwards: data entry as a
section offset instead of an RVA → 1 red; type entries unsorted → 1; resource
data-directory slot zeroed → 2; group written in file form → 1; `.icns` element
length excluding its head → 2; project PNG ignored → 2; layout row stops refusing
an iconless app → 1; plist stops emitting the key → 2; MSI drops
`ARPPRODUCTICON` → 1.

Nine more at the oracle level, each asserting exit 1 **and** the specific
message: emptied resource data directory; first `RT_ICON` not a PNG; group entry
declaring 128 for a 256 image; deleted `.icns`; stripped `CFBundleIconFile`
(dangling-reference arm in both directions); corrupted `ic07`; emptied Icon
table; blanked `Icon_`; removed `ARPPRODUCTICON`.

## Readers

Each artifact is read back by tools that are not ours. `.exe`: a CPython `struct`
walk plus inflate of every PNG, with `objdump -p` required to agree on types and
leaf count; on the workstation also `wrestool`, `icotool`, Pillow and
ImageMagick. `.icns`: a CPython element walk in CI, Pillow locally for all ten
elements — `icns2png` reads only 3 of 10 and is NOT relied on. `.msi`: `msiinfo`
for the Icon row, `Icon_` and `ARPPRODUCTICON`, plus an `msiinfo extract` stream
compared byte-for-byte against the launcher. The full table is in a comment
below. Authority runs: `iconutil` on macOS PASSED; the Windows shell and the
Start-menu icon were not run.

## Two defects found on the way

1. **`promisify(execFile)` under the GJS-built CLI** resolves with the bare
   stdout string — the polyfill has no `promisify.custom`. The first rasterizer
   worked under Node and crashed under GJS. Now `spawnToCompletion`.
   `schemas.ts` has the same shape and survives only because it ignores the
   result; that latent copy is named and left.
2. **`msiinfo export <msi> Icon` writes the binary column into the cwd**
   (IDT semantics). It dropped a 15 KiB launcher into the repo root during the
   e2e. Oracle and suite now export from scratch directories.

## Counts, verified independently of the authoring pass

    committed bundles reproduce byte-identically from source at HEAD   EXIT=0
    @gjsify/cli unit          1843 tests · 4538 assertions   EXIT=0
    tests/e2e/ship-macos      41/41
    tests/e2e/ship-windows    34/34

During the pass, also green: `ship-layout` 39, `ship` 44, `ship-signing` 21,
`ship-from-stage` 20, and `ship-msi` 26 inside the CI image (msitools is not on
the workstation). `oxfmt --check`, `oxlint` and `tsc --noEmit` clean.

## Not run, and what would settle it

A Windows install of a build from this branch. Pass = the Start-menu entry shows
the same icon as the taskbar button, Explorer shows it on the `.exe`, and
Add/Remove Programs shows it. Also unverified: whether Windows 11 draws the small
**PNG-compressed** entries (16–64) at BMP quality. The 256 entry is documented
PNG; the smaller ones are PNG because one encoder is one code path. If they look
wrong, the fix is a BMP encoder for the small sizes in `ico.ts` alone.

Learn6502 needs **no change** — it already declares the scalable SVG through
`gjsify.ship.icon`, which is exactly the input this seam renders from.

https://claude.ai/code/session_01EiXBPydVkcbfcCBb2S95wV


JumpLink:
## Readers, and where each one lives

| artifact | in CI | workstation only | authority |
|---|---|---|---|
| `.exe` `.rsrc` | CPython `struct` walk + inflate of every PNG; `objdump -p` must agree on types and leaf count | `wrestool`, `icotool`, Pillow, ImageMagick | Windows shell — **not run** |
| `.icns` | CPython element walk + inflate | Pillow (all ten); `icns2png` reads only 3 of 10 and is **not** relied on | `iconutil` on macOS — **PASS** |
| `.msi` | `msiinfo` Icon row + `Icon_` + `ARPPRODUCTICON` + stream `cmp` byte-equal to the launcher | — | Start-menu icon — **not run** |

PR #1677, opened 2026-09-14
fix(node-gi): select the fontconfig pango backend, and make it reach pango

### LINKED ISSUE #1668: gtk-runtime: non-Latin scripts render as tofu on win32 and darwin (Pango picks the platform backend)
## What happens

Any script whose coverage lies outside the platform font backend's own fallback chain
renders as empty boxes in a `--app node` GTK application on Windows and macOS, even
though the system carries a face for it and the bundle's fontconfig is configured to
find it.

Measured with Learn6502 0.8.0, whose Tamil translation is complete:

| platform | default backend | `PANGOCAIRO_BACKEND=fc` |
| --- | --- | --- |
| macOS 15.7.9 x86_64 | Tamil = tofu | **renders correctly** |
| Windows 11 x64 | Tamil = tofu | **renders correctly** |
| Linux (GJS, system GTK) | renders correctly | — |

## What it is not

- Not the translation, and not gettext: the UI switches languages correctly, the
  strings are the Tamil ones, only the glyphs are missing.
- Not a general fallback failure: **Japanese renders perfectly** on the same bundle,
  same run, same font map. Hebrew and every Latin script are fine. So the fallback
  chain works — it just does not reach this script.
- Not a missing font: `/System/Library/Fonts/Supplemental/Tamil Sangam MN.ttc` is
  present on macOS, `C:\Windows\Fonts\Nirmala.ttc` on Windows, and `fc-list :lang=ta`
  finds Tamil families on the macOS guest.
- Not a missing configuration: the win32 bundle already ships
  `etc/fonts/fonts.conf` carrying `<dir>WINDOWSFONTDIR</dir>`, and
  `gtk-runtime.js` already exports `FONTCONFIG_FILE`/`FONTCONFIG_PATH` pointing at
  it. The configuration is right; it is simply not the backend Pango uses.

## Cause

PangoCairo picks the platform backend by default — `pangowin32` (DirectWrite) on
Windows, CoreText on macOS. `gtk-runtime.js` already documents the DirectWrite half
for *faces* (ADR 0038 § W1-W5: a `FONTCONFIG_FILE` naming a directory of faces moves
the win32 map by zero families). The consequence for *script fallback* was not
followed through: those backends resolve a font for the requested family, and their
fallback for an uncovered script does not reach the faces fontconfig would.

## The fix, and why it is not a one-liner

The mechanism is one environment default in `packages/node-gi/node-gi/gtk-runtime.js`,
next to the `FONTCONFIG_FILE` lines that are already there:

```js
if (existsSync(fontsConf)) {
    setIfUnset('FONTCONFIG_PATH', join(bundle.dir, 'etc', 'fonts'));
    setIfUnset('FONTCONFIG_FILE', fontsConf);
    setIfUnset('PANGOCAIRO_BACKEND', 'fc');   // <- this
}
```

But it swaps the entire font map on both platforms for every consumer, which is a
rendering change, not a bug fix, and it needs to be treated as one:

- text quality changes (DirectWrite/ClearType and CoreText hinting give way to
  FreeType) — this wants a side-by-side visual comparison on both platforms, not
  just a green suite;
- ADR 0038's measured win32 rows describe the DirectWrite map and stop being true;
- `adwaitaUiFontAvailability()` currently expects `Adwaita Sans Text` on Windows
  (DirectWrite exposes the `opsz`/STAT name) and `absent` on macOS (CoreText cannot
  register a face at run time). Under fc both become the plain declared family —
  an improvement, since the bundled Adwaita face would finally be usable on macOS,
  but it moves assertions that exist today;
- the win32 and darwin GTK legs have to run, and `@gjsify/gtk-host` on win32 is
  still a `continue-on-error` probe (#1446), so that leg's verdict has to be read
  rather than trusted.

Deliberately not rushed into 0.51.1/0.51.2 alongside a consumer release for exactly
these reasons.

## Reproducing

Any `--app node` GTK app with a Tamil (or other non-covered script) catalogue:

```
LANGUAGE=ta ./eu.jumplink.Learn6502          # boxes
LANGUAGE=ta PANGOCAIRO_BACKEND=fc ./eu.jumplink.Learn6502   # correct
```

Use a script whose glyphs are visibly distinct — and pick the discriminator with
care: Japanese looks like a fine test and passes either way.

JumpLink:
Fixes #1668, on macOS **and** on Windows. An earlier revision of this branch concluded the
Windows half was unfixable here; that conclusion was wrong and § *The wrong diagnosis* below
keeps the trail rather than deleting it.

> **Draft until the two windowing legs have run.** The cause is established from the sources
> (§ *The real cause*), and the win32 claim above is the prediction that follows from it — but
> `windows-gtk-windowing` and `macos-gtk-windowing` are the only oracles, and they have not
> reported on this revision yet. If they contradict it, the claim moves, not the measurement.

Non-Latin text renders as tofu — empty boxes — in every shipped `--app node` app on macOS and
Windows. Measured 2026-09-12 with Learn6502 0.8.0 on real hardware, with the variable exported
by hand before launch (nothing in the product set it then):

| Platform | default backend | `PANGOCAIRO_BACKEND=fc` |
|---|---|---|
| macOS 15.7.9 | Tamil = tofu | **correct** |
| Windows 11 | Tamil = tofu | **correct** |
| Linux (GJS) | correct | — |

It was not the translation (the UI switched language, only the glyphs were missing), not a broken
fallback chain, not a missing font (`Tamil Sangam MN.ttc` / `Nirmala.ttf` present throughout) and
not missing configuration (the win32 bundle ships `etc/fonts/fonts.conf`, and the loader already
points `FONTCONFIG_FILE` at it).

`pangocairo-fontmap.c` builds the FIRST backend compiled in — coretext → win32 → fc — and
`PANGOCAIRO_BACKEND` is the only override. So the bundle drew through CoreText or
pangowin32/DirectWrite, whose script fallback does not reach those faces, and every
`FONTCONFIG_*` variable the loader sets was being read by nobody.

## The change

Two parts, because on Windows setting the variable is not enough to have it read.

**1.** One line in `packages/node-gi/node-gi/gtk-runtime.js`, beside the `FONTCONFIG_*` lines it
is supposed to make effective:

```js
setIfUnset('PANGOCAIRO_BACKEND', 'fc');
```

Scoped to darwin+win32 and to the bundle, not unconditional. `maybeWireGtkWindowingEnv()` has
already returned on anything else, so no platform branch is added. Linux was measured: it builds
a `PangoCairoFcFontMap` with nothing set, fc being the only backend compiled in there. And fc is
a safe choice only where something hands fontconfig both a configuration and faces — which a
windowing bundle does and an arbitrary host's Pango does not owe us. `setIfUnset`, so an operator
pinning `coretext` or `win32` wins.

**2.** `mirrorWindowingEnvIntoCrt()` in `gi.js`, which replays the loader's recorded writes
through `g_setenv()` on the first `requireGi()`. Win32 only, and § *The real cause* is why.

The comment that earns part 1 records the reasoning error it replaces. The old one said
*"`XDG_DATA_DIRS` already reaches `<bundle>/share/fonts` wherever fontconfig drives Pango … so on
darwin the set above is enough"* — on darwin fontconfig did not drive Pango, and nothing made it.

## The wrong diagnosis, kept because it was well-evidenced

CI run 34873488108, `Windows windowing proof`, on the bundle built from this branch:

```
font map PangoCairoWin32FontMap (PANGOCAIRO_BACKEND=fc); unknown glyphs latin=0 tamil=5 ...
```

The variable was set and the process still built the **win32** map. That was written up as
*"gvsbuild's pango has no FreeType/fontconfig cairo backend to select, so the variable is inert
there"*, and pushed into the code comment, ADR 0038, `status/open-todos.md`, the bundle README
and the PR body. It is false.

`bin/pangocairo-1.0-0.dll` from `GTK4_Gvsbuild_2026.6.0_x64.zip` (range-fetched via the ZIP
central directory — 33 KB, not the 300 MB archive):

- registers `PangoCairoFcFontMap` **and** `PangoCairoWin32FontMap`;
- imports `pangoft2-1.0-0.dll` and `fontconfig-1.dll`;
- carries ` win32 fontconfig` — what `pango_cairo_font_map_new()` prints for an unknown value —
  beside the three literals it compares against: `win32`, `fc`, `fontconfig`.

Verified from the other side too: `fc` and `fontconfig` are both accepted (exact match, two
aliases), while a prefix such as `f`/`fo`/`fon` is rejected with `Unknown $PANGOCAIRO_BACKEND
value` and a NULL map.

**Two search artefacts hid it, and both are worth naming.** A `PangoCairo[A-Za-z]*FontMap` regex
cannot match the digits in `Win32`, so the type list looked shorter than it was; and `strings`
defaults to a 4-character minimum, one more than `fc` has, so the alias was invisible in the very
dump meant to settle it.

**The evidence against the wrong reading was already in that log.** An invalid value is LOUD — a
`g_critical` naming every compiled-in backend, and a NULL font map. The run printed no such line
and got a working map. "The platform refused the value" and "the value never arrived" are
distinguishable, and the run had already distinguished them.

## The real cause: a Windows process has two environments

| environment | written by | read by |
|---|---|---|
| Win32 block | `SetEnvironmentVariableW()` | `GetEnvironmentVariableW()`, `g_getenv()`, children |
| C runtime copy | `_putenv()` / `_wputenv()` | `getenv()` |

The DLL loader reads the Win32 block too, which is why `maybePrependGtkRuntimeDllPath()` has
always worked from an in-process `process.env.PATH` write.

`pango_cairo_font_map_new()` reads plain `getenv("PANGOCAIRO_BACKEND")`. Node's `process.env`
setter is `uv_os_setenv()`, which on win32 is `SetEnvironmentVariableW()` and nothing else
(`deps/uv/src/win/util.c`; `node_env_var.cc` `RealEnvStore::Set`).

So the loader's write reached the Win32 block — `process.env` read it back, `g_getenv()` saw it,
a child process inherited it — and `getenv()` returned NULL, which takes the `!backend` branch
and builds the platform map. Every GLib-side variable the loader sets (`GSETTINGS_SCHEMA_DIR`,
`GDK_PIXBUF_MODULE_FILE`, `XDG_DATA_DIRS`, `GIO_MODULE_DIR`, `GST_PLUGIN_SYSTEM_PATH`) goes
through `g_getenv()` and has always worked, which is why nothing else in the win32 bundle ever
pointed at this. `FONTCONFIG_FILE` and `FONTCONFIG_PATH` are the other two that do not:
fontconfig reads them with `getenv()` as well, so `etc/fonts` on win32 had two reasons to look
unread and now has none.

The repair is `g_setenv()`, because glib already solved this — its win32 implementation calls
`_wputenv()` first and `SetEnvironmentVariableW()` second, and says why in its own source. That
needs a loaded GLib, which does not exist when the loader runs at module top level, so the replay
happens on the first `requireGi()`: after the addon loads, and before anything can build a font
map or read a fontconfig configuration, since both are reachable only through a namespace.

## The 2026-09-12 contradiction, resolved

The comment asserted both *"on real Windows 11, `fc`: renders"* and *"on win32 it is REFUSED"*.
The gvsbuild pin was already `2026.6.0` on 2026-09-12, so the published Learn6502 0.8.0 bundle
and this tree's bundle ship the same `pangocairo-1.0-0.dll` — at most one could be true of the
backend.

Both are true, and the difference is the only thing that ever differed. On 2026-09-12 the
variable was in the process **before launch**, so the C runtime picked it up at startup and
`getenv()` answered `fc`. In CI it was written **in-process by the loader**, so it never reached
that copy. Same backend, same value, different reach. Neither observation is withdrawn; together
they are what names the mechanism.

## The trade-off, stated

**This swaps the font map for every consumer of the bundle.** Text is rasterised by FreeType
instead of CoreText or ClearType, so hinting and subpixel rendering change for ALL text, Latin
included. `status/open-todos.md` argued from exactly that cost — *"it changes text rendering for
the whole application, which is not a decision a runtime bundle may take for its consumer"* — and
this PR overrules it: the bundle was already taking that decision silently, by shipping a Pango
whose compiled-in ordering picked the platform map. A different rasteriser is a preference and one
variable gives it back; a glyph that never arrives is a defect with no way back.

**The upside beyond the bug, on darwin**: the bundled Adwaita typeface becomes usable there.
`add_font_file` is a vfunc the CoreText map does not implement, so the darwin bundles carried
~7.3 MB of faces no process could reach and `adwaitaUiFontAvailability()` could only answer
`absent`. On an fc map `initFonts()` registers them like everywhere else. The `add_font_file`
handover stays wired on BOTH platforms regardless: it is what any process on a platform map still
needs, and nothing in the loader can know which map a given process gets.

## The test, and why its oracle discriminates

`packages/node-gi/node-gi/test/font-script-coverage.test.mjs` +
`test-programs/pango-script-coverage.program.mjs`.

The oracle is `pango_layout_get_unknown_glyphs_count()` — Pango's own tofu counter, the number
that produces the boxes a user reports, not a `!== null` check. `load_font()` answers a font for
every request including the one that will draw boxes.

1. **The counter is proved in both directions on every platform, before anything is concluded.**
   Latin must count 0 (not stuck high — no usable face at all) and the SAME Tamil text with
   `pango_attr_fallback_new(FALSE)` must count > 0 (not stuck at 0). Measured on Fedora 44 /
   Pango 1.57.1: `latin=0 tamil=0 tamil-no-fallback=5`, and with a fontconfig configuration that
   hides the Tamil faces, `tamil=5` — one per character.
2. **Which backends exist is asked of the process, not inferred from the OS.**
   `PANGOCAIRO_BACKEND=<unknown>` makes pango print its compiled-in list and return a NULL map
   (measured: *"Available backends are: fontconfig"*). This is the probe ADR 0038 § 3 already
   recommended to a human holding a Mac, turned into a check.
3. **The loader-set value is measured against the same value in the LAUNCH environment.** This is
   the arm the first diagnosis did not have. It separates "this pango cannot build that map" from
   "the value did not reach the library" — the two readings of one log line — and it is what would
   have caught the mistake on the day it was made.
4. **Where fontconfig can be selected, the claim carries its own negative control.** The map must
   be the Fc one, Tamil must render, and the measurement is repeated in a child process with the
   platform's own backend pinned back — the test FAILS if that control does not show the tofu. So
   it cannot pass both with and without the fix. It also proves the `setIfUnset` half: the child's
   explicit value wins. Where fontconfig genuinely is not compiled in, that gap is ASSERTED rather
   than skipped, so the branch retires itself.
5. **Which map a process built is read with `g_type_from_name()` over all three private map
   types**, all of them rather than the first match, and against the set registered BEFORE the map
   was built, so the delta names the map. The first version reported `find()` over a fixed
   coretext → win32 → fc order, which is how a win32 name could stand in a log while an fc map
   existed beside it. Not from the wrapper either: node-gi resolves `constructor.$gtype` to the
   nearest introspected ancestor and answers `PangoFontMap` for all three (gjs answers
   `unknown_PangoCairoFcFontMap` for the same read — a divergence that would have made a wrong
   control look right).
6. **A sixth case measures the host nobody can rent: a Mac with no Homebrew.** Choosing fontconfig
   on darwin puts the whole font supply behind a configuration the bundle does NOT ship — unlike
   win32, where `etc/fonts/fonts.conf` travels in the tarball. The test points `FONTCONFIG_FILE`
   at a path that does not exist, which makes fontconfig fall back to the configuration compiled
   into the library, and asserts text still draws. Mechanism verified on Linux: *"Cannot load
   default config file"* on stderr, built-in fallback still resolving `/usr/share/fonts`,
   `latin=0 tamil=0`.

The probe also prints both environment views (`process.env` and `g_getenv()`) beside the map it
got, so a run where every view agrees and the map does not says so in one line instead of being
inferred.

Japanese is deliberately absent. It rendered correctly in the same window of the same run in which
Tamil was empty boxes, because both platform maps carry a CJK fallback — a test built on it is
green before and after the fix.

Wired into `macos-gtk-windowing` and `windows-gtk-windowing`, the only oracles this change has,
and it runs on the Fedora `npm test` leg for the counter half.

## Also in this PR

`tests/e2e/ship-layout` asserted the darwin `fontGap` wording that had already been rewritten —
the one red check on the branch. The wording is corrected (it claimed GTK is never built against
fontconfig on darwin, in the same string that describes reading faces through fontconfig) and the
assertion follows it.

## Docs

- **ADR 0038 § Amendment 3** — the decision, the measurement, the wrong diagnosis with the two
  search artefacts that produced it, the environment split, the resolved contradiction and the
  trade-off. § W1-W5 and the darwin research are NOT withdrawn: they remain the correct account
  of the DirectWrite and CoreText maps, and § W1-W5's "moves the map by zero families" holds for
  a reason unrelated to reach — that map is filled only from the DirectWrite system collection.
- **`guides/bundled-fonts.md`** — both platforms move from `declined` to `registered`, `declined`
  kept as the platform-map outcome, and the troubleshooting rows folded into one.
- `docs/ship-formats.md`, the darwin and windows `Layout.fontGap` messages `gjsify ship` prints,
  the win32 bundle README, `packages/node-gi/AGENTS.md` (one bullet for the environment split),
  and the stale claim in `windowing.test.mjs`.
- `status/open-todos.md`: the win32 `etc/fonts` deletion proposal is REVERSED — that payload is
  load-bearing now — and the darwin entry records what is still unmeasured, which is a shipped
  `.app` on a Mac with no Homebrew.


JumpLink:
Reviewed. One question the ADR leaves open, chased down and **answered in the fix's favour** — recording it because the next reader will ask it too.

The loader sets `FONTCONFIG_PATH`/`FONTCONFIG_FILE` only `if (existsSync(fontsConf))`, and the darwin bundle ships no `etc/fonts/fonts.conf`: `packages/node-gi/scripts/build-gtk-runtime-darwin.mjs` mentions neither, where the win32 builder mentions `fonts.conf` three times. So on macOS this PR selects `fc` and hands fontconfig **no configuration at all**. The 2026-09-12 hand-measurement does not settle that, because it ran on a Homebrew-equipped VM and passed an explicit `FONTCONFIG_FILE`; and a macOS CI runner has Homebrew preinstalled, so a green `macos-gtk-windowing` would not have settled it either — the end user's Mac is the one machine with neither.

It is fine, and here is why. `libfontconfig.1.dylib` from `@gjsify/gtk-runtime-darwin-x64@0.51.1` carries a compiled-in fallback configuration:

```
<fontconfig>  <dir>/System/Library/Fonts</dir>
              <dir>/Library/Fonts</dir>
              <dir>~/Library/Fonts</dir>
              <dir>/System/Library/AssetsV2/com_apple_MobileAsset_Font7</dir>
              <dir prefix="xdg">fonts</dir> …
```

fontconfig scans those recursively, so `/System/Library/Fonts/Supplemental/Tamil Sangam MN.ttc` is reached with no config file present anywhere. The baked `/usr/local/…` paths in the same dylib are `<include ignore_missing="yes">` and cache locations, not a dependency.

Worth a line in ADR 0038 § Amendment 3: on darwin the faces are found through fontconfig's **built-in** fallback, not through anything the bundle ships — which also means adding an `etc/fonts/fonts.conf` to the darwin bundle later would *narrow* what is scanned unless it repeats those four directories.

Nothing else found. The win32 gate sits with the data (`windowingEnvWrites()` returns `[]` off win32) rather than with the consumer, `setIfUnset` records only its own writes so the `overwrite: true` mirror cannot clobber an operator's value, and `layout.ts` changes comments and one message string with no packaging behaviour attached.

Happy to merge once the two macOS legs report.

JumpLink:
Both macOS legs now report, and they contradict the darwin half. Recording what they said, because the test's refusal is the interesting part.

`Script-coverage proof (#1668)` fails on **darwin-arm64 AND darwin-x64**, at the same assertion, while `Windows windowing proof` passes it:

```
the coretext map draws Tamil with 0 unknown glyphs, so this test passes with AND
without the PANGOCAIRO_BACKEND line and proves nothing. That is a finding, not a
pass: the platform font map now reaches the script it did not reach on
2026-09-12, and the line in gtk-runtime.js needs re-justifying against a fresh
measurement.
```

That is the test doing its job — it declines to be vacuously green — and it is the case the draft note was held open for. The claim moves, not the measurement.

**What makes it sharper: `macos-15-intel` is macOS 15, the same major as the machine the original tofu was measured on.** On 2026-09-12 I measured Learn6502 0.8.0 on macOS 15.7.9 x86_64 by hand: default backend → every Tamil glyph an empty box, `PANGOCAIRO_BACKEND=fc` → renders correctly, screenshots both ways. The darwin GTK bundle is unchanged on this branch (`build-gtk-runtime-darwin.mjs` is untouched), so both runs use the same pango. Two macOS 15 machines, opposite answers.

Three candidates, none of them established yet:

1. The CI runner carries fonts a stock Mac does not (Xcode, Homebrew casks), so CoreText's fallback reaches Tamil there and not on a user's machine.
2. The probe and the rendered UI disagree — the probe counts unknown glyphs through one path, GTK draws labels through another, and CoreText's *fallback* may cover what the *requested family* does not.
3. macOS updated something between 15.7.9 and the runner's 15.x.

I have a bare macOS 15.7.9 VM with that exact app still installed and can run the probe program on it, which is the one machine neither CI nor a developer laptop provides. Say the word and I will, but the outcome is yours to act on: if (1), the line stays and the test needs a host-capability gate rather than a hard failure; if (2), the probe is measuring the wrong thing and the Windows pass needs re-reading too; if (3), the line is still right for older macOS and wrong to justify from a runner.

Windows is unaffected either way — that leg passes the same proof.

Separately: I pushed `test(gtk-host): exempt the iso-codes warning the fc backend uncovers` for the darwin-x64 shipped-closure failure. `GtkFontChooserDialog` builds its language filter from iso-codes and only reaches for it once fontconfig drives Pango; the path it names is a Linux one macOS does not have. Scoped to the message like `missingPortal` beside it, measured on the same head (arm64 silent, x64 warns), pattern checked against both messages and three controls.

PR #1708, opened 2026-09-19
ci: probe the superseded-run repair job

JumpLink:
Throwaway probe for #1705. Not for merge.

The first run on this commit concluded FAILURE against a body line of 178 characters that was
current at that moment — the sequential case, which no in-flight rule can reach. This edit
makes the description pass, so the next run is `current-pass` and `clear-superseded` should
restart the red one with a real `actions: write` token.


JumpLink:
Closing as intended: this PR existed only to exercise the privileged `clear-superseded` job against a `base.sha` that carries the script, before it first runs on `main`. It did its job — it found a read-back race no fixture had (312 ms after the POST, still `run_attempt: 1`, annotated "0 of 1 restarted" while the restart had in fact happened). The fix is in #1705.

PR #1712, opened 2026-09-19
feat: parse Blueprint in the build, spawn nothing

JumpLink:
ADR 0053 clause 5 makes the in-repo parser authoritative for the build "when it reports none
across the corpus". It reports none. `@gjsify/vite-plugin-blueprint` stops spawning
`blueprint-compiler` and calls `parseBlueprint` + `emitGtkBuilderXml` from `@gjsify/blueprint`,
with all five `EmitOptions` seams and **no fallback to the binary** — a silent one would put back
the dependency clause 7 exists to remove and would make a build's output depend on the host.

Clause 7's deletion list is deliberately NOT started here. `resolve-compiler.ts`, its spec, the
`oxlint-disable` in `loading-stack.ts` and the fence gate's skip are all still in the tree, and
each is its own change. One PR, one thing.

## The measurement, re-run on this branch

`blueprint-compiler 0.20.4` (Fedora 44) against the parser at `@girs` 5.2.0, over every tracked
`.blp` — the whole set, because stage A refuses any tracked `.blp` no corpus entry covers:

```
node scripts/check-blueprint-corpus.mjs --require-oracle
git ls-files '*.blp' | wc -l        -> 85
```

| | files |
|---|---:|
| tracked `.blp` | 85 |
| byte-equal (48 rule files + all 12 real build inputs) | 60 |
| divergent — `corpus/divergences.mjs` entries | 0 |
| refused by us, **compiled by the oracle** | 7 |
| refused by the oracle and by us | 18 |
| **in the dangerous direction** (we emit, the oracle rejects) | **0** |

Identical to the last pre-#1710 measurement, entry for entry. The oracle verdicts are not read
off the manifest: stage B re-runs the binary over each refused file and fails if the recorded
verdict disagrees.

## macOS and Windows now run a Blueprint e2e suite. They never did before

An earlier draft of this section said two OS legs "were green while testing nothing" because both
Blueprint e2e suites skipped on `resolveBlueprintCompiler() === null`. **The conclusion was right
and the mechanism was wrong, which made the claim unfalsifiable in the worst way.** Measured: the
only reference to either suite in `.github/` is a comment at `main.yml:1412`; they run in the
`e2e` job alone, `runs-on: ubuntu-latest` with `container: ghcr.io/gjsify/ci-fedora`, whose image
bakes `blueprint-compiler` in. `macos-suites.yml` ran exactly one e2e — `ship-signing` — and
`windows-suites.yml` ran per-package `test:node` and no e2e at all.

So on CI those guards were dead code: they only ever fired for a developer without the compiler
locally. macOS and Windows said nothing about Blueprint not because a guard skipped the suite but
because **no job ran the suite**.

The guards are still deleted — a skip for a dependency the build no longer has cannot fire — and
the coverage is now real rather than asserted:

- `macos-suites.yml` and `windows-suites.yml` each gained one step running
  `tests/e2e/library-blueprint`, the suite whose entire subject is that a `--library` build
  compiles `.blp` at all. Both workflows already build the workspace CLI the suite drives, so it
  is one step and no new setup.
- **win32 is the leg with the most to say.** The flip moved path handling out of a spawned
  process into JavaScript, and that job deliberately strips the git-bash POSIX utilities from
  `PATH`. The old path could never have run there at all: `blueprint-compiler` on Windows means an
  MSYS2 install shipping a shebang script Windows cannot execute.
- `create-app` is deliberately **not** added: nine templates scaffolded and npm-installed, at a
  15-minute-per-template budget those workflows do not have. Its three `.blp` templates exercise
  the same transform `library-blueprint` drives directly. It stays Fedora-only and both its
  comment and this section say so.

Locally, with a `blueprint-compiler` on `PATH` that exits 66 and prints a trap line:

```
node --test tests/e2e/library-blueprint/run.mjs   ->  pass 2, fail 0, skipped 0, trap silent
```

That is a Linux run and proves only that the suite passes without the binary. **The macOS and
win32 evidence is this PR's own CI**, which is the point of adding the steps rather than restating
the claim.

The malformed fixture still fails through the right parser: `template {` is "found `{`, expected
`$`" from `BlueprintSyntaxError`, naming the `.blp`, and never "Using declarations must have an
initializer" — the discriminator that suite was written for.

## The user-visible consequence, stated plainly

After this PR a `.blp` using one of the 7 constructs the oracle compiles **fails the build** where
it used to succeed: `internal-child`, a file-level `translation-domain`, a multi-step `.parent`
lookup chain, an inline `menu` as a property value, a response flag, an untyped closure, and a
type from a namespace `@girs` publishes no vocabulary for (`Gio`, `Gdk`, `GObject`). That is
clause 3's declared design — the safe direction against output that is plausible and wrong — but
it is a behaviour change, and none of the 85 files here uses one.

So the refusal had to become something a build can print. The parse half already threw
`BlueprintSyntaxError` with `file`/`line`/`column` as fields; the emit half threw
`new Error('blueprint: line N: ...')` and named **no file at all** — survivable while its only
reader was a harness that knew which file it had handed in, not survivable as the thing shown to
someone with twelve `.blp` open. New `BlueprintEmitError`, exported beside it, carrying `file` and
`line` as fields:

```
/abs/path/window.blp:6: `Gio.ListStore` names a namespace this resolver has no vocabulary for ...
```

**And the messages themselves were written for the wrong reader.** While the parser ran in shadow,
every refusal was read by someone holding the corpus harness, so four of the seven explained
themselves in the parser's own terms — "`Value` in ast.d.mts has no menu member", "`ExtensionEntry`
… has no field for it". Someone who installed a build plugin has never seen `ast.d.mts`, and none
of the seven said the build no longer calls `blueprint-compiler` — so the obvious next move was to
reinstall a tool that cannot help. All seven now name the construct, give one action, and carry one
shared sentence saying the GNOME compiler is not involved and pointing at `corpus/refused/`, which
ships in the tarball and is held by the gate so it cannot rot. The implementation reason for each
stays, as a comment beside its `throw`. The `Gio.ListStore` case gained the workaround it never
had — `$GListStore`, the extern form, measured in `rules/35-extern-real-class.blp` against the
reference compiler's own golden.

## Guards broken on purpose, because a check never seen fail is a check never tested

1. **An emit refusal that names no file.** Reverted the `try { }` site to the old bare `Error`:
   stage E fails with "the error does not name line 4". It passed before this PR only because
   `namesLine` had a second arm matching a bare `line N:`. That arm is gone — nothing produces it
   now, and an arm that can only fire for the error this stage exists to refuse is not a check.
2. **The prefix strip in `namesConstruct`, widened for the two-field location.** Changed the try
   refusal's text so it no longer contains "try": with the widened strip, stage E fails by name;
   with the old column-mandatory strip it passes **green**, because the path
   `refused/expression-try-empty.blp` contains "try". The structural hole is all **16** refusals
   the emitter raises — none of their prefixes gets stripped — but it is only VACUOUS where the
   path happens to carry the construct name, which is **3 of the 16**: `expression-item-in-bind`
   (`item`), `expression-try-empty` (`try`), `expression-cast-literal` (`cast`). The other 13 go
   red under either regex. `null-value` is the near miss: its `names` is `` `null` `` with
   backticks, which its path does not carry.
3. **A dropped seam.** Removed `enumOrFlagsTypeOf` from the plugin's `SEAMS`: `gjsify run check`
   fails on `Required<EmitOptions>`. A type and not a test, because no single `.blp` moves bytes
   for all five.
4. **The plugin's wiring.** Dropped `SEAMS` at the call site: the new `plugin.spec.ts` fails with
   `<property name="orientation">vertical</property>` against the golden's `1` — the exact
   silent-wrong-output case, and the one thing the corpus gate structurally cannot see, since it
   hands in its own five seams.

## No second proof gate

`scripts/check-blueprint-corpus.mjs` already covers every tracked `.blp` and refuses one no entry
covers. Nothing here duplicates it. The one gap it cannot close is this plugin's own wiring — it
imports the parser directly — so `plugin.spec.ts` holds the join on ONE file whose bytes move when
the wiring is wrong, not a second corpus.

Proof the binary is not reached: the same trap as above. `gjsify run build` and a full `--app gjs`
build of `showcases/gtk/adw-blueprint-layout` (2 `.blp`) both complete with the trap silent and
the compiled XML in the bundle.

## Three things the brief did not predict, each measured

1. **`private: true` DOES block this.** `scripts/verify-published-closure.mjs --phase pre-release`
   — a step of the required `Detect runtime-triplet drift` check — refuses a release-pinned edge
   from a published package to a private target: *"dead on arrival ... npm would resolve the NAME
   to whatever unrelated package owns it"*. So `@gjsify/blueprint` is un-privated, tier 1, on the
   train, with `files`, a `status/status.json` entry, and its first publish queued in
   `status/pending-npm-bootstrap.json`.
2. **`BlueprintEmitError` carries two fields, not three.** `ast.d.mts` gives every node a `line`
   and no column (`grep -n column src/ast.d.mts` returns only `BlueprintSyntaxError`'s own
   declaration), so a column here would be invented and would point at a place nothing measured.
   Adding columns to the AST is a parser change with 40+ construction sites — not this PR.
3. **It is 23 sites, not 24.** 17 in `emit-xml.mjs` and 6 of the 7 in `resolve-ident.mjs`. The
   24th is `merged()`, which fires while the `@girs` vocabularies are combined at import time,
   about the dependency and not about anyone's `.blp`. It has no file and no line to carry and
   stays a plain `Error`, with the reason in a comment beside it.

`where` therefore changed from a formatted `line N` string to a `SourceLocation { file, line }`
across the five seams. Consumers pass function references, so no call site outside the package
moved; `execa` left the plugin's dependencies with the spawn it existed for.

## What skipping the queued bootstrap actually does

A first draft of the ledger entry said this name was alphabetically first and would stall the
train. **Both halves were wrong** and the correction is worth more than the claim: measured, it is
12th of the 215 non-private `@gjsify/*` names — the same 215 `verify-published-closure` counts as
publishable — and the release does not walk them alphabetically at all. `npm:publish:prebuilt` is
`gjsify foreach --topological … gjsify publish --tolerate-untrusted-new`, which emits
`@gjsify/blueprint` before its dependent (measured with the same `foreach --topological`), and
`--tolerate-untrusted-new` returns `skipped-untrusted-new` with **exit 0** for a name OIDC cannot
create — that flag exists precisely so one un-bootstrapped package does not break the loop.

So the real failure mode is the opposite of a stall, and worse: the parser is skipped, the plugin
is **published behind it**, pinning a name npm does not have on a required `dependencies` edge —
every consumer install of the plugin fails, on every package manager.
`verify-published-closure`'s post-release phase catches it, which is after the tarball is on the
registry. The bootstrap is a BEFORE for that reason.

## Verified

`check-blueprint-corpus --require-oracle`, `check-blueprint-census`,
`check-blueprint-corpus-counts` (bidirectional — deleting the counts from `open-todos.md` failed
it too, so they are restated and checked), `check-doc-fences`, `audit-runtimes --check`,
`verify-published-closure --phase pre-release`, `verify-committed-bundles`,
`check-agent-context-size --check`, `gjsify run build`, `gjsify run check`, `gjsify format
--check`, `gjsify lint`, and both packages' own `check` + `test`.

## Docs reconciled rather than overwritten

ADR 0053 line 347 and 386-387 said clause 5 was unsatisfied pending `Gtk.SizeGroup.mode`. Stale:
the ledger is empty AND three rule files still probe `SizeGroup`
(`23-widget-reference-list.blp`, `29-enum-non-widget.blp`, `38-null-object-id.blp`), so the
silence is an answer and not a missing question. Both passages are dated in place rather than
deleted — what they record is what a shadow run that still reported looked like — and Amendment 3
carries the flip, the behaviour change, the error class and the publishing consequence.


JumpLink:
### The `@girs` weight, which the 2026-09-16 survey left unmeasured

§ 1.3 of `docs/reports/2026-09-16-parallel-day-survey.md` says the flip "moves the dependency
rather than removing it — today the plugin needs a *binary on PATH*, after the flip it needs
*pinned `@girs` packages in `node_modules`*", and that this was **unmeasured** and "must be
measured before the flip, not after". Measured on this branch:

| `@girs` package | installed | of which the parser loads |
|---|---:|---:|
| `gtk-4.0` | 6.4 MB | 124 KB |
| `adw-1` | 2.6 MB | 56 KB |
| `webkit-6.0` | 892 KB | 36 KB |
| `gtksource-5` | 716 KB | 24 KB |
| `shumate-1.0` | 480 KB | 16 KB |
| **total** | **12 MB** | **256 KB** |

Two things that shrink the headline. The parser imports only each package's `./vocabulary`
subpath, so a bundler reaches 256 KB of the 12 MB. And a project that has a `.blp` at all already
depends on `@girs/gtk-4.0` and usually `@girs/adw-1` for its own types — both templates checked
declare them — so 9 MB of the 12 is already on disk before this PR. The genuinely new install is
`gtksource-5` + `shumate-1.0` + `webkit-6.0`, **2.1 MB**, and that is a download rather than a
platform toolchain: the thing it replaces is an MSYS2 install on Windows.

JumpLink:
### Two skip-guards this PR made stale, and they were the OS legs

Found while checking the flip's own claim rather than by reading the diff: both Blueprint e2e
suites skipped on `resolveBlueprintCompiler() === null`, which is every macOS and Windows runner.
`tests/e2e/library-blueprint` skipped **whole** — the suite whose entire subject is that a
`--library` build compiles `.blp` at all — and `tests/e2e/create-app` skipped three of nine
templates, the three with a widget tree. The two runners the flip exists for were the two
reporting nothing about it.

Both conditions are deleted rather than re-pointed, per the root-cause rule: a skip-guard for a
dependency the build no longer has is a check that cannot fire. `create-app`'s `skipReason` had
no other writer, so its seven reads went with it.

Proof, with a `blueprint-compiler` on `PATH` that exits 66 and prints a trap line:

```
node --test tests/e2e/library-blueprint/run.mjs   ->  pass 2, fail 0, skipped 0, trap silent
```

The malformed fixture still fails through the right parser: `template {` is
`found `{`, expected `$`` from `BlueprintSyntaxError`, naming the `.blp`, and never
`Using declarations must have an initializer` — which is the discriminator that suite was
written for.

JumpLink:
### Where my false sentence came from, and it is still in five files

Corrected in `54a278e`. But I did not invent "alphabetically later, stalls the train" — I read it.
It is the repo's own claim, in five places, and **root `AGENTS.md` is one of them, so it is loaded
into every agent's context on every turn**:

| file | line |
|---|---|
| `AGENTS.md` | 112 — "skipping it stalls the release train for every alphabetically later package" |
| `docs/publishing.md` | 9 — "breaks the serialized publish loop … for every alphabetically-later package (OIDC exchange 404s, workflow exits 1 — v0.4.20 incident: `@gjsify/tls-native` unbootstrapped → 60+ packages stuck at 0.4.19)" |
| `docs/runtime-platform-axes.md` | 23 |
| `docs/prebuilds.md` | 24 |
| `docs/examples-showcases.md` | 23 |

**The incident is real and correctly recorded for its time. Both of its mechanisms were removed
afterwards, and none of the five sentences was updated.** Measured:

```
git show v0.4.20:package.json   ->  npm:publish = gjsify foreach -v --no-private … \
                                      gjsify publish --tolerate-republish …
```

At v0.4.20 (2026-05-21) the loop had **neither** `--topological` **nor**
`--tolerate-untrusted-new`, so an un-bootstrapped name did exit 1 and did stall everything after
it in enumeration order. Then:

- `fb3034a205` **2026-05-22** — `feat(cli): --tolerate-untrusted-new for first-publish Trusted
  Publisher gap (#246)`. The day after the incident, i.e. the fix for it. An un-bootstrapped name
  now returns `skipped-untrusted-new` with exit 0 and the loop continues.
- `b53b5cf7d6` **2026-08-04** — `ci: order the publish sweep, verify the closure`. The loop became
  `--topological`, so "alphabetically later" stopped naming the order even in principle.

So today the sentence is wrong twice, and wrong in the *reassuring* direction: nothing stalls, the
dependent gets published against a name npm does not have, and the only thing that notices is
`verify-published-closure`'s post-release phase — after the tarball is on the registry.

I have deliberately **not** swept the five here. Correcting the root `AGENTS.md` line changes a
rule agents follow and touches the exact byte ceiling in `status/agent-context-budget.json`, which
is a counter every open PR shares; and a repo-wide publishing-doc correction under a
`feat:`-titled Blueprint PR is not where a reviewer would look for it. Say the word and I will do
it, here or as its own change.

Worth keeping either way: this is a rule whose *reason* was preserved (the v0.4.20 incident, with
its package name and its 60+ stuck packages) while the rule itself went stale, because the two
commits that fixed it did not know which prose cited it. The incident survived and the mechanism
did not.

JumpLink:
### Three corrections landed, and the first one was a real hole

**1. macOS and Windows never ran either Blueprint e2e suite.** You are right and my evidence was a
Linux run. Confirmed independently: the only `.github/` reference to either is a comment at
`main.yml:1412`; both run in the `e2e` job alone (`runs-on: ubuntu-latest`, container
`ghcr.io/gjsify/ci-fedora`, compiler baked in). `macos-suites.yml` ran one e2e, `ship-signing`;
`windows-suites.yml` ran no e2e. So the guards I deleted were dead code on CI and fired only for a
local developer — deleting them bought nothing.

**Fixed by making it true, not by weakening it.** `be40a69` adds one step to each of
`macos-suites.yml` and `windows-suites.yml` running `tests/e2e/library-blueprint`. Both workflows
already build the workspace CLI that suite drives, so it is a step and no new setup. `create-app`
is NOT added — nine templates scaffolded and npm-installed is a budget neither has — and its
comment now says it stays Fedora-only instead of claiming an OS would speak up. All four claim
sites are restated; the darwin and win32 evidence is this PR's own CI, which is the point of
adding the steps.

**2. `namesConstruct`: 3, and the other number was wrong too.** Measured with the two regexes side
by side over all 25 refusals: vacuous under the old strip is exactly your three —
`expression-item-in-bind` (`item`), `expression-try-empty` (`try`), `expression-cast-literal`
(`cast`). The emitter refuses **16**, not the 15 I wrote. Both are now in the comment and the PR
body, kept apart on purpose: the structural hole is all 16 prefixes going unstripped, the vacuity
is 3, and reporting the second as the first would make the regex look like a three-file fix.
`null-value` is the near miss — its `names` is `` `null` `` with backticks, which its path does not
carry.

**3. The messages spoke the parser's language.** Exactly the four you named. All seven now read as
`<construct> is outside the subset this parser holds. <one action>.` plus one shared sentence
saying `blueprint-compiler` is not involved and pointing at `corpus/refused/` — which ships in the
tarball, holds one named fixture per refused construct, and is held by the gate, so the pointer
cannot rot. The `ast.d.mts` reasoning is kept as a comment beside each `throw`. `Gio.ListStore`
gained the workaround it never had: `$GListStore`, corpus-backed by
`rules/35-extern-real-class.blp` against the reference compiler's own golden.

```
refused/inline-menu.blp:4:15: an inline `menu` as a property value is outside the subset this
parser holds. Declare the menu at the top level, give it an id, and point the property at it:
`menu myMenu { … }` beside the object, then `menu-model: myMenu`. This build parses Blueprint in
process (`@gjsify/blueprint`) and never calls `blueprint-compiler`, so a file the GNOME compiler
accepts can still be refused here; the package's `corpus/refused/` names every construct the
subset does not hold.
```

Merge on a green run of `ffccbbe`, not on the green shown for an earlier head — noted, and the
`e2e` steps I just added are exactly the kind that can only be judged on this head.

And thank you for the `"${B}:path"` note. The same shape bit me twice today in zsh: `$var` not
word-splitting, which makes a loop over a list variable iterate once and look like a red finding.

JumpLink:
### The OS legs, measured on this head (`ffccbbe`) rather than claimed

Three jobs now run `tests/e2e/library-blueprint`, and all three ran it — not skipped, which is the
thing worth checking, since a suite that skips exits 0 too. Read out of each job's own log:

| job | step | result |
|---|---|---|
| Node-pillar suites (win32-x64 / cmd.exe) | success | `tests 2 · pass 2 · fail 0 · skipped 0` (1.05 s) |
| Node-pillar suites (darwin-arm64 / gjs + node) | success | `tests 2 · pass 2 · fail 0 · skipped 0` (0.48 s) |
| Node-pillar suites (darwin-x64 / gjs + node) | success | — |

I nearly filed the win32 result as a false green: the API reports that step as 1 second, which for
a suite that runs a real `gjsify build --library` reads like a skip. It is not — the fixture is one
3-line TypeScript module and one `.blp`, and `skipped 0` is in the log. Worth writing down because
the reflex was right even though the alarm was wrong.

So **Blueprint now compiles on Windows and macOS in this repository's CI for the first time**, and
on the win32 job specifically, which strips the git-bash POSIX utilities from `PATH` before it
starts. The old path could not have run there at all: `blueprint-compiler` on Windows means an
MSYS2 install shipping a shebang script Windows cannot execute — the 268-line resolver and its
237-line spec exist for exactly that, and clause 7 deletes them next.

That is the evidence for the section you asked me to make true rather than weaken. The remaining
`GJS` workflow jobs are still running on this head; nothing should merge before they report.

